PR #5, opened 2018-09-03
chore: keep proxy exports files structure for backwards compatibility

emmenko:
We don't want to expose the internal structure anymore when you import components. Components now live under `src/components` etc. The structure might also change further in the future.

However, the current usage of the `ui-kit` assumes that the package has a flat folders structure. To keep the imports backwards compatible, we have "proxy export" files that mimic the "old" structure and point to the new locations. This way the migration is smoother when upgrading to new versions.

At some point we will cleanup all those deprecated folders.


emmenko:
NOTE: we have a folder `materials/images` with `svgs` that are actually not used by any ui-kit component. We should remove this from the `ui-kit` and move those into the packages that actually use them. For backwards compatibility, we will keep them here until such migration is done.

lufego:
ideally, we should move the folder `materials`outside of components

emmenko:
> ideally, we should move the folder materials outside of components

It's already out yes:
```
- src
  - components
  - materials
  - utils
```

PR #6, opened 2018-09-03
refactor: rename svg files that are transformed into react components react.svg

emmenko:
We explicitly name SVG files that need to be transformed into React component as `.react.svg` to simplify the webpack loader setup


.storybook/webpack.config.js
```diff
@@ -41,59 +41,56 @@ const uikitWebpackConfig = {
...
         ],
       },
+      // For svg icons, we want to get them transformed into React components
+      // when we import them.
       {
-        test: /\.svg$/,
-        oneOf: [
-          // For svg icons, we want to get them transformed into React components
-          // when we import them.
+        test: /\.react\.svg$/,
```
emmenko: Again, those changes are just temporary until I fix it upstream

emmenko:
Since this is blocking the migration, I'll go ahead and merge this

Issue #11, opened 2018-09-04
SelectInput does not reset when value is undefined


dferber90:
When using `SelectInput`, and when it had a value before, then the value is not removed when the `value` prop changes to `undefined`.

@islam3zzat notified me about this, and we looked into it together. He will provide a fix for it.

PR #19, opened 2018-09-04
Set `SelectInput` value to `null` if it was `undefined`

islam3zzat:
fixes https://github.com/commercetools/ui-kit/issues/11

According to React select docs https://react-select.com/props#prop-types
> type ValueType = OptionType | OptionsType | null | void

In this PR we default the value to `null` if `value`  was not available.

To reproduce 
https://codesandbox.io/s/0ozvz9vvvl

### DIFF
--- a/src/components/inputs/select-input/select-input.js
+++ b/src/components/inputs/select-input/select-input.js
@@ -70,14 +70,23 @@ export class SelectInput extends React.Component {
       option => (has(option, 'value') ? option : option.options)
     );
 
+    const selectedOptions = optionsWithoutGroups.filter(
+      option =>
+        props.isMulti
+          ? props.value.includes(option.value)
+          : has(option, 'value') && option.value === props.value
+    );
+
+    // We default to null, `type ValueType = OptionType | OptionsType | null | void`
+    /**
+     * we default to null
+     * > type ValueType = OptionType | OptionsType | null | void
+     * https://react-select.com/props#prop-types
+     */
     return {
       selectedOptions: props.isMulti
-        ? optionsWithoutGroups.filter(option =>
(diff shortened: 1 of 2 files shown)


src/components/inputs/select-input/select-input.js
```diff
@@ -70,14 +70,23 @@ export class SelectInput extends React.Component {
...
     return {
       selectedOptions: props.isMulti
-        ? optionsWithoutGroups.filter(option =>
-            props.value.includes(option.value)
-          )
-        : optionsWithoutGroups.find(
-            option => has(option, 'value') && option.value === props.value
-          ),
+        ? selectedOptions
+        : selectedOptions[0] || null,
```
dferber90: I find this harder to read than it was before. I especially dislike that `selectedOptions` (the constant) doesn't actually hold the same value as `selectedOptions` (the state).

Could you use the approach we had when pairing?

src/components/inputs/select-input/select-input.js
```diff
@@ -70,14 +70,23 @@ export class SelectInput extends React.Component {
...
+          ? props.value.includes(option.value)
+          : has(option, 'value') && option.value === props.value
+    );
+
+    // We default to null, `type ValueType = OptionType | OptionsType | null | void`
+    /**
+     * we default to null
+     * > type ValueType = OptionType | OptionsType | null | void
+     * https://react-select.com/props#prop-types
+     */
```
dferber90: This comment looks a bit weird. Did you forget to delete the first part of it?

I would frame the comment more about the use-case we do this for (supporting changing a selected value back to nothing when `undefined` is passed / no related option exists).

PR #20, opened 2018-09-04
Enable SelectInput user to override default components

pa3:
Duplicate of xxx, which didn't make it to get into master before separation of UI-Kit.


src/components/inputs/select-input/select-input.js
```diff
@@ -77,7 +107,7 @@ export class SelectInput extends React.Component {
           )
         : optionsWithoutGroups.find(
             option => has(option, 'value') && option.value === props.value
-          ),
+          ) || [],
```
dferber90: This should be `null`, as we're in the single-value mode here. 

@islam3zzat was working on this as well. More information in https://github.com/commercetools/ui-kit/pull/19.

emmenko:
@pa3 no links to our private repo please. Also it does not make sense to link JIRA issues.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools-frontend/ui-kit",
-  "version": "1.0.0-beta.30",
+  "version": "1.0.0-beta.30-pa3",
```
emmenko: What is this?

pa3:
I changed the initial approach with using named exports for `components` and `defaultComponents` to one which adds them as static fields of `*SelectInput`.

src/components/inputs/async-creatable-select-input/README.md
```diff
@@ -60,6 +60,7 @@ It supports the same properties as `react-select`, except for `onChange` and `on
...
 | `onBlur`               | `function` |    -     | -                      | -       | Called with a fake event when input is blurred. The event's `target.name` will be the `name` supplied in props. In case `props.isMulti` is `true`, the name will have `.0` appended which helps with the formik integration.                                                                                                                                      |
+| `components`           | `object`   |    -     | -                      | -       | Overrides for `AsyncCreatableSelectInput` conponents , see [what components you can override](https://react-select.com/components)                                                                                                                                                                                                                                |
```
dferber90: We usually document the static exports in the README file as well. Could you add them?

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -147,5 +179,9 @@ export class AsyncCreatableSelectInput extends React.Component {
 }
 
 const Wrapped = injectIntl(AsyncCreatableSelectInput);
-Wrapped.isTouched = AsyncCreatableSelectInput.isTouched;
+addStaticFields(Wrapped, {
+  ...defaultComponents,
+  ...customizedComponents,
+  isTouched: AsyncCreatableSelectInput.isTouched,
+});
```
dferber90: I wonder how this would influence the tree-shaking capabilities. But these components are going to be bundled with `react-select` anyways, so it probably doesn't matter at all.

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -53,6 +62,31 @@ export class AsyncCreatableSelectInput extends React.Component {
...
+      MultiValue: PropTypes.func,
+      MultiValueContainer: PropTypes.func,
+      MultiValueLabel: PropTypes.func,
+      MultiValueRemove: PropTypes.func,
+      Option: PropTypes.func,
+      Placeholder: PropTypes.func,
+      SelectContainer: PropTypes.func,
+      SingleValue: PropTypes.func,
+      ValueContainer: PropTypes.func,
+    }),
```
dferber90: It would probably also have been fine to just use `PropTypes.object` as this component is only forwarding this prop and not using it itself. `react-select`s prop-types check will warn about invalid props anyways.

pa3:
Updated READMEs with details of static properties.

src/components/inputs/async-select-input/async-select-input.js
```diff
@@ -139,5 +147,9 @@ export class AsyncSelectInput extends React.Component {
 }
 
 const Wrapped = injectIntl(AsyncSelectInput);
-Wrapped.isTouched = AsyncSelectInput.isTouched;
+addStaticFields(Wrapped, {
+  ...defaultComponents,
+  ...customizedComponents,
```
dferber90: Wouldn't these effectively overwrite each other in case they use the same name?

To me it would be a bit clearer if we used names again for the static exports, so:

```js
addStaticFields(Wrapped, {
  defaultComponents,
  components,
  isTouched: /*  ..*/
})
```

I also likes that this groups the components, so that it's clear which ones we provided, and which ones are official from `react-select`. And I like that this does not pollute the statics too much.

What do you think?

pa3: @dferber90 I am personally think `<SelectInput.Option>` looks way better in code then `<SelectInput.defaultComponents.Option>`. But that's good that you actually raised that question, because, now I think we should not export components from `components` at all, because anyways we are not letting to customise them by doing that:

```js
<AsyncSelect
...
  components={{
    ...this.props.components,
    ...customizedComponents,
  }}
...
/>
```

So why not ending up with something like:

```
addStaticFields(Wrapped, {
  ...omit(defaultComponents, Object.keys(components),
  isTouched: /*  ..*/
})
```

pa3:
Ping! @dferber90

src/components/inputs/async-select-input/async-select-input.js
dferber90: I would have done this the other way around as

Assuming that `customizedComponents` are the components we overwrite by default, I would have assumed we do:

```jsx
<AsyncSelect
// ...
  components={{
    ...customizedComponents,
    ...this.props.components,
  }}
// ...
/>
```

Then users can even overwrite our customization with the default again in case they need to.


Regarding the exports, I think users could do this instead

```js
const { Option } = SelectInput.defaultComponents
const CustomOption = () => <Option />
```

I'm open to all approaches! We can always adjust it as we go. Feel free to either merge or use the way I suggested above.

PR #22, opened 2018-09-05
chore: move dependancies to peer and dev deps

montezume:
#### Summary

If we have too many big dependancies, our bundle size will be very big. This PR aims to start this movement by moving some "easy decision" deps to peer and dev.

Feel free to suggest others that could follow similar treatment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,13 +42,8 @@
     "lodash.sortby": "4.7.0",
     "lodash.uniq": "4.5.0",
     "lodash.without": "4.4.0",
-    "moment": "2.22.2",
-    "moment-timezone": "0.5.21",
     "prop-types": "15.6.2",
-    "react": "16.4.2",
-    "react-intl": "2.4.0",
     "react-required-if": "1.0.3",
-    "react-router-dom": "4.3.1",
     "react-select": "2.0.0",
     "react-textarea-autosize": "7.0.2",
     "react-virtualized": "9.20.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -88,16 +82,31 @@
...
     "storybook-readme": "4.0.0-beta1",
+    "styled-components": "3.3.3",
     "stylelint": "9.5.0",
     "stylelint-config-standard": "17.0.0",
     "stylelint-order": "0.8.1",
     "svg-url-loader": "2.3.2",
     "url-loader": "1.1.1"
+  },
+  "peerDependencies": {
+    "moment": "2.22.2",
```
tdeekens: I would specify ranges here. Otherwise you'll get warnings a lot on the consumer's end. E.g. `"react-dom": "^15.6 || ^16.0",`

emmenko: Yes, we should have ranges. We can also have it like e.g.`"react": ">=16"`

package.json
```diff
@@ -88,16 +83,29 @@
...
+    "react-router-dom": "4.3.1",
     "react-value": "0.2.0",
     "storybook-readme": "4.0.0-beta1",
     "stylelint": "9.5.0",
     "stylelint-config-standard": "17.0.0",
     "stylelint-order": "0.8.1",
     "svg-url-loader": "2.3.2",
     "url-loader": "1.1.1"
+  },
+  "peerDependencies": {
```
emmenko: 👌

PR #28, opened 2018-09-05
[WIP] [MCD-] Buttons refactoring - Primary Action Button

lufego:
As described in [link_to_issue], here we start the buttons refactoring.

This PR:

- comes up with a new API, still compatible with old one (minimum effort to migrate)
- creates tokens for buttons


lufego:
Since I've changed the structure a bit, I need to fixes the tests. CI is failing in the meantime. Feel free to review anyways

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+
+// This holds the styles of the button, which is wrapped by a `button` (defined as `BaseButtonWrappr`) with the `button` element logic
+export const BaseButtonContent = props => (
+  <span className={props.styles}>
+    <Spacings.Inline scale="xs" alignItems="center">
+      {props.icon &&
+        // FIXME: add proper tone when tones are refactored
+        React.cloneElement(props.icon, {
+          theme: props.isDisabled ? 'grey' : 'white',
+          size: props.size === 'big' ? 'big' : 'medium',
```
islam3zzat: can we use `props.size` directly

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+      </button>
+    );
+  }
+}
+
+export default BaseButtonWrapper;
+
+// This holds the styles of the button, which is wrapped by a `button` (defined as `BaseButtonWrappr`) with the `button` element logic
+export const BaseButtonContent = props => (
+  <span className={props.styles}>
```
islam3zzat: should the prop name be`className`, also is it possible to avoid passing `className`

src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,76 @@
...
+
+PrimaryActionButton.displayName = 'PrimaryActionButton';
+
+PrimaryActionButton.propTypes = {
+  id: PropTypes.string,
+  name: PropTypes.string,
+  ref: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  ariaLabel: PropTypes.string.isRequired,
+  onClick: PropTypes.func,
```
islam3zzat: `onClick` is required in readme, but not here

src/components/buttons/primary-action-button/primary-action-button.spec.js
```diff
@@ -25,7 +25,7 @@ describe('rendering', () => {
         buttonWrapper = wrapper.find(BaseButtonWrapper);
       });
       it('should render a <BaseButtonWrapper/>', () => {
-        expect(wrapper).toRender(BaseButtonWrapper);
+        expect(wrapper).toRender('BaseButtonWrapper');
```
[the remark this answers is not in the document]
lufego: just removed it again on latest commit

src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,76 @@
...
+      dataAttr={{
+        ...dataProps,
+        ...props.dataAttr,
+      }}
+      isDisabled={props.isDisabled}
+      isToggled={props.isToggled}
+      type={props.type}
+    >
+      <BaseButtonContent
+        styles={classnames(
```
islam3zzat: should we use `className`?

src/components/buttons/primary-action-button/primary-action-button.spec.js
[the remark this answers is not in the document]
lufego: > Why

was just trying to verify why it was failing without reading the logs. bad idea 😅

lufego:
tests are now green

src/components/buttons/primary-action-button/primary-action-button.spec.js
```diff
@@ -0,0 +1,184 @@
...
+      });
+
+      describe('setting the button as disabled', () => {
+        beforeEach(() => {
+          props = createProps({ isDisabled: true });
+          wrapper = shallow(<PrimaryActionButton {...props} />);
+          buttonWrapper = wrapper.find(BaseButtonWrapper);
+        });
+
+        it('should pass the `isToggled` prop', () => {
```
islam3zzat: should pass the `isDisabled ` prop

src/components/buttons/base-button-wrapper/base-button-wrapper.js
lufego: in this case we cannot because the prop `small` would render a `small` icon. In this case we want to render rather a medium or big icon

src/components/buttons/primary-action-button/primary-action-button.js
lufego: I wanted to differentiate from `className`since this is being passed down to the component

src/components/buttons/primary-action-button/primary-action-button.spec.js
lufego: good catch 🦅👁

lufego: its actually in the assertion below, not this one

src/components/buttons/base-button-wrapper/base-button-wrapper.js
lufego: we need to reuse the classes here

src/components/buttons/primary-action-button/primary-action-button.story.js
```diff
@@ -0,0 +1,34 @@
...
+import * as icons from '../../icons';
+import PrimaryActionButton from './primary-action-button';
+import Readme from './README.md';
+
+const iconNames = Object.keys(icons);
+
+storiesOf('Buttons', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withReadme(Readme))
+  .add('1. Primary Action Button', () => (
```
islam3zzat: is the `1. ` intentional

lufego: Yes. The idea is to organize the buttons better. In the end with something like:

image

lufego:
@dferber90 do you want to review this?

If there's anything else, I'd like to get this merged and proceed with the other buttons

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+        id={this.props.id}
+        name={this.props.name}
+        disabled={this.props.isDisabled}
+        onClick={this.props.onClick}
+        type={this.props.type}
+        ref={this.props.ref}
+        className={styles.reset}
+        // Accessibility
+        role="button"
+        aria-label={this.props.ariaLabel}
```
dferber90: I would make it so that our buttons accept all `aria` props and forward them by default, just like we do for `data-`.

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+        aria-pressed={this.props.isToggled}
+        {...this.props.dataAttr}
```
dferber90: I would just filter these  using `filterDataAttributes(this.props.data)`.

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+    type: PropTypes.oneOf(['button', 'reset', 'submit']),
+  };
+
+  render() {
+    return (
+      <button
+        id={this.props.id}
+        name={this.props.name}
+        disabled={this.props.isDisabled}
+        onClick={this.props.onClick}
```
dferber90: What about all the other `on` events? I have a feeling it might be best to just spread all props and then overwrite the specific ones (or set defaults depending on the attribute).

src/components/buttons/primary-action-button/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Usage
+
+```js
+import PrimaryActionButton from '@commercetools-frontend/ui-kit/buttons/primary-action-button';
+```
+
+#### Description
+
+Primary action buttons are used for a primary action on a page. You must also pass an
```
dferber90: I would write 

> The `PrimaryActionButton` is used for ..

src/components/buttons/primary-action-button/primary-action-button.spec.js
```diff
@@ -0,0 +1,184 @@
+import React from 'react';
```
dferber90: I would write all new tests using `react-testing-library`.

src/components/buttons/primary-action-button/primary-action-button.story.js
dferber90: I don't like the numbering. There is no real order of them.

src/components/buttons/base-button-wrapper/base-button-wrapper.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Spacings from '../../../materials/spacings';
+import styles from './base-button-wrapper.mod.css';
+
+// This holds the logic of the `button` element, which wraps a `span` (defined as `BaseButtonContent`) with styles
+class BaseButtonWrapper extends React.Component {
```
dferber90: Why is this called a wrapper? A wrapper is usually around something. This seems more like a base element.

dferber90:
> comes up with a new API, still compatible with old one (minimum effort to migrate)

If it is compatible with the old one, why is there a need for migration in the first place?


> As we had a `AccessibleButton` before, that was a bit misleading. Ideally, all our components should be "accessible" (or to have some level of semantics and accessibility). Therefore, there's no need to name them like that. Instead of it, I've created a "wrapper" which holds all of the logic of a button, leaving the styling part to the `ButtonContent` itself. 

I think this is how the `AccessibleButton` was supposed to be used as well.

> PR

What is your plan for this PR as it is targeting master and you're requesting reviews? Do you want to merge the buttons into `master` one-by-one? This seems like a weird way. I would expect one big "bang" which adds all buttons at once (and probably the Link-components as well).

lufego:
> If it is compatible with the old one, why is there a need for migration in the first place? Is there a breaking change or not? This sentence is confusing me, could you explain?

As stated between parenthesis: minimum effort to migrate. That means that for the primary button, there's no breaking changes

> I think this is how the AccessibleButton was supposed to be used as well.

Yes, I just wanted to remove this abstraction of `Accessible` to be just a `Wrapper`

> What is your plan for this PR as it is targeting master and you're requesting reviews? Do you want to merge the buttons into master one-by-one? This seems like a weird way. I would expect one big "bang" which adds all buttons at once (and probably the Link-components as well).

No worries, I've just created a branch called `feature/new-buttons` so we can merge the buttons one-by-one on it

src/components/buttons/base-button-wrapper/base-button-wrapper.js
lufego: not sure I understand. Would you be able to provide a code example of exactly what you mean here?

src/components/buttons/base-button-wrapper/base-button-wrapper.js
lufego: its already being filtered in <PrimaryActionButton />

dferber90:
I'm currently putting together another branch based on this one to experiment with how I'd do the buttons.

> Make `Link` a separate component (not belonging to buttons)

FYI I think the links should be a group of components - not a single one, just like buttons.

dferber90:
I played around a bit and if we remove all the props which we could just pass-through anyways then they become really tiny:

```
export const BaseButtonWrapper = () => (
  <button
    role="button"
    aria-pressed={this.props.isToggled}
    {...this.props}
    className={styles.reset}
  >
    {this.props.children}
  </button>
);
```

```
export const BaseButtonContent = props => (
  <span className={props.styles}>
      {props.icon &&
        // FIXME: add proper tone when tones are refactored
        React.cloneElement(props.icon, {
          theme: props.isDisabled ? 'grey' : 'white',
          size: props.size === 'big' ? 'big' : 'medium',
        })}
      <span>{props.children}</span>
    </Spacings.Inline>
  </span>
);
```

At this point it doesn't seem like it's even worth having them anymore.

Additionally, as mot all buttons support icons and not all buttons support `isToggled`, I'd remove the `BaseButtonContent` and `BaseButtonWrapper` components. That abstraction doesn't seem very useful anymore. Their only left benefit is to avoid having to repeat the code. I'd rather repeat the code for each button as there are not many anyways than having the different cases of all buttons end up in the `BaseButtonContent` and `BaseButtonWrapper` components.

You can try this out for yourself by trying to write the `PrimaryActionButton` without any underlying abstraction. To me that seems a ton simpler.

---

Things I would do:
- break the old API (e.g. no more `buttonAttributes`): If people have to migrate, then I'd rather have them migrate once instead of knowing that more changes will follow
- embrace the `button` element: forward all props passed to `<PrimaryActionButton />` which the button isn't using itself to the regular `button`. Only set the ones we need ourselves and provide sensible defaults.

lufego:
> I played around a bit and if we remove all the pass-through props (or renamed props) from the BaseButtonContent and BaseButtonWrapper components then they become really tiny:

Thanks for playing around and for your comment. But shouldn't we be consistent and expose the expected props as well as properly validate them? That's also why we would have a README file documenting the button behavior and props, and just not expect that people know exactly how buttons work

> At this point it doesn't seem like it's even worth having them anymore.
Additionally, as mot all buttons support icons and not all buttons support isToggled, I'd remove the BaseButtonContent and BaseButtonWrapper components. 

I'd say all buttons should support this state, as well as `isHovered` or something like this. The reason is not to **show business logic** - we wan't to get rid of that (e.G active discount) but to keep a state in a fresh loaded screen (e.G if redirecting from a page and you want to keep a button menu opened)

>  Their only left benefit is to avoid having to repeat the code. I'd rather repeat the code for each button as there are not many anyways than having the different cases of all buttons end up in the BaseButtonContent and BaseButtonWrapper components.

Why would you repeat a code which look exactly the same for all buttons? If you ever need to change the icon tone for example, you just need to target one place instead of seven

> You can try this out for yourself by trying to write the PrimaryActionButton without any underlying abstraction. To me that seems a ton simpler.

Not sure if I completely understand it. My only concern is not being consistent on the props exposure. Is that ok? we just "render" whatever props are passed?

----

I know the way you suggested would make things easier to migrate, but it wouldn't really solve the problem with the button's API

dferber90:
> what do you mean the `button` component? is there a `button` component? 🤔

Yes, when you write `<button />` you are rendering the button component which React provides.

See its validation [here](https://user-images.githubusercontent.com/1765075/45487718-7d7e3780-b75f-11e8-8536-50272b20435f.png).

> Here we have the styles separated from the functionality. I don't think a button functionality would change, unless W3C comes with a new specificity

Isn't that supporting what I'm saying? If the buttons never change, why would we build a foundation for it? We probably don't need to make any adjustments anyways.

My point is not that the button itself changes, but that more buttons might be added which can't build on top of the foundation we provided, for example the [`PrimaryActionDropdown`](
https://user-images.githubusercontent.com/1765075/45487849-ce8e2b80-b75f-11e8-9118-e8499a7ad7ce.png) which also is a button for the most part. Would we then try to adapt the foundation (and deal with edge-cases int here) or recreate it for that specific case?

I guess my overall point is that it feels premature to extract commons when we are just introducing the buttons. When things have settled and we see that they all end up being the same, then we can extract. Doing it now seems premature. If you are certain that this makes sense already, then go for it.

> That's a bit different. react-select is a whole bundled third party library with particular functionality. It can suffer modifications at any given time. Buttons are native elements and self-limited.

Same as above. If buttons never change, why not just forward the props?

> The migration for ui-kit consumers. The problems I referred, are described in the button's refactoring issue

I still don't get which problem you are referring to.


> me: If it is compatible with the old one, why is there a need for migration in the first place? Is there a breaking change or not? This sentence is confusing me, could you explain?
> > @lufego As stated between parenthesis: minimum effort to migrate. That means that for the primary button, there's no breaking changes


I am also confused because you argued to be able to break the API, but now you are stating that there is no breaking change. But I see many breaking changes based on the prop-types alone:

```js
// PrimaryButton (old one)
PrimaryButton.propTypes = {
  buttonAttributes: PropTypes.object,
  iconLeft: PropTypes.node,
  isToggleButton: PropTypes.bool.isRequired,
  isToggled: PropTypes.func /* redacted */,
  isDisabled: PropTypes.bool,
  onClick: PropTypes.func.isRequired,
  size: PropTypes.oneOf(['big', 'small']),
  tone: PropTypes.oneOf(['urgent', 'primary']),
};
```

```js
// PrimaryActionButton (new one)
PrimaryActionButton.propTypes = {
  id: PropTypes.string,
  name: PropTypes.string,
  ref: PropTypes.string,
  ariaLabel: PropTypes.string.isRequired,
  onClick: PropTypes.func.isRequired,
  dataAttr: PropTypes.object,
  isDisabled: PropTypes.bool,
  isToggled: PropTypes.bool,
  icon: PropTypes.node,
  type: PropTypes.oneOf(['button', 'reset', 'submit']),
  size: PropTypes.oneOf(['small', 'big']),
};
```

Breaking changes just based on the prop-types:
- No more support for `data-` attributes, have to use `dataAttr: {}` instead
- No more support for `label`, have to use `ariaLabel` instead (but no support for other aria attributes)
- No more support for `buttonAttributes` (can pass some attributes directly instead, but only `onClick` and `type` - no `onMouseOver` etc)
- `iconLeft` renamed to `icon`
- `isToggleButton` does not exist anymore
- No more support for `tone` (use separate buttons instead)

New features:
- `id` prop
- `ariaLabel` prop
- `dataAttr` prop
- `type` prop

Maybe we have a different understanding of what a breaking change is. I am lost.

lufego:
to repeat the code or not to repeat? I'd like to "hear" others opinions. Any thoughts @commercetools/merchant-center?

tdeekens:
To my personal experience sharing the common things very low is very hard and often leads to more complexity than it actually tried to remove (in the code duplication). I've seen this with common styles (not through design tokens) or with very common components. Often I/we failed to either find a good common denominator or introduced one where there should not be one. To me this is essentially also reflected in type systems. Often you one tries to find these super generic super types which entail everything and have type discriminators as enums to be distinguished. While all discriminated types in the end don't share too much and have loads of disjoint properties. Like property `A` only holds for type `Foo` while `B` only holds for type `Bar` and `C` and `B` only hold for type `Baz` while `C` actually has to be in range of `n..m` to be a valid property.

This sounds like a strange detour but at least in my brain maps to these very generic components. Like the "lowest common button", the "lowest common input". I don't mind code repetition or the other way round: I came to learn that code repetition is actually not too bad. Code should easy to delete. That's far more important than being DRY. If folks are not able to determine if code is editable/deletable they will not touch it or introduce bugs. 

This is obviously all just a trade-off like everything in tech but I'd not fear repetition too much. I'd fear overly clever or premature abstraction hoping and anticipating that change comes to then have an impact on 3-4 things upstream. Eventually that change might never come downstream but upstream in the component tree which then makes everything further down even more entangled or complex.

montezume:
I'm not against breaking changes. Like @dferber90 mentioned, this is why we have versioning. As someone who just spent about 6 hours changing hundreds and hundreds of imports due to breaking changes (that I helped to introduce), I would like us to take a `respectful` approach with breaking changes to those who implement the UI-Kit. So would you mind, please, explaining, why these changes are required? It's going to take awhile for all of the teams in the MC to implement these changes to buttons. So we at least owe it to them to justify this decision.

dferber90:
Topics being discussed:
- introduce breaking changes or not (the name change and the disappearing “type” prop used for styling will happen anyways it seems)
- what should the new api look like 
- to use a common component for all buttons or not

Issue #2, opened 2018-08-31
Setup release script


emmenko:
We can let travis do the publishing to npm every time we push a new `tag`.

https://docs.travis-ci.com/user/deployment/npm/

Additionally, we can also decide to do "canary" releases (e.g. for `master` branch).

adnasa:
I'd like to pair with someone on this.

emmenko:
Is the task clear on what needs to be done?

emmenko:
We can pair together if no one jumps in

dferber90:
It would be nice if whoever tackles could add a `PUBLISHING.md` or a section in `README.md` about how we publish!

emmenko:
Yeah, I would put it in `CONTRUBUTING.md`.

---

The general idea for publishing is that we only do that whenever we **create a `git tag`**.

dferber90:
The job of the release should also be to create the netlify deployment. We currently deploy to `uikit.commercetools.com` straight from `master`, but this is confusing as it leads to `uikit.commercetools.com` being a different version than what is available under the `latest` dist-channel on npm.

emmenko:
You can't "disable" the `production` branch from being deployed.

<img width="717" alt="image" src="https://user-images.githubusercontent.com/1110551/45094357-57cbaf80-b11b-11e8-83ea-a7f06199d926.png">

What we can do (maybe, I will have to try that first) is to rename the `production` branch to a “random” one and deploy using the Netlify CLI

emmenko:
> We could also create a `development` branch and set it as the main one in GitHub

FYI: 👆 suggestion from @dferber90

adnasa:
have you guys considered [gitflow](https://datasift.github.io/gitflow/IntroducingGitFlow.html) for this?

adnasa:
I want to make clear that I won't have time for this in the coming weeks.
anyone is free to pick this up.

emmenko:
I'll take this over then since I'm off for the next couple of weeks and I'd like to get this done by tomorrow

PR #30, opened 2018-09-06
Add bundling

### LINKED ISSUE #7: Ship transpiled code (webpack or rollup)

dferber90:
So far we were shipping untranspiled/unbundled code. This meant that all consumers of UI-Kit had to have the right webpack loaders in place in order to consume UI-Kit, which is quite bad of course. If we were to update any of our loaders, all consumers would have had to update them as well.

This PR makes it so that the UI-Kit ships as a standalone library, which can be consumed by any project which has the `peerDependencies` like `react` and `react-dom` of UI-Kit installed as regular dependencies.

### Usage

The available commands were moved around a bit. The main updates are:
- `yarn start`: an alias for `yarn storybook:start`
- `yarn build`: New command to build the ui-kit for production to `dist`
- `yarn storybook:start`: Starts the storybook locally
- `yarn storybook:build`: Builds storybook for production

#### Main export
Consumers of UI-Kit can use it by importing components from the main export:

```js
import { PrimaryButton } from '@commercetools-frontend/ui-kit'
```

#### Additional exports

Additionally, we make it possible to consume some utility functions by reaching in.

```js
import { filterDataAttributes } from '@commercetools-frontend/ui-kit/utils'
```

These are provided by additional `entry` points in the webpack config. They are not related to the folder structure inside the `src` folder.

### Publishing

The build gets put into the `dist` folder. As we don't want to force consumers to use ugly imports containing `dist` like `@commercetools-frontend/ui-kit/dist/utils`, we are doing a workaround to prevent this.

We copy all files which we want to be included in the npm module to the `dist` folder. Currently, this is the `README.md` and the `package.json`. We then publish from the `dist` folder.

You can try this out by running `yarn build; cd dist; yarn pack` (`yarn pack` is like `yarn publish`, except that it doesn't publish the created module). Then inspect the contents of the generated `.tar.gz` file which represent the module that would be published on npm.

Copying of `package.json` to `dist` and publishing from there lets consumers avoid the `dist` in imports as the folder is no longer part of the module.

`package.json` getting copied to `dist` is also why the `main` field in `package.js` contains `index.js` instead of `dist/index.js`.

This preparation should make it fairly easy to set up the release script in #2.

### `thread-loader`

I did some experimentation and it turns out that the build time is 20s faster on the initial build, and 4s faster on following builds (with filled cache) when not using `thread-loader`, so I removed it.

### Todo
- adapt netlify to execute the right command

### Possible Future Improvements:
- avoid compiling SVG
- extract CSS
- verify treeshaking works

Closes #7

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -33,9 +33,6 @@ module.exports = (storybookBaseConfig, configType) => {
         // improves compile time on larger projects
         Object.assign(
           {},
-          {
-            loader: require.resolve('thread-loader'),
-          },
           // Keep workers alive only for development mode
           configType === 'PRODUCTION'
             ? {}
--- a/package.json
+++ b/package.json
@@ -7,15 +7,17 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "src/index.js",
+  "main": "index.js",
   "scripts": {
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "precommit": "lint-staged",
     "prepublish": "node ./scripts/generate-icon-exports.js",
-    "prebuild": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
-    "prestart": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
-    "build": "build-storybook -o .public",
-    "start": "start-storybook -p 9002",
+    "prestorybook:build": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
+    "prestorybook:start": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
+    "build": "NODE_ENV=production webpack --config webpack.config.js",
+    "start": "yarn storybook:start",
+    "storybook:start": "start-storybook -p 9002",
+    "storybook:build": "build-storybook -o .public",
     "i18n:build": "mc-scripts extract-intl --build-translations --output-path=$(pwd)/i18n 'src/components/**/!(*.spec).js'",
     "lint": "jest --projects jest.eslint.config.js jest.stylelint.config.js",
     "lint:js": "jest --config jest.eslint.config.js",
@@ -67,7 +69,9 @@
     "@svgr/webpack": "2.2.1",
     "babel-loader": "8.0.2",
     "browserslist": "4.0.2",
+    "clean-webpack-plugin": "0.1.19",
     "common-tags": "1.8.0",
+    "copy-webpack-plugin": "4.5.2",
     "css": "2.2.3",
     "css-loader": "1.0.0",
     "enzyme": "3.5.0",
@@ -88,6 +92,7 @@
     "moment": "2.22.2",
     "moment-timezone": "0.5.21",
     "omit-empty": "0.4.1",
+    "peer-deps-externals-webpack-plugin": "1.0.4",
     "postcss": "7.0.2",
     "postcss-color-mod-function": "2.4.3",
     "postcss-import": "12.0.0",
@@ -108,8 +113,8 @@
     "stylelint-config-standard": "17.0.0",
     "stylelint-order": "0.8.1",
     "svg-url-loader": "2.3.2",
-    "thread-loader": "1.2.0",
-    "url-loader": "1.1.1"
+    "url-loader": "1.1.1",
+    "webpack-cli": "3.1.0"
   },
   "peerDependencies": {
     "moment": ">2.2",
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as filterDataAttributes } from './filter-data-attributes';
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,233 @@
+const path = require('path');
+const CleanWebpackPlugin = require('clean-webpack-plugin');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
+const PeerDepsExternalsPlugin = require('peer-deps-externals-webpack-plugin');
+const postcssImport = require('postcss-import');
+const postcssPresetEnv = require('postcss-preset-env');
+const postcssReporter = require('postcss-reporter');
+const postcssCustomProperties = require('postcss-custom-properties');
+const postcssCustomMediaQueries = require('postcss-custom-media');
+const postcssPostcssColorModFunction = require('postcss-color-mod-function');
+
+const browserslist = {
+  development: ['chrome', 'firefox'].map(
+    browser => `last 2 ${browser} versions`
+  ),
+  production: ['>1%', 'not op_mini all', 'ie 11'],
+};
+
+const sourceFolders = [
+  path.resolve(__dirname),
+  path.resolve(__dirname, '../examples'),
+  path.resolve(__dirname, '../src'),
+];
+
+module.exports = {
+  entry: {
+    main: './src/index.js',
+    utils: './src/utils/index.js',
+  },
+  mode: 'production',
+  output: {
+    path: path.join(__dirname, 'dist'),
+    library: 'McUiKit',
+    libraryTarget: 'umd',
+    filename: chunkData =>
+      chunkData.chunk.name === 'main' ? 'index.js' : '[name]/index.js',
+  },
+  devtool: 'source-map',
+  plugins: [
+    new CleanWebpackPlugin(['dist']),
+    // We copy everything we want to be in the bundle to "dist",
+    // as we publish from "dist".
+    //
+    // We do this to enable consumers to import without having "lib"
+    // in the path, and to avoid having to bundle onto the main folder.
+    //
+    // This is also why the "main" field in package.json says
+    // "index.js" instead of "dist/index.js", as package.json gets copied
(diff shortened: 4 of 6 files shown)


.storybook/webpack.config.js
```diff
@@ -33,9 +33,6 @@ module.exports = (storybookBaseConfig, configType) => {
         // improves compile time on larger projects
         Object.assign(
           {},
-          {
-            loader: require.resolve('thread-loader'),
-          },
```
dferber90: Turns out builds were faster when not using threads.

src/utils/index.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as filterDataAttributes } from './filter-data-attributes';
```
dferber90: This is exported for now to show that we can import by reaching in, like

```js
import { filterDataAttributes } from '@commercetools-frontend/ui-kit/utils'
```

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+    main: './src/index.js',
+    utils: './src/utils/index.js',
+  },
+  mode: 'production',
+  output: {
+    path: path.join(__dirname, 'dist'),
+    library: 'McUiKit',
+    libraryTarget: 'umd',
+    filename: chunkData =>
+      chunkData.chunk.name === 'main' ? 'index.js' : '[name]/index.js',
```
dferber90: This enables spitting out the nested build like

```
- dist
  - index.js
  - utils
    - index.js
```

We can adapt this function in case we want to export more things.

Note that the `filterDataAttributes` function will now exist twice. Once in the `main` bundle and once at this point. This isn't optional yet, as it increases the bundle size unnecessarily.

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+                  // Keeps ID's of svgs so they can be targeted with CSS
+                  { cleanupIDs: false },
+                ],
+              },
+            },
+          },
+        ],
+      },
+      // For normal svg files (not icons) we should load the file normally
+      // and simply use it as a `<img src/>`.
```
montezume: do we have normal SVG files in ui-kit?

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+    // This is also why the "main" field in package.json says
+    // "index.js" instead of "dist/index.js", as package.json gets copied
+    // to "dist" as well.
+    new CopyWebpackPlugin([
+      { from: './package.json' },
+      { from: './README.md' },
+    ]),
+    // Adds peerDependencies as externals
+    // See: https://webpack.js.org/guides/author-libraries/#externalize-lodash
+    new PeerDepsExternalsPlugin(),
```
dferber90: This plugin is responsible for excluding the `peerDependencies` from the build.

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+            loader: require.resolve('svg-url-loader'),
+            options: { noquotes: true },
+          },
+        ],
+      },
+      // "url" loader works like "file" loader except that it embeds assets
+      // smaller than specified limit in bytes as data URLs to avoid requests.
+      // A missing `test` is equivalent to a match.
+      {
+        test: /\.png$/,
```
montezume: I would remove this IMO we shouldn't be bundling images with UI-kit.

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+    // to "dist" as well.
+    new CopyWebpackPlugin([
+      { from: './package.json' },
+      { from: './README.md' },
+    ]),
+    // Adds peerDependencies as externals
+    // See: https://webpack.js.org/guides/author-libraries/#externalize-lodash
+    new PeerDepsExternalsPlugin(),
+  ],
+  module: {
```
dferber90: The `module` part was copied from `.storybook/webpack.config.js`. I only removed the `.md` loaders as we don't need them here. Further optimizations may be possible.

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+module.exports = {
+  entry: {
+    main: './src/index.js',
+    utils: './src/utils/index.js',
+  },
+  mode: 'production',
+  output: {
+    path: path.join(__dirname, 'dist'),
+    library: 'McUiKit',
+    libraryTarget: 'umd',
```
dferber90: @montezume asked (moving the discussion here for threaded conversation):

> Was wondering if you considered also creating an ES bundle as well?

We could do that, but I am not sure how it would work with CSS modules. I guess we would still need to run webpack, but we just would transpile less?

.storybook/webpack.config.js
emmenko: Really?

package.json
```diff
@@ -7,15 +7,17 @@
...
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "precommit": "lint-staged",
     "prepublish": "node ./scripts/generate-icon-exports.js",
-    "prebuild": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
-    "prestart": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
-    "build": "build-storybook -o .public",
-    "start": "start-storybook -p 9002",
+    "prestorybook:build": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
+    "prestorybook:start": "node ./scripts/generate-base-colors.js && node ./scripts/generate-base-shadows.js && node ./scripts/generate-colors-for-story.js",
+    "build": "NODE_ENV=production webpack --config webpack.config.js",
```
emmenko: 💯

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+
+module.exports = {
+  entry: {
+    main: './src/index.js',
+    utils: './src/utils/index.js',
+  },
+  mode: 'production',
+  output: {
+    path: path.join(__dirname, 'dist'),
+    library: 'McUiKit',
```
emmenko: Nitpick: simply `UiKit`? Or `CTUiKit`? Not sure it's only specific for mc...

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+  path.resolve(__dirname),
+  path.resolve(__dirname, '../examples'),
+  path.resolve(__dirname, '../src'),
+];
+
+module.exports = {
+  entry: {
+    main: './src/index.js',
+    utils: './src/utils/index.js',
+  },
```
emmenko: Can we also add other entry points like `materials`, `hocs`, etc?

webpack.config.js
emmenko: Can you add this explanation to the code?

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+    new CleanWebpackPlugin(['dist']),
+    // We copy everything we want to be in the bundle to "dist",
+    // as we publish from "dist".
+    //
+    // We do this to enable consumers to import without having "lib"
+    // in the path, and to avoid having to bundle onto the main folder.
+    //
+    // This is also why the "main" field in package.json says
+    // "index.js" instead of "dist/index.js", as package.json gets copied
+    // to "dist" as well.
```
emmenko: Hmm...is this temporary?

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+    new CopyWebpackPlugin([
+      { from: './package.json' },
+      { from: './README.md' },
+    ]),
```
emmenko: PS: there are other files that should be copied as well (license, etc)

webpack.config.js
emmenko: Yes we do

webpack.config.js
emmenko: Good point

webpack.config.js
```diff
@@ -0,0 +1,233 @@
...
+      // A missing `test` is equivalent to a match.
+      {
+        test: /\.png$/,
+        use: [require.resolve('url-loader')],
+      },
+      // "postcss" loader applies autoprefixer to our CSS
+      // "css" loader resolves paths in CSS and adds assets as dependencies.
+      // "style" loader turns CSS into JS modules that inject <style> tags.
+      {
+        test: /\.mod\.css$/,
```
emmenko: How should we deal with css? Should we extract them into a separate file?

emmenko:
FYI: the netlify deployment is failing because the build command changed. Maybe we should use the `netlify.toml` config instead of having the values set in the UI

.storybook/webpack.config.js
dferber90: Yes, really. Maybe it depends on the machine, but for me it was faster. See the `thread-loader` section in the PR description.

webpack.config.js
dferber90: Yes, but I'm not too sure what they would look like yet. I'd like to do it in follow-ups when we actually need them.

webpack.config.js
dferber90: I have not found a way to publish a subfolder. If we remove this, then we end up with `dist` in our imports.

If you know of a better way to achieve this, then I'm all ears 🐰

webpack.config.js
dferber90: I get the license, but I'm not sure there is much 

[etc](https://user-images.githubusercontent.com/1765075/45157440-7dbb8780-b1e1-11e8-96d8-abebc166a249.png)? Which other files would you like to see included?  :)

dferber90:
Turns out Webpack can't produce ESM 🤦‍♂️

webpack.config.js
emmenko: - LICENSE
- CHANGELOG (#31)
- i18n

Issue #8, opened 2018-09-03
Setup changelog/release template


emmenko:
I'd like to have a good and readable structure with emoji and so on to describe the release changes.

dferber90:
@emmenko Can you provide some additional information of what we want to achieve here?

emmenko:
I think it's fine for now, closing...

PR #31, opened 2018-09-06
Setup changelog using conventional changelog

emmenko:
Ref #8 

This is a first attempt on having a process to work with releases and changelog.

The idea is that **before** tagging a release, you update the changelog so that the changes are included in the release.

Personally, I like to have a mix of "auto-generated" changelogs and manual content. Having only an auto-generated list of commits does not help much in giving the **context** to the release. Often times you also want to include some tips, examples, migration instructions and so which you can only do manually.

So my proposal is to use `convention-changelog` to **get the changelog** and then manually add some more info (if necessary).
The script `yarn changelog` will output to the terminal the missing changelog, including the unlreleased commits.
Then you would manually copy-paste the output into the `CHANGELOG.md` and the Release page.

WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,24 @@
+<a name="Unreleased"></a>
+# [Unreleased](https://github.com/commercetools/ui-kit/compare/v1.0.0-beta.30...df570e9) (2018-09-06)
+
+
+### Bug Fixes
+
+* avoid machine-based paths in generated files ([004cd38](https://github.com/commercetools/ui-kit/commit/004cd38))
+* **select-input:** default selectInput to null ([635d48e](https://github.com/commercetools/ui-kit/commit/635d48e))
+* **select-input:** fix typo, and update code documentation ([5892497](https://github.com/commercetools/ui-kit/commit/5892497))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+### Features
+
+* **typography/text:** to accept isItalic for text detail and body ([df570e9](https://github.com/commercetools/ui-kit/commit/df570e9))
+
+
+
+<a name="1.0.0-beta.30"></a>
+# [1.0.0-beta.30](https://github.com/commercetools/ui-kit/compare/3ba7996...v1.0.0-beta.30) (2018-09-03)
```
emmenko: This release contains a huge amout of commits, probably due to the migration. In any case I felt like unnecessary to have them listed here. I added a couple of comments below instead 👇

package.json
```diff
@@ -11,6 +11,7 @@
   "scripts": {
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "precommit": "lint-staged",
+    "changelog": "conventional-changelog -p angular -u",
```
montezume: angular 😆

emmenko: 🤷‍♂️

PR #40, opened 2018-09-12
Optimize bundle sizes 🚀

emmenko:
Currently our bundle sizes are `~2M`. 🙃

Those changes will reduce the bundle sizes by 75% 🎉

```
$ ll dist/ui-kit*
-rw-r--r--  1 emmenko  staff   469K 12 Sep 10:07 dist/ui-kit.cjs.js
-rw-r--r--  1 emmenko  staff   560K 12 Sep 10:07 dist/ui-kit.esm.js
```

Notable changes:
- define `dependencies` as `external` (so they won't be bundled)
- make `react-select` a `peerDependencies` (even though it was also defined as `external`, it was injecting "a lot of crap" into the bundle)
- `lodash.mapvalues` is supposed to be a `dependency`
- `uglify` the CJS bundle (there is also an additional command `build:dev` to generated the bundles without uglify, useful for inspecting the bundle)
- removed `uuid` and replaced it with `Math.random`


rollup.config.js
```diff
@@ -36,72 +37,85 @@ const postcssPlugins = [
...
-      {
-        file: `dist/${pkg.main}`,
-        format: 'cjs',
-      },
-    ],
+    output: {
+      file: `dist/${pkg.module}`,
+      format: 'esm',
+    },
+    external: Object.keys(pkg.dependencies).concat(pkg.peerDependencies),
```
tdeekens: I'd dump a comment and a ref to the article on top. Otherwise it feels like natural to remove this in 1-2 years :)

src/components/fields/text-field/text-field.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import has from 'lodash.has';
-import uuid from 'uuid';
```
tdeekens: Did we remove it from the pkg.json too? I can't see that change.

rollup.config.js
```diff
@@ -36,72 +37,85 @@ const postcssPlugins = [
...
+    plugins: [...basePlugins, execute('node scripts/bundle-copy.js')],
+  },
+  {
+    input: 'src/index.js',
+    output: {
+      file: `dist/${pkg.main}`,
```
tdeekens: This is clever.

rollup.config.js
```diff
@@ -36,72 +37,85 @@ const postcssPlugins = [
...
+    exclude: ['**/*.mod.css'],
+    include: ['**/*.css'],
+    plugins: postcssPlugins,
+  }),
+  replace({
+    'process.env.NODE_ENV': JSON.stringify('production'),
+  }),
+  resolve(),
+  svgrPlugin({
+    include: ['**/*.react.svg'],
```
tdeekens: Maybe a comment for others that this is our way of targeting svgs which mean to become react components?

rollup.config.js
```diff
@@ -36,72 +37,85 @@ const postcssPlugins = [
...
+    exclude: ['node_modules/**/*.css'],
+    modules: true,
+    importLoaders: 1,
+    localIdentName: '[name]__[local]___[hash:base64:5]',
+    plugins: postcssPlugins,
+  }),
+  // this is used for importing both vendor css (from node_modules) and for
+  // importing our global css which uses tokens that require postcss plugins
+  // (such as color-mod) to be compiled properly.
+  postcss({
```
tdeekens: Hah, the good times of inverse exclude/include rules :).

src/components/fields/text-field/text-field.js
emmenko: It wasn't there in the first place 🤷‍♂️

tdeekens: Then I'd recommend using `depcheck` on the package once. It's awesome and is how I managed to clean up the deps on the mc-fe without much manual work.

src/components/fields/text-field/text-field.js
```diff
@@ -12,6 +11,11 @@ import ErrorMessage from '../../messages/error-message';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import messages from './messages';
 
+const randomId = () =>
+  Math.random()
+    .toString(36)
+    .substring(2);
```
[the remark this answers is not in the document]
dferber90: I pushed a commit adding this now 0b9cf93.

rollup.config.js
```diff
@@ -36,72 +44,104 @@ const postcssPlugins = [
...
+      'node_modules/flatpickr/dist/l10n/de.js': ['German'],
+    },
+  }),
+  // To convert JSON files to ES6
+  json(),
+  // To automatically externalize `dependencies` and `peerDependencies`
+  // so that they do not end up in the bundle.
+  // See also https://medium.com/@kelin2025/so-you-wanna-use-es6-modules-714f48b3a953
+  peerDepsExternal({
+    includeDependencies: true,
```
emmenko: FYI: I found out that this does exactly the same as manually doing `external: Object.keys(pkg.dependencies).concat(Object.keys(pkg.peerDependencies))`

rollup.config.js
```diff
@@ -36,72 +44,104 @@ const postcssPlugins = [
   postcssReporter(),
 ];
 
-export default [
+// This list includes common plugins shared between each output format.
+// The list is sorted alphabetically.
```
montezume: why sorted alphabetically? It's important that we have the correct order.

rollup.config.js
```diff
@@ -36,72 +44,104 @@ const postcssPlugins = [
...
+      'node_modules/react-is/index.js': ['isValidElementType'],
+      'node_modules/flatpickr/dist/l10n/de.js': ['German'],
+    },
+  }),
+  // To convert JSON files to ES6
+  json(),
+  // To automatically externalize `dependencies` and `peerDependencies`
+  // so that they do not end up in the bundle.
+  // See also https://medium.com/@kelin2025/so-you-wanna-use-es6-modules-714f48b3a953
+  peerDepsExternal({
```
montezume: peer deps external should go first. [ref](https://www.npmjs.com/package/rollup-plugin-peer-deps-external)

montezume:
Can you at least explain why you reordered the plugins?

PR #42, opened 2018-09-12
Remove proxy exports 🙌

emmenko:
NOTE: the #40 should be merged first

Since we have almost migrated the imports in our codebase, we can remove the `proxy-exports` workarounds in the next release.

I also re-arranged a bit some other parts:
- `materials` is a top level folder and only contains `*.mod.css` files
- `images` is now a top level folder
- `constraints` and `spacings` have been moved to `src/components`
- messages are now scoped as a `import { i18n } from '@commercetools-frontend/ui-kit;` export


dferber90:
Should we also tacke the Icons in this one?

emmenko:
I'd prefer to do it separately. We can however provide the named export `Icons` already so that the migration is easier. Then remove the top level icon exports in a next major release.

WDYT?

emmenko:
image

.npmignore
```diff
@@ -0,0 +1 @@
+*.story.js
```
dferber90: Why do we need this? If both, `.npmignore` and `.gitignore` are present, then it will use only the `.npmignore`! So this change undoes the exclusion done by `.gitignore`

>  In simpler terms, npm prefers the .npmignore file if it is there, but will fall back to the .gitignore file. So if both are present, it will use the .npmignore file.

Source: https://stackoverflow.com/a/24942436

This also means that we would be exposing any sensitive files excluded from `.gitignore` as we'd be bundling them into the package as mentioned in [For the love of god, don’t use .npmignore](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d) whose title I found pretty funny.

I'd use the [`files`](https://docs.npmjs.com/files/package.json#files) field instead, which we seem to be using as well?

dferber90: Are we no longer publishing from `dist` by copying the `package.json` over or why did you feel that we need this? The stories shouldn't end up there in any case?

dferber90:
Yeah, let's do the `Icons` separately then.

.npmignore
emmenko: Ah good point. I just wanted to ignore stories to be published as well

emmenko: Yes, we don't need to "publish from dist" anymore. It's just that we have some `*.story.js` files in the `materials` folder and I thought it would be good to not publish those

dferber90: I’m not even sure it works at the moment because of https://github.com/npm/npm/pull/11669. I’m on the phone now so I could not verify.

dferber90: Thanks for explaining. I googled a bit and this seems like the best option after all.

emmenko:
To confirm:

image

src/index.js
```diff
@@ -23,6 +25,8 @@ export { default as FieldLabel } from './components/field-label';
 
 export { default as TextField } from './components/fields/text-field';
 
+export { Icons };
```
emmenko: This is the new `Icons` export

src/index.js
```diff
@@ -23,6 +25,8 @@ export { default as FieldLabel } from './components/field-label';
...
+// TODO: this type of export is deprecated and should be removed in the next major release
 export * from './components/icons';
```
emmenko: This is kept for backwards compatibility and will be removed in the next major release

src/index.js
```diff
@@ -68,3 +71,5 @@ export { default as Text } from './components/typography/text';
 
 export { default as withMouseDownState } from './hocs/with-mouse-down-state';
 export { default as withMouseOverState } from './hocs/with-mouse-over-state';
+
+export { i18n };
```
emmenko: This is also now its own named export

src/index.js
montezume: I wonder if it would be worth following up here with possibly code splitting this?

package.json
```diff
@@ -7,14 +7,24 @@
...
+  "main": "dist/ui-kit.cjs.js",
+  "module": "dist/ui-kit.esm.js",
+  "files": [
+    "dist",
+    "i18n",
+    "images",
+    "materials/*.mod.css",
+    "materials/**/*.mod.css",
+    "CHANGELOG.md",
+    "README.md",
```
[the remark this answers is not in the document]
dferber90: Pushed a commit now fixing this.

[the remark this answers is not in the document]
tdeekens: Shit. Good catch!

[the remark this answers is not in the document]
emmenko: Thanks

PR #44, opened 2018-09-13
chore: react-select not as a peer dependency

emmenko:
Making `react-select` a peer dependency requires that the consumer uses the latest version `2.0.0` of the library.

I think it's fine to make it a package dependency for now.


emmenko:
I also noticed that we were not importing components from the main export for `react-select`, so I changed that as well.

PR #48, opened 2018-09-14
chore: new approach for publishing docs

emmenko:
The current way of publishing the docs was not working well because of the way travis checks out the git repo.

image

This new approach is more "_brutal_" but it should do its job

image

### DIFF
--- a/scripts/publish-docs.js
+++ b/scripts/publish-docs.js
@@ -1,7 +1,3 @@
-if (!process.env.TRAVIS_TAG) {
-  throw new Error('This script can only be executed when a git tag is created');
-}
-
 const shelljs = require('shelljs');
 
 const docsOrigin = 'origin-docs';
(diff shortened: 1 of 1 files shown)


scripts/publish-docs.js
```diff
@@ -16,10 +12,16 @@ if (process.env.TRAVIS === 'true') {
   shelljs.exec(`git config --global user.email "${userEmail}"`);
   shelljs.exec(`git config --global user.name "${userName}"`);
   shelljs.exec(`git remote add ${docsOrigin} ${remoteUrl} > /dev/null 2>&1`);
+} else {
+  shelljs.exec(
+    `git remote add ${docsOrigin} git@github.com:commercetools/ui-kit.git > /dev/null 2>&1`
```
emmenko: This part is in case you run the script locally, in which case you don't need the `GH_TOKEN` and can simply use the ssh remote.

dferber90: This seems like a useful comment in the script :)

scripts/publish-docs.js
```diff
@@ -16,10 +12,17 @@ if (process.env.TRAVIS === 'true') {
...
-// This assumes that git is checked out at this branch tag.
-shelljs.exec(`git merge ${process.env.TRAVIS_TAG}`);
-shelljs.exec(`git push --force-with-lease ${docsOrigin} ${docsBranch}`);
+// Make sure there is no existing docs branch locally
+shelljs.exec(`git branch -D ${docsBranch} > /dev/null 2>&1`);
+// Checkout a new local branch for `latest/docs`
+shelljs.exec(`git checkout -b ${docsBranch}`);
+shelljs.exec(`git push --force ${docsOrigin} ${docsBranch}`);
+// Switch back to the previous branch (e.g. `master`)
+shelljs.exec(`git checkout -`);
```
dferber90: Would it make sense to hard-code master? Do we need the dynamic nature of this? Can it happen that we want to switch back to anything other than master?

emmenko: I'll remove this actually, I think it just causes confusion

Issue #49, opened 2018-09-14
Add silent reporters on CI


tdeekens:
**Description**

I think silent reporters during test runs on CI are nice to not bloat the console when e.g. just stylelint'ing.

Maybe this is something nice to pick up on the side whenever somebody has some spare time. We have silent reporters setup in internal repositories so those could serve as a blueprint.

PR #56, opened 2018-09-14
chore: CI silent jest reporting

montezume:
Let's make jest silent like we have it on mc.

I think this is all we need? Let me know if it's missing something.

Closes [49](https://github.com/commercetools/ui-kit/issues/49)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
   directories:
     - node_modules
 
-script: yarn run jest --projects jest.*.config.js --maxWorkers=4
+script: yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
 
 before_deploy: yarn build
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -11,7 +11,7 @@ cache:
   directories:
     - node_modules
 
-script: yarn run jest --projects jest.*.config.js --maxWorkers=4
+script: yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
```
tdeekens: Don't we have to add it as a dep? Otherwise it will, if at all, be only there through somebody else requiring it.

montezume: @tdeekens it's a dep of `commercetools-frontend/jest-preset-mc-app` though

tdeekens: I don't have too strong opinions then. I don't expect us to drop our preset but I often feel direct deps can't hurt.

dferber90: I'd vote for a direct dep as we use it directly.

PR #57, opened 2018-09-14
chore: update translations

montezume:
#### Summary

LocalizedMultiLineTextField is missing German translations. Here they have been added.


i18n/de.json
```diff
@@ -9,20 +9,21 @@
...
+  "UIKit.LocalizedMultilineTextInput.show": "Alle Sprachen anzeigen",
+  "UIKit.LocalizedTextInput.collapse": "Sprachen verbergen",
+  "UIKit.LocalizedTextInput.expand": "Alle Sprachen anzeigen",
+  "UIKit.LocalizedTextInput.missingRequiredField": "Pflichtfeld*, Geben Sie mindestens einen Wert an",
   "UIKit.MoneyInput.chevronLabel": "Dropdown öffnen/schliessen",
   "UIKit.MultilineTextInput.collapse": "Ausblenden",
   "UIKit.MultilineTextInput.expand": "Erweitern",
   "UIKit.SelectInput.noOptionsMessageWithInputValue": "",
   "UIKit.SelectInput.noOptionsMessageWithoutInputValue": "",
+  "UIKit.TextField.missingRequiredField": "",
```
dferber90: What about this one?

I guess nobody is using it yet, so it's not a problem. We might soon reuse our error message, so it won't be necessary to repeat the common ones for every field. I'm fine with leaving it empty.

montezume: I could just make it the same as `UIKit.LocalizedTextInput.missingRequiredField`? 🤷‍♂️

Issue #17, opened 2018-09-04
Background of Inputs uses non-existant token


dferber90:
Background of inputs should stay white when focused.

We basically need to re-apply the changes from: https://github.com/commercetools/merchant-center-frontend/pull/5346 which already had fixes for these.

PR #67, opened 2018-09-17
fix: use correct token for selected input background

montezume:
#### Summary

To close [17](https://github.com/commercetools/ui-kit/issues/17), I applied the changes that were leftover from the earlier closed PR.


dferber90:
I still had it on my list as well. Thanks for following it up!

PR #69, opened 2018-09-17
Feature/primary action button

lufego:
TO SEE THE FULL DESCRIPTION, GO TO https://github.com/commercetools/ui-kit/pull/28

----

In order to provide the raised issues within https://github.com/commercetools/ui-kit/pull/28, this PR:

- brings **no breaking changes** to the primary action button.
- adds design tokens only to the PrimaryActionButton
- adds deprecation prop checking

----

Changes:
- No more support for `buttonAttributes` (it was not being used anyways?)
- No more support for `tone` (use separate buttons instead - has a deprecation warning if passed)
- No more support for `isToggleButton` (is enough to pass `isToggled` - has a deprecation warning if passed)

New features:
- `id` prop (optional)
- `name` prop (optional)
- `ariaLabel` prop (optional for this first step of migration)
- `type` prop (with default of `button`)
- Supports `aria-*`attributes

```js
// PrimaryButton (old one)
PrimaryButton.propTypes = {
  buttonAttributes: PropTypes.object,
  iconLeft: PropTypes.node,
  isToggleButton: PropTypes.bool.isRequired,
  isToggled: PropTypes.func /* redacted */,
  isDisabled: PropTypes.bool,
  onClick: PropTypes.func.isRequired,
  size: PropTypes.oneOf(['big', 'small']),
  tone: PropTypes.oneOf(['urgent', 'primary']),
};
```

```js
// PrimaryActionButton (new one)
PrimaryActionButton.propTypes = {
  id: PropTypes.string,
  name: PropTypes.string,
  ariaLabel: PropTypes.string,
  onClick: PropTypes.func.isRequired,
  isDisabled: PropTypes.bool,
  isToggled: PropTypes.bool,
  iconLeft: PropTypes.node,
  type: PropTypes.oneOf(['button', 'reset', 'submit']),
  size: PropTypes.oneOf(['small', 'big']),
};
```


src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,88 @@
...
+  isToggled: PropTypes.bool,
+  isDisabled: PropTypes.bool,
+  onClick: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['big', 'small']),
+  type: PropTypes.oneOf(['button', 'reset', 'submit']),
+
+  // DEPRECATED
+  tone(props, propName, componentName) {
+    return new Error(
+      `Invalid prop \`${propName}\` supplied to \`${componentName}\`. \`${propName}\` is deprecated. Render a different button instead`
```
tdeekens: What different button?

lufego: just improved a bit. Does it look good now?

lufego:
> Just so I get this right: all code will have to be refactored to use the PrimaryActionButton over the current PrimaryButton?

That was an issue raised couple of times before. So this PR tries to introduce it with no breaking changes. That means, for the PrimaryActionButton, only the import/name needs to be changed. There are some cases where `PrimaryButton` (old component) is being rendered with tone="urgent". The `PrimaryActionButton` has a prop deprecation warning to cover this cases, in case this is not reviewed before migrating. Maybe we can add such message in the change log?

src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,88 @@
...
+};
+const getSizeClassName = size => styles[size];
+
+const PrimaryActionButton = props => {
+  const dataProps = {
+    'data-track-component': 'PrimaryActionButton',
+    ...filterDataAttributes(props),
+  };
+  const ariaProps = {
+    'aria-label': props.ariaLabel,
```
lufego: Maybe we can do `'aria-label': props.ariaLabel || props.label` just to make sure we won't loose any accessibility with the migration?

src/components/buttons/primary-action-button/README.md
```diff
@@ -0,0 +1,51 @@
+# Buttons: Primary Action Button
+
+## Usage
+
+```js
+import PrimaryActionButton from '@commercetools-frontend/ui-kit/buttons/primary-action-button';
```
montezume: should be updated to 

```
import { PrimaryActionButton } from '@commercetools-frontend/ui-kit';

```

lufego:
@tdeekens @montezume your comments have been addressed. This is ready for another round of reviews. Thanks guys!

montezume:
Can you add a named export to `src/index.js`? Since this is supposed to be replacing the current `PrimaryButton`, should that export be removed as well? Also, do we want to think about grouping the button exports together? WDYT?

tdeekens:
Awesome. Thanks a lot the the explanation @lufego! I think @dferber90 should head first with a round of review. Then @montezume or I pick up right after. Thanks!

src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,93 @@
...
```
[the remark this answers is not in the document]
lufego: The idea is to turn the ariaLabel as required, so we can assure accessibility. I just didn't put it as required now, to avoid breaking changes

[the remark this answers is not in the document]
lufego: Alright. Just changed to `aria-label`

lufego:
I've edited the aria-label a bit in order to keep it more similar to what it is at the moment (so we don't lose accessibility in this transition) and we can make it required later on - I've added a comment for this

src/components/buttons/primary-action-button/primary-action-button.js
```diff
@@ -0,0 +1,96 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import classnames from 'classnames';
+import filterDataAttributes from '../../../utils/filter-data-attributes';
+import filterAriaAttributes from '../../../utils/filter-aria-attributes';
```
dferber90: This import occurs twice.

lufego: Which one? one is for Aria and the other one for Data

dferber90: Ah my bad. All good then!

PR #70, opened 2018-09-18
feat(select-inputs): let user override customized components

pa3:
#### Summary

Let user override pre-styled components in `*SelectInput`-family of components.

#### Description

My use case for that is to show `SearchIcon` in `AsyncSelectInput` instead of default `DropdownIndicator`.

### DIFF
--- a/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
+++ b/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
@@ -94,8 +94,8 @@ export class AsyncCreatableSelectInput extends React.Component {
               'react-select-warning': this.props.hasWarning,
             })}
             components={{
-              ...this.props.components,
               ...customizedComponents,
+              ...this.props.components,
             }}
             classNamePrefix="react-select"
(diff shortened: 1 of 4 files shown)


src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -94,8 +94,8 @@ export class AsyncCreatableSelectInput extends React.Component {
               'react-select-warning': this.props.hasWarning,
             })}
             components={{
-              ...this.props.components,
               ...customizedComponents,
+              ...this.props.components,
```
tdeekens: Maybe I am just not getting the naming here. Customized are the ones you pass in? `props.components` are the ones the parent component passes as a default? Or the other way round? I guess that's why we swap the spread?

pa3: Other way around. `customizedComponents` are those which customizied (compared to the defaults used by `react-select`)  by ui-kit, and `props.components` are coming from the user of ui-kit `SelectInput`.

Issue #41, opened 2018-09-12
Add warning state to localized inputs e.g. <LocalizedTextInput>


tdeekens:
### Summary

All simple inputs support both error and warning states. In recent discussions with design we learned that we also need warning states on localised inputs. Just like their error messages.

### Description

Given an e.g. <LocalizedTextInput> is used I want to pass `warnings` just like `errors` which are rendered under the respective localisation in orange tone.

### Alterantives

Showing the warning above the input itself. This however does not work when e.g. wanting to recommend (not enforce through errors) that a single localisation is unique across a set of entities.

### Visuals

![image](https://user-images.githubusercontent.com/1877073/45417387-5a318a80-b682-11e8-85db-f06af789a1be.png)

dferber90:
@tdeekens Thanks for bringing it up! I was curious whether we have a use-case for the warnings and left them out on purpose initially. Now that we have one, I'm totally fine with adding it.

One open question: When an error and a warning is passed, would you expect both to be displayed? Or should the component only show the error in that case? The border colour of the input would currently be red (from the error), as the `TextInput` gives precedence to the `hasError` prop.

@lufego Do you wanna pick this one up?

tdeekens:
> Now that we have one, I'm totally fine with adding it.

It's rather a theoretical one. I also asked for more use cases and no others are known yet. 

I think it's decent here to keep track of this at least.

> When an error and a warning is passed, would you expect both to be displayed? 

Personally, I think errors should take precedence over warnings. Warnings can after all be skipped.

lufego:
Question: is the user able to save the custom view even tho he has this warning?

lufego:
Cool, so I'll work on that later today

tdeekens:
No need yet. You can if you want to but it's not an urgent request.

PR #77, opened 2018-09-18
feat(localized-multiline-text): adds warning prop

lufego:
Closes https://github.com/commercetools/ui-kit/issues/41


src/components/inputs/localized-multiline-text-input/translation-input.js
```diff
@@ -88,6 +89,11 @@ export default class TranslationInput extends React.Component {
     // amount of lines it should have when collapsed
     const contentExceedsShownRows =
       this.state.contentRowCount > TranslationInput.MIN_ROW_COUNT;
+    const getCommands = props => {
+      if (props.error) return <div>{this.props.error}</div>;
```
dferber90: The function is receiving a `props` argument, but it is using `this.props`. This seems wrong.

Generally, passing plain `props` around is usually a code smell. The function receives too much information. It also stops ESLint from finding issues with the props.

I would recommend to use a `do` expression below instead. See next comment.

src/components/inputs/localized-multiline-text-input/translation-input.js
```diff
@@ -137,13 +143,7 @@ export default class TranslationInput extends React.Component {
...
         </div>
         <div className={styles.commandsContainer}>
-          <div className={styles.commandsLeft}>
-            {this.props.error ? (
-              <div>{this.props.error}</div>
-            ) : (
-              this.props.languagesControl
-            )}
-          </div>
+          <div className={styles.commandsLeft}>{getCommands(this.props)}</div>
```
dferber90: I would turn this into:

```jsx
<div className={styles.commandsLeft}>
  {
    do {
      if (this.props.error) <div>{this.props.error}</div>
      if (this.props.warning) <div>{this.props.warning}</div>
      this.props.languagesControl;
    }
  }
</div>
```

src/utils/localized.js
```diff
@@ -58,6 +58,10 @@ export const createLocalizedDataAttributes = (props, language) =>
 export const getHasErrorOnRemainingLanguages = (errors, selectedLanguage) =>
   Boolean(errors) && without(Object.keys(errors), selectedLanguage).length > 0;
 
+export const getHasWarningOnRemainingLanguages = (warnings, selectedLanguage) =>
```
dferber90: As far as I can tell, these two functions are exactly the same.

We could also rename one to `getHasIssueOnRemainingLanguages` and then either pass in the warnings or the errors. Then we avoid the duplication.

lufego: IMO, having `getHasIssueOnRemainingLanguages` is too generic and harder to understand the abstraction. Having them separate helps to understand it better. In the end, I'd prefer to duplicate if its in favour of keeping the code understandable

src/components/inputs/localized-multiline-text-input/translation-input.js
lufego: Is it supposed to be a do while? apparently it doesn't work as you wrote

dferber90: Sorry, I had assumed you were already familiar with `do`-expressions. They were not enabled yet for this repository (#81 was not merged). I now created & merged #87 which enables them for this repository. You need to rebase on top of master.

In case they don't work, you can always use an iife instead.

src/utils/localized.js
dferber90: Would a code comment help instead?

lufego: I'd keep the code separated. If something changes in the warning/error validation, its better to have them as is.

dferber90: 👍

src/components/inputs/localized-multiline-text-input/translation-input.js
lufego: This is the behaviour I got when I use the suggested code:

image

Using an IIFE for now

dferber90: The provided code wasn't quite correct. I pushed 0810ae0610b6e513db2bffe18b6fba18b7b73883 which adds it.

It should have been:

```js
            {do {
              if (this.props.error) <div>{this.props.error}</div>;
              else if (this.props.warning) <div>{this.props.warning}</div>;
              else this.props.languagesControl;
            }}
```

src/components/messages/warning-message/README.md
```diff
@@ -0,0 +1,23 @@
...
+import WarningMessage from '@commercetools-frontend/ui-kit/messages/error-message';
+
+<WarningMessage>This is a duplicate of {`${duplicate}`}</WarningMessage>;
+```
+
+#### Properties
+
+| Props      | Type   | Required | Values | Default | Description                               |
+| ---------- | ------ | :------: | ------ | ------- | ----------------------------------------- |
+| `children` | `node` |    ✅    | -      | -       | Warning message, either as string or node |
```
dferber90: Sidenote: The bug in the library used by `markdown-loader` where empty cells were not possible was now fixed. We can use empty cells again. I didn't try it yet, but I read it in the release notes. So we don't need to fill empty cells with `-` anymore.

src/components/inputs/localized-multiline-text-input/translation-input.js
```diff
@@ -139,11 +139,11 @@ export default class TranslationInput extends React.Component {
...
-            {(() => {
-              if (this.props.error) return <div>{this.props.error}</div>;
-              if (this.props.warning) return <div>{this.props.warning}</div>;
-              return this.props.languagesControl;
-            })()}
+            {do {
+              if (this.props.error) <div>{this.props.error}</div>;
+              else if (this.props.warning) <div>{this.props.warning}</div>;
+              else this.props.languagesControl;
+            }}
```
tdeekens: At least now nobody needs to google: "IIFE in React" cause one fails every time to put the right amount of `(` and `{`s. :)

PR #79, opened 2018-09-19
add react testing library

dferber90:
- adds `react-testing-library`
- adapts one test as an example
  - this component is quite lowlevel, so the benefits of `react-testing-library` don't show as much


.eslintrc.yaml
```diff
@@ -1,2 +1,11 @@
...
 extends: "@commercetools-frontend/eslint-config-mc-app"
+settings:
+  import/ignore:
+    # included by default, but replaced if explicitly configured
+    - node_modules
+    # We use `export * from 'react-testing-library';` in `test-utils`, so that
+    # tests can do `import { render, fireEvent }` from '../test-utils'`.
+    # Because we export the whole namespace, the `import/ignore` rule can't find
+    # the named exports. So we disable the rule for imports of that file.
+    - test-utils.js
```
dferber90: I force-pushed these changes. We now have a `test-utils` which exports a configured `render`. This is not necessary for the tests shown in this PR, but it is necessary for later tests which use `IntlProvider`. We can extend the basic `render` fn as needed. For now, I found it important to get the setup right from the beginning, so that all future tests use the `render` fn from `test-utils`.

src/test-utils.js
```diff
@@ -0,0 +1,18 @@
...
+const customRender = (node, ...options) =>
+  render(
+    <IntlProvider locale="en" messages={messages}>
+      {node}
+    </IntlProvider>,
+    ...options
+  );
+
+// re-export everything
+export * from 'react-testing-library';
```
dferber90: All other exports are reexported so that we can do

```
import { render, fireEvent } from '../test-utils'
```

in tests. This means that we don't even need to import from `react-testing-library` in tests anymore. This is the recommended way of `react-testing-library`.

PR #92, opened 2018-09-25
Fix id generation in NumberField and TextField

dferber90:
The `id` generation when no `id` prop is provided was broken in `NumberField` and `TextField`. The tests were not detecting this either.

This PR fixes the generation, the tests and introduces `do` expressions to replace the iife.

### DIFF
--- a/src/components/fields/text-field/README.md
+++ b/src/components/fields/text-field/README.md
@@ -8,7 +8,7 @@ states.
 ## Usage
 
 ```js
-import TextField from '@commercetools-frontend/ui-kit/fields/text-field';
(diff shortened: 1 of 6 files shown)


src/components/fields/text-field/README.md
```diff
@@ -8,7 +8,7 @@ states.
 ## Usage
 
 ```js
-import TextField from '@commercetools-frontend/ui-kit/fields/text-field';
+import { TextField } from '@commercetools-frontend/ui-kit';
```
montezume: 👍

PR #93, opened 2018-09-25
Add MoneyField

dferber90:
Adds `MoneyField` just like we already have `TextField` and `NumberField`.

### DIFF
--- a/src/components/fields/money-field/money-field.js
+++ b/src/components/fields/money-field/money-field.js
@@ -94,17 +94,12 @@ class MoneyField extends React.Component {
   });
 
   render() {
-    // These could differ in theory, so we keep them separate for now.
-    // We could examine the error and the current value to determine
-    // whether the error happened on the currency or the amount input for some
-    // errors like "missing" or "isHighPrecision".
-    //
-    // But we need to ensure to only "hide" errors for amount/currency when
-    // we are certain about all errors.
-    const hasCurrencyError =
-      MoneyInput.isTouched(this.props.touched) && hasErrors(this.props.errors);
-    const hasAmountError =
-      MoneyInput.isTouched(this.props.touched) && hasErrors(this.props.errors);
+    // We could determine the errors of amount and currencyCode separately
+    // and forward hasCurrencyError / hasAmountError depending on the error.
+    // This would work for example for the known "missing" error.
+    // Doing so would lead to the correct part of the MoneyField being marked
+    // with a red border instead of the complete field.
+    // This is something we can do later / when somebody asks for it.
     const hasAnyErrors =
       MoneyInput.isTouched(this.props.touched) && hasErrors(this.props.errors);
     return (
@@ -129,9 +124,9 @@ class MoneyField extends React.Component {
             onBlur={this.props.onBlur}
             isDisabled={this.props.isDisabled}
             onChange={this.props.onChange}
-            hasCurrencyError={hasCurrencyError}
+            hasCurrencyError={hasAnyErrors}
             hasCurrencyWarning={this.props.hasCurrencyWarning}
-            hasAmountError={hasAmountError}
+            hasAmountError={hasAnyErrors}
             hasAmountWarning={this.props.hasAmountWarning}
             {...filterDataAttributes(this.props)}
           />


src/components/fields/money-field/money-field.js
```diff
@@ -0,0 +1,149 @@
...
+  static propTypes = {
+    // MoneyField
+    id: PropTypes.string,
+    horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
+    errors: PropTypes.shape({
+      missing: PropTypes.bool,
+    }),
+    renderError: PropTypes.func,
+    isRequired: PropTypes.bool,
+    touched: PropTypes.shape({
```
dferber90: Our other fields accept `isTouched` instead. I did not want to accept `isTouched` here as it would be harder on the consumer.

The touched object looks like this: `touched: { somePrice: { amount: true, currencyCode: true } }`.

We want consumers to be able to do:

```jsx
<MoneyField name="somePrice" touched={formik.touched.somePrice} />
```

If we wanted to accept an `isTouched` prop we'd have to do:

```jsx
<MoneyField
  name="somePrice"
  isTouched={MoneyField.isTouched(formik.touched.somePrice)}
/>
```

which seems awkward.

---

I am thinking of changing `isTouched` in `TextField` and `NumberField` to be just `touched` as well.

dferber90: On a general note, I like the is/has/can prefix convention for variables, but it's not as useful for component props. The components will emit prop-type warnings anyways in case a wrong thing gets passed in.

Further, it looks nicer when rendering components without providing the values:

```diff
- <Button isDisabled={true} isReadOnly={true} />`
+ <Button disabled readonly />
```

This would obviously be a huge change and one that I would not do without a providing a codemod. I'm just mentioning it for now. I don't plan to take any action aside from maybe opening an issue for discussion.

src/components/fields/money-field/money-field.js
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  render() {
+    // These could differ in theory, so we keep them separate for now.
+    // We could examine the error and the current value to determine
+    // whether the error happened on the currency or the amount input for some
+    // errors like "missing" or "isHighPrecision".
+    //
+    // But we need to ensure to only "hide" errors for amount/currency when
+    // we are certain about all errors.
```
dferber90: Maybe I should simplify the code instead and use a simple

```
const hasAnyErrors =
      MoneyInput.isTouched(this.props.touched) && hasErrors(this.props.errors);
```

The rest is unnecessary for now and only there in case we want to extend it - but we can add it back once we extend it.

I'll add a commit removing this unused stuff again.

src/components/fields/money-field/money-field.js
tdeekens: > I am thinking of changing isTouched in TextField and NumberField to be just touched as well.

I wouldn't have minded the isTouched to be honest. Just suggesting to do the backflip the other way round. We introduce isTouched for now, keeping all fields consistent, to then refactor it to `touched` with all fields in one go?

tdeekens: > I like the is/has/can prefix convention for variables, but it's not as useful for component props

I don't like it either but followed along for 16 months now. I think it should be a one of refactor for all components in UIKit?

dferber90: Yep, that's what I have in mind.

dferber90: Maybe there is a misunderstanding. If we use wanted to use `isTouched`, then every consumer would have had to do:

```js
<MoneyField
  isTouched={MoneyField.isTouched(formik.touched.price)}
/>
```

Whereas now they can do

```js
<MoneyField touched={formik.touched.price} />
```

Accepting the whole `touched` object further keeps the possibility to use the information in the `touched` state. For example, when we receive a `missing` error we can inspect the `value` and the `touched` state to display an error accordingly - something we could not do properly when we only accept a boolean `isTouched` as we'd lose the information of whether `amount` or `currencyCode` was touched.

dferber90: > > I am thinking of changing isTouched in TextField and NumberField to be just touched as well.
>
> I wouldn't have minded the isTouched to be honest. Just suggesting to do the backflip the other way round. We introduce isTouched for now, keeping all fields consistent, to then refactor it to `touched` with all fields in one go?

I don't fully get the question. `TextField` and `NumberField` have one input only. `MoneyField` uses two "inputs" (the currency dropdown and the number input).

`TextField` and `NumberField` have one touched state only, while `MoneyField` has two (one for currency, one for amount).

I was thinking whether it would make sense to keep the name consistent and always accept a `touched` prop on fields. It would accept a boolean for `TextField` and `NumberField`, but it would accept an object (`{ amount, currencyCode }`) for `MoneyField`.

For now it is like this:

```js
<TextField isTouched={true} />
<NumberField isTouched={true} />
<MoneyField touched={{ amount: true, currencyCode: false }} />
```

Eventually, it could become

```js
<TextField touched={true} />
<NumberField touched={true} />
<MoneyField touched={{ amount: true, currencyCode: false }} />
```

tdeekens: Okay okay. `touched` is fine.

dferber90: Cool :) I didn't get your question so I just tried to give the answer to every possible question :P

Using `touched` from the beginning avoids the refactoring on the consumer and allows us to do nicer UX inside `MoneyField` right away. There are only two fields with `isTouched` so far and both are not used yet, so refactoring the fields is the lesser evil :) 

I wrote the comment initially to find out whether we're okay with always using `touched` even if it contains a boolean, as this is kind of against our current prop naming conventions. Sorry that this wasn't more clear. If we're okay with it, I'd rename `isTouched` in `NumberField` and `TextField` now.

tdeekens: Alright, then. Fine here and lets refactor the other two then before used :)

src/components/fields/money-field/money-field.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  state = {
+    // We generate an id in case no id is provided by the parent to attach the
+    // label to the input component.
+    id: this.props.id,
+  };
+
+  static getDerivedStateFromProps = (props, state) => ({
+    id: do {
```
tdeekens: Ouh. Yeah!

PR #97, opened 2018-09-25
break: use touched instead of isTouched for all fields

dferber90:
This was previously called `isTouched`. We call it `touched` now to have a consistent prop name for all fields. Some fields might accept objects as `touched`  and then `isTouched`  would not make sense anymore.


dferber90:
As discussed in #93 

cc @tdeekens

tdeekens:
💯

PR #103, opened 2018-09-27
feat(money-field): add high-precision price badge

dferber90:
- Removes the ability to pass in a badge manually
- Adds the `showHighPrecisionBadge` prop which will show the "high precision price" badge in case a high-precision price was entered.

hppbadge

### DIFF
--- a/src/components/fields/money-field/money-field.form.story.js
+++ b/src/components/fields/money-field/money-field.form.story.js
@@ -19,28 +19,39 @@ storiesOf('Examples|Forms/Fields', module)
   .addDecorator(withReadme(Readme))
   .add('MoneyField', () => {
     const currencies = ['EUR', 'USD', 'AED', 'KWD'];
+    const horizontalConstraint = select(
+      'horizontalConstraint',
+      ['xs', 's', 'm', 'l', 'xl', 'scale'],
+      'm'
+    );
     return (
       <Section>
         <IntlProvider locale="en">
           <Formik
             initialValues={{
-              price: { currencyCode: currencies[0], amount: '' },
+              price: { ...MoneyField.parseMoneyValue() },
+              pricePerTon: { ...MoneyField.parseMoneyValue() },
             }}
             validate={values => {
-              const errors = { price: {} };
+              const errors = { price: {}, pricePerTon: {} };
               if (MoneyField.isEmpty(values.price)) errors.price.missing = true;
               else if (MoneyField.isHighPrecision(values.price))
                 errors.price.unsupportedHighPrecision = true;
+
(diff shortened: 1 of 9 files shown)


src/components/fields/money-field/money-field.form.story.js
```diff
@@ -66,6 +73,20 @@ storiesOf('Examples|Forms/Fields', module)
                     }
                   }}
                 />
+                <MoneyField
+                  title="Price per ton (High Precision)"
+                  description={'How much is a ton of fish?'}
```
tdeekens: I like this.

src/components/fields/money-field/money-field.js
```diff
@@ -111,7 +122,20 @@ class MoneyField extends React.Component {
             description={this.props.description}
             onInfoButtonClick={this.props.onInfoButtonClick}
             hintIcon={this.props.hintIcon}
-            badge={this.props.badge}
+            badge={
+              this.props.showHighPrecisionBadge &&
```
tdeekens: `shouldShowHighPrecisionBadge`. I know it gets long. Maybe these prefixes for React props are something we should really drop.

lufego:
Code looks good. I have a user related question to design team tho. What if the user doesn't know what a "high precision price" is about? maybe we can add an info button or a tooltip on hover explaining it?

dferber90:
> Code looks good. I have a user related question to design team tho. What if the user doesn't know what a "high precision price" is about? maybe we can add an info button or a tooltip on hover explaining it?

Nice catch. I spoke with Filip. The text in the designs is coupled tightly to the views. We will keep it as a pure indicator for now. We will open a modal eventually which will explain the meaning of a high-precision price for the specific content. To support this, the `MoneyField` will get an `onHighPrecisionBadgeClick`. Then the consumers can show the modal with any text they like.

src/components/fields/money-field/money-field.js
dferber90: Should I do it or not? I thought the `show`  prefix is as a good enough indicator. But it does sound like a function now that you mention it.

dferber90: Maybe just `hasHighPrecisionBadge`. The potentially confusing part is that the badge will only be shown when the value holds a high precision price. But the `show` prefix has the same problem.

dferber90:
Handed off to @filippobocik for UX review (can't assign as he's not part of the repo yet).

dferber90:
I'll merge now. We'll address UX feedback in a new PR. I want to do a refactoring now which would be inconvenient without having this component in.

PR #106, opened 2018-09-27
Add MultilineTextField

dferber90:
Adds the `MultilineTextField` component.


dferber90:
I'll merge now. We'll address UX feedback in a new PR. I want to do a refactoring now which would be inconvenient without having this component in.

PR #108, opened 2018-09-28
Fix lost updates of MoneyInput for Formik validation

dferber90:
Validation was still being called with invalid values in case the user enters a perfectly formatted amount before selecting a currency.

<a href="https://cl.ly/81ddf56d1348" target="_blank"></a>

This workaround can also be removed once https://github.com/jaredpalmer/formik/pull/939 is released.


tdeekens:
Should we have issues on the repo? Like Tech Debt ones or something? The code comments are awesome but I fear us forgetting about them...

PR #112, opened 2018-10-01
UI-KIT-37: Tags styling issue when dropdown isMulti

lufego:
#### Summary

When adding isMulti to the dropdowns, the component was getting a bigger height.

Now we have:

- Made the Tag component smaller (from 27px to 26x height) by introducing a token for that
- Removing the padding from the original library when the component is on `isMulti` mode

----

#### Description

**Before**
image

**After**
image

With multi **options** (after)
image


materials/tokens/sizes.mod.css
```diff
@@ -22,4 +24,11 @@
 
   /* Height */
   --token-size-height-input: 32px;
+
+  /* ==========================================================================
+   Tags
+   ========================================================================== */
+
+  /* Height */
+  --token-size-height-tags: 26px;
```
dferber90: We have `--token-size-height-input` with  `input` being singular.
We now add `--token-size-height-tags` with `tags` being plural.
We seem to use `--token-border-radius-tags` in plural as well though.

Can you elaborate why you went for plural for tags but singular for inputs?

PR #113, opened 2018-10-01
adds `CriticalActionButton`

lufego:
#### Summary

Adds `CriticalActionButton` 

#### Description

TBD

### DIFF
--- a/src/components/buttons/critical-action-button/README.md
+++ b/src/components/buttons/critical-action-button/README.md
@@ -15,12 +15,11 @@ aria label for accessibility reasons.
 
 ```js
 <CriticalActionButton
-  icon={<AddIcon />}
+  iconLeft={<AddIcon />}
   onClick={() => alert('Products deleted')}
-  {...{ 'aria-label': 'Deletes all products' }}
->
-  {'Delete all'}
-</CriticalActionButton>
+  aria-label="Deletes all products"
+  label="Delete all"
+/>
 ```
 
 ### Dos and don'ts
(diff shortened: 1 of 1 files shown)


src/components/buttons/critical-action-button/README.md
```diff
@@ -0,0 +1,49 @@
+# Buttons: Critical Action Button
+
+## Usage
+
+```js
+import { CriticalActionButton } from '@commercetools-frontend/ui-kit';
```
adnasa: Why do we have two different __usage__ sections?

lufego: This is a copy and past of other components. I see many of them contains two section actually. We can fix them in a bulk in separate PR. Thanks for pointing this out

adnasa: that would be great :)

src/components/buttons/critical-action-button/critical-action-button.js
```diff
@@ -0,0 +1,96 @@
...
+
+const CriticalActionButton = props => {
+  const dataProps = {
+    'data-track-component': 'CriticalActionButton',
+    ...filterDataAttributes(props),
+  };
+  const ariaProps = {
+    // TODO: Turn `aria-label` into its own prop and make it required
+    // in order to make the component more accessible. Currently, for backwards
+    // compatibility reasons, `aria-label` contains the `label` prop value.
```
adnasa: 👍 good comment

src/components/buttons/critical-action-button/critical-action-button.spec.js
```diff
@@ -0,0 +1,181 @@
...
+          });
+          wrapper = shallow(<CriticalButton {...props} />);
+        });
+
+        it('should overwrite the Icon theme to be white', () => {
+          expect(wrapper.find(AddBoldIcon)).toHaveProp('theme', 'white');
+        });
+      });
+    });
+    describe('setting the button as active', () => {
```
adnasa: when is toggled

src/components/buttons/critical-action-button/critical-action-button.spec.js
```diff
@@ -0,0 +1,181 @@
...
+      beforeEach(() => {
+        props = createProps({ isToggled: true });
+        wrapper = shallow(<CriticalButton {...props} />);
+      });
+
+      it('should render the button in the active state', () => {
+        expect(wrapper.find('div')).toContainClass('active');
+      });
+    });
+    describe('setting the button as not active', () => {
```
adnasa: when is not toggled

src/components/buttons/critical-action-button/critical-action-button.spec.js
```diff
@@ -0,0 +1,181 @@
...
+      beforeEach(() => {
+        props = createProps({ isToggled: false });
+        wrapper = shallow(<CriticalButton {...props} />);
+      });
+
+      it('should not render the button in the non active state', () => {
+        expect(wrapper.find('div')).not.toContainClass('active');
+      });
+    });
+    describe('setting the button as disabled', () => {
```
adnasa: when is disabled

src/components/buttons/critical-action-button/critical-action-button.js
```diff
@@ -0,0 +1,96 @@
...
+  isDisabled: PropTypes.bool,
+  onClick: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['big', 'small']),
+  type: PropTypes.oneOf(['button', 'reset', 'submit']),
+
+  // DEPRECATED
+  tone(props, propName, componentName, ...rest) {
+    if (props.tone) {
+      return new Error(
+        `Invalid prop \`${propName}\` supplied to \`${componentName}\`. \`${propName}\` is deprecated. Render a different button instead (e.G \`CriticalActionButton\` if tone is \`urgent\`)`
```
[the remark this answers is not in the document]
lufego: True. This needs to be removed

src/components/buttons/critical-action-button/critical-action-button.js
```diff
@@ -0,0 +1,88 @@
...
+const CriticalActionButton = props => {
+  const dataProps = {
+    'data-track-component': 'CriticalActionButton',
+    ...filterDataAttributes(props),
+  };
+  const ariaProps = {
+    // TODO: Turn `aria-label` into its own prop and make it required
+    // in order to make the component more accessible. Currently, for backwards
+    // compatibility reasons, `aria-label` contains the `label` prop value.
+    'aria-label': props.label,
```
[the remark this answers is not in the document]
lufego: Exactly. We want to keep the "accessibility" for this component. It means that by default (if the prop `props['aria-label']` is not provided) it will take `props.label` as default.

src/components/buttons/critical-action-button/critical-action-button.js
```diff
@@ -0,0 +1,88 @@
...
+  label: PropTypes.string.isRequired,
+  iconLeft: PropTypes.node,
+  isToggled: PropTypes.bool,
+  isDisabled: PropTypes.bool,
+  onClick: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['big', 'small']),
+  type: PropTypes.oneOf(['button', 'reset', 'submit']),
+
+  // DEPRECATED
+  isToggleButton(props, propName, componentName, ...rest) {
```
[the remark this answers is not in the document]
lufego: The idea is to avoid showing business logic through button states. We've introduced `ToggleBar` for that. We still have the `isToggled` to show application logic (if a `SecondaryActionButton` get clicked to display some filters list for example, and clicking it again will hide them)

src/components/buttons/critical-action-button/critical-action-button.mod.css
```diff
@@ -0,0 +1,74 @@
...
+  background: none;
+  color: inherit;
+  border: none;
+  padding: 0;
+  font: inherit;
+  cursor: pointer;
+  outline: inherit;
+}
+
+ /* States */
```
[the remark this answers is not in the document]
lufego: Because it lists `default`, `hover`, `active` and `disabled`

lufego:
@dferber90 thanks for your feedback. Your comments have been addressed. I noticed that most of your "catches" needs to also be updated on the `PrimaryActionButton`

src/components/buttons/critical-action-button/critical-action-button.js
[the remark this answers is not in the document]
dferber90: I don't get the vague difference of "business logic" and "application logic".  In the end it's still a button that either supports toggling or not, it seems. But maybe the difference is that now all buttons can potentially be toggled, but they will always look the same when they are not toggled (no matter whether they support toggling or not). I'm fine with this then 👍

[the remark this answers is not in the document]
lufego: > I don't get the vague difference of "business logic" and "application logic". In the end it's still a button that either supports toggling or not, it seems.

I'd say that **business logic** is all of the things that belongs to the client's business. Lets take an example of wether a discount is active or not. That is not up to the application's frontend, that belongs to our customer's store logic. 

The **application logic**, is more of how we take the decision to show things in the application (e.g UX decisions on the dos and dont's). A button is supposed to execute an action, and bringing the abstraction of showing a business logic together with it (remember the example of showing an active discount) just makes the component and the user experience more complex. That's why we segregated the logic for business logic for the `ToggleBar` and we keep the buttons to be as simple as to call actions or toggle internal feature states.

I hope the difference is more clear now

> But maybe the difference is that now all buttons can potentially be toggled, but they will always look the same when they are not toggled (no matter whether they support toggling or not)

Every default state of a button looks different in terms of color, because they mean different things (one is a primary action, the other is a critical action etc). The "toggle/active" state is a visual representation of a real-world state when a button is wether pressed or not. Some design approaches deal with it by for example adding inner shadows (to create some 3D spatial simulation just like in real life), others darkens a color etc. I think every button (if its a button) should support that.

PR #114, opened 2018-10-01
Add SelectField

dferber90:
Adds `SelectField`.

Needs #119 merged first.


src/components/fields/select-field/select-field.js
```diff
@@ -0,0 +1,176 @@
...
+              'description',
+              'onInfoButtonClick',
+              'hintIcon',
+              'badge',
+              'hasRequiredIndicator',
+              'htmlFor',
+              'errors',
+              'isVisible',
+              'renderError',
+              // data attributes are forwarded
```
tdeekens: `...omit(this.props, Object.keys(SelectField.propTypes)`?

montezume: in this case doing something like `const { hintIcon, badge, ..., ...rest } = this.props` would have been more readable, but against our standards I suppose.

dferber90: That's pretty clever! We need to omit additional props as well like `hasError` (as we do not want to support it).

To me the better approach (compared to what we're currently doing) is to be strict about the props supported in `SelectInput` and to remove the spread there. Then we can also be explicit about the forwarded props here.

I think accepting any props is a bad approach as the API is not clearly defined and prop-spreading is generally hard to deal with.

I'd do this in a follow-up PR then though (I added it to my list).

dferber90: > in this case doing something like `const { hintIcon, badge, ..., ...rest } = this.props` would have been more readable, but against our standards I suppose.

I like the explicitness of `omit` over "abusing" destructuring for omitting purposes, especially as we might need to declare some variables which we then won't use (like `hasError`).

dferber90: I now created #119 which explicitly declares all props accepted by `SelectInput`. I will stop the spread in `SelectField` (here) and forward the explicitly as well.

Then we don't need this ugly omission anymore.

src/components/fields/select-field/README.md
```diff
@@ -0,0 +1,83 @@
+# SelectField
+
+#### Description
+
+A controlled input component with validation states and a label getting a selection from the user.
```
lufego: maybe:
> A controlled input component with validation states and a label getting a `select value` from the user.

src/components/fields/select-field/README.md
```diff
@@ -0,0 +1,83 @@
...
+| `components`            | `objectOf(func)`   |    -     | -                                  | -       | Map of components to overwrite the default ones, see [what components you can override](https://react-select.com/components)                                                                                                                                                                |
+| `filterOption`          | `func`             |    -     | -                                  | -       | Custom method to filter whether an option should be displayed in the menu                                                                                                                                                                                                                   |
+| `inputId`               | `string`           |    -     | -                                  | -       | The id of the search input                                                                                                                                                                                                                                                                  |
```
lufego: what would this be used for? maybe we can add the use case here?

src/components/fields/select-field/README.md
```diff
@@ -0,0 +1,83 @@
...
+| `isClearable`           | `bool`             |    -     | -                                  | -       | Is the select value clearable                                                                                                                                                                                                                                                               |
+| `isDisabled`            | `bool`             |    -     | -                                  | -       | Is the select disabled                                                                                                                                                                                                                                                                      |
+| `isOptionDisabled`      | `func`             |    -     | -                                  | -       | Override the built-in logic to detect whether an option is disabled                                                                                                                                                                                                                         |
```
lufego: can we add this in the bottom with an explanation more in depth? (arguments etc)

src/components/fields/select-field/README.md
dferber90: Sorry, this is outdated by now. `react-select` offers `id` (for the container) and `inputId` (for the input element).

As we want to be able to use labels to focus the search input we need to provide a way to set the `id` of the `input`. To avoid confusion we offer `id` (the id of the input)  and `containerId` (the id of the container) as props. They are mapped to `id` and `inputId` internally.

```
<Select
  id={props.containerId}
  inputId={props.id}
```

dferber90: I force-pushed an update to the docs.

src/components/fields/select-field/README.md
dferber90: All functions are undocumented. The `react-select` docs offer enough reference IMHO. I don't want to repeat them multiple times (`SelectInput`, `AsyncSelectInput`, `CreatableSelectInput`, `AsyncCreatableSelectInput`, `SelectField`, `AsyncSelectField`, `CreatableSelectField`, `AsyncCreatableSelectField`).

src/components/fields/select-field/README.md
dferber90: This is based on `SelectInput` where it is already that way.

src/components/fields/select-field/README.md
lufego: So maybe we can point out this documentation in here?

src/components/fields/select-field/README.md
lufego: no worries. was just a minor suggestion. looks good enough

src/components/fields/select-field/select-field.js
```diff
@@ -0,0 +1,182 @@
...
+  static NoOptionsMessage = SelectInput.NoOptionsMessage;
+  static MultiValue = SelectInput.MultiValue;
+  static MultiValueContainer = SelectInput.MultiValueContainer;
+  static MultiValueLabel = SelectInput.MultiValueLabel;
+  static MultiValueRemove = SelectInput.MultiValueRemove;
+  static Option = SelectInput.Option;
+  static Placeholder = SelectInput.Placeholder;
+  static SelectContainer = SelectInput.SelectContainer;
+  static SingleValue = SelectInput.SingleValue;
+  static ValueContainer = SelectInput.ValueContainer;
```
adnasa: 🔥

src/components/fields/select-field/select-field.js
```diff
@@ -0,0 +1,182 @@
...
+    horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
+    errors: PropTypes.shape({
+      missing: PropTypes.bool,
+    }),
+    renderError: PropTypes.func,
+    isRequired: PropTypes.bool,
+    touched: (props, ...rest) =>
+      props.isMulti
+        ? PropTypes.arrayOf(PropTypes.bool, ...rest)(props, ...rest)
+        : PropTypes.bool(props, ...rest),
```
adnasa: 🔥 hot damn!

src/components/fields/select-field/select-field.js
```diff
@@ -0,0 +1,182 @@
...
+
+    // SelectInput
+    'aria-label': PropTypes.string,
+    'aria-labelledby': PropTypes.string,
+    isAutofocussed: PropTypes.bool,
```
adnasa: huh... [guess I was the confused one](http://www.future-perfect.co.uk/grammar-tip/is-it-focussed-or-focused/)

src/components/fields/select-field/select-field.js
```diff
@@ -0,0 +1,182 @@
...
+    // We generate an id in case no id is provided by the parent to attach the
+    // label to the input component.
+    id: this.props.id,
+  };
+
+  static getDerivedStateFromProps = (props, state) => ({
+    id: do {
+      if (props.id) props.id;
+      else if (state.id) state.id;
+      else sequentialId();
```
adnasa: 😍

src/components/fields/select-field/README.md
lufego: Alright. No worries. I just noticed this because I got a bit lost trying to understand it

PR #115, opened 2018-10-01
Add missing peer deps

dferber90:
This resolves some issues when running `yarn` by adding missing peer dependencies.


package.json
```diff
@@ -60,6 +60,8 @@
     "dom-helpers": "3.3.1",
     "downshift": "2.2.2",
     "flatpickr": "4.5.2",
+    "graphql": "^0.13.0",
```
montezume: Maybe I'm stupid but how come we need graphql?

dferber90: Yes, that's true. We are not using it ourselves. We are also not using `pegjs` for example. 

If these explicit dependencies are not listed, we end up with:

```
warning "@commercetools-frontend/jest-preset-mc-app > jest-transform-graphql@2.1.0" has unmet peer dependency "graphql-tag@^2.0.0".
warning "@commercetools-frontend/mc-scripts > graphql-tag@2.9.2" has unmet peer dependency "graphql@^0.9.0 || ^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0".
```

To clarify the reason for this: I was experimenting with visual regression testing on the side and ended up getting an old version of `babel` unless I listed them explicitly. I am adding them now because it seems like a good idea anyways.

montezume: Hm, how about making them devDependancies then? Since they are used by two deps that are devDeps?

dferber90: Great, good catch 👍

PR #119, opened 2018-10-01
SelectInput: explicitly declare props

dferber90:
### Summary

- tightens API of `SelectInput` (drops support for currently unused props)
- adapts story and docs to updated props
- ensures values are shown in correct order
- shortens the docs of statically exported components
- moves `SelectInput` out of beta status

### Details
So far we've been lax on the props we accept for `SelectInput`. We forwarded all props to `react-select` and just excluded the ones intended for `SelectInput` itself.

This has been good in the beginning when the usage of `SelectInput` was still unclear. Now that things have become more concrete, it's time to be stricter on the API.

This PR removes the generic pass-through of props and instead only forwards supported props. It also updates the documentation to mention all supported properties instead of referring to the `react-select` documentation.

In theory we should be able to replace `react-select` with any other implementation without having to touch the consumers. It is thus good to be explicit on the supported properties. This also makes it easier for consumers to see what we're supporting officially.

We can extend what `SelectInput` forwards to `react-select` in case any consumer ever needs to use a property supported by `react-select` but not currently forwarded by `SelectInput`.

### Fix
The selected values had been shown in the order they appear in the `options` instead of the order they were selected in. This is quite confusing when using the component and so I changed it. They are now shown in the order they were selected in.

### Note

This is a breaking change, but it should not affect the MC apps. I looked through the usages and the removed props were not used by any of our apps.


src/components/inputs/select-input/select-input.js
```diff
@@ -157,18 +231,29 @@ export class SelectInput extends React.Component {
...
-                      messages.noOptionsMessageWithInputValue,
-                      { inputValue }
-                    ))
+            onChange={selectedOptions =>
+              // selectedOptions is either an array, or a single option
+              // depending on whether we're in multi-mode or not (isMulti)
+              this.props.onChange({
+                target: {
+                  name: this.props.name,
+                  // eslint-disable-next-line no-nested-ternary
```
lufego: can this nested ternary be moved in a if statement or something else?

src/components/inputs/select-input/select-input.js
```diff
@@ -34,17 +33,76 @@ export class SelectInput extends React.Component {
...
+    // react-select base props
+    //
+    // Currently unsupported props are commented out. In case you need one of
+    // these props when using UI Kit, you can submit a PR and enable the
+    // prop. Don't forget to add it to the story, docs and other select input
+    // components as well!
+    //
+    // See https://react-select.com/props#select-props
+    'aria-label': PropTypes.string,
+    'aria-labelledby': PropTypes.string,
```
lufego: maybe just use `filterAriaAttributes` util?

src/components/inputs/select-input/select-input.js
```diff
@@ -99,11 +166,10 @@ export class SelectInput extends React.Component {
       <Constraints.Horizontal constraint={this.props.horizontalConstraint}>
         <div {...filterDataAttributes(this.props)}>
           <Select
-            {...omit(this.props, [
-              'horizontalConstraint',
-              'hasError',
-              'hasWarning',
-            ])}
+            aria-label={this.props['aria-label']}
+            aria-labelledby={this.props['aria-labelledby']}
```
lufego: maybe just use `filterAriaAttributes` util?

dferber90: See https://github.com/commercetools/ui-kit/pull/120#discussion_r221872796

src/components/inputs/select-input/select-input.js
dferber90: It can become 

```js
                  value: do {
                    if (!selectedOptions) selectedOptions;
                    else if (this.props.isMulti)
                      selectedOptions.map(option => option.value);
                    else selectedOptions.value;
                  },
```

but I'm not sure this reads more clearly. I'd keep it. With prettier guaranteeing a good formatting of the nested ternary it isn't confusing to read (to me).

adnasa: personally I would have preferred the `do` expression here, but this is fine for me

lufego: personally, I'd also have preferred the do expression

PR #120, opened 2018-10-01
AsyncSelectInput: explicitly declare props

dferber90:
- explicitly declares props of `AsyncSelectInput`
- moves `AsyncSelectInput` out of beta
- fixes an issue in `SelectInput` which was still using `autoFocus` instead of `isAutofocussed`

See #119 for more information. This is basically the same change, but for `AsyncSelectInput`.

### DIFF
--- a/src/components/inputs/async-select-input/async-select-input.js
+++ b/src/components/inputs/async-select-input/async-select-input.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
-import omit from 'lodash.omit';
 import {
   components as defaultComponents,
   Async as AsyncSelect,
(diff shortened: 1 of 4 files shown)


tdeekens:
👌🏿💯

src/components/inputs/async-select-input/async-select-input.js
```diff
@@ -60,29 +101,19 @@ export class AsyncSelectInput extends React.Component {
...
       <Constraints.Horizontal constraint={this.props.horizontalConstraint}>
         <div {...filterDataAttributes(this.props)}>
           <AsyncSelect
-            {...omit(this.props, [
-              'horizontalConstraint',
-              'hasError',
-              'hasWarning',
-            ])}
+            aria-label={this.props['aria-label']}
+            aria-labelledby={this.props['aria-labelledby']}
```
lufego: maybe just use `filterAriaAttributes` util?

dferber90: There is no point in forwarding additional `aria` attributes to `AsyncSelect` (the one from `react-select`) as it won't do anything with them anyways. This way we're explicit about the ones we support.

Using `filterAriaAttributes` would set the wrong expectation that we're supporting any kind of `aria` props.

PR #121, opened 2018-10-02
CreatableSelectInput: explicitly declare props

dferber90:
- Explicitly declares props of `CreatableSelectInput`
- Moves `CreatableSelectInput` out of beta
- Adapts story and README

### DIFF
--- a/src/components/inputs/creatable-select-input/creatable-select-input.js
+++ b/src/components/inputs/creatable-select-input/creatable-select-input.js
@@ -6,7 +6,6 @@ import {
   components as defaultComponents,
   Creatable as CreatableSelect,
 } from 'react-select';
-import omit from 'lodash.omit';
 import Constraints from '../../constraints';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import addStaticFields from '../../../utils/add-static-fields';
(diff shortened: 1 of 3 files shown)


src/components/inputs/creatable-select-input/creatable-select-input.js
```diff
@@ -83,11 +116,10 @@ export class CreatableSelectInput extends React.Component {
       <Constraints.Horizontal constraint={this.props.horizontalConstraint}>
         <div {...filterDataAttributes(this.props)}>
           <CreatableSelect
-            {...omit(this.props, [
-              'horizontalConstraint',
-              'hasError',
-              'hasWarning',
-            ])}
+            aria-label={this.props['aria-label']}
+            aria-labelledby={this.props['aria-labelledby']}
```
lufego: maybe just use `filterAriaAttributes` util?

src/components/inputs/creatable-select-input/creatable-select-input.js
```diff
@@ -135,25 +181,37 @@ export class CreatableSelectInput extends React.Component {
...
-                inputValue === ''
-                  ? this.props.intl.formatMessage(
-                      messages.noOptionsMessageWithoutInputValue
-                    )
-                  : this.props.intl.formatMessage(
-                      messages.noOptionsMessageWithInputValue,
-                      { inputValue }
-                    ))
-            }
+            isValidNewOption={this.props.isValidNewOption}
```
lufego: isNewOptionValid?

src/components/inputs/creatable-select-input/creatable-select-input.js
dferber90: See https://github.com/commercetools/ui-kit/pull/120#discussion_r221872796

src/components/inputs/creatable-select-input/creatable-select-input.js
dferber90: Those props are based on the `react-select` props and I would not want to deviate from them to avoid confusion.

See https://react-select.com/props#creatable-props

src/components/inputs/creatable-select-input/creatable-select-input.js
```diff
@@ -135,25 +181,37 @@ export class CreatableSelectInput extends React.Component {
                   }
                 : undefined
             }
+            onChange={(value, info) =>
+              // selectedOptions is either an array, or a single option
+              // depending on whether we're in multi-mode or not (isMulti)
```
adnasa: ... and? what do we do with this information when it comes to `onChange`?
Or was this comment meant to clarify that `value` is either __one of__ options or __the__ option provided depending on `isMulti`?

dferber90: The `value` argument used to be called `selectedOptions`. This was confusing as `selectedOption` wasn't always an array (as the name misleadingly indicated). That's why the comment was added.

The comment is only there for the next person looking at this code to understand that `value` can be an array or a single value depending on `isMulti`. We are just forwarding the value but I felt like it is good to know anyways.

PR #122, opened 2018-10-02
AsyncCreatableSelectInput: explicitly declare props

dferber90:
- move `AsyncCreatableSelectInput` out of beta
- explicitly declare props
- adapt docs and spec

Same as #119 but for `AsyncCreatableSelectInput`.

### DIFF
--- a/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
+++ b/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
-import omit from 'lodash.omit';
 import {
   components as defaultComponents,
   AsyncCreatable as AsyncCreatableSelect,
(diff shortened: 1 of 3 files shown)


src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -60,31 +99,27 @@ export class AsyncCreatableSelectInput extends React.Component {
...
       <Constraints.Horizontal constraint={this.props.horizontalConstraint}>
         <div {...filterDataAttributes(this.props)}>
           <AsyncCreatableSelect
-            {...omit(this.props, [
-              'horizontalConstraint',
-              'hasError',
-              'hasWarning',
-            ])}
+            aria-label={this.props['aria-label']}
+            aria-labelledby={this.props['aria-labelledby']}
```
lufego: maybe just use the `filterAriaAttributes` util?

dferber90: See https://github.com/commercetools/ui-kit/pull/120#discussion_r221872796

PR #124, opened 2018-10-02
adds `filterAriaAttributes` util

lufego:
This is already contained in the buttons branch. 

As we are covering accessibility in most of our components now, maybe its a good idea to get this merged before the branch `feature/new-buttons` gets merged in master


adnasa:
Doesn't this block #113 since we are using it there?

lufego:
@adnasa No, this is already implementend inside `feature/new-buttons` branch and `CriticalActionButton` is pointing to that branch

Issue #116, opened 2018-10-01
Use custom configuration (remove presets)


dferber90:
This application uses the presets we usually use to build the MC apps. However, the setup is a bit different and we thus incur dependencies of things we don't need (see #115) and are limited in what we can configure (it's not possible to turn module mocking of `.css` files off in Jest for example, as we can't overwrite the preset).

The presets are a good fit when working on MC apps, but not such a good fit for the UI Kit. Using the presets limits the flexibility (I ran into issues with the Jest setup), and incurs unnecessary dependencies and peer dependencies.

The exact dependency requirements listed in the presets can also prevent us from upgrading other dependencies (as it happened in #111).

Of course the presets have some advantages, but they seem to not outweigh the downsides we are facing. Again: Presets are great for MC apps, but problematic for UI Kit as it isn't one. I would like to have more flexibility in our setup of UI Kit, as all the consumers should care for is the final build anyways.

I propose to get rid of:
- `@commercetools-frontend/babel-preset-mc-app`
  - configures transforms we're not using (e.g. GraphQL)
- `@commercetools-frontend/jest-preset-mc-app`
  - has been getting in the way of the visual regression testing setup
  - it configures things we don't need (like GraphQL)
- `@commercetools-frontend/mc-scripts` (maybe, we only use `mc-scripts extract-intl`)
  - it forces a dependency on `graphql-tag` which forces `graphql`, both things we don't need

I would use a custom configuration instead.

I would keep `@commercetools-frontend/eslint-config-mc-app`.

PR #125, opened 2018-10-02
feat(build-process): move babel preset to internal script

montezume:
#### Summary

Second step in #116 , we kill the babel-preset-mc-app dependency by moving the get babel-preset functionality into ui-kit itself.

Next up we move the eslint and jest config internally, and we can kill off more dependencies.


dferber90:
Thanks, this is a step in the right direction. I'd want to remove the babel preset completely at some point and just use a regular configuration instead. We can do that in a second step.

dferber90:
I've pushed some commits on top of this branch. I will create a fresh PR for it, because the scope increased quite a bit.

Issue #136, opened 2018-10-04
Badges


dferber90:
It would be cool to add some badges (Prettier, npm install size, latest version on npm, ...).

dferber90:
Closed by #141.

We can add a size badge later.

The current size reported in `bundlephobia` seem off as they are super low. My best guess is that as we produce an `esm` build, it `bundlephobia` reports the size of `import {} from '@commercetools-frontend/ui-kit`, so basically the size of anything that can't be treeshaken (and without the static assets in the `images` folder). This would explain the super low size.

PR #141, opened 2018-10-07
Add badges

NoobTW:
#### Summary

Add badges for #136.

#### Description

Add badges like:

- npm latest version
- npm minified size
- GitHub issues
- GitHub forks
- GitHub stars
- License

Awesome badges are created from [shields.io](https://shields.io).


montezume:
Hey! Thanks for the contribution.

Would you mind

* Moving the badges underneath the title and subtitle. Like [here](https://github.com/styled-components/styled-components)
* Aligning 'center' the badges.

My one concern is using the bundle size badge, as we have some sort of issue now with bundlephobia showing the bundle size way smaller than in reality. What say you @commercetools/ui-kit ?

tdeekens:
We can leave them out for now. Let's wait for two days if @NoobTW gets to it otherwise we change the alignment ourselves.

NoobTW:
How about this?

I've also removed the size badge.

montezume:
@NoobTW do you want to update your branch and merge?

PR #151, opened 2018-10-08
Redo TimeInput

dferber90:
Recreates `TimeInput` without `flatpickr`.

## Advantages of new `TimeInput`

timeinput

UX:
- shows time depending on locale
- allows user to enter time using the keyboard
- it is possible to enter time in the 24h format or 12h format

Code improvements:
- does not use flatpicker anymore, so it’s A LOT cheaper to initialize (I dropped all optimizations we made, where initialized it on click only, and not at all on mobile)
- renamed `isInvalid` to `hasError` as we use that for all inputs
- added `isAutofocussed`
- removed unused `mode` prop
- removed unused `timeZone` prop
- `value` can now always be converted to 24h format using `TimeInput.to24h(value)`
- added `name` for easier use with Formik

## Problems of [old `TimeInput`](https://uikit.commercetools.com/?selectedKind=Inputs&selectedStory=TimeInput&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)

old-timeinput

It has the UX limitations that:
- ~~it is not possible to enter any value with the keyboard~~
- no matter the locale, the value is always entered and shown in 12hr format (AM/PM)
- The documentation was wrong
- ~~users can only adjust the time by a specificity of 5 minutes~~
- users can not specify time more precise than minutes even though the API supports precision up to milliseconds

Technically it had the problems that:
- It had support for props that had no effect on it or even broke it when unsupported props were passed
- it is really heavy to initialize so we had to defer initialization for Modified Products where many `TimeInput` fields can be present at the same time


dferber90:
Tests are failing on CI because of something related to `locale` apparently. The tests are succeeding on my machine. I'm investigating.

dferber90:
Fixed the failing test. Ready for review now :)

dferber90:
Ok, that wasn't very long until a new requirement surfaced.

The input should support a higher precision than currently shown. It should support up to `14:00:00.000`, as [the API supports the same precision](https://docs.commercetools.com/http-api-types#time).

I'll add this before putting the component into review again.

dferber90:
The `TimeInput` now supports high precision times as well. It will only show the high-precision time (eg `14:00:05.895`) when seconds or milliseconds are entered. Otherwise it will show the basic time (eg `14:00`).

dferber90:
high-precision-time

High precision timing for ya'll

montezume:
@dferber90 so much precision, so much wow 😮 👍

src/components/inputs/time-input/time-input-body.js
```diff
@@ -0,0 +1,104 @@
...
+
+const getIconTheme = (isDisabled, isMouseOver) => {
+  if (isDisabled) return 'grey';
+  if (isMouseOver) return 'orange';
+  return 'black';
+};
+
+export const ClearSection = props => (
+  <div
+    onClick={props.isDisabled ? undefined : props.onClear}
```
tdeekens: I've seen `null` being passed in that case. I don't mind but was wondering if there is a specific reason.

src/components/inputs/time-input/time-input-body.js
```diff
@@ -0,0 +1,104 @@
...
+    value: PropTypes.string,
+    isAutofocussed: PropTypes.bool,
+    isDisabled: PropTypes.bool,
+    isReadOnly: PropTypes.bool,
+    hasError: PropTypes.bool,
+    onClear: PropTypes.func,
+    onChange: PropTypes.func.isRequired,
+    onBlur: PropTypes.func.isRequired,
+    placeholder: PropTypes.string,
+    horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
```
tdeekens: Does it have a default constraint? Should it?

src/components/inputs/time-input/time-input.spec.js
```diff
@@ -1,33 +1,152 @@
...
+      expect(TimeInput.to24h('10:2:3.5')).toEqual('10:02:03.500');
+      expect(TimeInput.to24h('10:3.5')).toEqual('');
+      expect(TimeInput.to24h('1300:00.000')).toEqual('');
+      expect(TimeInput.to24h('1300')).toEqual('');
+      expect(TimeInput.to24h('300')).toEqual('');
+      expect(TimeInput.to24h('13:00.000')).toEqual('');
+      expect(TimeInput.to24h('15:09.300')).toEqual('');
+      expect(TimeInput.to24h('10:3.5')).toEqual('');
+    });
+  });
```
tdeekens: Nice unit tests! Just one thought: should we group them somehow if possible? 

```txt
describe: with padded time
describe: with unpadded time
describe: with second accuracy
describe: with high precision time
```

Just a thought as it may document the "spec" for the component a bit better.

src/components/inputs/time-input/time-input.spec.js
```diff
@@ -1,33 +1,152 @@
...
+        // See https://github.com/jsdom/jsdom/issues/1626#issuecomment-253795231
+        // and https://github.com/nodejs/node/issues/8818
+        value: expect.stringMatching(/^12:03( PM)?$/),
+      },
+    });
+  });
+
+  it('should have focus automatically when isAutofocussed is passed', () => {
+    const { container } = render(<TimeInput {...baseProps} isAutofocussed />);
+    expect(container.querySelector('input')).toHaveFocus();
```
tdeekens: ❤️

src/components/inputs/time-input/time-input-body.js
dferber90: The `time-input.js` itself already sets `scale` as the default `horizontalConstraint`, so the `time-input-body` does not need one 👍

src/components/inputs/time-input/time-input-body.js
dferber90: I'm not aware of a difference and don't have any arguments for either side. I usually use `undefined` though.

If the target component was a custom component (not a "native" one like a `div`), then passing `null` would lead to the default prop not overwriting it if I'm not mistaken. Hence, I usually use `undefined`. In this case it does not matter though.

src/components/inputs/time-input/time-input.js
```diff
@@ -1,354 +1,187 @@
...
+//   13:00
+//   3:00
+//   3 PM
+//   14:5 am
+//   13:00:00.000
+//   13:00:60
+//   13:00:59.908
+// Returns an array containing
+//   [hours, minutes, seconds, milliseconds]
+// or null
```
dferber90: This is wrong, it returns an object now.

src/components/inputs/time-input/time-input.js
```diff
@@ -1,354 +1,187 @@
...
+    hasMilliseconds: Number(milliseconds) !== 0,
+    amPm,
+  };
 };
 
-export class TimeInput extends React.PureComponent {
+// This component lets the user select a time.
+//
+// The `value` of this component and the value of the event `onChange` gets
+// called with is always a 24hr format time string, or an empty string.
```
dferber90: This is outdated, the value can be any string in any format. Only `TimeInput.to24h(value)` will return the 24h format or an empty string. 

The comment needs to be adjusted.

src/components/inputs/time-input/time-input.js
```diff
@@ -1,354 +1,187 @@
...
-    .utcOffset();
-
-  return timeZoneOffset - localTimeOffset;
-};
-
-const addFlatpickrOffset = (value, timeZone) =>
-  moment(value)
-    .add(getFlatpickrOffset(value, timeZone), 'minutes')
-    .toISOString();
+import { TimeInputBody } from './time-input-body';
```
dferber90: This is kinda weird. I want to change it to a default export instead.

dferber90:
This was approved by @coolmilo as well, so we're good to go!

src/components/inputs/time-input/time-input.js
```diff
@@ -1,354 +1,187 @@
...
+  if (amPm && Number(hours) > 12) return null;
+  if (!amPm && Number(hours) > 23) return null;
+  if (Number(seconds) > 59) return null;
+  if (Number(milliseconds) > 999) return null;
+
+  return {
+    hours: Number(hours) + (amPm === 'pm' ? 12 : 0),
+    minutes: Number(minutes),
+    seconds: Number(seconds),
+    milliseconds: Number(milliseconds),
```
dferber90: The parsing of milliseconds is actually wrong. When a user enters `00:00:00.75` we parse 75 milliseconds, but it should be 750. We need to use right-pad to fill up to 3 digits with 0s before parsing the number.

The bug isn't noticeable because we use rightPad before displaying the data in `format24hr`, but the `parseTime` function itself is broken.

emmenko:
@dferber90 anything else blocking this? Can you give us a status update on how to proceed here?

dferber90:
See the last comment:

> I'm thinking of waiting with the merge though, so that DateInput, DateTimeInput and TimeInput can be released at once.

I want to merge this along with #166 so that they can be released at the same time. That's also why the "Awaiting release" label is on this PR. Otherwise consumers would have to do the migration in two steps if we release in between because of a bugfix or something alike.


#166 is still missing the implementation of the new designs, but I've started that yesterday.

PR #152, opened 2018-10-09
refactor: replace invariant with tiny-invariant

dferber90:
#### Summary

This pull request replaces `invariant` with [`tiny-invariant`](https://github.com/alexreardon/tiny-invariant). A simpler, more lightweight solution. Not the biggest of our problems but it was on my list. Essentially `tiny-invariant` as the same API but does not allow placeholders in the string. Something we never used an if would use template literals for anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "dom-helpers": "3.3.1",
     "downshift": "2.2.3",
     "flatpickr": "4.5.2",
-    "invariant": "2.2.4",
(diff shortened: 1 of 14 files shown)


montezume:
Since we are doing

```
  peerDepsExternal({
    includeDependencies: true,
  }),
```

in the rollup config, the MC will need to add tiny-invariant as a dep, right?

PR #156, opened 2018-10-09
feat: bundle small deps, explicitly peer dep others.

montezume:
#### Summary

Now we are treating every single dependancy as a peer dep. This isn't sustainable, since users don't want to install a small lodash module everytime we add one. It only makes sense to treat large dependencies as peer dependencies. 

#### Approach

`flatpickr`, `react-select`, `react-textarea-autosize`, `react-virtualized` and `styled-components` are moved to peer deps, the rest stay as deps, and remove the rollup config that treats all deps as peer deps.

This increases the bundle size from 614kb -> 813kb, but I think the trade off is worth it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "common-tags": "1.8.0",
     "dom-helpers": "3.3.1",
     "downshift": "2.2.3",
-    "flatpickr": "4.5.2",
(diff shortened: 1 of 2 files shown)


montezume:
@dferber90 anything from your side?

PR #166, opened 2018-10-15
Redo `DateInput`, `DateTimeInput`, `DateRangeInput`

dferber90:
Drops usage of Flatpickr for `DateInput`, `DateTimeInput`, `DateRangeInput`.

Uses a solution built on top of `react-select` instead, inspired by https://react-select.com/advanced#experimental.


i18n/core.json
```diff
@@ -7,8 +7,35 @@
   "UIKit.CreatableSelectInput.createLabel": "Create \"{inputValue}\"",
   "UIKit.CreatableSelectInput.noOptionsMessageWithInputValue": "No options",
   "UIKit.CreatableSelectInput.noOptionsMessageWithoutInputValue": "No options",
-  "UIKit.DateInput.labelRange": "to",
-  "UIKit.DateTimeInput.labelRange": "to",
+  "UIKit.DateInput.lastDay": "[Yesterday], Do MMM YYYY",
```
montezume: this is where it might make sense to add a US and UK locale 🤔

dferber90: Might be an idea for another PR :)

dferber90:
The styling has been applied now. What's left is some refactoring to make the code nicer, but I'm pretty happy with how it turned out overall.

This PR adds a lot of lines, but don't forget that we drop a whole 12.7kb (minified & gzipped) dependency in exchange for that! And with some refactoring we can get the number of lines of this PR down as well. Right now some things are repeated for every component.

dferber90:
Improved styling is applied now, Spanish translations have been cleaned up and some common styling was moved out. Next up is adding some tests, and then this should be good to go - finally!

src/components/inputs/date-input/date-input.mod.css
```diff
@@ -1,3 +1,81 @@
...
+  border-color: var(--token-border-color-input-focus);
+}
+
+.daysHeaderItem {
+  color: var(--color-gray-60);
+  font-weight: bold;
+  cursor: default;
+  font-size: var(--token-font-size-small);
+  display: inline-block;
+  width: 12%;
```
tdeekens: How do these values come about? Could we rather `calc` them?

src/components/inputs/date-time-input/date-time-input.spec.js
```diff
@@ -1,33 +1,263 @@
...
+  });
+
+  it('should have an HTML name', () => {
+    const { container } = render(<DateTimeInput {...baseProps} name="foo" />);
+    expect(container.querySelector('[name="foo"]')).toBeTruthy();
   });
-  it('should render', () => {
-    expect(shallow(<DateTimeInput {...createTestProps()} />)).toMatchSnapshot();
+
+  it('should allow changing the value by typing in english', () => {
```
tdeekens: `typing in english`?

montezume: isn't the idea that you type in German format and you get US format back? Maybe the test could be renamed to reflect this? BTW there's nothing English about US date formatting 😠

src/components/inputs/date-input/date-input.mod.css
dferber90: This is for styling the inputs. The calendar dropdown has a flexible width which depends on the `horizontalConstraint`. The days always fill the whole calendar.

image

image

The percentages are used to format the days in the calendar. The specific values ensure that the days are always centered and fill the whole screen.

The reason for these values is that: there are 7 days in a week. 7*12% = 84%. There are 6 spaces between the 7 days. Each day has 1% margin on the left and on the right: 6 * 2 * 1% = 12%. This totals 96%. Then we add 2% spacing to the left and 2% spacing to the right of the calendar and we end up with a calendar having 100% width.

There is not really a way to `calc` them as so many rely on each other and it only adds up to 100 for certain combinations.

src/components/inputs/date-time-input/date-time-input.spec.js
dferber90: Yeah, true. I'll change that 👍 

"typing in english" was meant to indicate "typing a date in US format"

src/components/inputs/date-input/date-input.js
```diff
@@ -1,353 +1,455 @@
...
+  return (
+    <CalendarConnector.Consumer>
+      {({ month, setMonth, locale }) => {
+        // const days = ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
+        // This code is modifying `days`, so we need to create a fresh array
+        // to avoid manipulating the moment data
+        const days = [...moment.localeData(locale).weekdaysMin()];
+        const firstDayOfWeek = moment.localeData(locale).firstDayOfWeek();
+        // Rearrange so that week starts at Sunday/Monday depending on locale
+        Array.from({ length: firstDayOfWeek }).forEach(() => {
```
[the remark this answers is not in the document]
dferber90: Oh, nice 👍 I'm going to use that!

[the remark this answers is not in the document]
dferber90: Done in bb789ae8b08a35501654c4335045edfe76d9fbda 👍  I also extracted the generation of weekdays into its own utils function.

src/components/inputs/date-time-input/date-time-input.js
```diff
@@ -1,349 +1,837 @@
...
+      <CalendarConnector.Consumer>
+        {({ selectRef, month, setMonth }) => (
+          <SelectComponents.SelectContainer
+            {...this.props}
+            innerProps={{
+              ...this.props.innerProps,
+              // Let react-select handle the keyDown event unless the menu
+              // is already open. When the menu is open, we want enhance
+              // the UX by customizing the keyboard usage.
+              onKeyDown: event => {
```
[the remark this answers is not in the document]
dferber90: In theory it could be reused (maybe). There are slight nuances in keyboard usage between the different components, so the handlers vary a little bit. It could be that it is the same handler at the moment, but the they could vary if we add more functionality - they are not guaranteed to always stay the same. So I would rather duplicate them.

Another thing which would bother me about extracting the function is that it is using the context. It would feel a bit weird to extract a non-pure function which accesses the context.

src/components/inputs/date-input/date-input.js
```diff
@@ -1,353 +1,455 @@
...
-      return moment(value)
-        .subtract(getFlatpickrOffset(value, timeZone), 'minutes')
-        .toISOString();
-    }
-    case 'date':
-      return moment(value).format('YYYY-MM-DD');
-    default:
-      return value;
-  }
+const setFocus = (selectRef, option) => {
```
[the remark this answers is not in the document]
dferber90: Done in 0b727d829434c4a0e242e47ded9905f859c1e720 👍

dferber90:
@pa3 @montezume I made the changes you requested. Anything else?

src/index.bundlespec.js
```diff
@@ -9,11 +9,15 @@ describe('exports', () => {
...
   });
 
-  it('should have the translations', () => {
-    // This test ensures some translations exist, assuming the others would
-    // work as well then
-    expect(i18n.en['UIKit.DateInput.labelRange']).toEqual('to');
-    expect(i18n.de['UIKit.DateInput.labelRange']).toEqual('bis');
-    expect(i18n.es['UIKit.DateInput.labelRange']).toEqual('hasta');
+  it('should have translations', () => {
+    expect(Object.keys(i18n.en).length > 0).toBe(true);
```
montezume: hm, does this need to be part of this PR?

dferber90: The reason this is here is because until now the tests were using `UIKit.DateInput.labelRange` to check whether the translations in the final bundle work. But in this PR we are dropping the `UIKit.DateInput.labelRange` message completely, so this test broke. To me it makes sense to change it here. Could be its own PR as well though, but 🤷‍♂️

dferber90:
I looked into downshift and rebuilding the functionality there today, because I feel quite bad about the code here. It looks very promising and I came very far with the PoC. It might happen that this PR is merged and replaced immediately - or not merged at all.

dferber90:
The `downshfit` exploration is going perfectly smooth so far and there is much less code involved, and it's way less hacky. I'll close this for now and hope the switch to `downshift` stays smooth.

Thanks everyone for helping me get through this mess 👍

PR #167, opened 2018-10-16
docs: add global locale knob

dferber90:
So far we've been redefining `IntlProvider` for almost every story. Sometimes we added a knob to be able to change the locale. This is repetitive, cumbersome and in some cases unnecessary as we already have global `intl` decorator which wraps all stories in `IntlProvider` anyways.

I now refactored our global `intl` decorator to include a knob for selecting a locale. This means all stories now have a "global locale" selector and they don't need to define `IntlProvider` anymore.

Check out `.storybook/decorators/intl/intl.js` for the changed setup. The other changes are removing the unnecessary `IntlProvider` from all stories.

image

<sup>Intermediate loss of functionality: One thing to note is that `DateInput`, `DateTimeInput` and `DateRangeInput` were accessing the selected locale directly and adding it as a key to the components in order to rerender them in case the locale changes. This was done because the date-inputs have the flaw that they do not update when the locale changes. This is bad and no longer working. We would instead need to access apply `injectIntl` to the story and then pass `this.props.intl.locale` as the key. I didn't do that since our new date inputs will work with a changing locale ouf the box. The workaround in the meantime is to select a locale and then to reload the page, but I doubt anybody would need about this soon.</sup>

### DIFF
--- a/src/components/fields/async-creatable-select-field/async-creatable-select-field.form.story.js
+++ b/src/components/fields/async-creatable-select-field/async-creatable-select-field.form.story.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { Formik } from 'formik';
-import { IntlProvider } from 'react-intl';
 import { storiesOf } from '@storybook/react';
 import omitEmpty from 'omit-empty';
 import { action } from '@storybook/addon-actions';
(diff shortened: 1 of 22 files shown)


src/components/fields/async-creatable-select-field/async-creatable-select-field.form.story.js
```diff
@@ -82,71 +81,69 @@ storiesOf('Examples|Forms/Fields', module)
...
-                    { value: 'whales', label: 'Whales' },
-                  ]}
-                  loadOptions={searchText =>
-                    delay(delayTimeMs).then(() =>
-                      options.filter(option =>
-                        option.label
-                          .toLowerCase()
-                          .startsWith(searchText.toLowerCase())
-                      )
+        <Formik
```
dferber90: All that happened here and from here on downwards is that I removed the wrapping `<IntlProvider>` component. I didn't touch anything else.

src/components/inputs/date-time-input/date-time-input.story.js
```diff
@@ -13,41 +12,38 @@ storiesOf('Inputs', module)
   .addDecorator(withKnobs)
   .addDecorator(withReadme(Readme))
   .add('DateTimeInput', () => {
-    const locale = select('locale', ['de', 'en'], 'en');
```
dferber90: This is one of the work-around cases where `locale` was passed as a key to force the component to rerender. The new input components will no longer need this workaround, so I removed it.

This means they date inputs will ignore the changing locale in stories (as they would in features) until the new date inputs replace them.

dferber90:
Merged without reviews as @lufego is sick today and this is a small internal change only.

PR #168, opened 2018-10-16
Use custom configuration (remove presets) 

### LINKED ISSUE #116: Use custom configuration (remove presets)

dferber90:
This PR drops `@commercetools-frontend/jest-preset-mc-app`, `@commercetools-frontend/eslint-config-mc-app` and `@commercetools-frontend/babel-preset-mc-app` in favor of custom dependencies as discussed in #116.

On top of the reasons stated in #116, this further enables us to experiment with Visual Regression Testing more easily, as we now have full control over the setup.

This PR also upgrades `eslint`, which has so far been blocked by an issue in `eslint-config-mc-app` (see https://github.com/commercetools/merchant-center-application-kit/issues/14).

Closes #116


jest.test.config.js
```diff
@@ -1,5 +1,42 @@
...
+  rootDir: rootPath,
+  setupFiles: [
+    'raf/polyfill',
+    '<rootDir>/test/setup-tests.js',
+    'jest-localstorage-mock',
+  ],
   setupTestFrameworkScriptFile: '<rootDir>/scripts/setup-test-framework.js',
+  snapshotSerializers: ['enzyme-to-json/serializer'],
+  testEnvironment: 'jsdom',
+  testURL: 'https://mc.commercetools.com/',
```
montezume: do we need this?

jest.test.config.js
```diff
@@ -1,5 +1,42 @@
...
+  testPathIgnorePatterns: ['node_modules', 'cypress'],
```
montezume: do we need cypress?

jest.test.config.js
```diff
@@ -1,5 +1,42 @@
...
+  transform: {
+    '^.+\\.js$': '<rootDir>/test/transform-babel-jest.js',
+  },
+  transformIgnorePatterns: [
+    // This option tells Jest to ignore specific folders from being transpiled
+    // (e.g. with babel).
+    // However we need to instruct jest to actually transpile some packages.
+    // NOTE: this might not be necessary anymore once we ship packages already
+    // transpiled.
+    'node_modules/(?!(@commercetools-frontend)/)',
```
montezume: not sure we need this?

test/setup-tests.js
```diff
@@ -0,0 +1,66 @@
...
+
+// eslint-disable-next-line no-console
+const logError = console.error;
+global.console.error = (...messages) => {
+  logOrThrow(logError, 'error', messages);
+};
+
+// Avoid unhandled promise rejections from going unnoticed
+// https://github.com/facebook/jest/issues/3251#issuecomment-299183885
+// In Node v7 unhandled promise rejections will terminate the process
```
montezume: I'm not sure we need this for UI-kit

jest.test.config.js
dferber90: We don't need this exact value but it's good to define a value for it. It sets the value you get in JSDOM when reading `window.location.href` for example.

jest.test.config.js
dferber90: There is no cypress in ui-kit. I'll remove it 👍

test/setup-tests.js
dferber90: It's debatable. I'd like to ensure our tests fail in case of an unhandled rejection. I'm not sure if we need this setup for it or if there is another way.

If we decide to take this approach, we definitely need a script on CI which checks for the existence of `fail-tests-because-there-was-an-unhandled-rejection.lock` and fails CI in case it exists. Good catch 👍

jest.test.config.js
dferber90: We are not using any `@commercetools-frontend` packages anymore, so we can remove this 👍

PR #171, opened 2018-10-17
MoneyInput: align dropdown caret & support isTouched(x, {all: true})

dferber90:
Fixes the alignment of the caret in `MoneyInput` when no currency is selected.

**Before**
bildschirmfoto 2018-10-17 um 17 48 10

**After**
bildschirmfoto 2018-10-17 um 17 47 58

It is weird that only the caret can be clicked, and not the whole element. But I think this is for another time.

---

This PR also introduces `MoneyInput.isTouched(x, { all: true })` which only returns `true` in case all fields have been touched. Usually it returns `true` if at least one field has been touched.

Right now it is possible to make a difference between the left (currency) and the right (amount) when displaying errors. It is possible to have the red border on the left or right only. This introduces a lot of overhead for users of `MoneyInput` for a questionable gain in UX. Maybe we can get rid of this separation and just always make the input red? The displayed error message should clarify the error good enough for the user to figure out what went wrong in any case.

cc @filippobocik @Shecki ?

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -218,7 +218,11 @@ export default class MoneyInput extends React.Component {
     return moneyValue && moneyValue.type === 'highPrecision';
   };
 
-  static isTouched = touched => touched && Object.values(touched).some(Boolean);
+  static isTouched = (touched, { all = false } = {}) =>
+    touched &&
+    (all
+      ? Object.values(touched).every(Boolean)
+      : Object.values(touched).some(Boolean));
 
   static propTypes = {
     id: PropTypes.string,
--- a/src/components/inputs/money-input/README.md
+++ b/src/components/inputs/money-input/README.md
@@ -209,15 +209,17 @@ return (
           onBlur={() => setFieldTouched('somePrice')}
           isDisabled={isSubmitting}
           onChange={value => setFieldValue('somePrice', value)}
-          hasAmountError={touched.somePrice && Boolean(errors.somePrice)}
+          hasAmountError={
+            MoneyInput.isTouched(touched.somePrice) && Boolean(errors.somePrice)
+          }
           horizontalConstraint="l"
         />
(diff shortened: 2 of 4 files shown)


src/components/inputs/money-input/money-input.js
```diff
@@ -218,7 +218,11 @@ export default class MoneyInput extends React.Component {
     return moneyValue && moneyValue.type === 'highPrecision';
   };
 
-  static isTouched = touched => touched && Object.values(touched).some(Boolean);
+  static isTouched = (touched, { all = false } = {}) =>
+    touched &&
+    (all
```
tdeekens: Where is my friggn test? 💋

filippobocik:
I would be fine with the whole input field having the red border, if we could display the error message below. To my knowing, we couldnt do this, when the inputs are in a table. If this could be solved somehow, I wouldnt mind having the red border around the whole field
image

tdeekens:
> if we could display the error message below. 

Inputs themselves do not have error messages. Fields to combine an input an an error message. Correct me if I am wrong.

dferber90:
@tdeekens Yeah that is true. The concern was more about the table not rerendering when an error pops up (a longstanding, known issue/assumption). But now that I think of it somemore, I guess it isn't even a problem if we pass Formik/Formik's errors as a prop to the table, just so that it rerenders when formik changes.

@filippobocik I think we should be able to do display errors for that case in the table, yep.

dferber90:
@tdeekens Yeah that is true. The concern was more about the table not rerendering when an error pops up (a longstanding, known issue/assumption). But now that I think of it somemore, I guess it isn't even a problem if we pass Formik/Formik's errors as a prop to the table, just so that it rerenders when formik changes.

@filippobocik I think we should be able to do display errors for that case in the table, yep.

dferber90:
I'll split this PR into the fix for the styling and into the corrections of the `MoneyInput` API.

PR #172, opened 2018-10-18
chore: improve linters config

emmenko:


lint-staged.config.js
```diff
@@ -23,10 +23,12 @@ const formatCss = [
 ];
 
 module.exports = {
-  '*.md': formatMarkdown,
   '{.github,.storybook,examples,materials,src}/**/*.md': formatMarkdown,
-  '*.js': formatJs,
+  './*.md': formatMarkdown,
+
   '{.storybook,examples,materials,scripts,src}/**/*.js': formatJs,
+  './*.js': formatJs,
```
dferber90: Can you explain the difference? We are applying `formatJs` to all top-level `js` files and to all `js` files in `.storybook,examples,materials,scripts,src`. Could we just apply it to all of them, as I'm not aware which ones  we need to actually exclude?

I have the same questions for the markdown-files.

emmenko: - `./*.js` matches only top level js files
- `{.storybook,examples,materials,scripts,src}/**/*.js` matches nested js files

I'm not sure what is confusing you...we're using the same setup in all repos

jest.stylelint.config.js
```diff
@@ -3,6 +3,8 @@ module.exports = {
   displayName: 'stylelint',
   moduleFileExtensions: ['css'],
   testMatch: [
+    '<rootDir>/.storybook/*.css',
+    '<rootDir>/.storybook/**/*.css',
```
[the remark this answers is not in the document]
emmenko: Ah, are you sure? I wasn't 100% sure so I put defined it twice...but if `**/*` includes the other one then we can remove it of course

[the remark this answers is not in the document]
dferber90: I tested it locally and `minimatch("f/bar.js", "f/**/*.js")` returns `true` 👍

lint-staged.config.js
dferber90: I would like to find out which files (or if any) we need to exclude, because we could use `**/*.js` otherwise.

dferber90:
@emmenko I rebased and simplified the setup some more. I'd be ready to merge this now if you're okay with these changes.

Issue #170, opened 2018-10-17
Rethink export and consumption of css vars


montezume:
#### Problem

How we export and consume css variables leaves a bit to be desired...

#### Current situation

```css
@import '@commercetools-frontend/ui-kit/materials/colors/base-colors.mod.css';
```

#### Possible way forward

`postcss-custom-media` has introduced `importFrom` and exportTo` functionality.

For example, for `base-colors.mod.css` (which is already generated from JS), we could instead have these generated to a json file, and then use importFrom to consume them. 

I would propose doing a bit of research into this and making a spike out of it.

tdeekens:
Nice idea. I've seen that somewhere that colors and variables in general where exported as JSON, JS and CSS variables.

montezume:
@tdeekens yeah would be great to give consumers more options

dferber90:
Having our tokens as JSON/JS is great for ui-kit itself as well. It is possible to customize `react-select` with CSS-in-JS, but we need to fall back to overwriting its styles in CSS right now as we don't have access to the design tokens from JS!

emmenko:
Who would like to do a spike on this? Should we plan something?

dferber90:
@montezume is on it

dferber90:
@montezume should we close this one?

montezume:
@dferber90 I think we have done everything. The only thing we didn't deal with is exporting the [media queries](https://github.com/commercetools/ui-kit/blob/master/materials/media-queries.mod.css). There is a similar way with [postcss-custom-media](https://github.com/postcss/postcss-custom-media) but I'm not sure if it's worth the trouble. Would be nice to remove that need for V6 that @emmenko found though.

PR #173, opened 2018-10-18
feat(bundling): lower bundle size by discarding comments 

montezume:
#### Summary
While working on #170, I found a quick way to improve our bundle.
We are using [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) to import css variables from other files. This is a common approach in most of our `mod.css`

```
@import '../../../../materials/colors/base-colors.mod.css';

h1.my-class {
   color: var(--color-green);
}
```

and the current output is as follows.

```

/* stylelint-disable comment-empty-line-before */

/* THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
  This file is created by 'scripts/generate-base-color.js' script. The real
  colors's values should be updated in '../materials/colors/decisions/base-colors.json';
  Also, This file is transformed to JSON version in order to list the colors
  in the UIKit. Color groups are indicated by a comment preceeding the group
  of colors. */

/* Base Colors */

:root {

}

h1.my-class {
   color: green
}

```

This is not ideal, as we end up with these comments (and empty root elements), everywhere. Leading to wasted bytes.

#### Approach

Use [postcss-discard-comments](https://github.com/ben-eb/postcss-discard-comments) to remove the comments. However, if we add this to the end of the plugin list, we will notice that we have removed these comments, but we still have empty root elements like this

```
:root{
}

:root{
}

:root{
}

:root{
}

:root{
}

:root{
}

.number-input-mod_transition-standard__2b_2_{
  -webkit-transition:all 0.2s ease;
  transition:all 0.2s ease;
  -webkit-transition:all 0.2s ease;
  transition:all 0.2s ease;
}
```

One root element per import in the file. We can solve this by placing our discard comment plugin BEFORE the postcss-custom-properties plugin.

Now the output is

```
.number-input-mod_transition-standard__2b_2_{
  -webkit-transition:all 0.2s ease;
  transition:all 0.2s ease;
  -webkit-transition:all 0.2s ease;
  transition:all 0.2s ease;
}
```

Beautiful, isn't it? Lowers the bundle size (esm) to 710KB from 813Kb.

TL:DR -> I lowered the bundle size by over 100kb by removing some css comments 😆 🤣

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -12,6 +12,7 @@ import replace from 'rollup-plugin-replace';
 import postcssCustomProperties from 'postcss-custom-properties';
 import postcssCustomMediaQueries from 'postcss-custom-media';
 import postcssPostcssColorModFunction from 'postcss-color-mod-function';
+import postcssDiscardComments from 'postcss-discard-comments';
 import peerDepsExternal from 'rollup-plugin-peer-deps-external';
 import svgrPlugin from '@svgr/rollup';
 import pkg from './package.json';
@@ -35,6 +36,12 @@ const postcssPlugins = [
     browsers: browserslist.production,
     autoprefixer: { grid: true },
   }),
+  // we need to place the postcssDiscardComments BEFORE postcssCustomProperties,
+  // otherwise we will end up with a bunch of empty :root elements
+  // wherever there are imported comments
+  // see https://github.com/postcss/postcss-custom-properties/issues/123
+  // and https://github.com/commercetools/ui-kit/pull/173
+  postcssDiscardComments(),
   postcssCustomProperties({
     preserve: false,
   }),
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -34,6 +35,11 @@ const postcssPlugins = [
     browsers: browserslist.production,
     autoprefixer: { grid: true },
   }),
+  // we need to place the postcssDiscardComments BEFORE postcssCustomProperties,
+  // otherwise we will end up with a bunch of empty :root elements
+  // wherever there are imported comments
+  // see https://github.com/postcss/postcss-custom-properties/issues/123
```
montezume: this comment is important and I tried to explain it's relevance in the PR description.

dferber90: Thanks for the comment! I'd also add a link to this PR to the comment as the PR's description does such a nice job of explaining what's going on.

PR #176, opened 2018-10-18
Use postcss-custom-properties to import postcss variables

montezume:
#### Summary 

As a first step in #170, after talking to @dferber90, I think it makes sense to replace our current use of css imports. 

#### Approach

Use [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) `importFrom` property to import our css variables, and remove all of the `@import`s in our css files. So now, when you are developing in UI-Kit, you don't use @import in css anymore. If you need to add a new file with css variables, you need to adjust both the `rollup.config.js` and the `webpack.config.js`. Although we could extract the list of paths with css variables to one file and consume it in both places to make it simpler 🤔.

#### Consumption changes

No changes, user can continue to consume our variables as they are currently. 

#### TODO

* Update documentation? This is an internal change, as the way of consuming our materials css files remains unchanged for now.

#### Next Steps

* Use exportTo to export javascript and css files that can be consumed in the MC and other apps using either @import, or the same `importFrom` using our exported files.
* Use the exported javascript materials to remove the `select.css` and style the `select-input`s with javascript.
* Get rid of the scripts that generate css files from json, and just directly import our postcss variables from json.

Bundle size seems to have gone down by ~30kb, which is cool.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -108,6 +108,7 @@ module.exports = (storybookBaseConfig, configType) => {
               }),
               postcssCustomProperties({
                 preserve: false,
+                importFrom: ['materials/colors/base-colors.mod.css'],
               }),
               postcssCustomMediaQueries(),
               postcssPostcssColorModFunction(),
@@ -147,6 +148,7 @@ module.exports = (storybookBaseConfig, configType) => {
                   }),
                   postcssCustomProperties({
                     preserve: false,
+                    importFrom: ['materials/colors/base-colors.mod.css'],
                   }),
                   postcssCustomMediaQueries(),
                   postcssPostcssColorModFunction(),
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -10,6 +10,8 @@ const materialSources = [
   'materials/colors/base-colors.mod.css',
   'materials/spacings.mod.css',
   'materials/borders.mod.css',
+  'materials/transitions.mod.css',
+  'materials/sizes.mod.css',
 ];
 
 const browserslist = {
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -7,11 +7,18 @@ const postcssCustomMediaQueries = require('postcss-custom-media');
 const postcssPostcssColorModFunction = require('postcss-color-mod-function');
 
 const materialSources = [
-  'materials/colors/base-colors.mod.css',
-  'materials/spacings.mod.css',
   'materials/borders.mod.css',
-  'materials/transitions.mod.css',
+  'materials/colors/base-colors.mod.css',
+  'materials/shadows/base-shadows.mod.css',
   'materials/sizes.mod.css',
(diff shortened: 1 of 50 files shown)


.storybook/webpack.config.js
```diff
@@ -6,6 +6,21 @@ const postcssCustomProperties = require('postcss-custom-properties');
...
+  'materials/sizes.mod.css',
+  'materials/spacings.mod.css',
+  // tokens
+  'materials/tokens/backgrounds.mod.css',
+  'materials/tokens/borders.mod.css',
+  'materials/tokens/fonts.mod.css',
+  'materials/tokens/shadows.mod.css',
+  'materials/tokens/sizes.mod.css',
+  'materials/transitions.mod.css',
+];
```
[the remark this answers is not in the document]
montezume: Well what @dferber90 and I talked about is getting rid of the scripts that generate these files and grabbing all of the variables from commented javascript files. I'm not sure how much it makes sense to invest in improving this array since my plan would be to remove it in the next or after-next iteration of #170

rollup.config.js
```diff
@@ -36,6 +51,8 @@ const postcssPlugins = [
     browsers: browserslist.production,
     autoprefixer: { grid: true },
   }),
+  postcssCustomMediaQueries(),
+  postcssPostcssColorModFunction(),
```
emmenko: Nitpick: I think this is a leftover name typo

rollup.config.js
```diff
@@ -102,6 +118,9 @@ const plugins = [
   postcss({
     exclude: ['**/*.mod.css'],
     include: ['**/*.css'],
+    // literally the only reason we have this, is because of the `select.css` file
+    // or else we would not need to include any plugins as this would
+    // just be vendor (plain) css.
```
emmenko: Is there any way to get rid of it as well?

montezume: Yes, what I would like to do is in the next step, export a javascript file with all of our css variables, and then remove `select.css` and replace it by importing this file into [select-input](https://github.com/commercetools/ui-kit/blob/0a0722a2d28d13901ed8d2664df39658c29918c4/src/components/inputs/select-input/select-input.js) and passing the styles with the design tokens into react-select there.

montezume:
@emmenko I would like to test it locally on the MC. The deploy link looks good to me though.

emmenko:
Btw, what’s the plan for doing something similar “outside” of the uikit? Eg: Styling a custom view or something. Is it supposed to work the same way?

dferber90:
I just noticed that the changed `yarn.lock` is also where the 2000 removed LoC were coming from. After committing `yarn.lock` generated  with the integrity hashes, it's only 188 removed LoC.

tdeekens:
> After committing yarn.lock generated with the integrity hashes, it's only 188 removed LoC.

Yeah those come since yarn@1.10. I mentioned in Slack already that we should all update. Otherwise one adds the integrity hashes while the next removes them again.

PR #178, opened 2018-10-19
Add optional MoneyField example

dferber90:
In https://github.com/commercetools/ui-kit/pull/175#issuecomment-430995949 it occurred to me that we've so far only shown how to validate required `MoneyField`s.

This adds an example of how to validate an optional `MoneyField`. Notice that we do not error when this field is filled out partially, as it is not required. Supplying either only currency or only an amount will be treated as leaving the field blank. UX could want to show an error in this case, but then we'd also need a way to unset the currency, which is not possible now.

cc @filippobocik


dferber90:
Good idea! Let's do this in another step!

dferber90:
Good idea! Let's do this in another step!

PR #179, opened 2018-10-19
Drop styled-components peerDep (Redo CollapsibleMotion)

### LINKED ISSUE #21: Make styled-components a devDependency
We are currently only using the `keyframes` feature of `styled-components`, and only for a single component. It doesn't make much sense to bundle all of `styled-components` for this.

We should make `styled-components` a `devDependency` and restrict its usage to stories of storybook!

dferber90:
As described in #21 we were including all of  `styled-components` and making it a peer dependency just for the sake of one single component called `CollapsibleMotion`.

The task of `CollapsibleMotion` is not an easy one, as it needs to animate a components height from `0` to `auto` and vice versa. This is a really difficult animation, as browsers don't allow to animate to `auto` - because it is really expensive. There is [one long article](https://css-tricks.com/using-css-transitions-auto-dimensions/) which goes into detail of different approach to this problem and their downsides. None of them are pretty.

`CollapsibleMotion` solves this problem in a clever way as it dynamically injects keyframes to be able to animate. It also sets the height to the measured height and then, only on the last frame of the animation sets it back to `auto`. It thereby simulates the browser animating between `0` and `auto`.

We were using `styled-components` to dynamically inject the generated keyframes into the DOM. Including `styled-components` for this single animation is overkill as styled-components [weighs in at 16.2kb minified+gzipped](https://bundlephobia.com/result?p=styled-components@3.4.10).

This PR
- drops `styled-components` and recreates the dynamic keyframes injection ~~manually~~ *using `emotion`*
- drops `styled-components` as a peer dependency 🔥 
- refactors `CollapsibleMotion` to the new lifecycle and generally simplifies it
- refactors `registerContentNode` to the use `React.createRef` (no breaking change though)
- adds a story for `CollapsibleMotion`
- rewrites the tests in `react-testing-library`

Closes #21

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "classnames": "2.2.6",
     "common-tags": "1.8.0",
     "dom-helpers": "3.3.1",
-    "downshift": "2.2.3",
+    "downshift": "3.1.1",
     "is-touch-device": "1.0.1",
     "lodash.flatmap": "4.5.0",
     "lodash.has": "4.5.2",
@@ -185,14 +185,14 @@
     "rollup-plugin-replace": "2.1.0",
     "rollup-plugin-url": "2.0.1",
     "shelljs": "0.8.2",
-    "storybook-readme": "4.0.0-beta1",
+    "storybook-readme": "4.0.2",
     "style-loader": "0.23.1",
-    "styled-components": "3.4.10",
+    "styled-components": "4.0.2",
     "stylelint": "9.6.0",
     "stylelint-config-standard": "18.2.0",
     "stylelint-order": "1.0.0",
     "url-loader": "1.1.2",
-    "webpack": "4.20.2"
+    "webpack": "4.21.0"
   },
   "peerDependencies": {
     "flatpickr": ">4.5.0",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "test:watch": "jest --config jest.test.config.js --watch"
   },
   "dependencies": {
-    "@sindresorhus/string-hash": "1.0.0",
     "classnames": "2.2.6",
     "common-tags": "1.8.0",
     "dom-helpers": "3.3.1",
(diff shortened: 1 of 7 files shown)


dferber90:
I converted the tests to `react-testing-library`. This is ready for review now 🔍

montezume:
First off, damn good work. You fought the 🐉 ⚔️. 

Just one small thing though, it seems like when the panel renders for the first time, it animates open. I can see this in the deploy preview [here](https://deploy-preview-179--mc-uikit.netlify.com/?selectedKind=Panels%2FCollapsiblePanel&selectedStory=Controlled&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybook%2Factions%2Factions-panel), whereas currently, it doesn't do the animation on render.

#### Current behaviour (doesn't animate when you open the page)
current


#### New behaviour (animates on render)
 
animate-when-shouldnot

package.json
```diff
@@ -97,6 +97,7 @@
     "babel-eslint": "10.0.1",
     "babel-jest": "23.6.0",
     "babel-loader": "8.0.4",
+    "babel-plugin-emotion": "9.2.11",
```
montezume: nice

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -1,27 +1,9 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import invariant from 'tiny-invariant';
-import stringHash from '@sindresorhus/string-hash';
-import styleInject from 'style-inject';
+import { keyframes } from 'emotion';
```
tdeekens: Can we add this as a proper dependency now that we depend on it top-level so we don't just transiently live by it through react-select?

montezume: can add as regular dep and add it as an external to rollup

dferber90: It has been added as a `peerDependency` (and as a `devDependency`). `react-select` also depends on `emotion` and is a `peerDependency` itself.

tdeekens: Ok.

dferber90:
Major shoutouts to @montezume for suggesting to just replace `import { keyframes } from 'styled-components'`  with `import { keyframes } from 'emotion'`.

This might be the most hours we've wasted on what ended up being a single LoC change 🤦‍♂️

PR #181, opened 2018-10-22
feat: extract css variables to js and css files

montezume:
*** THIS CONSTITUTES A BREAKING CHANGE 😱 ***

In MC, the @imports will have to either be removed, or to be changed to the new singular css file.

#### Summary

This is (as far as I can tell), the simplest way to extract the materials vars to js and css without refactoring the materials folder and making a whole bunch of decisions with regards to the tokens.

#### Approach.

* Create an `index.css` file in materials that imports all of the variables we want to expose.
* Use `postcss-custom-properties` to extract those into a js file (`materials/materials.js`) and css file (`materials/materials.css`)
* Format that file with prettier.
* Refactor the `materials` folder and move all internal files to the `internal` directory.
* Only expose `media-queries.mod.css`, `materials.css` and `materials.js` through npm
* Amend the `rollup` and `webpack` configs to use the extracted variables file instead of the array of files.
* Update readme


scripts/extract-materials.js
```diff
@@ -0,0 +1,53 @@
...
+      const exportedCss = fs.readFileSync(`${exportPath}.css`, 'utf8');
+      // format the files with prettier
+      fs.writeFileSync(
+        `${exportPath}.js`,
+        prettier.format(exportedJs, prettierConfig)
+      );
+      fs.writeFileSync(
+        `${exportPath}.css`,
+        prettier.format(exportedCss, { ...prettierConfig, parser: 'scss' })
+      );
```
emmenko: Nice!

Should we maybe inject a comment about those files being autogenerated?

montezume:
@emmenko Good call! Also would make sense to update the README

materials/exports/vars.js
```diff
@@ -0,0 +1,113 @@
+module.exports = {
+  customProperties: {
```
tdeekens: Just curious: is this something we can set? Why is it "wrapped"?

scripts/extract-materials.js
tdeekens: Nice to format them but to be fair: I don't mind too much how an autogenerated file is formatted. :)

montezume: @tdeekens it's either this or to add it to the prettier ignore 😆

materials/exports/vars.js
montezume: @tdeekens this is the shape that [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) generates and expects for it's importFrom. 

There's an example [here](https://github.com/postcss/postcss-custom-properties/blob/master/test/import-properties.js) as well

.storybook/webpack.config.js
```diff
@@ -123,7 +108,7 @@ module.exports = (storybookBaseConfig, configType) => {
               }),
               postcssCustomProperties({
                 preserve: false,
-                importFrom: materialSources,
+                importFrom: 'materials/exports/vars.css',
```
montezume: I don't like these names but I'm not feeling very creative 😞

materials/exports/vars.js
tdeekens: Thanks.

scripts/extract-materials.js
tdeekens: True.

package.json
```diff
@@ -19,8 +19,9 @@
     "dist",
     "i18n",
     "images",
-    "materials/*.mod.css",
-    "materials/**/*.mod.css",
+    "materials/materials.css",
```
montezume: maybe should just call them index.css and index.js?

emmenko: How about `materials/imports.css`?

emmenko: I'm also fine with the `index`, I don't mind much as long as it's proper documented

.storybook/webpack.config.js
dferber90: I would rename `imports.js/css` to `custom-properties.js/css`. That name seems more appropriate. "custom properties" is more or less a fancy name for "css variables". And that's what we're exposing here.

scripts/extract-materials.js
```diff
@@ -0,0 +1,63 @@
...
+
+const relativePath = path.join(__dirname, '../materials/internals');
+const exportPath = path.join(__dirname, '../materials');
+const importPath = path.join(__dirname, '../materials/internals/index.css');
+
+const css = fs.readFileSync(importPath, 'utf8');
+
+const commentDoNotModify = `/* THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+  This file is created by the 'scripts/extract-materials.js' script. The real
+  values of the variables should be updated in '${relativePath}' */ \n
```
dferber90: This resolves to the actual path on the system, so for example `/Users/dom/repos/ui-kit/materials/internals`. This is bad because it would change every time this file gets generated on a different machine. We need to print the path relative to the root-folder, or just hardcode it.

montezume:
Thanks for the help 👍

PR #182, opened 2018-10-22
refactor(*-select-input): use JS-based design tokens

dferber90:
Redoes the date inputs with CSS-in-JS instead of CSS-based styling.

This is in preparation of the date-pickers styling, which will use CSS-in-JS - but it also allows us to get rid of the only global CSS file. We soon have css-modules left only (after dropping `flatpickr`), and we can then drop the setup for the global css from webpack and rollup.

~~This PR uses a manually created tokens-file in JS, which we want to generate automatically in the future (probably even before merging this branch).~~ *This PR now uses the auto-generated `customProperties` thanks to #181*

**Breaking change**: It is no longer possible to overwrite  styles of any of our `*SelectInput` and `*SelectField` components as they will not have classnames anymore. If you have a case for styling them, please file an issue. We will probably enable passing `styles` to our components to allow overwrites instead of continuing to rely on global CSS.


dferber90:
There was this snippet which I couldn't really figure out how to convert. I assume the option group headings will be missing a border-top now.

```css
.react-select .react-select__group:not(:first-child) .react-select__group-heading {
  border-top: 1px var(--token-border-color-separator) solid;
}
```

montezume:
@dferber90 I'm wondering why we don't just apply that to the groupStyles instead of to the groupHeading. IMHO it looks better without the margins.

Code like 

```
const groupStyles = () => base => ({
  ...base,
  padding: 0,
  '&:not(:first-child)': {
    borderTop: `1px ${tokens['--token-border-color-separator']} solid`,
  },
});
```
new
screen shot 2018-10-23 at 12 19 16 pm
old
screen shot 2018-10-23 at 12 19 27 pm



otherwise we would need to somehow import groupHeaderStyle and I'm really not that familiar with emotion...

dferber90:
@montezume Great suggestion! I amended the last commit to include that change.

I also rebased this on top of `master`, resolved all conflicts and switched to using the `custom-properties.json` file.

montezume:
Did you want to remove the lines from the webpack and rollup config in this PR as the 'plugins' [here](https://github.com/commercetools/ui-kit/blob/master/rollup.config.js#L111) is no longer required

dferber90:
@montezume pushed the changes he requested. This is ready to be merged now. We'll do the UX/UI review afterwards.

PR #183, opened 2018-10-22
Rebuild MoneyInput on top of react-select

dferber90:
### Motivation
I noticed that `MoneyInput` was built on top of `downshift`, but we weren't using `downshift` properly. We were only using `isOpen` and `toggleMenu`. So we didn't benefit from ~~any~~ *most* of `downshift`s a11y-features, nor did we benefit from the keyboard navigation it offers. We were using it as a glorified version of `setState(prevState => ({ isOpen: !prevState.isOpen }))`.

`downshift` is a library to enable building *"simple, flexible, WAI-ARIA compliant enhanced input React components"*. They also state in their documentation: *"NOTE: The original use case of this component is autocomplete, however the API is powerful and flexible enough to build things like dropdowns as well."*

### Replacing with  `react-select`
In the case of `MoneyInput` we were using `downshift` as a dropdown, but barely so (as mentioned above). It dawned on me that we already have a perfectly reasonable way to build select-components: `react-select`. So I dropped  `downshift` in favour of `react-select` and it made everything a ton easier. I dropped a couple of custom components and the  `MoneyInput` is now the least complex it has ever been.

Additionally, `MoneyInput` was the only consumer of  `downshift`, so I dropped  the dependency on `downshift` completely. Another 6kb gone.

Note that none of this is meant to be against `downshift` itself. It is a wonderful library and really useful. We were using it wrong, and we already had a more suitable dropdown component in our application - so there was no need to keep `downshift` around.

### Comparison

| Before | After |
| ---- | ---- |
| downshift-based | react-select-based |

💥 Note the amount of components we don't need anymore. Over 560 LoC gone, while the component itself gained a11y and keyboard navigation.

🔍 ~~The tests became outdated. There is already another branch where they get converted to  `react-testing-library`: https://github.com/commercetools/ui-kit/pull/178/files#diff-1d08be2727662e35fa4162f51a3fe737R4. I will wait for that PR to be merged before and fix the tests when rebasing.~~ *Tests are updated now.*


tdeekens:
> So we didn't benefit from any of downshifts a11y-features, nor did we benefit from the keyboard navigation it offers. We were using it as a glorified version of setState(prevState => ({ isOpen: !prevState.isOpen })).

This is sadly true for most to any dropdown in our application. I wanted to give it a swing lately but had to dive into other topics.

src/components/inputs/money-input/money-input.mod.css
```diff
@@ -64,124 +64,3 @@
...
-
-.option-wrapper:last-child {
-  border-radius: 0 0 var(--token-border-radius-input)
-    var(--token-border-radius-input);
-}
-
-.option-wrapper:hover {
-  /* FIXME: apply token for dropdown option hover */
-  background: var(--color-gray-90);
-}
```
tdeekens: 💯

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -508,65 +491,65 @@ describe('rendering', () => {
...
+  //           persist: expect.any(Function),
+  //           target: expect.objectContaining({
+  //             name: `foo.currencyCode`,
+  //             value: 'USD',
+  //           }),
+  //         })
+  //       );
+  //     });
+  //   });
+  // });
```
tdeekens: This is WIP?

src/components/inputs/money-input/money-input.js
```diff
@@ -329,34 +372,55 @@ export default class MoneyInput extends React.Component {
...
+        o => o.value === this.props.value.currencyCode
+      );
+      if (matchedOption) matchedOption;
+      // ensure an option is found, even when the currencies don't include
+      // the money value's currencyCode
+      else if (this.props.value.currencyCode.trim() !== '')
+        ({
+          label: this.props.value.currencyCode,
+          value: this.props.value.currencyCode,
+        });
```
tdeekens: For myself: this is whenever a currency code was set which is now removed from the projects possible currencies.

src/components/inputs/money-input/money-input.spec.js
dferber90: Hehe no worries :) Better eager than late/never :P I already refactored these tests to `react-testing-library` on another branch (for which a PR is open). I didn't merge anything today due to GitHub being flaky. I need to merge my open branches in the proper order, then rebase and then these tests might actually only need few changes.

tdeekens: Same here. I do not dare to merge anything almost.

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -338,11 +338,11 @@ describe('MoneyInput', () => {
   });
 
   it('should show the passed value', () => {
-    const { getByLabelText } = render(
+    const { getByLabelText, getByTestId } = render(
```
dferber90: Look at these beautiful changes of the spec. This PR completely ripped out `downshift` and replaced it with `react-select`. This is a fundamental change in the implementation - and yet the tests remain mostly the same.

Even though I did all the changes above, this spec was able to find a mistake: The select dropdown wasn't usable with a label - because I was passing `id` instead of `inputId` to `Select`. This is so impressive. Testing the DOM instead of the implementation/React components is such an excellent idea.

This is why I love `react-testing-library`.

src/components/inputs/money-input/money-input.js
```diff
@@ -2,13 +2,47 @@ import React from 'react';
...
+}) => {
+  const selectStyles = createSelectStyles({ hasWarning, hasError });
+  return {
+    ...selectStyles,
+    control: (base, state) => ({
+      ...selectStyles.control(base, state),
+      borderTopRightRadius: '0',
+      borderBottomRightRadius: '0',
+      borderRight: '0',
+      width: '72px',
```
qmateub: I'm wondering if we can extract this attributes to a external `const commonAttributes` or something like that. I know that you are not using them all over the place but maybe improves the readibility  😋

src/components/inputs/money-input/money-input.js
```diff
@@ -2,13 +2,47 @@ import React from 'react';
...
+import DropdownIndicator from '../../internals/dropdown-indicator';
 import isNumberish from '../../../utils/is-numberish';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import Contraints from '../../constraints';
 import styles from './money-input.mod.css';
 import currencies from './currencies.json';
-import Currency from './currency';
-import CurrencyDropdown from './currency-dropdown';
+import createSelectStyles from '../../internals/create-select-styles';
+import * as vars from '../../../../materials/custom-properties.json';
```
tdeekens: Nitpick: feels like the import name is not really descriptive. Just fearing a bit that it "spreads". Maybe `materialProperties` would be better? Not sure. It's also not `cssVariables` or `cssProperties` I suppose.

src/components/inputs/money-input/money-input.js
```diff
@@ -328,59 +370,55 @@ export default class MoneyInput extends React.Component {
...
+      hasNoCurrencies,
+    });
+    const options = this.props.currencies.map(currencyCode => ({
+      label: currencyCode,
+      value: currencyCode,
+    }));
+
+    const option = do {
+      const matchedOption = options.find(
+        o => o.value === this.props.value.currencyCode
```
tdeekens: Nitpick: I thought we do not use one letter abbreviations :)? I don't mind too much, at some level one just runs out of names. I remember using `optionCandidate` the other day which I still don't know if it's better.

src/components/inputs/money-input/money-input.js
```diff
@@ -328,59 +370,55 @@ export default class MoneyInput extends React.Component {
...
-          )}
+          <Select
+            inputId={MoneyInput.getCurrencyDropdownId(this.props.id)}
+            name={getCurrencyDropdownName(this.props.name)}
+            value={option}
+            isDisabled={this.props.isDisabled || hasNoCurrencies}
+            isSearchable={false}
+            components={{ DropdownIndicator, ClearIndicator }}
+            options={options}
+            placeholder=""
```
qmateub: The placeholder is empty because is the currency selector and we don't want anything there right?

package.json
```diff
@@ -57,7 +57,6 @@
     "classnames": "2.2.6",
     "common-tags": "1.8.0",
     "dom-helpers": "3.3.1",
-    "downshift": "3.1.1",
```
montezume: 💯

src/components/inputs/money-input/money-input.js
```diff
@@ -2,13 +2,47 @@ import React from 'react';
...
+      ...selectStyles.control(base, state),
+      borderTopRightRadius: '0',
+      borderBottomRightRadius: '0',
+      borderRight: '0',
+      width: '72px',
+    }),
+    singleValue: base => ({
+      ...base,
+      color: do {
+        if (hasError) vars['--token-font-color-error'];
```
montezume: so good 👍

src/components/inputs/money-input/money-input.js
dferber90: I was also struggling to find a good name.

Those are all the `customProperties` we provide in CSS. In CSS we write  `var(--color-red)`, and in JS we can now write `vars['--color-red']`. I liked the similarity. I had `customProperties['--color-red']` at first.

With this explanation, would you be okay with keeping it?

src/components/inputs/money-input/money-input.js
dferber90: Yeah, I ran out of names :( I like `optionCandidate`!

PR #184, opened 2018-10-23
fix: replace styled-components with react-emotion

montezume:
#### Summary

Storybook in production was broken since we removed `styled-components`. 

#### Approach

`react-emotion` is pretty much a plug and play replacement for how we are using sc. Let's use that instead.


rollup.config.js
```diff
@@ -76,10 +76,12 @@ const plugins = [
   resolve(),
   // For shimming nodejs builtins
   builtins(),
-  // To automatically externalize `peerDependencies`
+  // To automatically externalize `peerDependencies` and `dependencies`
   // so that they do not end up in the bundle.
   // See also https://medium.com/@kelin2025/so-you-wanna-use-es6-modules-714f48b3a953
-  peerDepsExternal(),
+  peerDepsExternal({
+    dependencies: true,
```
montezume: if this is a reasonable default for [KCD](https://github.com/kentcdodds/kcd-scripts/blob/8337bbd16fa4f964f688975f3c510d4705c94932/src/config/rollup.config.js#L48), it's good enough for us?

montezume: Yes, I misunderstood how it worked 😢. I think it's better now. If we had UMD bundles, it would make sense to bundle non peer deps.

PR #190, opened 2018-10-26
fix(*-select-input): align indicator icons of clearable select-inputs

dferber90:
Fixes alignment of indicator icons in `SelectInput`, `CreatableSelectInput`, `AsyncSelectInput` & `AsyncCreatableSelectInput`  after it broke in #182.

Before
bildschirmfoto 2018-10-26 um 15 11 50

After
bildschirmfoto 2018-10-26 um 15 12 44

I also changed both icons to have size `medium`. Previously one was `small` and the other was `medium`.

### DIFF
--- a/src/components/internals/clear-indicator/clear-indicator.js
+++ b/src/components/internals/clear-indicator/clear-indicator.js
@@ -2,12 +2,23 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { CloseIcon } from '../../icons';
 
-const ClearIndicator = props => (
-  <div className="react-select__clear-indicator" {...props.innerProps}>
-    {/* FIXME: add proper tone when tones are refactored */}
-    <CloseIcon theme={props.isDisabled && 'grey'} size="medium" />
-  </div>
-);
+const ClearIndicator = props => {
+  const {
+    getStyles,
+    innerProps: { ref, ...restInnerProps },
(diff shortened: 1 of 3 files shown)


src/components/internals/clear-indicator/clear-indicator.js
```diff
@@ -2,12 +2,23 @@ import React from 'react';
...
+      {...restInnerProps}
+      ref={ref}
+      style={getStyles('clearIndicator', props)}
+    >
+      <div>
+        <CloseIcon theme={props.isDisabled && 'grey'} size="medium" />
+      </div>
+    </div>
+  );
+};
```
dferber90: This restyled component enables us to keep overwriting the styles through the `styles={}` prop on `Select` itself. Previously the `ClearIndicator` would just ignore all custom styling.

src/components/internals/create-select-styles.js
```diff
@@ -72,7 +72,6 @@ const indicatorSeparatorStyles = () => base => ({
 const dropdownIndicatorStyles = () => base => ({
   ...base,
   color: vars['--token-font-color-default'],
-  marginRight: vars['--spacing-8'],
```
dferber90: There wasn't really a point in defining `marginRight` when we overwrite it with 0 in the next line.

PR #191, opened 2018-10-26
feat(typography): add `title` prop to all typography components

pa3:
tooltip

`Text.*` components provides a way to truncate long texts by setting `truncate` prop to true. And when using this feature it's a very common scenario to show full version of truncated text in a tooltip on hover. Which leads to code like this all around the place:

```js
<div titile={longText}>
  <Text.Detail truncate>{longText}</Text.Detail>
</div>
```

I suggest to add `tooltip` prop to all the `Text` component to prevent adding unnecessary markup elements. So that example above would look like:

```js
<Text.Detail truncate title={longText}>{longText}</Text.Detail>
```

The only thing I am not sure about is whether it worth adding `title` prop to `Text.Wrap` component, as it not provide truncation behaviour. But I added it to make API across all the `Text.*` components consistent.

### DIFF
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -3,6 +3,15 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import styles from './text.mod.css';
 
+const nonEmptyString = (props, propName, componentName) => {
+  const value = props[propName];
+  if (typeof value === 'string' && !value)
+    return new Error(
+      `Invalid prop '${propName}' supplied to '${componentName}'. Expected it to be nonempty string, but it was empty.`
+    );
+  return null;
+};
+
 const Headline = props => {
   const HeadlineElement = props.elementType;
   return (
@@ -20,7 +29,7 @@ Headline.displayName = 'TextHeadline';
 Headline.propTypes = {
   elementType: PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired,
   children: PropTypes.node.isRequired,
-  title: PropTypes.string,
+  title: nonEmptyString,
   truncate: PropTypes.bool,
 };
 
@@ -51,7 +60,7 @@ Subheadline.propTypes = {
     'negative',
   ]),
   children: PropTypes.node.isRequired,
-  title: PropTypes.string,
+  title: nonEmptyString,
   truncate: PropTypes.bool,
 };
 
@@ -63,7 +72,7 @@ const Wrap = props => (
 Wrap.displayName = 'TextWrap';
 Wrap.propTypes = {
   children: PropTypes.node.isRequired,
-  title: PropTypes.string,
+  title: nonEmptyString,
 };
 
 const Body = props =>
@@ -106,7 +115,7 @@ Body.propTypes = {
     'inverted',
   ]),
   children: PropTypes.node.isRequired,
-  title: PropTypes.string,
+  title: nonEmptyString,
   truncate: PropTypes.bool,
 };
 
@@ -138,7 +147,7 @@ Detail.propTypes = {
     'warning',
   ]),
   children: PropTypes.node.isRequired,
-  title: PropTypes.string,
+  title: nonEmptyString,
   truncate: PropTypes.bool,
 };
 
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -10,6 +10,7 @@ const Headline = props => {
       className={classnames({
         [styles.truncate]: props.truncate,
       })}
+      title={props.title}
     >
       {props.children}
     </HeadlineElement>
@@ -19,6 +20,7 @@ Headline.displayName = 'TextHeadline';
 Headline.propTypes = {
   elementType: PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired,
   children: PropTypes.node.isRequired,
+  title: PropTypes.string,
   truncate: PropTypes.bool,
 };
 
@@ -31,6 +33,7 @@ const Subheadline = props => {
         [styles[`${props.tone}`]]: props.tone,
         [styles.truncate]: props.truncate,
       })}
+      title={props.title}
     >
       {props.children}
     </SubheadlineElement>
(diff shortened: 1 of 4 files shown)


tdeekens:
> But I added it to make API across all the Text.* components consistent.

Personally I appreciate keeping them consistent.

Is this the styling the design teams wants for any or only this type of tooltips or are they fine with diverging with the internal `<Tooltip />` (I don't like that component).

src/components/typography/text/text.spec.js
```diff
@@ -109,6 +119,9 @@ describe('<Subheadline>', () => {
     it('should have "primary" class', () => {
       expect(wrapper).toContainClass(styles.primary);
     });
+    it('should has `title` prop', () => {
+      expect(wrapper.prop('title')).toBe('tooltip text');
```
dferber90: I would suggest `toHaveProp`? It gives slightly better error messages.

You could (but only if you feel like it) rewrite the whole thing with `react-testing-library`. Absolutely no worries if not though! :)

dferber90:
> I heard that UX had some more specific requirements, like only showing the tooltip when the text actually got truncated. Personally, I'd be fine with this already as-is without the additional complexity.

To clarify this some more.. If we accept a `title` prop, I'd make the text always show the title. If we want to build custom "only show title when text is truncated", then we should built that on top of (and outside of) the `title` prop. It should then be achieved by not passing  `title` in case the text did not get truncated.

But let's wait for UX's take on this, and whether it's still a requirement anymore at all.

montezume:
I think it makes sense to strip out the title if it's undefined. Else, if you have a title on a parent element, it doesn't seem to work.

The pink background is the padding of the parent, the white is the Text element.

Current behaviour (can't see title from parent when you hover over the sibling who has empty title attribute)
not-stripped

With title stripped out (can see title from parent when you hover over the sibling who has empty title attribute)

stripped-out

src/components/typography/text/text.js
```diff
@@ -31,6 +33,7 @@ const Subheadline = props => {
         [styles[`${props.tone}`]]: props.tone,
         [styles.truncate]: props.truncate,
       })}
+      title={props.title}
```
montezume: can you strip this out if it's not defined?

mariabarrena:
I would start saying that personally I believe the UI of our current tooltips needs to be improved and we should do that whenever we review those for the UI kit. Being said that, doing a distinction between customised and browser based tooltips is a common practice. 
I would keep the customised ones for those cases where there's something that needs to be explained or provides extra info. They are most likely to appear in cases where the user does not expect them and therefore we can play a bit with the timing when they are being showed to make them appear faster than the browser based one.
In the case of truncation, the chances the user expect to get one are higher so it is more probable the user stops the mouse in the truncated text to trigger the tooltip so we can and should use the browser based ones as we are already doing in the PIM search for truncated content. 
Does this answer the questions @dferber90 and @tdeekens ?

dferber90:
@mariabarrena Thanks for you answer! Your response didn't answer the question of whether we want to always show the title when hovered, or only to show it in case the text got too long.

> I heard that UX had some more specific requirements, like only showing the tooltip when the text actually got truncated. Personally, I'd be fine with this already as-is without the additional complexity.

I found [this snippet](https://stackoverflow.com/a/13259660) on StackOverflow. It would allow us (after converting it to React) to only show the tooltip in case the text actually was too long and got the ellipses in case we want to show the tooltip/title only when the text is too long.

```js
$('.mightOverflow').bind('mouseenter', function(){
    var $this = $(this);

    if(this.offsetWidth < this.scrollWidth && !$this.attr('title')){
        $this.attr('title', $this.text());
    }
});
```

I'm not sure whether we should build it into the Text components themselves or whether we should rather use another component built on top of `Text` instead.

src/components/typography/text/text.js
pa3: Looks like it's not passed to DOM in case it's `undefined`. Behaviour you were observing happens if `title` is an empty string. So, I will add fallback to `undefined` if has a falsy value.

mariabarrena:
@dferber90 I would display the tooltip only if the text is truncated, unless we can select which texts can have tooltips and which ones not. Otherwise it can be annoying to see tooltips anywhere. But I guess in cases where we have responsive screens, it is hard to know if a text is truncated or not. Am I wrong?

montezume:
@mariabarrena It's only going to show up when the developer explicitly sets that `title` prop. It won't put these all over the MC. It can be done through logic in the feature about whether to set title or not.

mariabarrena:
Then it is fine for me to show it either if the text is truncated or not, if that helps you guys a bit :) @montezume @dferber90

src/components/typography/text/text.js
```diff
@@ -10,6 +10,7 @@ const Headline = props => {
       className={classnames({
         [styles.truncate]: props.truncate,
       })}
+      title={props.title || undefined}
```
dferber90: I would either introduce a function to clarify why the `undefined` fallback is used.

```js
// This function avoids empty strings as titles as browsers
// will show an empty title otherwise
const getOptionalTitle = title => title || undefined
```

```jsx
title={getOptionalTitle(props.title)}
```

This helps the next dev understand what's going on here.


Or, I would just do `title={props.title}`. It's the parent's fault if they screw this up and it's not something we should fix for them? We could add a warning in `propTypes` in case `title` is an empty string and encourage to pass `undefined` instead.

montezume: @dferber90 yeah I would go for the second.

Explaining it is just explaining how react works...

pa3:
Reverted fallback to `undefined`. Introduced custom PropType validation instead:

screen shot 2018-10-29 at 14 20 04

To recap: tooltip will be show if it's passed, and now shown otherwise. No logic to check if text was actually truncated (one suggested by @dferber90) is added. @mariabarrena, all good by you?

src/components/typography/text/text.js
```diff
@@ -3,13 +3,23 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import styles from './text.mod.css';
 
+const nonEmptyString = (props, propName, componentName) => {
+  const value = props[propName];
+  if (typeof value === 'string' && !value)
+    return new Error(
+      `Invalid prop '${propName}' supplied to '${componentName}'. Expected it to be nonempty string, but it was empty.`
```
montezume: This is cool 👍

tdeekens: Nice. 💯

Issue #187, opened 2018-10-24
Finish basic bundle-testing setup (includes Visual Regression Testing)


dferber90:
### Problem

So far we didn't run any tests against the produced bundles. Our testing was spread to:
- manual testing inside storybook (which uses the Webpack configuration)
- regular `.spec.js` Jest tests which test the source-files, but not the final bundle (so the setup is different)

But none of these were able to test the final bundle we produce with Rollup. This left us pretty blind
- when updating dependencies related to Rollup
- when making changes to the build process

We also didn't have a proper way to ensure the styling of our components has no regressions.

#### Testing the build

I suggest to test the produced bundle using a different Jest confing and different tests. Instead of calling these files `.spec.js` I would call them `.bundle.spec.js` to indicate that they are run against the bundle instead of the source files. Maybe we can rename `.spec.js` to `.source.spec.js`. Not too sure about the naming. Maybe just  `.bundlespec.js` and `.spec.js` (or `.sourcespec.js`).

Tests against the produced build can verify that:
- the build is actually valid
- the build contains the exported `customProperties`, `i18n` and so on

#### Adding a Sandbox

We can add a sandbox (for lack of a better name) which is a web-app which uses the produced bundle just like a consumer would. We can do manual testing there.

Later on, this can be used as the basis for the Sketch export of our React components.

#### Adding Visual Regression Testing

We can further add Visual Regression Testing to ensure there are no visual regressions in our application. This will be immensely useful when making changes to the Rollup/PostCSS setup, but also when modifying components.

### Getting there

To enable these solutions, we'd need to:
- add more npm `scripts` to run the different things
- add another `jest` config which runs the `bundle-spec` files
- set up Visual Regression Testing

### State

I already experimented with all of the solutions above on https://github.com/commercetools/ui-kit/tree/df-add-sandbox. Some cleanup is left, but it works for the most part. I already showed it to @montezume to get some feedback, and he's on board!

I want to split that branch up into:
- adding a sandbox
- adding tests for the created bundle
- adding visual-regression-testing

Right now it's all entangled in one PR, but these are not related to each other. I will create different PRs soon to show the addition of these testing-setups step-by-step.

### Todo list

For the current entangled branch, these things are left to do:
- split addition of sandbox (the localhost:8080 & wepback thing) and the addition of VRT
- enable having `.bundlespec.js` files spread across the folders instead of containing them within sandbox
- use two different test-util files (react-testing-library might work differently depending on each one)
[...]

PR #192, opened 2018-10-27
Add playground

dferber90:
Adds a playground which allows testing the final build manually.

image

The playground can be started with `yarn playground`. It will test the production bundle of `ui-kit` - which needs to be built with `yarn build` or `yarn build:watch` first.

### Playground Contents

The playground contents themselves are currently pretty minimal. We can render more components, make the components interactive or add routing there at some point. For now I'm happy just having a place where we can try components out without on their own.

### Future

In the future, we might use the playground setup to export the UI Kit to Sketch using [`html-sketchapp`](https://github.com/brainly/html-sketchapp). We had a working proof-of-concept a while back but we didn't merge it back then because there were some problems with it still.

This relates to #187.

### DIFF
--- a/playground/README.md
+++ b/playground/README.md
@@ -1,0 +1,25 @@
+# Playground
+
+The playground offers a way to try out the final bundle in a web application, just like consumers of `@commercetools-frontend/ui-kit` would.
+
+## Staring the playground
+
+As the playground uses the produced bundle of UI Kit, you need to continuously build the bundle using `yarn build:watch`.
+
+Then you can start the playground in another terminal with `yarn playground`.
+
+## Building the playground
+
+Building the playground should not be necessary, so there is no script/config for it.
+
+## Using `@commercetools-frontend/ui-kit` inside the playground.
+
+This package (`@commercetools-frontend/ui-kit`) is aliased to `ui-kit` inside the playground. So you can do the following to import from UI Kit:
+
+```js
+import { PrimaryButton, MoneyInput, i18n } from 'ui-kit';
+```
+
+## Does the playground use the CJS or the ESM build?
+
+The playground uses the ESM build (`dist/ui-kit.esm.js`).
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "storybook:build": "build-storybook -o .public",
     "build": "rimraf dist && cross-env NODE_ENV=production rollup --config",
     "build:watch": "rimraf dist && cross-env NODE_ENV=production rollup --config --watch",
+    "playground": "webpack-dev-server --config playground/webpack.config.js --watch",
     "start": "yarn storybook:start",
     "storybook:start": "start-storybook -p 9002",
     "docs:publish": "node scripts/publish-docs.js",
@@ -77,6 +78,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.1.2",
+    "@babel/plugin-proposal-class-properties": "7.1.0",
     "@babel/plugin-proposal-do-expressions": "7.0.0",
     "@babel/plugin-proposal-export-default-from": "7.0.0",
     "@babel/plugin-proposal-export-namespace-from": "7.0.0",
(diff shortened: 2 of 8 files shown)


tdeekens:
Nice! 💯

emmenko:
> We can render more components, make the components interactive or add routing there at some point

Btw: I’d love to start having soon a proper (interactive) documentation website, alongside the app-kit stuff. Maybe we can use that instead for testing the production build. We can also have a playground or examples space as part of that, and not something separate.

Two birds with one stone.

Let’s think about it 😉

tdeekens:
> alongside the app-kit stuff. Maybe we can use that instead for testing the production build.

I agree that eventually a documentation website might be nice. I am struggling to see how it replaces this adequately. Moreover, I already like the storybook readmes and would focus on component coverage instead first.

Furthermore, testing a production build on a documentation site feels a bit counter intuitive and as if feels as if it's too late then. I want to know if the bundle works before I publish (canary or latest) with feedback from CI. That's how I understood the intention of this PR.

playground/README.md
```diff
@@ -0,0 +1,25 @@
+# Playground
+
+The playground offers a way to try out the final bundle in a web application, just like consumers of `@commercetools-frontend/ui-kit` would.
+
+## Staring the playground
```
montezume: `Starting` the playground?

dferber90:
> I’d love to start having soon a proper (interactive) documentation website, alongside the app-kit stuff

I didn't hear a complaint about our current documentation so far. Are you missing something from the docs? The storybook itself is already interactive. I also don't get why the Storybook wouldn't be considered "proper".

I don't want to build a documentation website only because it would be cool to have. We also didn't plan a documentation website for Q4, and it was not mentioned as being important so far.

If we build a documentation website now, then there's just one more thing to update when working on components which will slow component development down. And given that most of our components are unstable and still changing a lot, it feels premature to build more polished documentation at this stage.

That's why I agree with @tdeekens that the README's in Storybook & stories are good enough for now - and that component coverage is more important than beautiful docs at the moment.

Ultimately - when the components and remaining setup have matured - I'd love to have polished docs of course! At that point, it might be better to replace Storybook with a doc website though to prevent maintaining documentation & examples in two places.

> We can also have a playground or examples space as part of that, and not something separate.

The playground is rarely used in practice. I found it useful to debug why Visual Regression Tests were broken in some cases - especially while working on the VRT setup. It helps to determine whether the VRT setup is broken, or whether the component itself is broken. The playground also helps to see what components look like when the page itself doesn't have any styles. It's good to determine what UI Kit offers "out of the box" and what the minimal setup of UI Kit looks like for consumers.

Furthermore, the playground can be improved to allow us to export the UI Kit to Sketch soon (as stated in the README). This is something which would not be possible to do with the docs website.

I wasn't sure whether to even add the playground at the moment, but for the rare case that the VRT setup needs debugging I found it useful enough. It's not something which is going to be used a lot though. I also like to keep it around as a "stepping-stone" for the UI-Kit-to-Sketch export which will require something very similar.

> I want to know if the bundle works before I publish (canary or latest) with feedback from CI. That's how I understood the intention of this PR.

Almost! This PR can be used to test this manually, but the playground will not run on CI. Automated tests of the produced bundle are added  in #194. They don't require a playground though :)

tdeekens:
> Almost! This PR can be used to test this manually, but the playground will not run on CI. Automated tests of the produced bundle are added in #194. They don't require a playground though :)

I understand. Sorry, I failed to make myself clear. I mixed two PRs but their intend relates. I wanted to say that testing the bundles is done on CI and gives assurance that the build output passes a sanity check. This PR is obviously for manual checking.

emmenko:
As usual communicating feedback and intentions in plain text just doesn't work 😔
I wish there was a way of leaving voice comments instead 🗣📣

---

Nothing is planned or something but I expressed my intention of having a proper documentation website for developing MC/UI applications since half a year ago, and part of that includes the UI-Kit obviously.
Our current storybook is good enough for now, no doubt about that. We also have more important tasks to work on.
But...if you were to start working with the UI-Kit as a developer, there is not much documentation material to help you getting started and understand the intentions behind the different UI-Kit parts.
I'm sure we are all aware of that.

Nevertheless, at some point I would love to have something like that for the following reasons:
- help people getting started
- have a nicer website that just looks more professional
- describe our Vision with the UI-Kit
- document our Design System intentions and values
- build the website using the UI-Kit itself
- interactive components API documentation (_below some examples on top of my mind_)
  - https://react-kawaii.now.sh/#/Components
  - https://atlaskit.atlassian.com/packages/core/button
- have examples and tutorials (e.g. "how to build a form")

Anyway, this is not the place to talk about that.

I just wanted to hint that in the future we can include those things into one tool.

PR #194, opened 2018-10-27
Add bundle tests setup

dferber90:
So far it was not possible to test the produced ui-kit bundle. This left us blind when errors happened within the bundling process (e.g. when updating the rollup config).

This PR adds a way to test the produced `dist/ui-kit.esm.js` bundle.

There are now two types of tests, which can both be placed anywhere in the code-base:
- `*.spec.js`: The regular tests for the sources
- `*.bundlespec.js`: Tests for the produced bundle

### `*.spec.js`

- These tests use Jest to transpile the source code of the file-under-test
- It is possible to mock other source files in these tests
- These tests can not catch errors in the build process
- These tests can not test the final bundle

### `*.bundlespec.js`

- These tests use the produced final bundle (they don't need transpilation on modern envs)
- It is **not** possible to mock parts of ui-kit
- These tests can catch errors in the build process
- These tests can only test the final bundle
- These tests **may not** import source files

### Future

This is the basic setup for testing the produced bundle. I'll open a PR which adds Visual Regression Testing (VRT) on top of this setup soon. VRT will also run on top of the produced bundle.

Relates to #187

### DIFF
--- a/src/index.bundlespec.js
+++ b/src/index.bundlespec.js
@@ -4,7 +4,16 @@ describe('exports', () => {
   it('should export i18n for three languages', () => {
     expect(Object.keys(i18n)).toEqual(['en', 'de', 'es']);
   });
+
   it('should export custom-properties', () => {
     expect(customProperties).toBeTruthy();
+  });
+
+  it('should have the translations', () => {
+    // This test ensures some translations exist, assuming the others would
+    // work as well then
+    expect(i18n.en['UIKit.DateInput.labelRange']).toEqual('to');
+    expect(i18n.de['UIKit.DateInput.labelRange']).toEqual('bis');
+    expect(i18n.es['UIKit.DateInput.labelRange']).toEqual('hasta');
   });
 });
--- a/src/index.bundlespec.js
+++ b/src/index.bundlespec.js
@@ -1,5 +1,4 @@
 import { i18n, customProperties } from 'ui-kit';
-// import { i18n, customProperties } from '..';
 
 describe('exports', () => {
   it('should export i18n for three languages', () => {
--- a/src/index.bundlespec.js
+++ b/src/index.bundlespec.js
@@ -1,0 +1,11 @@
+import { i18n, customProperties } from 'ui-kit';
+// import { i18n, customProperties } from '..';
+
+describe('exports', () => {
+  it('should export i18n for three languages', () => {
+    expect(Object.keys(i18n)).toEqual(['en', 'de', 'es']);
+  });
+  it('should export custom-properties', () => {
+    expect(customProperties).toBeTruthy();
+  });
+});
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -58,16 +58,20 @@ rules:
           'componentDidUpdate',
         ]
     'function-paren-newline': 0
+    # ui-kit is a mapped module in bundle tests, so it will not be found
+    # settting ignorePackages will disable the extension warning for it.
     'import/extensions': [
       'error',
-      'always',
+      'ignorePackages',
       {
         js: 'never',
       },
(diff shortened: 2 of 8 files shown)


dferber90:
For some reason `yarn build && yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter` works locally but fails on Travis CI.

It works when I replace `ui-kit` for `..` in the import in `index.bundlespec.js`:

```diff
- import { i18n, customProperties } from 'ui-kit';
+ import { i18n, customProperties } from '..';
```

However, I'd want to avoid having to use relative paths. Maybe something about `moduleNameMapper` is wrong. I also had trouble with it in watch-mode before.

I tried various combinations but couldn't make it work. Instead, I'll use [babel-plugin-module-rewrite](https://www.npmjs.com/package/babel-plugin-module-rewrite) to rewrite the `ui-kit` import to a relative path pointing to the root directory which solves the problem nicely.

We can remove that once we figure out how to do this with `jest` itself.

emmenko:
How about you actually `npm pack` and install the tarball as a dependency and run the tests against that?

You can have a look [here](https://github.com/commercetools/nodejs/blob/master/scripts/install-npm-packages-for-integration-tests) how we run the integration tests for the nodejs packages.

That way we also test that npm builds a correct package.

dferber90:
Nice idea, but that doesn't work with continuously building the bundle right? The current flow works wonderfully locally. It's just that `moduleNameMapper` seems somehow broken on CI (and I previously ran into problems with `moduleNameMapper` in Jest's watch-mode as well).

emmenko:
Yeah, that would be mostly useful to run on CI

src/index.bundlespec.js
```diff
@@ -0,0 +1,11 @@
+import { i18n, customProperties } from 'ui-kit';
+// import { i18n, customProperties } from '..';
```
montezume: can we remove this commented out line?

src/index.bundlespec.js
```diff
@@ -0,0 +1,11 @@
...
+
+describe('exports', () => {
+  it('should export i18n for three languages', () => {
+    expect(Object.keys(i18n)).toEqual(['en', 'de', 'es']);
```
montezume: maybe we should check if the shape is correctish? because we have had issues in the past where we go from exporting
```
de: {
   property: value
}
```

to 
```
de: {
 default ...
}
```

dferber90:
I noticed something is still wrong on this PR. The `ui-kit` rewrite to an relative path still fails at times. I'll look into it again before the merge.

src/index.bundlespec.js
dferber90: We can do this in another step. This PR is intended to get the setup up and running. That test was just the simplest one I could come up with.

I'm not too what a test would look like for this. Would it expect that each language doesn't have the  `default` key?

montezume: Yeah, let's leave it to another PR. I guess the simplest way would just be to check like

expect(i18n.de.UIKit.DateInput.labelRange).toEqual('to');

dferber90:
The issues seem to have come from caching issues. It is working fine now. I added 150e2bb to ensure tests are run against recent bundles. This aims to prevent WTF moments when one forgets to `yarn build:watch` before starting the tests.

src/index.bundlespec.js
dferber90: Good idea 👍 I added these now.

The tests would fail in the unlikely case this translation changes, but it's worth ensuring that it works. I also thought about `expect(i18n.en['UIKit.DateInput.labelRange']).toEqual(expect.any(String));` or using a regex, but then I just went for the explicit tests as it is so unlikely to change.

PR #195, opened 2018-10-27
Add visual regression testing setup

dferber90:
Adds the most basic setup for Visual Regression Testing.

To try it out, run `yarn build:watch` and then in another terminal, run `yarn test:bundle` (or `test:bundle:watch`).

Future PRs can then:
- introduce a separate `test-util` file for VRT (we already have one for source tests but we should not reuse it as the setup might be different between source and bundle tests)
- add custom eslint rules to ensure the `.bundlespec.js` tests don't import components directly (and uses the right test-utils)


dferber90:
Short update on the state:

The setup works perfectly fine locally, but there are some issues with [running Puppeteer on TravisCI](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-travis-ci) and there is currently no way to see the diff of a failed visual regression test as the files are not accessible on TravisCI.

### Running Puppeteer on TravisCI

We need to pass `--no-sandbox` to `puppeteer.launch()`. This can be done in `jsdom-screenshot` using

```js
await generateImage({ launch: { args: ['--no-sandbox'] } })
```

But this means we'd have to repeat the setup for every test. And we should rather invoke `--no-sandbox` on CI only anyways:

```js
await generateImage({ launch: { args: process.env.CI ? ['--no-sandbox'] : [] } })
```

Repeating this for every `generateImage` is obviously quite inconvenient. So instead, I want to add a `setConfig` method to [`jsdom-screenshot`](https://github.com/dferber90/jsdom-screenshot). We could then do

```js
// in scripts/setup-test-framework-for-bundle.js
import { setConfig } from "jsdom-screenshot";

setConfig({ launch: { args: process.env.CI ? ['--no-sandbox'] : [] } })
```

Then the tests could be as simple as 

```js
await generateImage({ launch: { args: process.env.CI ? ['--no-sandbox'] : [] } })
```

again.


### Upload diff images from failed tests

In [`jest-image-snapshot`](https://www.npmjs.com/package/jest-image-snapshot#upload-diff-images-from-failed-tests) is a section called *Upload diff images from failed tests* which as [an example](https://github.com/americanexpress/jest-image-snapshot/blob/HEAD/examples/image-reporter.js) of how to use a jest-runner to upload diff images from failed tests. We can use this approach to upload the failed tests to AWS S3. There might be other ways to do this as well though.

dferber90:
I've got the image upload to AWS S3 working (I think, CI is currently not starting the build), but there is an unfortunate limitation. Due to security restrictions, the [secure env variables are not available for PRs from forks](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions) which totally makes sense. However, this also means we can't upload VRT diffs of PRs coming from forks.

One way around this would be to upload to a different host which does not require auth (no secure env variables necessary).

dferber90:
Setting up the reporter to upload the diff-images worked, and as we expected it is about the font, [see diff image](https://visual-regression-tests.s3.amazonaws.com/null/src/components/buttons/primary-button/__image_snapshots__/__diff_output__/primary-button-bundlespec-js-has-no-visual-regressions-1-diff.png).

Overall I'm pretty happy with how this is turning out so far! I'm sure the font issue can be solved as well.

dferber90:
I modified the jest-runner to create an overview of the failed Visual Regression Tests to avoid having to check each image individually.

The link in the logs looks like this

image

And the actual page looks like [this](https://visual-regression-tests.s3.amazonaws.com/df-add-bundle-tests/447392512/index.html):

image

dferber90:
So `jsdom-screenshot` takes screenshots using `puppeteer`. `puppeteer` launches headless chrome, whose font rendering depends on the host OS. This is where the discrepancy comes from as I created the initial screenshots by running the tests on Mac OS, but CI runs them on Linux.

Seems like running the tests inside a Docker container would ensure consistent font rendering. Not too sure how to approach this though.

Would Docker need to be added to `jsdom-screenshot`, is that even possible!? Or would we run our bundle tests in Docker? I wanna talk to you peeps about this to get some direction here.

tdeekens:
That‘s exactly why our Cypress VRTs also failed. On an issue I reported the maintainers suggested to either inject a global style-sheet to overwrite fonts to something OS independent (e.g Arial). I am not sure and never tested if that helps or to run tests (as you also state) in a Docker container. It all kind of makes sense that the struggles here are the same (sadly). I wonder how anyone gets a decent and stable result. Maybe font overwriting works cause I doubt anybody spins up a Docker container.  

Thanks for the efforts.

tdeekens:
Find my old issue in the repo again. Just for completeness. https://github.com/palmerhq/cypress-image-snapshot/issues/18

dferber90:
Upon further consideration, it seems better to use an external service to do VRT.

Having the tests defined locally has these downsides:
- they can not run in parallel, so they are slow
- we'd need to add docker so that they are consistent across machines
- even with docker, we're not guaranteed consistent rendering

PR #203, opened 2018-10-29
Add prop to trigger hovered state of `Radio.Option` and `Checkbox`

pa3:
### Description

Sometimes it's needed to trigger `hovered` state of switches when `onMouseEnter` happened somewhere up the hierarchy, like in the following case:

screen shot 2018-10-29 at 10 51 14

, where Radio.Option should be highlighted when `Card` component is hovered. 

I can see two options:
* Let client code to mess with `Switches.*` components internals, buy introducing CSS selectors for parent components which will change the color of switches' svgs.
* Introduce `isHovered` prop on our `Switches.*` components.

This PR takes second approach to solve the issue. 

### Stuff I am not sure about

* Maybe I have overlooked a third and more elegant approach. Any suggestions are welcome!
* For now I didn't rewrite tests to use `react-testing-library` as changes all about visual appearance. Once [Visual Regression Testing](https://github.com/commercetools/ui-kit/issues/187) will be there, I will come with another PR to rewrite enzyme-tests for touched components.


src/components/switches/radio/README.md
```diff
@@ -40,11 +40,12 @@ import { Radio } from '@commercetools-frontend/ui-kit/switches';
...
-| `isDisabled` | `bool`   |    -     | -      | `false` | Disables the Radio.Option within the Radio.Group              |
-| `value`      | `string` |    ✅    | -      | -       | The value of the option passed to `onChange` in `Radio.Group` |
+| Props        | Type     | Required | Values | Default | Description                                                                                                      |
+| ------------ | -------- | :------: | ------ | ------- | ---------------------------------------------------------------------------------------------------------------- |
+| `children`   | `string` |    -     | -      | -       | The descriptive text of the Radio.Option                                                                         |
+| `isDisabled` | `bool`   |    -     | -      | `false` | Disables the Radio.Option within the Radio.Group                                                                 |
+| `isHovered`  | `bool`   |    -     | -      | `false` | Renders the Radio.Option in a hovered state. We need this, as hover can somethimes come from a parent component. |
```
dferber90: As far as I understand, the `isHovered` passed in here adds to the natural hovered state which uses `:hover`. So if any of them are true, the hovered state is shown, right?

The current documentation makes it seem as-if we always need to pass `isHovered`, when it is actually only needed seldom. I would change the wording to make it more obvious that this is an edge-case to prevent devs from reimplementing the `isHovered` for the case where the component supports it out of the box.

I would also add a note about this in the propTypes of the component's source code itself (not the docs).

src/components/switches/checkbox/checkbox.spec.js
```diff
@@ -84,6 +84,36 @@ describe('<Checkbox>', () => {
...
+        beforeEach(() => {
+          props = createTestProps({
+            isHovered: true,
+          });
+
+          wrapper = shallow(<Checkbox {...props} />);
+        });
+
+        it('should render checkbox in a hovered state', () => {
+          expect(wrapper.exists('.isHovered')).toBe(true);
```
dferber90: I would use `toHaveClassName` instead as It gives nicer error messages.. or maybe I'm missing some upsite of doing it like this?

pa3: I find `exists('.isHovered')` to give less coupling to component internals. It just check that some traces of hovered state are there, without bounding itself to specific DOM-element which is accepting this class.

dferber90: Ah that's what I thought! A drop on a hot stone (a badly translated german saying :D) in this case unfortunately, but a good thought! Can't wait to convert these to `react-testing-library` and VRTs.

Another hint would be to import the styles and then do `styles.isHovered`. This helps with the logic when reading the tests. But I don't care as this is going away. Just wanted to point it out in case you weren't aware already.

pa3:
@dferber90 updated wording of READMEs

PR #206, opened 2018-10-30
Money Input Fixes

### LINKED ISSUE #205: MoneyInput does not format on blur in Safari
This umbrella issue keeps track of 3 things
- not formatting properly in Firefox on blur
- inputs look different in Safari
- ellipses are shown for the currency code even for short currencies like EUR

`MoneyInput` does not format on blur in Firefox.

image

It does work in Safari, but it looks a bit broken in Safari (the border gets truncated). The ellipses happens for all browsers though, but it should not appear.

image

Further, the currency gets ellipses in some cases. It should always be visible.

montezume:
#### Approach
A few CSS fixes to fix the problems `ellipses are shown for the currency code even for short currencies like EUR` and `inputs look different in Safari`.

However, due to Firefox limitations, it's not possible to have a correct format on blur functionality on firefox, as firefox number inputs don't allow extra zeros. So the input has been changed to type "text"

* Fixes #205 
* Fixes focus so that the input acts as one input with regards to focus and error states.

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -338,14 +338,16 @@ export default class MoneyInput extends React.Component {
     // value to a number, as we want to keep a string!
     // The fake event does not contain the input type information, so Formik
     // will not convert the value to a number.
-    const fakeEvent = {
-      persist: () => {},
-      target: {
-        name: event.target.name,
-        value: event.target.value,
-      },
-    };
-    this.props.onChange(fakeEvent);
+    if (isNumberish(event.target.value)) {
+      const fakeEvent = {
+        persist: () => {},
+        target: {
+          name: event.target.name,
+          value: event.target.value,
+        },
+      };
+      this.props.onChange(fakeEvent);
+    }
   };
 
   handleAmountBlur = () => {
@@ -429,7 +431,7 @@ export default class MoneyInput extends React.Component {
             ref={this.amountInputRef}
             id={MoneyInput.getAmountInputId(this.props.id)}
             name={getAmountInputName(this.props.name)}
-            type="number"
+            type="text"
             value={this.props.value.amount}
             className={getAmountStyles({
               isDisabled: this.props.isDisabled,
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -356,19 +356,8 @@ export default class MoneyInput extends React.Component {
   };
 
   handleAmountChange = event => {
-    // We need to emit a fake event to stop Formik from auto-converting the
-    // value to a number, as we want to keep a string!
-    // The fake event does not contain the input type information, so Formik
-    // will not convert the value to a number.
     if (isNumberish(event.target.value)) {
-      const fakeEvent = {
-        persist: () => {},
-        target: {
-          name: event.target.name,
-          value: event.target.value,
-        },
-      };
-      this.props.onChange(fakeEvent);
+      this.props.onChange(event);
     }
   };
 
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -377,8 +377,7 @@ export default class MoneyInput extends React.Component {
       // When the user entered a value with centPrecision, we can format
       // the resulting value to that currency, e.g. 20.1 to 20.10
       if (String(formattedAmount) !== amount) {
-        // We need to emit a fake event to stop Formik from auto-converting the
-        // value to a number, as we want to keep a string!
+        // We need to emit an event with the now formatted value
         const fakeEvent = {
           persist: () => {},
           target: {
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -19,7 +19,7 @@ const createCurrencySelectStyles = ({
   hasError,
   hasNoCurrencies,
   isDisabled,
-  inputHasFocus,
+  hasFocus,
 }) => {
   const selectStyles = createSelectStyles({ hasWarning, hasError });
   return {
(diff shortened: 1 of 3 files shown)


src/components/inputs/money-input/money-input.js
```diff
@@ -31,6 +31,11 @@ const createCurrencySelectStyles = ({
     }),
     singleValue: base => ({
       ...base,
+      marginLeft: '0',
+      marginRight: '0',
+      maxWidth: '100%',
+      left: '50%',
+      transform: 'translate(-50%, -50%)',
```
montezume: this fixes the ellipsis problem.

src/components/inputs/money-input/money-input.mod.css
```diff
@@ -23,6 +23,7 @@
   outline: none;
   box-shadow: none;
   appearance: none;
+  margin-left: 0;
```
montezume: safari display problem

src/components/inputs/money-input/money-input.js
dferber90: Let's add a code comment so the next person reading this knows what's up.

dferber90: In `control` we set `width: '72px',`. This is what I did originally to get the `react-select` to be visible at all and to avoid the currency ellipses. Can we remove that now? Or would it maybe have been enough to just bump that to `75px`?

src/components/inputs/money-input/money-input.js
```diff
@@ -333,14 +338,16 @@ export default class MoneyInput extends React.Component {
...
-    this.props.onChange(fakeEvent);
+    if (isNumberish(event.target.value)) {
+      const fakeEvent = {
+        persist: () => {},
+        target: {
+          name: event.target.name,
+          value: event.target.value,
+        },
+      };
+      this.props.onChange(fakeEvent);
```
dferber90: Now that we're not using an `<input type="number" />` we avoid the problems formik introduced by trying to auto-convert the value to a number. I think we can just do

```js
if (isNumberish(event.target.value)) this.props.onChange(event)
```

and delete the comment as well!

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -327,9 +327,9 @@ describe('MoneyInput', () => {
     expect(getByLabelText('Amount')).toHaveAttribute('data-foo', 'bar');
   });
 
-  it('should render a number input', () => {
+  it('should render a text input', () => {
     const { getByLabelText } = render(<TestComponent />);
-    expect(getByLabelText('Amount')).toHaveAttribute('type', 'number');
+    expect(getByLabelText('Amount')).toHaveAttribute('type', 'text');
```
dferber90: IMHO we can delete this spec. The user doesn't care what kind of input we render as long as they can type. Even if we rendered a content-editable div, they wouldn't care. So I guess this spec can just go away completely.

I wasn't sure whether I should add it back then, and it's kinda annoying that we had to change the tests just because the implementation changed. Users don't care about the type here.

src/components/inputs/money-input/money-input.js
montezume: Just figured this was easiest way to fix the problem without changing the look or width of the input. Just replaced the margin on the left and the right, made the max width 100% instead, and centre aligned it.

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -345,15 +341,24 @@ describe('MoneyInput', () => {
     expect(getByTestId('money-input-container')).toHaveTextContent('EUR');
   });
 
-  it('should allow changing the amount', () => {
-    const onChange = jest.fn();
+  it('should allow changing the amount with numbers', () => {
+    let event;
+
+    const onChange = e => {
```
montezume: courtesy of @pa3

pa3: Just to shed some light on that: once creation of `fakeEvent` was removed from `handleAmountChange` the test started to check agains actual React Synthetic Events. By the time of expectation is executed Synthetic Event is already returned back to pool and all it fields is nullified, which leads to failing tests and lot's of warnings in console about us messing with Synthetic Events. 

That is just work-around, and maybe we'll be able to come with some utility function to deal with this kind of issue.

Weird stuff, that in theory, test should start passing once we add the body to jest.fn, like this:

```js
  it('should allow changing the amount', () => {
    const onChange = jest.fn(e => e.persist()); /// <====== here
    const { getByLabelText } = render(<TestComponent onChange={onChange} />);

    const event = { target: { value: '12' } };
    fireEvent.change(getByLabelText('Amount'), event);

    // onChange should be called with the updated amount
    expect(onChange).toHaveBeenCalledWith({
      persist: expect.any(Function),
      target: {
        name: 'some-name.amount',
        value: '12',
      },
    });
  });
```

But once I do this, jest starts to hang on trying to execute `toHaveBeenCalledWith` which looks crazy and I have no idea why is that happening.

montezume:
@dferber90 anything else from your side?

src/components/inputs/money-input/money-input.js
dferber90: I found a more natural way to do this without `transform`. I'll push the commit onto this branch. Setting `maxWidth: 'initial'` fixes the problem on its own luckily.

src/components/inputs/money-input/money-input.spec.js
dferber90: I got curious and looked into this a bit.

I dumbed the problem down and the following code is working for me:

```jsx
  it('should allow changing the amount with numbers', () => {
    const onChange = jest.fn();
    const { getByLabelText } = render(
      <TestComponent
        onChange={onChange}
        value={{
          currencyCode: 'EUR',
          amount: '13.00',
        }}
      />
    );

    fireEvent.change(getByLabelText('Amount'), { target: { value: '12' } });
    expect(onChange).toHaveBeenCalledTimes(1);
  });
```

This means the hanging is not because of the event itself. If we change the test to

```diff
-     const onChange = jest.fn();
+    const onChange = jest.fn(event => {
+      event.persist();
+    });
```

then it still works.

But it hangs again if we now change

```diff
-    expect(onChange).toHaveBeenCalledTimes(1);
+    expect(onChange).toHaveBeenCalledWith({ foo: true });
```

My assumption is that this happens because there are cycling references inside the event and so the `toHaveBeenCalledWith` follows this loop infinitely.

If we do this instead, the test will pass (and not hang)

```diff
-    expect(onChange).toHaveBeenCalledTimes(1);
+    expect(onChange.mock.calls[0][0].target.name).toBe('some-name.amount');
```

This would also work:

```diff
-    expect(onChange).toHaveBeenCalledTimes(1);
+    expect(onChange.mock.calls[0][0].target).toEqual(
+      expect.objectContaining({ name: 'some-name.amount' })
+    );
```

This brings us to the next problem. We can not verify the value with this approach. If we try to read `onChange.mock.calls[0][0].target.value`, then we will see the old value. This is because `MoneyInput` is a controlled component and the current test never passes the original value back in as a prop. So the amount input (what we see as `event.target`) will always keep the same value - even when `onChange` is called correctly, because our test is not passing the value back in - and because `event.target` is a reference to the DOM node itself - whose value never updated.

So.. what @pa3 came up with seems to be a good workaround, even if slightly awkward. Testing plain APIs of components doesn't seem to be a strength of `react-testing-library`, or maybe we just got something totally backwards.

PR #207, opened 2018-11-04
Add percy

dferber90:
This is an experiment to see how well [percy.io](https://percy.io/) would work for us.


dferber90:
Closing in favor of #267 which starts with a blank slate.

PR #214, opened 2018-11-05
chore: upgrade dependencies

dferber90:
Updates dependencies.

No changes worth mentioning.


dferber90:
> Didn't know we were on storybook@4 already

We've been on the v4alpha for a really really long time. The upgrade was pretty easy for us because of that :)

PR #221, opened 2018-11-09
feat: add MoneyInput value formatting

### LINKED ISSUE #215: MoneyInput separators
There was a wish that our `MoneyInput` should use thousand-separators. We were not able to do this previously as we were using `<input type="number" />` to drive the `MoneyInput`.

Since we are now using `<input type="text" />`, we can add the thousand-separators.

Given that the user locale is `en` and the amount is `1000.34` then when `MoneyInput` gets blurred
it can format it to `1,000.34`.

Given that the user locale is `de` and the amount is `1000.34` then when `MoneyInput` gets blurred
it can format it to `1.000,34`.

However, we should not mess with the user-input while they type. Further, we should probably always parse `1000,34` and `1000.34` as 100034 cents. In other words:  When there is only one separator, we should not rely on the user's locale to parse. We should always treat it as the lowest amount.

dferber90:
Adds thousand-separator formatting to `MoneyInput`.

money-format

This is still missing tests and coordination with UX.

Makes it easier to read big amounts by adding thousand-separators.

Parsing
- Since most users are not careful about how they enter values, we will parse both `.` and `,`
- When a value is `1.000,00` we parse it as `1000`
- When a value is `1,000.00` we also parse it as `1000`
- This means the highest amount always wins
- We do this by checking comparing the last position of `.` and `,`. Whatever is later is used as the decimal separator

**Breaking change** The `MoneyInput.parseMoneyValue` now requires callers to pass a `locale` as the second argument, so that the value can be formatted according to the users locale initially.

Closes #215

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -118,10 +118,6 @@ const createCurrencySelectStyles = ({
 // }
 // which equals 0.0123456 €
 
-// Parses the value returned from <input type="text" />'s onChange function
-// The input will only call us with parseable values, or an empty string
-// The function will return NaN for the empty string.
-//
 // Parsing
 // Since most users are not careful about how they enter values, we will parse
 // both `.` and `,` as decimal separators.
@@ -131,15 +127,15 @@ const createCurrencySelectStyles = ({
 // This means the highest amount always wins. We do this by comparing the last
 // position of `.` and `,`. Whatever occurs later is used as the decimal
 // separator.
-export const parseAmount = amount => {
-  const lastDot = String(amount).lastIndexOf('.');
-  const lastComma = String(amount).lastIndexOf(',');
+export const parseRawAmountToNumber = rawAmount => {
+  const lastDot = String(rawAmount).lastIndexOf('.');
+  const lastComma = String(rawAmount).lastIndexOf(',');
 
   const separator = lastComma > lastDot ? ',' : '.';
   const throwaway = separator === '.' ? ',' : '\\.';
 
-  // The amount with only one sparator
-  const normalizedAmount = String(amount)
+  // The raw amount with only one sparator
+  const normalizedAmount = String(rawAmount)
     .replace(new RegExp(`${throwaway}`, 'g'), '')
     .replace(separator, '.');
 
@@ -157,15 +153,15 @@ export const parseAmount = amount => {
 //  - no currency was selected
 //
 // This function expects the "amount" to be a trimmed value.
-export const createMoneyValue = (currencyCode, amount) => {
+export const createMoneyValue = (currencyCode, rawAmount) => {
   if (!currencyCode) return null;
 
   const currency = currencies[currencyCode];
   if (!currency) return null;
 
-  if (amount.length === 0 || !isNumberish(amount)) return null;
+  if (rawAmount.length === 0 || !isNumberish(rawAmount)) return null;
 
-  const amountAsNumber = parseAmount(amount);
+  const amountAsNumber = parseRawAmountToNumber(rawAmount);
   if (isNaN(amountAsNumber)) return null;
 
   const centAmount = amountAsNumber * 10 ** currency.fractionDigits;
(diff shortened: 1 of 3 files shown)


src/components/inputs/money-input/money-input.js
```diff
@@ -117,10 +118,33 @@ const createCurrencySelectStyles = ({
...
+//
+// This means the highest amount always wins. We do this by comparing the last
+// position of `.` and `,`. Whatever occurs later is used as the decimal
+// separator.
+export const parseAmount = amount => {
+  const lastDot = String(amount).lastIndexOf('.');
+  const lastComma = String(amount).lastIndexOf(',');
+
+  const separator = lastComma > lastDot ? ',' : '.';
+  const throwaway = separator === '.' ? ',' : '\\.';
```
tdeekens: Is this logic implying that the last separator always wins?

src/components/inputs/money-input/money-input.js
```diff
@@ -223,9 +249,14 @@ export default class MoneyInput extends React.Component {
...
     );
 
-  static parseMoneyValue = moneyValue => {
+  static parseMoneyValue = (moneyValue, locale) => {
     if (!moneyValue) return { currencyCode: '', amount: '' };
 
+    invariant(
+      typeof locale === 'string',
+      'MoneyInput.parseMoneyValue: A locale must be passed as the second argument'
+    );
```
tdeekens: 💯

src/components/inputs/money-input/money-input.js
```diff
@@ -175,27 +199,29 @@ export const createMoneyValue = (currencyCode, amount) => {
...
+const getAmountAsNumberFromMoneyValue = money =>
+  money.type === 'highPrecision'
+    ? money.preciseAmount / 10 ** money.fractionDigits
+    : money.centAmount / 10 ** currencies[money.currencyCode].fractionDigits;
+
+// gets called with a string and should return a formatted string
+const formatAmount = (value, currencyCode, locale) => {
   // fallback in case the user didn't enter an amount yet (or it's invalid)
-  const moneyValue = createMoneyValue(currencyCode, amount) || {
+  const moneyValue = createMoneyValue(currencyCode, value) || {
```
tdeekens: Should we have an `locale` invariant here too?

dferber90: `formatAmount` is currently only called from places where either
- `locale` is already checked
- the instance method where the locale is available as prop-type checks warn if not

So right now the `invariant` prevents mistakes from `ui-kit` consumers, but we don't get warnings in case it's used wrong inside this file itself. Not sure if it's worth adding a check. I don't have a strong opinion.

Maybe those checks should ALL be moved to `formatAmount`? Then we'd always run them in either case. I'll move them 👍

src/components/inputs/money-input/money-input.js
tdeekens: No it's good. It helped just wanted to make sure.

src/components/inputs/money-input/money-input.js
dferber90: On second thought, this would spread the checks across different functions which is less nice. I think I like the current version better.

dferber90:
💭 All this converting values back and forth really makes me wish for TypeScript in ui-kit

adnasa:
side note: naive question, will MoneyInput deal with roundings?

image

dferber90:
@adnasa Good question (but unrelated to the changes in this PR). Right now we have the problem that when the user enters `2.49` into the input, the parent is called correctly (`{ amount: "2.49", currencyCode: "EUR" }`).

But when the parent then calls `MoneyInput.convertToMoneyValue` with the value we provided, the result will be off:`MoneyInput.convertToMoneyValue({ amount: "2.49", currencyCode: "EUR" })` returns 

```
{
  "centAmount": 249.00000000000003,
  "currencyCode": "EUR",
  "fractionDigits": 2,
  "type": "centPrecision"
}
```

We've had this problem for a while. I'm going to try and tackle it separately.

adnasa:
thanks for the reply. just wanted to raise awareness!
ping me when you're on this, I'm glad to help

dferber90:
I opened #229 which fixes this 👍

PR #231, opened 2018-11-14
Add accepting `theme` in FieldLabel's  hintIcon

tdeekens:
#### Summary

The `<FieldLabel />` used by all fields currently defaults/overwrites the `theme` to be orange in any case. We recently received designs requiring a custom theme (according to the icon's available themes). 

#### Description

Refactors the `<FieldLabel>` so that a custom `theme` can be passed without being overwritten.

```js
hintIcon={<WarningIcon theme="black" />}
```

will be accepted after this PR. We already have application level code passing (mostly black) which will just work after upgrading to this in the next version of UIKit.

### DIFF
--- a/src/components/field-label/README.md
+++ b/src/components/field-label/README.md
@@ -30,6 +30,23 @@ import { FieldLabel } from '@commercetools-frontend/ui-kit';
 />
 ```
 
+The `hintIcon` also accepts a custom `theme` while defaulting to `orange` in the case above.
+
+```diff
+<FieldLabel
+  title={<FormattedMessage {...messages.title} />}
+  hasRequiredIndicator={true}
+  onInfoButtonClick={() => {}} />}
+  hint={<FormattedMessage {...messages.hint} />}
+- hintIcon={<WarningIcon />}
++ hintIcon={<WarningIcon theme="green" />}
+  description={<FormattedMessage {...messages.description} />}
+  badge={<FlatButton tone="primary" label="show" />}
+  htmlFor="sampleInput"
+  horizontalConstraint="m"
+/>
+```
+
 #### Properties
 
 | Props                  | Type               | Required | Values                             | Default | Description                                                                                                                                                                                                                                                           |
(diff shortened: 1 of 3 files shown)


src/components/field-label/README.md
```diff
@@ -30,6 +30,23 @@ import { FieldLabel } from '@commercetools-frontend/ui-kit';
...
+The `hintIcon` also accepts a custom `theme` while defaulting to `orange` in the case above.
+
+```diff
+<FieldLabel
+  title={<FormattedMessage {...messages.title} />}
+  hasRequiredIndicator={true}
+  onInfoButtonClick={() => {}} />}
+  hint={<FormattedMessage {...messages.hint} />}
+- hintIcon={<WarningIcon />}
++ hintIcon={<WarningIcon theme="green" />}
```
tdeekens: Found the `diff` type of illustrate best what the difference is between both examples.

src/components/field-label/field-label.js
```diff
@@ -41,7 +41,7 @@ export const FieldLabel = props => (
               {// FIXME: add proper tone when tones are refactored
               React.cloneElement(props.hintIcon, {
                 size: 'medium',
-                theme: 'orange',
+                theme: props.hintIcon.props.theme || 'orange',
```
tdeekens: That's all we need.

src/components/field-label/field-label.spec.js
```diff
@@ -114,6 +114,21 @@ describe('rendering', () => {
...
+          props = createTestProps({
+            hint: 'foo',
+            hintIcon: <WarningIcon theme="green" />,
+          });
+          wrapper = shallow(<FieldLabel {...props} />);
+          hintIconWrapper = wrapper.find(WarningIcon);
+        });
+
+        it('should set the icon theme', () => {
+          expect(hintIconWrapper).toHaveProp('theme', 'green');
```
tdeekens: Just making sure the prop is forwarded so the change can't be reverted without a test failing.

PR #232, opened 2018-11-14
feat(notifications/static): add ContentNotification component

pa3:
#### Summary

Add in-place notification panel component.

screenshot 2018-11-14 at 16 08 26


src/components/notifications/static/README.md
```diff
@@ -0,0 +1,22 @@
+# MoneyInput
```
dferber90: 👀

src/components/notifications/static/__snapshots__/static.spec.js.snap
```diff
@@ -0,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dferber90: We're trying to get rid of `enzyme` and snapshot tests in ui-kit. Could you remove this one?

src/components/notifications/static/static.spec.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { render } from 'react-testing-library';
```
dferber90: Although `react-testing-library` works fine in this case, I'd welcome if we could use `test-utils` (from `src/test-utils`). This will prepare the tests to be expanded (e.g. for different locales). And it will set a good example of using `test-utils`.

src/components/notifications/static/static.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+    expect(container.querySelector("[data-foo='bar']")).toBeInTheDocument();
+  });
+
+  it.each(['error', 'info', 'success', 'warning'])(
+    'should render notification of type `%s`',
+    type => {
+      const { container } = render(<Static type={type}>A message</Static>);
+
+      expect(container).toMatchSnapshot();
```
dferber90: This test will break whenever we change the implementation because it asserts the whole DOM. This is pretty inflexible and what we're trying to get away from by using `react-testing-library`.

Can you write this test in another way? What is the actual thing being tested here?

src/components/notifications/static/static.story.js
```diff
@@ -0,0 +1,27 @@
...
+import Section from '../../../../.storybook/decorators/section';
+import Readme from './README.md';
+import Static from './static';
+
+storiesOf('Notifications', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withReadme(Readme))
+  .add('Static', () => (
+    <Section>
+      <Static type="warning">
```
dferber90: Most of our stories are set up in a way where the component is rendered only once. The different states can be interacted with using the knobs. This helps designers figure out which knobs (props) exist and to communicate better with developers.

Can you change the component to be rendered once only and to use a knob for the `type` attribute?

I know that it's not nice to see all our notifications at once, but that's not what we're aiming for in the storybook documentation. The goal would be to have a dedicated documentation website which shows off the components better later on.

src/components/notifications/static/static.spec.js
pa3: Don't like this kind of tests too much, but don't know what else can be tested for this kind of dumb presentational component.

src/index.js
```diff
@@ -4,51 +4,79 @@ import '../materials/internals/grid.mod.css';
 import '../materials/internals/reset.mod.css';
 
 /* eslint-disable max-len,prettier/prettier */
-import * as i18n from '../i18n'
+import * as i18n from '../i18n';
```
dferber90: @emmenko originally turned of prettier for this specific file, as indicated by the `/* eslint-disable max-len,prettier/prettier */` comment. This file is also part of `.pretterignore`.

The intention of turning prettier of was to keep the exports as one-liners which makes them a bit easier to read (subjective).

Seems like your editor did not respect this setting and that's where the diff is coming from.

To me the inconvenience of turning prettier on/off for this one file outweighs the improved readability. I would be okay with enabling prettier for this file, removing the `/* eslint-disable prettier/prettier */` comment and removing this file from `.prettierignore`.

WDYT @emmenko ?

src/components/notifications/index.js
```diff
@@ -0,0 +1,3 @@
+import Static from './static';
+
+export default { Static };
```
dferber90: I would rename the component to something like `ContentNotification` and expose it as that.

Our naming pattern has mostly settled on prefixing the more specific types: `TextInput`, `NumberInput`, .. instead of `Input.Text`, `Input.Number` etc.

I would like to eventually have this consistent across ui-kit. So we can use `ContentNotification`. Then we can later have `PageNotification` and so on in case we ever decide to move those into ui-kit.

src/components/notifications/static/static.spec.js
dferber90: Yeah, me neither. I would not test that at all (as stated in https://github.com/commercetools/ui-kit/pull/232#discussion_r233527322).

What we're missing are Visual Regression Tests. This is what we really would like to test here. I'm trying to introduce it to ui-kit before I leave but it requires some coordination because it comes with a price-tag (as we need to use a paid service for it). I've just completely left out any tests which attempt to test the visuals (like classnames) recently. I want Visual Regression Testing and I don't want to deal with writing tests that attempt to catch these visual regressions (using classnames) which actually won't catch it in most cases (e.g. because somebody edits the css itself). 

TLDR: I'd just remove the tests if their only purpose is to test the styling.

src/components/notifications/static/README.md
pa3: 🙈

src/components/notifications/static/static.spec.js
pa3: But, I would think about switching to CSS-in-JS approach and using snapshots again. That can be a nice substitute to the full-fledged VRT screenshot-based approach, as combination of DOM-nodes and styles kind of garanties specific appearance. Buy paying the price of more tight coupling to the implementation, we will get some obvious advantages: better performance, and getting rid off all fonts related issues.

dferber90: > and getting rid off all fonts related issues. 

We will experiment with Percy soon who have the font issues and other stuff figured out. They run super fast too because they run in parallel. I would not add snapshot tests in the meantime.

src/index.js
dferber90: I opened #236 which removes the exclusion.

src/components/notifications/content-notification/README.md
```diff
@@ -0,0 +1,22 @@
...
+  Something went wrong here!
+</ContentNotification>;
+```
+
+#### Properties
+
+| Props      | Type     | Required | Values                                | Default | Description                                       |
+| ---------- | -------- | :------: | ------------------------------------- | ------- | ------------------------------------------------- |
+| `children` | `node`   |    -     | -                                     | -       | Message to be shown inside a notification panel.  |
+| `type`     | `string` |    -     | `success`, `info`, `warning`, `error` | -       | Type of notification, determining its appearance. |
```
dferber90: @mariabarrena Can we include some guidelines (Dos & Don'ts) here on when to use `ContentNotification`s over `SideNotification`s ? What's your take on it? :)

src/components/notifications/content-notification/content-notification.js
```diff
@@ -0,0 +1,50 @@
...
+    theme: PropTypes.string.isRequired,
+  };
+
+  render() {
+    const Icon = do {
+      if (this.props.type === 'error') ErrorIcon;
+      else if (this.props.type === 'info') InfoIcon;
+      else if (this.props.type === 'warning') WarningIcon;
+      else SuccessIcon;
+    };
```
dferber90: 💯

src/components/notifications/content-notification/README.md
dferber90: The `type` seems to be a required prop, but it's stated as optional here.

dferber90: I fixed this, merged #246 and rebased this branch 👍

PR #242, opened 2018-11-19
Adds <Avatar> component

qmateub:
#### Summary

Adds Avatar component based on the old one in application kit

https://github.com/commercetools/merchant-center-application-kit/tree/master/packages/application-shell/src/components/avatar

The new version improves:
* 🔨 Rename `size` prop to `scale` to be consistent with other UIKit components
* 🍰 Adds new `m` scale (needed for team members page)
* 🍰 Remove usage of `is-retina` in favor of `src` & `srcSet` from `img` element (kudos to @dferber90 for suggesting that)
* 💅 Small improvements


src/components/avatar/avatar.story.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
+import withReadme from 'storybook-readme/with-readme';
+import Section from '../../../.storybook/decorators/section';
+import Avatar from './avatar';
+import Readme from './README.md';
+
+storiesOf('Images', module)
```
qmateub: Not sure if this is the best path 🤔

src/components/avatar/avatar.story.js
```diff
@@ -0,0 +1,39 @@
...
+    };
+
+    return (
+      <Section>
+        <div
+          style={{
+            width: wrapperScale,
+            height: wrapperScale,
+          }}
+        >
```
qmateub: This wrapper div is for the avatar not taking the whole space in the Section component 👍

src/components/avatar/__snapshots__/avatar.spec.js.snap
```diff
@@ -0,0 +1,69 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
montezume: Could you remove the snapshots? AFAIK we are trying to keep them out of ui-kit 🙈

src/components/avatar/avatar.spec.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
dferber90: We want to avoid adding more `enzyme` tests to `ui-kit` Can you rewrite those to `react-testing-library`?

You can drop the tests which are concerned with visual stuff (like dom snapshots or class assertions) as we will move to Visual Regression Testing for these things.

src/components/avatar/avatar.story.js
dferber90: Let's put it top-level if there is no obvious path for now. We will group the components in a later step.

qmateub:
@dferber90 @montezume 

All addressed guys 🎉 

* No snapshots Malcolm :trollface: 
* Use `react-testing-library`
* Rename to `size` prop
* Proper usage of `srcset` (thanks Dom 🙏 )

Let me know if you guys miss something

src/components/avatar/avatar.story.js
```diff
@@ -0,0 +1,39 @@
...
+      else '100px';
+    };
+
+    return (
+      <Section>
+        <div
+          style={{
+            width: wrapperSize,
+            height: wrapperSize,
+          }}
```
tdeekens: Can we use a className over inline styles here?

qmateub: I mean yeah, but is for the `story` so I was feeling bad about having to define a style for this 🤣

montezume: Just use react emotion?

qmateub: > I didn't take a look myself. Does it even need a wrapper? 💭

Yep, it's needed just for the storybook because if not takes the whole width space 😅

qmateub: All good and fixed in 7208e4d

src/components/avatar/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+#### Properties
+
+| Props           | Type     | Required | Values                | Default | Description                                      |
+| --------------- | -------- | :------: | --------------------- | ------- | ------------------------------------------------ |
+| `gravatarHash`  | `string` |    ✅    | -                     | -       | Hash of the image to be showed                   |
+| `firstName`     | `string` |          | -                     | ''      | First name of the user in case there is no image |
+| `lastName`      | `string` |          | -                     | ''      | Last name of the user in case there is no image  |
+| `size`          | `string` |    ✅    | oneOf(['s', 'm', 'l]) | `s`     | The size of the Avatar component.                |
+| `isHighlighted` | `bool`   |    -     | -                     | `false` | Highlights the Avatar                            |
```
dferber90: @qmateub Can you explain what "highlights" means? Is it a different look for a mouseover state? When would we use `isHighlighted`?

src/components/avatar/avatar.js
```diff
@@ -0,0 +1,98 @@
...
+  if (size === 'l') 100;
+  else if (size === 'm') 48;
+  else 26;
+};
+
+const GravatarImg = props => (
+  <img
+    className={styles['gravatar-img']}
+    src={createGravatarImgUrl(props.hash, getSizeInPx(props.size))}
+    srcSet={oneLineTrim`
```
dferber90: `srcset` is usually spelled in lowercase. I think we should use lowercase as well, unless this is a React thing (which I don't think it is).

src/components/avatar/README.md
dferber90: Seems like it is indeed used for the hover effect. In that case I would rename it to `isHovered`. I guess the use-case for it is that we want to show the hovered version when something like a user card is hovered. So the element itself isn't hovered directly but it still needs to look as-if it was hovered?

We can also document that no hover styles will be applied unless this property is passed in.

src/components/avatar/avatar.mod.css
```diff
@@ -0,0 +1,53 @@
+:root {
+  --size-s: 26px;
+  --size-m: 48px;
+  --size-l: 100px;
```
dferber90: This is note for designers... Usually our design system should provide us with these sizes. We should not define "arbitrary" sizes for specific components.

I'm okay with it for now, as we can work on this incrementally. However I want us to be aware that what we're doing here is not the way it's supposed to be done.

cc @commercetools/design-team

src/components/avatar/avatar.js
dferber90:
```suggestion
    srcset={oneLineTrim`
```

PR #243, opened 2018-11-20
feat: add <ConnectorIcon>

NickDevG:
#### Summary
Adds new icon 

image


tdeekens:
We should likely start building the navigation we need to build in the app with other filler icons and then just replace them once UIKit was released just to unblock us.

emmenko:
image

This can be named e.g. `InvertedPagesIcon`, since we already have `PagesIcon`

image


image

This can be named e.g. `ConnectorIcon`, or whatever the name of this symbol is

dferber90:
Thanks for adding these and thanks for the name suggestions @emmenko!

Regarding `InvertedPagesIcon`: `Inverted` is usually used when the colors are inverted. Here the colors are the same, but the icon is mirrored/flipped. So I would propose either `FlippedPagesIcon` or `MirroredPagesIcon`.

I also wonder what the use-case for having the mirrored icon is.

NickDevG:
I see your point @dferber90, if @emmenko agrees I'll rename it as `MirroredPagesIcon`

dferber90:
> I also wonder what the use-case for having the mirrored icon is.

@NickDevG Do you know why we need a mirrored version of this icon?

NickDevG:
I just asked design about this. We can use the `PagesIcon` for the nav bar. It seems that they forgot to check the UIKit icons for this

PR #245, opened 2018-11-22
Date Inputs WIP

dferber90:
This is a WIP branch. I just want a deploy link. The PR is probably going to be split up completely different in the end. Safe yourself the trouble and don't review yet :)

This is missing tests and it's not possible to type dates at the moment. This PR also uses names like `DateCalendar` instead of `DateInput` as I didn't want to replace the old ones while developing.


dferber90:
I've opened #281  and #282 which add `DateInput` and `DateTimeInput` respectively. `DateRangeInput` will follow afterwards.

I'll close this PR as it's not needed anymore.

PR #260, opened 2018-11-26
refactor(icons): rename some icons, drop Icon suffix from export name

### LINKED ISSUE #259: Invalid icons names
**Describe the bug**
After fixing the naming for the latest new icons (#251), I noticed that most of them have invalid naming.

**Current behavior**
There are some icons that are named after the context where they are being used in the MC.

**Expected behavior**
Icons should be named according to what the icons itself represents, not the context.

---

If we decide to rename them, we should document properly which icons got renamed to what, to make it simple to migrate.

See [proposal naming down below](https://github.com/commercetools/ui-kit/issues/259#issuecomment-441624081)

emmenko:
Fixes #259 

- renamed icons [listed here](https://github.com/commercetools/ui-kit/issues/259#issuecomment-441624081)
- output generated exports to a `exports.js`
- make the `index.js` export a `default` object containing all the icons instead of doing it in the package entry point (so that we can also internally use this kind of import)
- dropped `Icon` suffix from name
  - only components `displayName` still have the `Icon` suffix, for better readability

image


emmenko:
Yeah, for me this is more of a bugifx that we should address now. I also didn't want to bring up other topics like grouping to avoid going off topic.
Grouping is also something important for sure but we should plan it properly.

src/components/buttons/flat-button/flat-button.spec.js
```diff
@@ -1,14 +1,14 @@
...
-import { AddIcon } from '../../icons';
+import Icons from '../../icons';
 import { FlatButton } from './flat-button';
 
 const createTestProps = props => ({
   type: 'primary',
   label: 'Accessibility text',
   onClick: jest.fn(),
-  icon: <AddIcon size="medium" />,
+  icon: <Icons.Add size="medium" />,
```
dferber90: I was following along on the original issue. I was behind the motion to rename the icons to more meaningful names (name them after what they are, not what they are used for). At some point scope of the issue seems to have been widened to also changing the naming scheme. It's been only 9 hours since the issue was opened, so here is my feedback even though you already did the work.

This naming scheme is the opposite of the direction the rest of our components are going in currently.

We have `TextInput` instead of `Inputs.Text`. We have `FlatButton` instead of `Buttons.Flat`. We have `CollapsiblePanel` instead of `Panels.Collapsible`. We have `FieldLabel` instead of `Labels.Field`. We have `ErrorMessage` instead of `Messages.Error`. We have `ContentNotification` instead of `Notifications.Content`.

So far all icons were fitting our naming pattern - now they are an exception (along with some other exceptions like `Spacings` and `Text` we still have).

I further dislike our `displayName` no longer matching the component name which adds a gotcha when looking at the React tree in devtools.

People are already familiar with it. If we merge this, everyone has to learn about the change.

We currently export `Icons` so that consumers can do `import { Icons } from '@ct/ui-kit'`. But I'd much rather enable consumers to just pick any component straight from the export (so `import { ArrowLeftIcon }` from `ui-kit`).

I also believe (but did not verify) that this might break treeshaking. Using separate exports at the top-level certainly sets us up for a path where we can use treeshaking, while having a "packed" single `Icons` export might not make this as easy/actually prevent treeshaking.

I don't see an advantage of changing the naming format but I do see the downsides listed above. For these reasons I'd rather keep the naming scheme as it was.

tdeekens:
Yep. I hope it is all fine and no boundaries were crossed just wanted to raise the thought. To me renaming (for sure) and re-naming how we export on Icons (very likey) make sense. Maybe there are diverging plans cause we didn‘t wait for feedback long and I hope nobody feels overlooked. I am saying this cause we all know that engineering is as much a social experiment than it is solving technical problems. Any way, good night and lets wait for other people‘s feedback.

src/components/buttons/flat-button/flat-button.spec.js
emmenko: > Maybe that is what @emmenko set out to do initially to change the names only

Yes I only wanted to fix the names. According to the feedback though it seemed that the `Icon` suffix was an annoying thing, so I went ahead and removed it. We can easily put it back though, it's not a problem.

src/components/icons/exports.js
```diff
@@ -0,0 +1,314 @@
...
+export const Truck = withThemeProp(withSizeProp(OrigTruck));
+Truck.displayName = 'TruckIcon';
+export const UserFilled = withThemeProp(withSizeProp(OrigUserFilled));
+UserFilled.displayName = 'UserFilledIcon';
+export const UserLinear = withThemeProp(withSizeProp(OrigUserLinear));
+UserLinear.displayName = 'UserLinearIcon';
+export const Verified = withThemeProp(withSizeProp(OrigVerified));
+Verified.displayName = 'VerifiedIcon';
+export const Warning = withThemeProp(withSizeProp(OrigWarning));
+Warning.displayName = 'WarningIcon';
```
tdeekens: Display names do not change

src/components/icons/exports.js
```diff
@@ -0,0 +1,314 @@
...
+export const World = withThemeProp(withSizeProp(OrigWorld));
```
tdeekens: Export names do change or I am lost: `import { WorldIcon } from ` won‘t work anymore.

src/components/buttons/flat-button/flat-button.spec.js
dferber90: > I'm not sure why you think this changed now.

The screenshot on this PR has the `Icon` suffix removed. And as this diff shows, the way to use an icon outside of the ui-kit now differs from the way to use it inside of ui-kit. I dislike this divergence.

There is also a `// TODO: this type of export is deprecated and should be removed in the next major release` comment from you which states that we want to drop the plain exports in favour of `Icons.X`.

There are only 9 places in 6 files doing `import { Icons } from '@ct/ui-kit`. Contrary, there are more than 300 results for `<XIcon />`. So we might as well drop `Icons.X` and refactor in an upcoming release. It doesn't seem worth the confusion it added.

> We can easily put it back though, it's not a problem.

Yeah, I would appreciate that 👍

tdeekens: I guess we moved to quickly from an idea (which raised the concern already) to just changing it. 

The naming in the Icons.XY changed giving the impression (also by the export cited on the issue) that this would be the preferred way of consuming icons. Which then leads to the inconsistency to fields, buttons or inputs not even having this kind of way of importing them. 

Maybe, and that‘s what I meant with common denominator, we should decide:

1. Do we actually want this Icons.XY export given we are not too sure what it means for tree shaking and it being inconsistent with other component groups
2. Should display names and component names actually diverge as they do here 

Given the review comment below I got a bit more worried about export actually being changed meaning this being a breaking change.

tdeekens:
This is GitHub telling me to go to bed. I don‘t want to know how many hours I sat infront of a screen today. 🤖

src/components/buttons/flat-button/flat-button.spec.js
emmenko: > The screenshot on this PR has the Icon suffix removed. 

I'm talking about the exports, not the names. You were all complaining about treeshaking being broken, but the changes here do not change anything in those regards.

src/components/icons/exports.js
emmenko: I'm not sure what you're confused about here. This was the all point of having the `Icon` suffix removed (which I only did because people were asking for it).

src/components/buttons/flat-button/flat-button.spec.js
tdeekens: Sorry, GitHub had issues and seems to have lost a long comment of mine while previously adding my last request for changes 10x hence the last comment. 

I think we just got distracted on the issue. 

1. Dicussion about namings (all good)
2. Discussing removing the Icon sufffix
2.1. Fear of inconsistent names
2.2. Suggestion to prefix the icon as a module Icon.XY
2.3. Fear of inconsistency 
2.4. Notice that the Icon export already existed
3. Implementation in proposed PR
3.1. Fear of previously existing and now recommended export breaking tree shaking

We are mixing most of those things

1. Tree shaking potentially is and was broken no matter if it is Icons.XY or Icons.XYIcon. Given that it would we should not offer the export as svgs are big
2. Display names did not change but the component name (even though generated) leading to divergence in DevTools
3. Generally different (but existing just not known) way of exporting icons through Icons.
4. The facts that this likely breaks being able to import as { XYIcon }

PR #261, opened 2018-11-26
chore(storybook): fix page title and favicon

### LINKED ISSUE #244: Update website page title
The current page title of the website shows

image

We should use the ct favicon and name the page properly (e.g. `UI Component library @commercetools`)

emmenko:
Fixes #244 

image


tdeekens:
Cool! Did we decide on the title? Does the Design Team have a preference?

emmenko:
I don't know, anything better than `Storybook` 😅

.storybook/manager-head.html
```diff
@@ -0,0 +1 @@
+<script>document.title = "UI component library - commercetools";</script>
```
dferber90: I would call it UI Kit to avoid introducing many terms which has shown to be confusing.

.storybook/manager-head.html
```diff
@@ -0,0 +1 @@
+<script>document.title = "UI Kit;</script>
```
dferber90: @emmenko A syntax error was introduces here after it got approved 🤓

PR #262, opened 2018-11-26
refactor(icons): invalid names

### LINKED ISSUE #259: Invalid icons names

emmenko:
Fixes #259 
Closes #260 

Renamed icons [listed here](https://github.com/commercetools/ui-kit/issues/259#issuecomment-441624081)

image


src/components/icons/svg/angle-thin-left.react.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: sketchtool 45.2 (43514) - http://www.bohemiancoding.com/sketch -->
+    <title>arrow-left</title>
```
emmenko: PS: the names here are still "wrong". Should we update them as well?

src/components/icons/svg/angle-thin-left.react.svg
```diff
@@ -0,0 +1,16 @@
...
+    <desc>Created with sketchtool.</desc>
+    <defs></defs>
+    <g id="Icons" stroke="none" stroke-width="1" fill-rule="evenodd">
+        <g id="MC-icon-set" transform="translate(-24.000000, -936.000000)" fill="#000000">
+            <g id="Directions" transform="translate(24.000000, 888.000000)">
+                <g id="Arrow-Left" transform="translate(0.000000, 48.000000)">
```
emmenko: Same here with the `id`

src/components/icons/svg/angle-thin-left.react.svg
tdeekens: Uff. Good one. No opinion 😐

emmenko: I went ahead and fixed the names: https://github.com/commercetools/ui-kit/pull/262/commits/f38e0124a19214bd2a286ea10214473406938ac6

I also noticed some inconsistency in the SVG structures. I already checked and notified the design team about it.

PR #267, opened 2018-11-28
Add percy

dferber90:
Adds percy

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -24,10 +24,14 @@ plugins:
   - 'prettier'
   - 'prefer-object-spread'
   - 'react'
+  - 'react-percy'
 overrides:
   - files: ['*.spec.js']
     rules:
       'react/display-name': 0
+  - files: ['*.percy.js']
+    env:
+      'react-percy/globals': true
 settings:
   import/ignore:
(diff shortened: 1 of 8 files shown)


dferber90:
@tdeekens We already have a paid account at the moment. It's cancellable any time. We've set it up two weeks ago to be able to take Percy for a test run (as our trial had already expired).

emmenko:
Who set that up? Just asking to know who the reference person is

dferber90:
@svenmueller set it up. We are currently on a tiny plan limited to 5 people. 3 of them are already taken (by @montezume, me and Sven). I'm thinking about reserving the other 2 seats for designers. Not sure how to best split them. We first want to prove Percy is worth it before moving to a bigger plan. The next plan with more seats is $400 per month for 10 users. Right now we're at $99 per month for 5.

Sven can eventually leave the org in Percy and give up his seat. But as the billing is tied to his account, we don't want to risk breaking anything until we actually got Percy up and running.

tdeekens:
> Sven can eventually leave the org in Percy and give up his seat

We can also work together and collaborate during testing. Same we do with LD.

emmenko:
@dferber90 got it, thanks for the info

PR #268, opened 2018-11-28
chore: add avatar export

montezume:
Looks like we forgot to include this export 🙈


montezume:
@dferber90 more tests running against the built bundle? 🤷‍♂️

qmateub:
@dferber90 @montezume 

Shouldn't this be my fault or part of the review? I mean, if I add the component I have to make sure that is available for uikit consumers right?

Happens the same with sync-action when you forget the main export 😓

PR #270, opened 2018-11-28
Remove `Icons.AX` export

tdeekens:
#### Summary

This pull request proposes to remove the `Icons.*` export.

#### Description

The export has been creating great confusion recently and I suggest to remove it for the following reasons:

1. It breaks tree shaking as all icons are imported (in case of SVG causing some unexpected bundle weight to consumers)
2. It's convenient but not really needed in most cases
   - Currently a mapping could be implemented by switching on strings and returning a specifically imported icon

Future work could entail adding a generic component to UIKit which looks like `<Icon name='circle' />` which does the mapping. However that will also break tree shaking for consumers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,6 @@
 import '../materials/internals/grid.mod.css';
 import '../materials/internals/reset.mod.css';
 import * as i18n from '../i18n';
-import * as Icons from './components/icons';
 
 export {
   default as AccessibleButton,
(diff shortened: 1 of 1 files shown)


tdeekens:
> Can we move away from import { Icons } from '@ct/ui-kit' internally (in case there are some cases left) 

I couldn't find any cases. The storybook uses a `import * as icons from './icons'` or something to be able to create element on them dynamically by name.

dferber90:
Nice! Seems good to go.

App-kit will require a migration, but @emmenko is already addressing that.

tdeekens:
Do I need another review? /cc @emmenko

tdeekens:
Got confirmation from @dferber90 that merging is fine. Thanks.

PR #271, opened 2018-11-28
test: add visual regression tests for buttons & avatar

dferber90:
Adds Visual Regression Tests for Avatar & Buttons.

Check out https://percy.io/commercetools-GmbH/ui-kit/builds/1204720 for the diff.

---

## Report

I got a better feeling of Percy now and it's pretty disappointing. Here are the major downsides I noticed.

#### Description

Even though I just started adding some components, we're running out of snapshots pretty quickly. Right now our plan allows up to 5k snapshots per month. We are already at ~750 snapshots today. This means we'd run out within a week (and that's a conservative assumption since I added the tests at the end of the day, not the beginning). Our plan currently only allows 5k snapshots per month, and the next bigger option is 10k. After that it gets pretty expensive. Both are likely to be exceeded really quickly.

#### Possible solution

To counteract this, we could decide to only take snapshots of components in their really primitive states instead of testing many variations (as we're doing now). This obviously lowers the value added by having Visual Regression Tests in the first place.

We were running snapshot tests for 2 different viewports and for 2 different Browsers (Chrome & Firefox). I believe this means taking one screenshot increases the screenshot count by 4 (1 screenshot x 2 browsers x 2 viewports). I already disabled Firefox after making this PR. I couldn't figure out a way to set the viewports globally (it is possible to set them per test though).

#### Description

A test run already takes ~13 minutes for ~250 snapshots. I guess it will only get worse while we add more tests. We would like to add visual regression tests to our inputs and fields as well. Let's say we end up with 1000 snapshots which Percy needs to process per test run. This means it would take almost an hour to complete the visual regression tests. And it means we'd run out of our 5k limit after 5 commits.

#### Possible solution

Basically the same as above. We need to reduce the amount of snapshots we take (by testing less variants; only one viewport; only one browser) in order to make the test runs fast. But even then we'd end up with a test duration of about 15 minutes based on my calculations (which might be totally wrong).

#### Description

The snapshots currently still have issues with the font as I haven't found a way to inject our global styles into Percy yet. Our components also render without the UI Kit context (no router). This leads to errors sometimes (e.g. when the component tries to render a `Link` or use `withRouter`).

#### Possible solution

I'm pretty sure we could write a wrapping component which injects the styles and renders a containing router. It's another thing to maintain though, as we already do almost the exact same thing for `render` from `test-utils`.

#### Description

And finally, it's not possible to interact with the component before taking a screenshot like we can when using `react-testing-library` (or the VRT style I "pioneered"). This means we can't e.g. open a date picker by clicking on it before taking a screenshot (at least not with the react integration). We can define the props the component should use when rendering, but that's it.

#### Possible Solution

I don't think it's possible to do anything about this at the moment.

#### Description

Right now Percy uses its own webpack configuration to run the Visual Regression Tests (see [`percy.config.js`](https://github.com/commercetools/ui-kit/blob/fe7b07dc741674a1cce54886974dedb96b91fb35/percy.config.js)). This means Percy can't notify us in case we screw up our build process (the rollup config) or Storybook (the webpack config) - as they are three separate configurations.

#### Possible Solution

I don't think it's possible to do anything about this at the moment. There is an integration for Percy with Storybook which we could try instead. However that integration seems to only take screenshots of the initial configuration of every story. This would mean one less configuration, but it's still not a real solution.

## Summary

I feel like Percy was more intended for complete static sites rather then testing individual UI components as the low amount of supported snapshots and the missing ability to interact with a component before taking a snapshot indicate :/ Maybe it's worth for me to pursue my initial prototype  after all (personally; it's probably not a good idea for the ui-kit yet as it's a huge distraction)  💭 

## Possible Actions

We can go one of two ways

### Pursue separate `percy.js` files

- We should reduce the number of snapshots we take (maybe only the regular state and a disabled state)
- We would need to add a context provider component (like `render` in `test-utils`) which every test needs to use to set up the global styles and things like `react-intl` and `react-router`

### Using the storybook integration

We could also move to the Storybook integration (if it allows testing multiple states without requiring an explicit story for them). Then we'd naturally reduce the amount of snapshots we take (as we don't have that many stories anyways). This would take care of the global styles (probably) and of the context setup. I'm not sure whether it's possible to test additional states (e.g. the disabled states) without introducing stories for them.

## Tomorrow

I'll try get a feeling for the storybook integration tomorrow and keep you posted. See #272.


test/percy/spec.js
```diff
@@ -0,0 +1,18 @@
...
+);
+
+Spec.propTypes = {
+  label: PropTypes.string.isRequired,
+  children: PropTypes.node,
+};
+
+Spec.displayName = 'Spec';
+
+export default Spec;
```
tdeekens: This is nice.

test/percy/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+We import the components from bundle built with rollup, to ensure that we are testing the **production** version.
+
+```js
+import { PrimaryButton } from '../../../dist/ui-kit.esm';
+import { Suite, Spec, screenshot } from '../../../test/percy';
+
+screenshot('PrimaryButton', () => (
+  <Suite>
+    <Case label="when disabled">
```
tdeekens: It's `Spec`, right?

dferber90: Done 👍

dferber90:
Huge kudos to @montezume for suggesting to put multiple states of one component into one screenshot. This improves runtime dramatically and also helps us stay under the otherwise rather tight limit of visual diffs per month. 👏 👏 👏

dferber90:
Huge kudos to @montezume for suggesting to put multiple states of one component into one screenshot. This improves runtime dramatically and also helps us stay under the otherwise rather tight limit of visual diffs per month. 👏 👏 👏

test/percy/README.md
```diff
@@ -1,24 +1,82 @@
...
+
+When writing visual regression tests we need to ensure that we stay within our monthly budget of visual diffs on Percy. Every time a screenshot is taken for a specific browser and a specific viewport it counts as one _visual diff_. Another factor to consider is that we want to ensure that Percy completes in a timely manner. Taking one snapshot takes around 3 seconds. If we take one snapshot for every state of every component, then it quickly adds up to over 45 minutes to complete a single test run. We therefore render one component in many states in a single snapshot. We also limit ourselves to one browser and one viewport size to reduce the number of visual diffs created per test run.
+
+### Why we can't interact with components before taking snapshots
+
+At the time of writing we can not interact with components before taking a screenshot. The only way to manipulate a component is to pass props. Interaction is not possible: for example, we can't open a dropdown by clicking on it before taking a screenshot. This is accepted as a downside for now. It doesn't seem like any Visual Regression Testing service is offering this capability yet (without falling back to Cypress).
+
```
tdeekens: This is super awesome.

PR #272, opened 2018-11-28
add percy storybook integration

dferber90:
This is an alternative to #271. It adds the Percy Storybook integration.

79 snapshots were created in ~3m30s (using one viewport and one browser). This doesn't account for the time it took to generate storybook though, which takes around 45s on my machine, totalling to 4m15s which is still okay.

It doesn't seem possible to test anything other than the stories itself (e.g. disabled/warning/error states). Everything we want a snapshot of would require its own story, which is pretty impractical as we also share the storybook with the ui-kit users as a reference. Too many stories would be overwhelming.

We could use a hybrid approach of testing the stories and manually writing tests (as in #271). But that kind of beats the purpose of the Storybook integration as we could just use `react-percy` for everything (as done in #271). It would result in two places to maintain instead of one.

Check out https://percy.io/commercetools-GmbH/ui-kit/builds/1204803 for the diff

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,11 +1,28 @@
-import { configure, setAddon, addDecorator } from '@storybook/react';
+import {
+  configure,
+  setAddon,
+  addDecorator,
+  getStorybook,
+} from '@storybook/react';
+import createPercyAddon from '@percy-io/percy-storybook';
 import { withOptions } from '@storybook/addon-options';
 import IntlDecorator from './decorators/intl';
-// import infoAddon from '@storybook/addon-info';
-
-// setAddon(infoAddon);
-
 import './main.mod.css';
+
+const { percyAddon, serializeStories } = createPercyAddon();
+setAddon(percyAddon);
+
+const srcStories = require.context('../src', true, /\.story\.js$/);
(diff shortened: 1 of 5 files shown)


.storybook/config.js
```diff
@@ -1,11 +1,28 @@
...
+import './main.mod.css';
 
-// setAddon(infoAddon);
+const { percyAddon, serializeStories } = createPercyAddon();
+setAddon(percyAddon);
 
-import './main.mod.css';
+const srcStories = require.context('../src', true, /\.story\.js$/);
+const materialsStories = require.context('../materials', true, /\.story\.js$/);
+const exampleStories = require.context('../examples', true, /\.story\.js$/);
```
tdeekens: Just wondering: example stories contain something from a src story and additional example things we do not really care about too much?

dferber90: Those are these stories which serve as an "overview" and where the code itself contains technical pointers:

image.

We had them all along, they're not related to the Visual Regression Testing (but there is no way to exclude stories afaik).

dferber90:
I spoke to @montezume and we agreed to continue with dedicated tests as proposed in #271.

PR #281, opened 2018-12-03
Redo DateInput using downshift

dferber90:
This is the new  `DateInput` based on  `downshift`.

| Before | After |
| ----- | ---- |
| image | image |

I will add follow-up PRs to replace `DateTimeInput` and to add `DateRangeInput` based on this one.

**Breaking Change** Previously `onChange` would be called with the value directly. Now it gets called with an event containing the value. This was done to be in line with how the other inputs work.


src/components/inputs/date-input/README.md
```diff
@@ -16,24 +16,21 @@ multiple date selection. It formats the selected date depending on the current l
...
+| Props                  | Type     | Required | Values                             | Default | Description                                                                                                                               |
+| ---------------------- | -------- | :------: | ---------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`                   | `string` |    -     | -                                  | -       | Used as the HTML `id` attribute.                                                                                                          |
+| `name`                 | `string` |    -     | -                                  | -       | Used as the HTML `id` attribute.                                                                                                          |
```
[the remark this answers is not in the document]
dferber90: Done 👍

src/components/inputs/date-input/date-input.js
```diff
@@ -1,357 +1,271 @@
-/*
-  TECHDEBT:
-  - Support timeZone in all cases, but only when timeZone is passed.
-*/
-
-import 'flatpickr/dist/themes/airbnb.css';
 import React from 'react';
 import PropTypes from 'prop-types';
-import requiredIf from 'react-required-if';
+import Downshift from 'downshift';
```
dferber90: Note that we're not importing `moment` into this component at all. All usage of `moment` is moved to the utils. This allows to use the `date-input` (and the other ones) with any date-library (like [`date-fns`](https://github.com/date-fns/date-fns)).

src/components/internals/calendar-day/calendar-day.js
```diff
@@ -0,0 +1,54 @@
...
+import styles from './calendar-day.mod.css';
+
+const DateCalendarDay = props => (
+  <li className={styles.wrapper}>
+    <div
+      className={classname(
+        do {
+          if (props.type === 'heading') styles.heading;
+          else if (props.type === 'spacing') styles.spacing;
+          else styles.day;
```
tdeekens: You're doing this so only one class can be applied, correct? Using the object `classnames` notion would yield something else?

src/components/internals/calendar-day/calendar-day.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import omit from 'lodash.omit';
+import classname from 'classnames';
```
tdeekens: `import classname_s_`?

src/components/internals/calendar-day/calendar-day.mod.css
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.spacing {
+  composes: base;
+}
+
+/* States */
+
+.highlighted {
+  background-color: #e6e6e6;
```
tdeekens: Just curious towards the long term vision of these colors? Are they gonna be a token?

src/components/internals/calendar-header/calendar-header.mod.css
```diff
@@ -0,0 +1,24 @@
...
+.month > * {
+  margin: 0 4px 0 0;
+}
+
+.year {
+  display: flex;
+}
+
+.year > * {
+  margin: 0 0 0 4px;
```
tdeekens: I am such a party pooper! Why not 6px?

src/components/internals/calendar-menu/calendar-menu.mod.css
```diff
@@ -0,0 +1,27 @@
...
+  border-color: var(--token-border-color-input-warning);
+}
+
+.menuBase.error {
+  border-color: var(--token-border-color-input-error);
+}
+
+.menu {
+  composes: menuBase;
+  padding-bottom: 10px;
```
tdeekens: Is this a sum of 4 and 6 on the `.container`? Should we have variables and a `calc` here?

src/components/internals/calendar-day/calendar-day.js
dferber90: Exactly, with `classnames` we get multiple things applied. I found this easier to read.

src/components/internals/calendar-day/calendar-day.mod.css
dferber90: Ah thanks. These should use tokens already. I skipped the tokens during implementation with the intention to add them before opening the PR... which I forgot about in excitement :P

dferber90: I replaced the plain values with tokens now. Thanks again for the reminder, I had already done it in other files but must have skipped this one.

src/components/internals/calendar-header/calendar-header.mod.css
dferber90: Both 4px and 6px are part of our design system. Why would you prefer 6px?

src/components/internals/calendar-menu/calendar-menu.mod.css
dferber90: It's not the sum of anything to my knowledge. We don't need to have a design token for everything, it's okay if edge-cases use their own values. A date-calendar is a one-off component imho. There is no token for a spacing of 10 either, so I hardcoded it.

PR #283, opened 2018-12-03
Update library `emotion` to v10

montezume:
#### Summary

[emotion](https://emotion.sh) has a new version out, v10, with breaking changes. In order to keep using this library for their `keyframes` functionality, and to stay on the bleeding edge 🔪, some changes were required to the `CollapsibleMotion` component.

#### Approach

No breaking changes.


src/components/collapsible-motion/collapsible-motion.story.js
```diff
@@ -1,6 +1,8 @@
+/** @jsx jsx */
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs, number, boolean } from '@storybook/addon-knobs';
+import { jsx, css } from '@emotion/core';
```
montezume: feels weird to have to do this, since CollapsibleMotion is supposed to be part of our API. I'm not sure how else to do this though. There must be a way 🤔

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -117,18 +107,30 @@ class CollapsibleMotion extends React.PureComponent {
...
-              this.props.children({
-                isOpen,
-                toggle: animationToggle,
-                containerStyles,
-                registerContentNode,
-              })
-            }
+            {({ animation, toggle: animationToggle, registerContentNode }) => (
+              <ClassNames>
+                {({ css }) => {
```
montezume: this is necessary to expose a className as a string to another component.

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -117,18 +107,40 @@ class CollapsibleMotion extends React.PureComponent {
...
+                  let animationClass;
+
+                  if (animation) {
+                    animationClass = css`
+                      animation: ${animation} 200ms forwards;
+                    `;
+                  }
+
+                  const containerClassName = css`
+                    height: ${do {
```
dferber90: I'd use a ternaries here.

src/components/collapsible-motion/collapsible-motion.spec.js
```diff
@@ -29,7 +29,7 @@ describe('uncontrolled mode', () => {
       expect.objectContaining({
         isOpen: true,
         // no animation here because the panel is already expanded
-        containerStyles: { height: 'auto' },
+        containerClassName: expect.any(String),
```
dferber90: The comment doesn't match the expectation anymore.

package.json
```diff
@@ -62,7 +62,7 @@
     "classnames": "2.2.6",
     "common-tags": "1.8.0",
     "dom-helpers": "3.4.0",
-    "emotion": "9.2.12",
+    "@emotion/core": "10.0.1",
```
emmenko: Shouldn't we also install `@emotion/styled`?

https://emotion.sh/docs/install#with-styled

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -117,18 +107,34 @@ class CollapsibleMotion extends React.PureComponent {
...
+                  }
+
+                  const containerClassName = css`
+                    height: ${isOpen ? 'auto' : '0'};
+                    overflow: ${isOpen ? 'hidden' : 'inherit'};
+                  `;
+
+                  return this.props.children({
+                    isOpen,
+                    containerClassName: cx(
```
emmenko: Is the breaking change really necessary? Is there a way to be backward compatible?

montezume: It would be possible to access the animation name through emotion, but it's not really how they recommend using the library. We could do something like this

```js
 const class = keyframes`
    0% { height: 0; overflow: hidden; }
    99% { height: 90px; overflow: hidden; }
    100% { height: auto; overflow: visible; }
  `;

const animationName = class.name;
const styles = {
   animation: `${class.name} 200ms forwards`
}

```

But if they change the implementation to longer expose this name attribute, I don't know if they would consider this a breaking change or not. I would guess no.

montezume: This being the `console.log` of the keyframes response

screen shot 2018-12-04 at 12 07 56 pm

As you can see, the name attribute seems to match the animation name. But would changing the animation name to not match the name attribute be a breaking change? 🤷‍♂️

montezume: Hm, then we could look into having a `next` branch and merge breaking changes into that? We could continue releases features and fixes from master, and then merge them back into next?

tdeekens: I am like being explicit about breaking changes and just making it a quick and short pain point. It feels (to my current understanding) that we can't really work a way around this again. So it'd rather accept it rather than working around it and postponing it. Am I missing something in that?

montezume:
Okay I updated this removing the breaking change @emmenko @dferber90

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -118,17 +117,37 @@ class CollapsibleMotion extends React.PureComponent {
...
-              })
-            }
+            }) => (
+              <ClassNames>
+                {({ css }) => {
+                  let animationStyle = {};
+
+                  if (animation) {
+                    // this is required so that the keyframes animation is actually injected into the DOM
+                    css`
```
montezume: as comment says, without this we don't inject anything into the dom re: keyframes

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -26,25 +26,22 @@ export class ToggleAnimation extends React.Component {
...
+    const containerStyles = props.isOpen
       ? { height: 'auto' }
       : { height: 0, overflow: 'hidden' };
 
     if (props.isOpen !== state.isOpen) {
-      containerStyles = {
-        ...containerStyles,
-        animation,
-      };
+      animation = props.isOpen
```
dferber90: We could make `animation` a `const` now that we can determine it by a ternary.

src/components/collapsible-motion/collapsible-motion.js
dferber90: Could you explain what "this" refers to in the comment? The `css` tagged template literal? Maybe you could clarify the comment.

src/components/collapsible-motion/collapsible-motion.js
montezume: it would be a nested ternary though?

```js

    const animation =
      props.isOpen !== state.isOpen
        ? props.isOpen
          ? createOpeningAnimation(state.fullHeight)
          : createClosingAnimation(state.fullHeight)
        : null;

```

src/components/collapsible-motion/collapsible-motion.js
montezume: Sure, calling css here returns a className, and injects the styles into the head. Without this, we don't end up with the keyframes in the document and the animation doesn't work. I think we have to do this weird workaround because generally you don't write components using emotion like this.

montezume:
@dferber90 what do you think about adding another property `containerClassName`, so that when we  deprecate using `CollapsibleMotion` with styles? That way we can remove a bit of the weirdness surrounding having to call `css` but not actually use it.

src/components/collapsible-motion/collapsible-motion.js
tdeekens: ...or a do expresso

dferber90:
@montezume I'd like to play with this PR tomorrow to get an understanding of what's going on. Right now I only have the rough picture. Maybe there's a way to make achieve the same without some of the additional stuff we had to add to make it work in the first place. We can have a call about it tomorrow 📞

src/components/collapsible-motion/collapsible-motion.js
dferber90: I spoke to @montezume and we can keep the current API (no breaking change) if we accept a bit of weirdness into the component (which is fine for me). We know that there's a way out in case it gets even weirder to keep supporting `containerStyles` but it still feels okay.

PR #288, opened 2018-12-06
Fix date input styling

dferber90:
| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |

Fixes `DateInput`, `DateTimeInput` and `TimeInput`.

### DIFF
--- a/src/components/internals/calendar-body/calendar-body.js
+++ b/src/components/internals/calendar-body/calendar-body.js
@@ -51,8 +51,8 @@ export default class CalendarBody extends React.PureComponent {
           <input
             ref={this.props.inputRef}
             className={classnames(styles['date-input'], {
-              [styles.error]: this.props.hasError,
-              [styles.warning]: this.props.hasWarning,
+              [styles.error]: !this.props.isDisabled && this.props.hasError,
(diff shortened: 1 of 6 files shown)


src/components/internals/calendar-body/calendar-body.js
```diff
@@ -51,8 +51,8 @@ export default class DateCalendarBody extends React.PureComponent {
           <input
             ref={this.props.inputRef}
             className={classnames(styles['date-input'], {
-              [styles.error]: this.props.hasError,
-              [styles.warning]: this.props.hasWarning,
+              [styles.error]: !this.props.isDisabled && this.props.hasError,
+              [styles.warning]: !this.props.isDisabled && this.props.hasWarning,
```
tdeekens: Since having used do expressions a bit I often feel I like them even more for their explicitness over using `classnames` at times. But that's nothing requesting a change here.

dferber90: Yeah, I'm totally on the same page. If I were to write this from scratch, I would have probably went for a `do` expression. Right now the logic of which styles to apply is spread to CSS as we use `:not()` excludes as well.

PR #290, opened 2018-12-06
chore(percy): don't run babel on built ui-kit.esm.js

### LINKED ISSUE #289: Percy shouldn't run babel on `ui-kit.esm.js`
See [issue](https://github.com/percy/react-percy/issues/107).

Right now, percy is running babel on our built file. We can either solve this with a PR to percy, or we can deal with it through ui-kit.

montezume:
Fixes #289 

I read [here](https://github.com/percy/react-percy/pull/83/files/1c417e388c73fdfe858502543c9f288b1fb5bdce#r139290092) that .babelrc files are respected by the percy config. 

Now it builds with no warnings

screen shot 2018-12-06 at 8 31 15 pm

Now I can sleep properly tonight 🤣 😆


.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "ignore": [
+    "**/dist/ui-kit.esm.js"
```
dferber90: Should we just ignore all if `dist` just in case we ever rename the file? I doubt we'd ever need to compile any of `dist` anyways? And maybe a code comment why we need `.babelrc` and why we ignore it?

montezume: yeah, that makes sense, especially since `react-percy` excludes `node_modules` from their babel config anyways. 👍

montezume: I updated it but I left out a comment. Sometimes babelrc is treated as a JSON and sometimes not, so I don't think it's worth chancing it...

PR #292, opened 2018-12-07
Fix for Checkbox and Radio hovered, disabled states

montezume:
#### Summary

We were using SVG ids to apply css styling to some of our svgs in the components `checkbox` and `radio`. However, at some point in the last period of time since we last checked to see if this worked, SVGO moved to making SVG id's unique, by prepending them with the file name.

We give id="borderAndContent", and SVGO outputs id="radio-option-default_react_svg__borderAndContent". 

#### Approach

Use regex to target the id. [ref](https://stackoverflow.com/questions/8903313/using-regular-expression-in-css)

Add VRT tests for the two components.

screen shot 2018-12-07 at 2 31 04 pm

### DIFF
--- a/src/components/switches/checkbox/checkbox.mod.css
+++ b/src/components/switches/checkbox/checkbox.mod.css
@@ -11,11 +11,11 @@
   align-items: center;
 }
 
-.checkboxWrapper svg [id='borderAndContent'] > [id='border'] {
+.checkboxWrapper svg [id$='borderAndContent'] > [id$='border'] {
   stroke: 1px var(--token-border-color-input-pristine) solid;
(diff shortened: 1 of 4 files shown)


src/components/switches/checkbox/checkbox.mod.css
```diff
@@ -11,11 +11,11 @@
   align-items: center;
 }
 
-.checkboxWrapper svg [id='borderAndContent'] > [id='border'] {
+.checkboxWrapper svg [id$='borderAndContent'] > [id$='border'] {
```
montezume: that's the fix

PR #293, opened 2018-12-07
Add even more visual regression tests

dferber90:
Adds even more visual regression tests. We are now only missing `CollapsiblePanel` and `Table` in our Visual Regression Tests.

This needs a review from designers to ensure that our baseline is correct. I can explain it in a call.


montezume:
yesss 🎉

PR #302, opened 2018-12-10
Don't run `yarn percy` when PR is opened by 🤖

montezume:
#### Summary

If `renovate` opens an average of 6 PRs per week, that means, 24 PRs per month.

24 * ~50 snapshots = 1200, or almost 25% of our current percy snapshot allowances.

These robots must be stopped.

🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖

### DIFF
--- a/scripts/run-percy.sh
+++ b/scripts/run-percy.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# skip percy builds when this is not a pull request or when it's not master
+if [[ $TRAVIS_PULL_REQUEST != 'false' || ${TRAVIS_BRANCH} == 'master' ]]
+  then
+  # skip percy builds when pull request is opened by a robot
+  regexp='^renovate\/'
+  if ! [[ ${TRAVIS_PULL_REQUEST_BRANCH} =~ ${regexp} ]]
+    then
+      yarn percy
+    else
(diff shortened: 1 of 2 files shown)


scripts/run-percy.sh
```diff
@@ -0,0 +1,16 @@
...
+# skip percy builds when this is not a pull request or when it's not master
+if [[ $TRAVIS_PULL_REQUEST != 'false' || ${TRAVIS_BRANCH} == 'master' ]]
+  then
+  # skip percy builds when pull request is opened by a robot
+  regexp='^renovate\/'
+  if ! [[ ${TRAVIS_PULL_REQUEST_BRANCH} =~ ${regexp} ]]
+    then
+      yarn percy
+    else
+      echo 'Skipping percy - PR opened by `renovate`'
```
tdeekens: Maybe add a note why we do this. It seems intuitive to run percy when deps update.

montezume: But we already run percy on `master`, so if someone were to merge one of these PRs, it will run.

montezume: comment updated [here](https://github.com/commercetools/ui-kit/pull/302/commits/c52f18d368154613f20df0988c942c12ba9aeb56)

tdeekens: Thanks.

PR #304, opened 2018-12-10
Typography (Text): adding data test prop to text components

NickDevG:
#### Summary
Adding `data-test` prop to every `Text` component.

#### Description
Adding `data-test` prop to avoid wrapping `Text` components for E2E tests

### DIFF
--- a/src/components/typography/text/README.md
+++ b/src/components/typography/text/README.md
@@ -26,6 +26,7 @@ Wraps the given text in the given HTML header `size`.
 | `children`    | `PropTypes.node` |    ✅    | -                    | -       | -                                                              |
 | `title`       | `String`         |    -     | -                    | -       | Text to show in a tooltip on hover over the element            |
 | `truncate`    | `Bool`           |    -     | -                    | `false` | Option for truncate content in case the screen has small width |
+| `dataTest`    | `String`         |    -     | -                    | -       | Prop for E2E testing                                           |
 
 #### Where to use
 
@@ -51,6 +52,7 @@ Wraps the given text in the given HTML header `size`.
 | `children`    | `PropTypes.node` |    ✅    | -                                                                 | -       |
 | `title`       | `String`         |    -     | -                                                                 | -       |
 | `truncate`    | `Bool`           |    -     | -                                                                 | `false` |
+| `dataTest`    | `String`         |    -     | -                                                                 | -       | Prop for E2E testing |
 
 #### Where to use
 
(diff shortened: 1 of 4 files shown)


src/components/typography/text/README.md
```diff
@@ -26,6 +26,7 @@ Wraps the given text in the given HTML header `size`.
 | `children`    | `PropTypes.node` |    ✅    | -                    | -       | -                                                              |
 | `title`       | `String`         |    -     | -                    | -       | Text to show in a tooltip on hover over the element            |
 | `truncate`    | `Bool`           |    -     | -                    | `false` | Option for truncate content in case the screen has small width |
+| `dataTest`    | `String`         |    -     | -                    | -       | Prop for E2E testing                                           |
```
montezume: instead of making a new prop, can you follow how it's done [here](https://github.com/commercetools/ui-kit/tree/master/src/components/inputs/text-input) in text-input?

Don't add it to the table instead add 
```
The component further forwards all data- attributes to the underlying component.
```

src/components/typography/text/text.js
```diff
@@ -43,6 +45,7 @@ const Subheadline = props => {
         [styles.truncate]: props.truncate,
       })}
       title={props.title}
+      data-test={props.dataTest}
```
montezume: can you use [filter-data-attributes](https://github.com/commercetools/ui-kit/blob/master/src/utils/filter-data-attributes.js) instead so that it supports all data attributes? You can just spread it like 
```
```

src/components/typography/typography.story.js
```diff
@@ -51,6 +51,7 @@ storiesOf('Typography/Text', module)
       <InlineColorWrapper width={'200px'}>
         <Text.Wrap
           title={text('title', 'Text to be shown as tooltip on hover')}
+          dataTest={text('dataTest', 'test-prop')}
```
montezume: can you remove this as well?

src/components/typography/text/text.spec.js
```diff
@@ -290,11 +322,19 @@ describe('<Body>', () => {
       });
     });
   });
+  describe('with dataTest', () => {
+    beforeEach(() => {
+      wrapper = shallow(<Text.Body data-test="prop-test">{'Body'}</Text.Body>);
```
dferber90: It would be nice to also test the `isInline` case and ensure data-props work with that as well since that uses a different rendering tree. If you have some time, feel free to test that case too. Not super important though.

Issue #308, opened 2018-12-11
Add withId HoC


dferber90:
Right now quite some of our components auto-generate an `id` in case the parent doesn't pass one in so that we can hook up labels to inputs. Unfortunately the code to do so is quite spread around the component. Usually we do

```js
import createSequentialId from '../../../utils/create-sequential-id';


// somewhere at the top
const sequentialId = createSequentialId('date-field-');

export default class Foo extends React.Component {
  state = {
    // We generate an id in case no id is provided by the parent to attach the
    // label to the input component.
    id: this.props.id,
  };

  static getDerivedStateFromProps = (props, state) => ({
    id: do {
      if (props.id) props.id;
      else if (state.id) state.id;
      else sequentialId();
    },
  });

  render () {
    return <input id={this.state.id} />
  }
}
```

That's a bit much setup and it's spread across different sections of the component.

We could improve this by creating a HoC which ensures there is an `id`. The usage would then look like this instead:

```js

class Foo extends React.Component {
  render () {
    return <input id={this.props.id} />
  }
}

export default withId(Foo)
```

The downside of this is that all the components would be wrapped in a HoC, which makes it less easy to see them in react's dev-tools.

dferber90:
Closing this as #311 showed that it's not easily done so far. And even if it was, then the additional HoC added to all inputs would be a bit inconvenient to work with.

PR #309, opened 2018-12-11
Add DateField component

dferber90:
#### Summary

Add `DateField` component.

*I did not add Visual Regression Tests yet as we're awaiting #293.*

#### Design

image

##### Technical

- Documentation

  - [x] Has `README.md` file
  - [ ] Explains use-cases
  - [x] Explains props
  - [x] Has story for Storybook

- Props

  - [ ] Has default props for labels (where applicable)

- Design

  - [ ] Uses design-tokens (instead of e.g. color variables)

- Testing

  - [x] Uses [react-testing-library](https://github.com/kentcdodds/react-testing-library) tests
  - [ ] Has visual regression tests

- Accessibility (optional)
  - [ ] Supports `aria` attributes

### DIFF
--- a/src/components/fields/date-field/date-field.js
+++ b/src/components/fields/date-field/date-field.js
@@ -1,0 +1,108 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import requiredIf from 'react-required-if';
+import Constraints from '../../constraints';
+import Spacings from '../../spacings';
+import FieldLabel from '../../field-label';
+import DateInput from '../../inputs/date-input';
+import createSequentialId from '../../../utils/create-sequential-id';
+import FieldErrors from '../../field-errors';
+import filterDataAttributes from '../../../utils/filter-data-attributes';
+
+const sequentialId = createSequentialId('date-field-');
+
+const hasErrors = errors => errors && Object.values(errors).some(Boolean);
+
+class DateField extends React.Component {
+  static displayName = 'DateField';
+
+  static propTypes = {
+    // DateField
+    id: PropTypes.string,
+    horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
+    errors: PropTypes.shape({
+      missing: PropTypes.bool,
+    }),
+    renderError: PropTypes.func,
+    isRequired: PropTypes.bool,
+    touched: PropTypes.bool,
+
+    // DateInput
+    name: PropTypes.string,
+    value: PropTypes.string.isRequired,
+    onChange: requiredIf(PropTypes.func, props => !props.isReadOnly),
+    onBlur: PropTypes.func,
+    onFocus: PropTypes.func,
+    isDisabled: PropTypes.bool,
+    placeholder: PropTypes.string,
+
+    // LabelField
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    hint: requiredIf(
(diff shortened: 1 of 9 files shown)


src/components/fields/date-field/date-field.js
```diff
@@ -0,0 +1,110 @@
...
+import filterDataAttributes from '../../../utils/filter-data-attributes';
+
+const sequentialId = createSequentialId('date-field-');
+
+const hasErrors = errors => errors && Object.values(errors).some(Boolean);
+
+class DateField extends React.Component {
+  static displayName = 'DateField';
+
+  static isEmpty = DateInput.isEmpty;
```
tdeekens: Didn't know we do this for all fields.

src/components/fields/date-field/date-field.js
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+  state = {
+    // We generate an id in case no id is provided by the parent to attach the
+    // label to the input component.
+    id: this.props.id,
+  };
+
+  static getDerivedStateFromProps = (props, state) => ({
+    id: do {
```
tdeekens: This is a nice use case for a HoC or hook.

src/components/inputs/date-input/date-input.js
```diff
@@ -25,8 +25,9 @@ import {
   parseInputToDate,
 } from '../../../utils/calendar';
 
-class DateCalendar extends React.Component {
-  static displayName = 'DateCalendar';
+class DateInput extends React.Component {
+  static displayName = 'DateInput';
+  static isEmpty = value => value === '';
```
tdeekens: Not need to `trim` or do anything more? Maybe even share `TextInput.isEmpty`?

src/components/fields/date-field/date-field.js
dferber90: It felt like a good idea to take the burden of determining the empty state away from the consumers. We don't have this functionality anywhere yet, but seems like it would be useful for all inputs & fields.

src/components/fields/date-field/date-field.js
dferber90: Yep, already on that :P I don't think it could be a hook, but I'm working on a HoC for it. See #308

src/components/fields/date-field/date-field.js
tdeekens: Sure. I am just arguing from a consumer perspective. My intuition would have "told" me to use the respective `*Input` for the `isEmpty` check and not the field itself. Not saying to change it. Other folks might have different expectations.

dferber90: Yeah we could do that as well. having to redeclare the statics on the fields, and having to document the statics on the field again is a bit of overhead for the ui-kit anyways. It might make sense to educate consumers of these components to use the statics of the underlying inputs instead.

Thanks for pointing it out 🥇 

cc @montezume

coolmilo:
@dferber90 at the moment my only concern is using the "xs" and "s" horizontalConstraint. I'm not sure if we are restricted to space that it would be useful.

coolmilo:
horizontalconstraint-small

dferber90:
Thanks for pointing that out @coolmilo. We plan to go through all components at once and fix it for all of them at the same time. So for now we keep adding the "unsupported" props.

See https://github.com/commercetools/ui-kit/issues/325

coolmilo:
Approved from my side, @dferber90 do you need @filippobocik to have a last look?

PR #310, opened 2018-12-11
chore: refactor `localized-text-input` to use rtl

montezume:
#### Summary

Refactors `LocalizedTextInput` to use `react-testing-library`. 

Kills some enzyme snapshots

Feature: Adds support for focusing `LocalizedTextInput` without a name prop by clicking it's label.


src/components/inputs/localized-text-input/__snapshots__/localized-text-input.spec.js.snap
```diff
@@ -1,145 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dferber90: 👋 👋

src/components/inputs/localized-text-input/localized-text-input.js
```diff
@@ -180,12 +183,20 @@ export default class LocalizedTextInput extends React.Component {
       hasErrorOnRemainingLanguages ||
       props.hideExpansionControls ||
       state.areLanguagesOpened;
-    return { areLanguagesOpened };
+
+    const id = do {
```
dferber90: Let's wait for #311 and use `withId` instead?

src/components/inputs/localized-text-input/localized-text-input.spec.js
```diff
@@ -1,267 +1,263 @@
...
+        value={this.state.value}
+        selectedLanguage={this.state.selectedLanguage}
+      />
+    );
+  }
+}
+
+const renderLocalizedTextInput = (props, options) => ({
+  ...render(<TestComponent {...props} />, options),
+});
```
dferber90: Nitpick: No need to create the object for now. We can do

```
const renderLocalizedTextInput = (props, options) =>
  render(<TestComponent {...props} />, options)
```

We might actually not use this function at all and just always do `render(<TestComponent {...props} />, options)` as this case here is pretty standard and boilerplate-free. It's up to you :)

src/components/inputs/localized-text-input/localized-text-input.js
dferber90: I closed #311 as it wasn't feasible yet. Let's continue the way it is right now then.

PR #311, opened 2018-12-11
add withId HoC

### LINKED ISSUE #308: Add withId HoC

dferber90:
We were repeating id generation in a few of our components and it felt like the right time to extract the logic.

I also noticed that we were generating the wrong ids for a bunch of our components and fixed it along the way (we were using the wrong prefix - a copy-paste error).

The only thing I dislike about this is that our inputs are now wrapped in another component, which makes it a tiny bit harder to access them in the dev-tools.

Closes #308

### DIFF
--- a/src/components/fields/async-creatable-select-field/async-creatable-select-field.js
+++ b/src/components/fields/async-creatable-select-field/async-creatable-select-field.js
@@ -5,15 +5,13 @@ import Constraints from '../../constraints';
 import Spacings from '../../spacings';
 import FieldLabel from '../../field-label';
 import AsyncCreatableSelectInput from '../../inputs/async-creatable-select-input';
-import createSequentialId from '../../../utils/create-sequential-id';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import FieldErrors from '../../field-errors';
-
-const sequentialId = createSequentialId('text-field-');
+import withId from '../../../hocs/with-id';
 
 const hasErrors = errors => errors && Object.values(errors).some(Boolean);
 
-export default class SelectField extends React.Component {
+class SelectField extends React.Component {
   static displayName = 'SelectField';
 
   static isTouched = AsyncCreatableSelectInput.isTouched;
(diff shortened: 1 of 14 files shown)


dferber90:
Tests are failing as I still need to hoist the static functions inside the HoC. I'm on it.

dferber90:
Unfortunately we still have some `enzyme` tests which we'd need to adapt to be able to test this setup. It's not worth changing the `enzyme` tests since we want to use replace them with `react-testing-library` anyways - in which case it would work out of the box.

After talking to @montezume we agreed that it's best to do nothing for now and wait until more tests are converted.

> The only thing I dislike about this is that our inputs are now wrapped in another component, which makes it a tiny bit harder to access them in the dev-tools.

This point still holds true and would be a bit annoying if we ever introduce `withId` in the future.

I will open another PR to fix the wrong names passed to `createSequentialId`.

PR #316, opened 2018-12-12
chore: refactor multiline-text-input to use rtl

montezume:
#### Summary

Refactors `multiline-text-input` to use `react-testing-library`.

#### Limitations

I couldn't figure out how to work out how to test the open / closed states of the component, mostly due to `react-textarea-autosized` not being able to calculate the nodeHeight in the js-dom environment. Do you have any ideas @pa3 ?

### DIFF
--- a/src/components/inputs/text-input/text-input.spec.js
+++ b/src/components/inputs/text-input/text-input.spec.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { render, fireEvent } from '../../../test-utils';
-import { TextInput } from '../../../index';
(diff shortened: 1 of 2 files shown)


src/components/inputs/text-input/text-input.spec.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { render, fireEvent } from '../../../test-utils';
-import { TextInput } from '../../../index';
+import TextInput from './text-input';
```
montezume: feels better not to have the risk of circular import here

PR #322, opened 2018-12-12
chore: refactor select-input tests to use rtl

montezume:
#### Summary

Move select-input tests to use RTL.

### DIFF
--- a/src/components/inputs/select-input/select-input.spec.js
+++ b/src/components/inputs/select-input/select-input.spec.js
@@ -28,7 +28,7 @@ class TestComponent extends React.Component {
   };
 
   state = {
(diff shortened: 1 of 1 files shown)


src/components/inputs/select-input/select-input.spec.js
```diff
@@ -1,147 +1,277 @@
...
+      expect(onChange).toHaveBeenCalledWith({
+        persist: expect.any(Function),
+        target: {
+          name: 'some-name',
+          value: ['mango'],
+        },
+      });
+      // open list again
+      onChange.mockReset();
+      onChange.mockClear();
```
dferber90: I don't think we need these resets at all.

PR #326, opened 2018-12-13
fix: remove random generate heights from VRTs for spacings

montezume:
#### Summary

Stack and Inline spacings were randomly generated heights, leading to percy tests failing.

Also, move all VRTs into one suite for spacings to save on some snaps. #snaps


src/components/spacings/spacings.percy.js
```diff
@@ -0,0 +1,225 @@
...
+};
+
+const InlineExample = ({ alignItems }) => (
+  <Stack>
+    {sizes.map(size => (
+      <Row key={size.name}>
+        <InlineColorWrapper>
+          <Scale>
+            <Spacings.Inset scale="s" alignItems="center">
+              <Text.Subheadline elementType="h4">
```
dferber90: Something to be aware of is that by reusing components in other vrt's it is possible that these "unrelated" tests break when the used components change. Not too big of a deal as long as we don't overdo it.

PR #328, opened 2018-12-13
TextField: using react testing library

NickDevG:
#### Summary
Converting tests from `enzyme` to `react-testing-library`

#### Description
image

### DIFF
--- a/src/components/fields/number-field/number-field.spec.js
+++ b/src/components/fields/number-field/number-field.spec.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import omit from 'lodash.omit';
 import { render, fireEvent } from '../../../test-utils';
 import NumberField from './number-field';
 
@@ -14,7 +13,7 @@ import NumberField from './number-field';
 class Story extends React.Component {
   static displayName = 'Story';
   static propTypes = {
-    onEvent: PropTypes.func.isRequired,
+    onChange: PropTypes.func.isRequired,
     value: PropTypes.string,
     id: PropTypes.string,
   };
@@ -25,7 +24,7 @@ class Story extends React.Component {
     value: this.props.value || '',
   };
   handleChange = event => {
-    this.props.onEvent(event);
+    this.props.onChange(event);
     this.setState({ value: event.target.value });
   };
   render() {
@@ -34,7 +33,7 @@ class Story extends React.Component {
         <label htmlFor={this.props.id}>NumberField</label>
         <NumberField
           id={this.props.id}
-          {...omit(this.props, 'onEvent')}
+          {...this.props}
           value={this.state.value}
           onChange={this.handleChange}
         />
@@ -44,14 +43,14 @@ class Story extends React.Component {
 }
 
 const renderNumberField = (customProps, options) => {
-  const onChange = jest.fn();
   const props = {
     title: 'foo',
+    onChange: jest.fn(),
     ...customProps,
   };
   return {
-    onChange,
-    ...render(<Story onEvent={onChange} {...props} />, options),
+    ...render(<Story {...props} />, options),
+    onChange: props.onChange,
   };
 };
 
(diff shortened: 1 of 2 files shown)


NickDevG:
I've tried to cover everything that you could actually see on the storybook. One thing is missing: the `hintIcon`

src/components/fields/text-field/text-field.spec.js
```diff
@@ -1,172 +1,187 @@
...
-  describe('when there are unknown errors', () => {
-    let props;
-    let wrapper;
-    beforeEach(() => {
-      props = createTestProps({
+describe('when showing an info button', () => {
+  it('should render an info button', () => {
+    const onInfoButtonClick = jest.fn();
+    const { container } = renderTextField({ onInfoButtonClick });
+    expect(container.querySelector('button')).toBeTruthy();
```
montezume: instead of using querySelector you can use getByLabelText `More Info` right?

montezume:
@NickDevG the snapshot tests take care of those cases anyways 😄 🎉 
<mg width="903" alt="screen shot 2018-12-13 at 2 29 28 pm" src="https://user-images.githubusercontent.com/2425013/49941690-852ab500-fee3-11e8-8caa-34c462e41034.png">

NickDevG:
I've added `NumberField` tests here as well

src/components/fields/number-field/number-field.spec.js
```diff
@@ -1,172 +1,210 @@
...
+// It overwrites the onChange function and places a label for the
+// input component. It also ensures an id so that the label can associate
+// the input. This allows tests to use getByLabelText.
+// It also makes sure the event's value passed to onChange flows back to the
+// component so that we can test it under real conditions.
+// As a convenience, we enable accessing a mocked onChange function.
+class Story extends React.Component {
+  static displayName = 'Story';
+  static propTypes = {
+    onEvent: PropTypes.func.isRequired,
```
dferber90: The comment above currently doesn't match the actual behaviour as the `onChange` function does not get overwritten. I would also be careful about introducing `onEvent` as it makes it seem as-if `onEvent` is a function supported by `NumberField`. If you keep this as `onChange` then you also don't need the `omit` below.

We use this pattern in tests of multiple components so people should be familiar with anyways. 

So could you do

```js
handleChange = event => {
  this.props.onChange(event);
  this.setState({ value: event.target.value });
}

<NumberField
  id={this.props.id}
  value={this.state.value}
  onChange={this.handleChange}
/>
```

src/components/fields/number-field/number-field.spec.js
```diff
@@ -1,172 +1,210 @@
...
+
+const renderNumberField = (customProps, options) => {
+  const onChange = jest.fn();
+  const props = {
+    title: 'foo',
+    ...customProps,
+  };
+  return {
+    onChange,
+    ...render(<Story onEvent={onChange} {...props} />, options),
```
dferber90: Nitpick because I'm afraid this will get copied around: I would recommend to put the spread first and then provide the `onChange` to avoid accidental overwrites (in other cases):

```js
return {
  ...render(<Story onEvent={onChange} {...props} />, options),
  onChange
}
```

src/components/fields/number-field/number-field.spec.js
```diff
@@ -1,172 +1,210 @@
...
+it('should have an HTML name', () => {
+  const { container } = renderNumberField({ name: 'foo' });
+  expect(container.querySelector('[name="foo"]')).toBeInTheDocument();
+});
+
+it('should call onFocus when the input is focused', () => {
+  const onFocus = jest.fn();
+  const { container } = renderNumberField({ onFocus });
+  container.querySelector('input').focus();
+  expect(container.querySelector('input')).toHaveFocus();
```
dferber90: The `Story` component renders a `label` which we can use to find the input instead of using `container.querySelector('input')`. This is closer to how a user would interact with the component and should therefore be preferred.

The *[Which get method should I use?](https://github.com/kentcdodds/dom-testing-library#faq)* section explains this nicely:

image

src/components/fields/number-field/number-field.spec.js
```diff
@@ -1,172 +1,210 @@
...
-    it('should mark the NumberInput as erroneous', () => {
-      expect(wrapper.find(NumberInput)).toHaveProp('hasError', true);
+      expect(
+        getByText('This field is required. Provide a value.')
+      ).toBeInTheDocument();
     });
-    it('should forward the error', () => {
-      expect(wrapper.find(FieldErrors)).toHaveProp('errors', props.errors);
+  });
+  describe('when has a custom error', () => {
```
dferber90: "when there is a custom error"?

src/components/fields/text-field/text-field.spec.js
```diff
@@ -1,172 +1,189 @@
...
-        renderError: jest.fn(key => key),
-        errors: { customError: 5 },
+        errors: { missing: true },
       });
-      wrapper = shallow(<TextField {...props} />);
-    });
-    it('should mark the NumberInput as erroneous', () => {
-      expect(wrapper.find(TextInput)).toHaveProp('hasError', true);
+      expect(
+        getByText('This field is required. Provide a value.')
```
dferber90: A tip to keep this test working even when the message changes would be to use a RegEx: `getByText(/field is required/i)`.

src/components/fields/text-field/text-field.spec.js
```diff
@@ -1,172 +1,189 @@
...
+      ).toBeInTheDocument();
     });
-    it('should forward the error', () => {
-      expect(wrapper.find(FieldErrors)).toHaveProp('errors', props.errors);
+  });
+  describe('when has a custom error', () => {
+    it('should render a custom error', () => {
+      const renderError = jest.fn(() => 'Custom error');
```
dferber90: Nitpick: This doesn't actually need the  `jest.fn` wrapper.

PR #339, opened 2018-12-17
Add  LocalizedMoneyInput Component

islam3zzat:
#### Summary

image

image


src/components/inputs/localized-money-input/messages.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default defineMessages({
+  missingRequiredField: {
+    id: 'UIKit.LocalizedMoneyInput.missingRequiredField',
+    description: 'Error message for missing required value',
+    defaultMessage: 'This field is required. Provide at least one value.',
+  },
+  show: {
+    id: 'UIKit.LocalizedMoneyInput.show',
+    description: '',
```
montezume: maybe should add descriptions for these?

src/components/inputs/localized-money-input/README.md
```diff
@@ -0,0 +1,173 @@
...
+The `parseMoneyValues` function will turn a [`MoneyValue`](https://docs.commercetools.com/http-api-types#money) into a value the LocalizedMoneyInput component can handle `({ amount, currencyCode })`.
+
+##### `LocalizedMoneyInput.isEmpty`
+
+The `isEmpty` function will return `true` for the crrency when the passed `LocalizedMoneyInput` value for this currency is empty (either has no currency or no amount, or does not exist at all).
+
+```js
+LocalizedMoneyInput.isEmpty([{ amount: '', currencyCode: 'EUR' }]); // -> { EUR: true }
+
+LocalizedMoneyInput.isEmpty({ amount: '5', currencyCode: 'EUR' }); // -> { EUR: false }
```
dferber90: @islam3zzat I know we spoke about this before. However this isn't quite what I had expected (or maybe I'm remembering it wrong). Here's my current take for how things are supposed to work. I'm happy to chat to you again to align on the approach.

Generally, `LocalizedMoneyInput` should only operate on the `value` it expects (instead of on an array with `amount` and `currencyCode`).

I would also make `LocalizedMoneyInput.isEmpty` the only way to check whether the value is empty. So no `isAllEmpty` and no `hasEmptyValue`.

---

To me, `LocalizedMoneyInput.isEmpty` should be designed to work with the `value` passed to `LocalizedMoneyInput` which is an object (not an array as shown some of the examples). So `LocalizedMoneyInput.isEmpty` should accept an object instead of an array.

We should be able to do `LocalizedMoneyInput.isEmpty({ EUR: "12.00", USD: "14.23", CAD: "" })`, which would return `false` as at least one non-empty currency-amount pair is provided. That way it works simliar to `TextInput.isEmpty` which also returns `false` when at least one translation is provided.

This means we can remove `isAllEmpty` and `hasEmptyValue`.

---

There might be a case where users need to provide a value for every currency. We can accept

```js
LocalizedMoneyInput.isEmpty(
  { EUR: "12.00", USD: "14.23", CAD: "" },
  { requireAllCurrenciesToBeFilled: true }
)
```

for such cases. It might also be useful for consumers to be able to determine the currencies with empty values. We can make this easier by providing a separate function which returns this information (either as an array or as an object; not sure which API is better):

```js
// as an array
LocalizedMoneyInput.getEmptyCurrencies({ EUR: "12.00", USD: "14.23", CAD: "" })
// -> ["CAD"]

// as an object
LocalizedMoneyInput.getEmptyCurrencies({ EUR: "12.00", USD: "14.23", CAD: "" })
// -> { EUR: false, USD: false, CAD: true }
```

I slightly prefer the first version as it's easier to do `emptyCurrencies.includes("CAD").


However I wouldn't build this functionality until we actually need it. So let's leave it out for now?

src/components/inputs/localized-money-input/localized-money-input.js
```diff
@@ -0,0 +1,307 @@
...
+
+export class LocalizedMoneyInput extends React.Component {
+  static displayName = 'LocalizedMoneyInput';
+
+  static propTypes = {
+    id: PropTypes.string,
+    name: PropTypes.string,
+    // then input doesn't accept a "currencies" prop, instead all possible
+    // currencies have to exist (with empty or filled strings) on the value:
+    //   { EUR: 12.00, USD: 14.23, CAD: 13.70 }
```
dferber90: The example given here doesn't use strings but actual numbers. However the prop-type requires strings to be present

```suggestion
    //   { EUR: "12.00", USD: "14.23", CAD: "" }
```

src/components/inputs/localized-money-input/localized-money-input.js
```diff
@@ -0,0 +1,307 @@
...
+      MoneyInput.convertToMoneyValue({
+        currencyCode,
+        amount: values[currencyCode],
+      })
+    );
+
+  static parseMoneyValues = (moneyValues = [], locale) =>
+    moneyValues.map(value => MoneyInput.parseMoneyValue(value, locale));
+
+  static isHighPrecision = (values = []) =>
```
dferber90: As our `props.value` accepts an object, I'd make `isHighPrecision` work with an object as well.

So we should be able to do

```js
LocalizedMoneyInput.isHighPrecision({ EUR: "12.00", USD: "14.23323", CAD: "" })
// -> true
```

---

Another issue with the current implementation is that the name is `isHighPrecision` which indicates a boolean would be returned, but it returns an object.

---

Similar to how I described `LocalizedMoneyInput.getEmptyCurrencies` above, we can offer a way for our consumers to find out which of the supplied prices are high-precision. We can offer this API for it:

```js
LocalizedMoneyInput.getHighPrecisionCurrencies({ EUR: "12.00", USD: "14.2334", CAD: "" })
// -> ["USD"]
```

> Sidenote: The implementation of `LocalizedMoneyInput.isHighPrecision` can then just be
> ```js
> LocalizedMoneyInput.isHighPrecision = value =>
>   LocalizedMoneyInput.getHighPrecisionCurrencies(value).length > 0
> ```

src/utils/localized.js
```diff
@@ -42,6 +42,14 @@ export const sortLanguages = (selectedLanguage, allLanguages) => {
   return [selectedLanguage, ...related.sort(), ...unrelated.sort()];
 };
 
+// sorts the currencies with the following priority:
+// - The selected currency is placed first (e.g EUR)
+// - All other currencies follow, sorted alphabetically as well
+export const sortCurrencies = (selectedCurrency, allCurrencies) => {
```
dferber90: It's kinda weird to have something about currencies in a file called `localized`. This file has so far been used for `LocalizedString` related stuff. I'd keep this functionality in the component itself.

src/components/inputs/localized-money-input/README.md
islam3zzat: You are right, the statics here should operate on the value it expects.

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -576,110 +576,54 @@ describe('LocalizedMoneyInput.parseMoneyValues', () => {
...
-    it('should return false if some have value', () => {
+    it('should return empty array', () => {
       expect(
-        LocalizedMoneyInput.hasEmptyValue([
-          { amount: '5', currencyCode: 'EUR' },
-          { amount: '', currencyCode: 'EUR' },
-        ])
-      ).toBe(true);
+        LocalizedMoneyInput.getEmptyCurrencies({ EUR: '5.22', USD: '31.88' })
+      ).toMatchObject([]);
```
tdeekens: Could also be `toHaveLength(0)` I presume?

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -576,110 +576,54 @@ describe('LocalizedMoneyInput.parseMoneyValues', () => {
...
-  describe('when value is all', () => {
-    it('should return true if all are empty', () => {
-      expect(
-        LocalizedMoneyInput.hasEmptyValue([
-          { amount: '21.33', currencyCode: 'EUR' },
-          { amount: '51.99', currencyCode: 'EUR' },
-        ])
-      ).toBe(false);
+        LocalizedMoneyInput.getEmptyCurrencies({ EUR: '', USD: '31.88' })
+      ).toMatchObject(['EUR']);
```
tdeekens: `toEqual` would work, right? or `toEqual(expect.arrayContaining([])` if you need it more flexible.

src/components/inputs/localized-money-input/README.md
```diff
@@ -128,16 +153,16 @@ Here are examples of `centPrecision` and `highPrecision` prices.
...
 
 ```js
-LocalizedMoneyInput.isEmpty([{ amount: '', currencyCode: 'EUR' }]); // -> { EUR: true }
+LocalizedMoneyInput.getEmptyCurrencies({ EUR: '', USD: '12.77' }); // -> ['EUR']
 
-LocalizedMoneyInput.isEmpty({ amount: '5', currencyCode: 'EUR' }); // -> { EUR: false }
+LocalizedMoneyInput.getEmptyCurrencies({ EUR: '12.77' }); // -> []
 ```
 
 ##### `LocalizedMoneyInput.isAllEmpty`
```
dferber90: This should go away still iirc.

src/components/inputs/localized-money-input/README.md
```diff
@@ -171,3 +196,121 @@ LocalizedMoneyInput.hasEmptyValue([
...
+import omitEmpty from 'omit-empty';
+import ErrorMessage from '@commercetools-frontend/ui-kit/messages/error-message';
+import { LocalizedMoneyInput } from '@commercetools-frontend/ui-kit';
+
+
+// the existing document, e.g. from the database
+const doc = {
+  prices: [
+    {
+      currencyCode: 'ALL',
```
dferber90: I would use a different currency in the example as *ALL*  sounds like the english word *all* which might be confusing.

src/components/inputs/localized-money-input/README.md
```diff
@@ -171,3 +196,121 @@ LocalizedMoneyInput.hasEmptyValue([
...
+    },
+    {
+      currencyCode: 'AMD',
+      centAmount: 3300,
+    }
+  ],
+};
+
+// A function to convert a document to form values.
+const docToFormValues = aDoc => ({
```
dferber90: Usually we just call this variable `doc` (instead of `aDoc`)

src/components/inputs/localized-money-input/README.md
```diff
@@ -171,3 +196,121 @@ LocalizedMoneyInput.hasEmptyValue([
...
+  const errors = { prices: {} };
+  Object.keys(formValues.prices).forEach(currency => {
+    errors.prices[currency] = {};
+  })
+  const emptyCurrencies = LocalizedMoneyInput.getEmptyCurrencies(formValues.prices);
+  // ['ERU', 'USD']
+  // THis form dosen't accept highprecision prices
+  const highPrecisionCurrencies = LocalizedMoneyInput.getHighPrecisionCurrencies(formValues.prices);
+  // ['CAD', 'USD']
+  emptyCurrencies.forEach(emptyCurrency => {
```
dferber90: The spacing in this example is a bit weird. We do

```
get empty currency errors
get high precision currency errors
add empty currency errors

add high precision currency errors
```

instead of

```
get empty currency errors
get high precision currency errors

add empty currency errors
add high precision currency errors
```

or

```
get empty currency errors
add empty currency errors

get high precision currency errors
add high precision currency errors
```

src/components/inputs/localized-money-input/README.md
```diff
@@ -171,3 +196,121 @@ LocalizedMoneyInput.hasEmptyValue([
...
+      setFieldTouched,
+      handleSubmit,
+      isSubmitting,
+    }) => (
+      <form onSubmit={handleSubmit}>
+        <LocalizedMoneyInput
+          value={values.prices}
+          onBlur={() => setFieldTouched('prices')}
+          isDisabled={isSubmitting}
+          onChange={value => setFieldValue('prices', value)}
```
dferber90: In the example, I'd rather do 

```
<LocalizedMoneyInput
  name="prices"
  onChange={formik.handleChange}
  onBlur={formik.handleBlur}
/>
```

instead of setting the name using `setFieldTouched` and `setFieldValue`.

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -0,0 +1,629 @@
...
+      container.querySelector('input[name="foo.USD.amount"]')
+    ).not.toBeInTheDocument();
+  });
+  it('should allow changing value of the input', () => {
+    const { container } = renderLocalizedMoneyInput({
+      name: 'foo',
+      selectedCurrency: 'CAD',
+    });
+    const event = { target: { value: '17.34' } };
+    const input = container.querySelector('input[name="foo.CAD.amount"]');
```
dferber90: Is there a way to find this input using a label or another approach than `querySelector`? We should strive not to use other getters first.

image

See [*"Which get method should I use?"*](https://testing-library.com/docs/faq)

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -0,0 +1,629 @@
...
+    ).toBeInTheDocument();
+  });
+  it('should allow changing the USD input', () => {
+    const { getByLabelText, container } = renderLocalizedMoneyInput({
+      name: 'foo',
+      selectedCurrency: 'CAD',
+    });
+    getByLabelText(/show all currencies/i).click();
+    const event = { target: { value: '12.98' } };
+    const USDInput = container.querySelector('input[name="foo.USD.amount"]');
```
dferber90: I'd rename this to `usdInput`. It's a regular variable so it should start lowercase. Otherwise the `USDInput.value` access later on looks like we're reading a static property from a class. It's pretty nitpicky, but I wouldn't want this naming pattern to start spreading.

src/components/inputs/localized-money-input/localized-money-input.percy.js
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
+import {
```
montezume: can you refactor this to match our current VRT tests? 

we need one .visualroute.js and one .visualspec.js 👍

src/components/inputs/localized-money-input/localized-money-input.percy.visualroute.js
```diff
@@ -1,18 +1,16 @@
...
+import { Suite, Spec } from '../../../../test/percy';
 
 const value = {
   EUR: '12.77',
   USD: '13.55',
   CAD: '19.82',
 };
 
-screenshot('LocalizedMoneyInput', () => (
+export const routePath = '/localized-money-input';
```
montezume: 👍

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -64,23 +64,21 @@ it('should forward data-attributes', () => {
...
 it('should have an HTML name', () => {
-  const { container } = renderLocalizedMoneyInput({
+  const { getByLabelText } = renderLocalizedMoneyInput({
     name: 'foo',
     selectedCurrency: 'CAD',
   });
-  expect(
-    container.querySelector('input[name="foo.CAD.amount"]')
-  ).toBeInTheDocument();
+  expect(getByLabelText('CAD')).toBeInTheDocument();
```
montezume: this doesn't really test that it has an HTML name though?

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -64,23 +64,21 @@ it('should forward data-attributes', () => {
...
 it('should call onBlur when input loses focus', () => {
   const onBlur = jest.fn();
-  const { container } = renderLocalizedMoneyInput({
+  const { getByLabelText } = renderLocalizedMoneyInput({
     name: 'foo',
     selectedCurrency: 'CAD',
     onBlur,
   });
-  const input = container.querySelector('input[name="foo.CAD.amount"]');
+  const input = getByLabelText('CAD');
```
montezume: 👍

src/components/inputs/localized-money-input/localized-money-input.spec.js
```diff
@@ -89,24 +87,21 @@ it('should call onBlur when input loses focus', () => {
...
     });
-    expect(
-      container.querySelector('input[name="foo.CAD.amount"]')
-    ).toBeInTheDocument();
-    expect(
-      container.querySelector('input[name="foo.USD.amount"]')
-    ).not.toBeInTheDocument();
+    expect(getByLabelText('CAD')).toBeInTheDocument();
+    // toThrow('Unable to find a label with the text of: USD"')
+    expect(queryByLabelText('USD')).not.toBeInTheDocument();
```
montezume: 👍

src/components/inputs/localized-money-input/localized-money-input.spec.js
islam3zzat: not directly, or indirectly :D
updated the test name

dferber90:
I pinged @filippobocik for a review of the Visual Regression Tests baseline.

@montezume This looks ready to be reviewed again.

src/components/inputs/localized-money-input/README.md
```diff
@@ -0,0 +1,286 @@
...
+  highPrecisionCurrencies.forEach(highPrecisionCurrency => {
+    errors.prices[highPrecisionCurrency].highPrecision = true;
+  });
+
+  return omitEmpty(errors);
+};
+const initialValues = docToFormValues(doc);
+const renderErrors = (errors) => {
+  Object.keys(errors.prices).reduce(currency => {
+    const error = errors.prices[currency]?.missing && (
```
tdeekens: Is it just the diff or is the formatting off?

src/components/inputs/localized-money-input/README.md
```diff
@@ -0,0 +1,286 @@
...
+      // ready to be consumed by the API
+      const nextDoc = formValuesToDoc(formValues);
+      console.log(nextDoc);
+    }}
+    render={({
+      values,
+      errors,
+      touched,
+      setFieldValue,
+      setFieldTouched,
```
tdeekens: We don't use half of these below, should we remove them given that this is a copy and paste basis?

src/components/inputs/localized-money-input/localized-money-input.js
```diff
@@ -0,0 +1,319 @@
...
+const sequentialId = createSequentialId('localized-money-input-');
+// sorts the currencies with the following priority:
+// - The selected currency is placed first (e.g EUR)
+// - All other currencies follow, sorted alphabetically as well
+export const sortCurrencies = (selectedCurrency, allCurrencies) => {
+  const remainingCurrencies = without(allCurrencies, selectedCurrency);
+  return [selectedCurrency, ...remainingCurrencies.sort()];
+};
+
+class LocalizedInput extends React.Component {
```
tdeekens: Should this not be called `LocalizedMoneyInput`?

src/components/inputs/localized-money-input/localized-money-input.js
```diff
@@ -0,0 +1,319 @@
...
+    // then input doesn't accept a "currencies" prop, instead all possible
+    // currencies have to exist (with empty or filled strings) on the value:
+    //   { EUR: '12.00', USD: '14.23', CAD: '13.70' }
+    value: PropTypes.objectOf(PropTypes.string).isRequired,
+    onChange: PropTypes.func,
+    selectedCurrency: PropTypes.string.isRequired,
+    onBlur: PropTypes.func,
+    onFocus: PropTypes.func,
+    hideExpansionControls: PropTypes.bool,
+    isDefaultExpanded: (props, propName, componentName, ...rest) => {
```
tdeekens: @dferber90 is this consistent with https://github.com/commercetools/ui-kit/pull/389?

src/components/inputs/localized-money-input/localized-money-input.js
dferber90: This is the single input which then gets reused to build the `LocalizedMoneyInput`. We do the same in `LocalizedTextInput`. It's fine this way (although in hindsight `MoneyInput` would have been a more appropriate name - but then it gets confusing with the `MoneyInput` we already have; so this might actually be the least confusing name).

src/components/inputs/localized-money-input/localized-money-input.js
dferber90: Oh nice catch! Yeah, this needs to be changed indeed! I totally missed it :/

src/components/inputs/localized-money-input/localized-money-input.js
tdeekens: Thanks for clarifying just wasn't sure and wanted to leave a note. Resolving.

src/components/inputs/localized-money-input/README.md
```diff
@@ -0,0 +1,286 @@
...
+// A function to convert a document to form values.
+const docToFormValues = doc => ({
+  // The parseMoneyValue function will turn a MoneyValue into a
+  // value the LocalizedMoneyInput component can handle ({currency: amount})
+  prices: LocalizedMoneyInput.parseMoneyValues(doc.prices),
+});
+
+// a function to convert form values back to a document
+const formValuesToDoc = formValues => ({
+  // The convertToMoneyValue function will turn a LocalizedMoneyInput
```
montezume: ~~turn~~ convert

src/components/inputs/localized-money-input/README.md
```diff
@@ -0,0 +1,286 @@
...
+});
+
+const validate = formValues => {
+  const errors = { prices: {} };
+  Object.keys(formValues.prices).forEach(currency => {
+    errors.prices[currency] = {};
+  })
+  const emptyCurrencies = LocalizedMoneyInput.getEmptyCurrencies(formValues.prices);
+  // ['ERU', 'USD']
+  // THis form dosen't accept highprecision prices
```
montezume: formatting

src/components/inputs/localized-money-input/localized-money-input.js
```diff
@@ -0,0 +1,319 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import without from 'lodash.without';
+import oneLine from 'common-tags/lib/oneLine';
+import { injectIntl } from 'react-intl';
+import Spacings from '../../spacings';
+import Constraints from '../../constraints';
+import {
+  createLocalizedDataAttributes,
+  getHasErrorOnRemainingLanguages,
```
montezume: did we decide where or not it makes sense to make these into their own functions, instead of reusing the language ones?

src/components/inputs/localized-money-input/localized-money-input.js
montezume: I believe it should be 

```js
    hideCurrencyExpansionControls: PropTypes.bool,
    isDefaultExpanded
```

right?

dferber90: > so `isDefaultExpanded` will be `defaultExpandCurrencies`?

Exactly!

And `hideExpansionControls` should be renamed to `hideCurrencyControls`.

> this doesn't sound like a boolean prop, or did I misunderstand the change here

That's true, it doesn't sound like a boolean prop.

There is an "unspoken" convention in React where props which start with `default` will be respected only as an initial value. Changes to that prop will be ignored. The `default` prefix is used when that prop results in an [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html) behaviour for that aspect. An example of this is `input` which can be used as a controlled component `<input value="foo" onChange={fn} />` or as an uncontrolled component `<input defaultValue="foo" />`.

I found the `default` prefix to be of higher importance than having the boolean prefix, as it's more important to signal that this is a prop which is only used to determine the initial state than it is to signal that it contains a boolean.

I could have named it `defaultAreCurrenciesExpanded` or something, but I felt that it's good enough and more readable as the names are right now.

islam3zzat: Thanks for the clarification

dferber90: What we want is:

```js
defaultExpandCurrencies // replaces isDefaultExpanded
hideCurrencyExpansionControls // replaces hideExpansionControls
```

islam3zzat: done

dferber90: There are still references to `isDefaultExpanded` in the README and in a spec file 🔍

islam3zzat: sorry forgot to push :D

src/components/inputs/localized-money-input/localized-money-input.js
dferber90: Maybe renaming the function would be good enough?

dferber90:
I merged as @islam3zzat didn't have the permissions to do so himself. Nice to have this in 🌈

islam3zzat:
Thanks @dferber90

PR #343, opened 2018-12-18
Temporarily disable percy

dferber90:
We ran over the limit now

> $ react-percy
> Compiling...
> Uploading 48 snapshots to Percy
> Error: This organization has exceeded the limits of the (unactivated) plan. Administrators can upgrade here: https://percy.io/organizations/commercetools-GmbH/billing

We can't upgrade using their UI as we are one a custom plan at the moment (since we signed up through GitHub). I already reached out to their support and requested to be moved to a different plan. We will reenable Percy once the bigger plan is activated or when the next month starts and the budget is reset in case we were not upgraded in the meantime.

cc @montezume FYI


montezume:
😢

PR #346, opened 2018-12-18
Add TimeField

dferber90:
Add `TimeField` component.

image

"Unrelated" improvements:
- adds `onFocus` to `TimeInput`
- shows default placeholders in `DateField` story


src/components/fields/time-field/time-field.spec.js
```diff
@@ -0,0 +1,156 @@
...
+      wrapper = shallow(<TimeField {...props} />);
+    });
+    it('should mark the NumberInput as erroneous', () => {
+      expect(wrapper.find(TimeInput)).toHaveProp('hasError', true);
+    });
+    it('should forward the error', () => {
+      expect(wrapper.find(FieldErrors)).toHaveProp('errors', props.errors);
+    });
+  });
+});
```
tdeekens: rtl migration is next?

dferber90: I'm really not sure what happened. I duplicated `TextField` as the basis of this. Maybe I duplicated things before I pulled `master` again. Not sure. This is here by accident. I'll add another commit to upgrade it. Thanks for catching it!

dferber90:
Fixed some more things on this PR now (see description).

Issue #325, opened 2018-12-12
Remove unsupported horizontalConstraints


dferber90:
Right now all of our components support all values of `horizontalConstraint` even if the constraint is too narrow for that particular component (e.g. in date inputs).

We want to go through all components and drop support of values for horizontalConstraint which don't make sense for that particular component (by dropping the values from the prop-types).

montezume:
# Remove unsupported horizontalConstraints

#### FieldLabel

Drop `xs`

<details>
<summary>Show screenshots</summary>

**xs**
![screen shot 2019-01-14 at 10 24 36 am](https://user-images.githubusercontent.com/2425013/51104588-dbdb2580-17e6-11e9-9f57-6f199550012e.png)
</details>

## Fields

#### Select fields

This covers `AsyncCreatableSelectField`,`AsyncSelectField`,  `CreatableSelectField` and `SelectInput`

Drop `xs`

<details>
<summary>Show screenshots</summary>
**xs**

![screen shot 2019-01-14 at 10 28 05 am](https://user-images.githubusercontent.com/2425013/51104723-278dcf00-17e7-11e9-8c2e-3009fcde64db.png)

</details>

#### Date inputs and fields 

This covers `DateField`,`DateRangeField` and `DateTimeField`.

Drop `xs`, `s`

<details>
<summary>Show screenshots</summary>
**xs**

![screen shot 2019-01-14 at 10 37 56 am](https://user-images.githubusercontent.com/2425013/51105138-7c7e1500-17e8-11e9-9cbe-d6327e576601.png)

**s**
![screen shot 2019-01-14 at 10 38 01 am](https://user-images.githubusercontent.com/2425013/51105147-7f790580-17e8-11e9-8ed0-8a60e5cebca4.png)

</details>

#### Multiline text fields

This covers `LocalizedMultilineField` and `MultilineTextField`
* (for any "localized" input, we need to drop xs and s because the bottom "show all languages" overlaps passed the S width anyways...

Drop `xs` and `s`


<details>
<summary>Show screenshots</summary>

**xs**

![screen shot 2019-01-14 at 10 49 01 am](https://user-images.githubusercontent.com/2425013/51105711-0aa6cb00-17ea-11e9-9ca6-f4fc3eb106aa.png)


**s**

![screen shot 2019-01-14 at 10 48 56 am](https://user-images.githubusercontent.com/2425013/51105716-0e3a5200-17ea-11e9-804e-0862cff5b807.png)

</details>

[...]

PR #348, opened 2018-12-18
Add DateRangeField

dferber90:
Add `DateRangeField`

image

### DIFF
--- a/src/components/fields/date-range-field/README.md
+++ b/src/components/fields/date-range-field/README.md
@@ -1,0 +1,59 @@
+# DateRangeField
+
+#### Description
+
+A controlled date input component for a date range.
+
+## Usage
+
+```js
+import { DateRangeField } from '@commercetools-frontend/ui-kit';
+
+<DateRangeField
+  title="Release Date"
+  value={['2018-09-20', '2018-09-24']}
+  onChange={event => alert(event.target.value)}
+/>;
+```
+
+#### Properties
+
+| Props                  | Type               | Required | Values                             | Default | Description                                                                                                                                                                                                                                                           |
+| ---------------------- | ------------------ | :------: | ---------------------------------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


src/components/fields/date-range-field/README.md
```diff
@@ -0,0 +1,59 @@
...
+| ---------------------- | ------------------ | :------: | ---------------------------------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`                   | `string`           |    -     | -                                  | -       | Used as HTML `id` property. An `id` is auto-generated when it is not specified.                                                                                                                                                                                       |
+| `horizontalConstraint` | `object`           |          | `xs`, `s`, `m`, `l`, `xl`, `scale` | `scale` | Horizontal size limit of the input fields.                                                                                                                                                                                                                            |
```
montezume: let's make our lives easier? #325

src/components/fields/date-range-field/date-range-field.js
```diff
@@ -0,0 +1,108 @@
...
+
+const hasErrors = errors => errors && Object.values(errors).some(Boolean);
+
+class DateRangeField extends React.Component {
+  static displayName = 'DateRangeField';
+
+  static propTypes = {
+    // DateRangeField
+    id: PropTypes.string,
+    horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
```
montezume: I would remove support for the unsupported one here but 🤷‍♂️

src/components/fields/date-range-field/README.md
dferber90: I want to do that in one go at the end instead of having to think about it for every component now. I want to tackle it once for all fields instead of ending up with a mix of some components only offering the values they support while others still offer all values.

It's also easier to communicate that we didn't do this at all so far and will tackle it once later instead of trying to communicate that we only did it for some and not for others and then consumers can never be sure which ones are supported and which ones aren't.

PR #349, opened 2018-12-18
Add type to buttons

### LINKED ISSUE #340: Support submit-on-enter for buttons
**Describe the bug**
When working on forms, it's normal to use buttons with `type="submit"` that will cause the form to be submitted on **Enter**. At the moment, our buttons do not support that.

**Expected behavior**
I can use a button that behaves like a "submit", so I can press **Enter** to submit the form.

dferber90:
This PR adds the `type` prop to multiple buttons (`FlatButton`, `IconButton`, `PrimaryButton`, `SecondaryButton`, `SecondaryIconButton`) to enable using them as submit/reset buttons in forms.

However the `FlatButton` already had a prop called `type` which only supported `primary` or `secondary` as values. That prop was renamed to `tone`. The naming is not ideal but good enough until we redo the buttons. The `PrimaryButton` also uses a `tone` prop at the moment. Note that the usage of `tone` is a temporary change and is supposed to be redone once the buttons get reorganised.

I further did not add a `type` prop to the `LinkButton` as it's not really a button in the first place. It always renders a link. We should have a *Link*  category and this should probably be a `ButtonLink` (as it looks like a button but is a link).

Closes #340

**BREAKING CHANGE**: The `type` prop of `FlatButton` was renamed to `tone`. A new `type` prop was added to `FlatButton` instead for which possible values are `submit`, `reset` and `button`.

### DIFF
--- a/src/components/buttons/accessible-button/README.md
+++ b/src/components/buttons/accessible-button/README.md
@@ -42,6 +42,7 @@ export default Button;
 | Property         | Type     | Required? | Description                                                                                                   |
 | :--------------- | :------- | :-------: | :------------------------------------------------------------------------------------------------------------ |
 | id               | String   |     -     | Used as the HTML `id` attribute.                                                                              |
+| type             | String   |     -     | Used as the HTML `type` attribute. Supported values are type: `submit`, `reset`, `button`.                    |
 | label            | String   |     ✓     | The text that should be used as `aria-label`                                                                  |
 | onClick          | Function |     ✓     | Callback that is called when the button is clicked                                                            |
 | isToggled        | Bool     |     -     | Set this prop to indicate a toggle button and it's state                                                      |
--- a/src/components/buttons/accessible-button/README.md
+++ b/src/components/buttons/accessible-button/README.md
@@ -42,7 +42,7 @@ export default Button;
 | Property         | Type     | Required? | Description                                                                                                   |
 | :--------------- | :------- | :-------: | :------------------------------------------------------------------------------------------------------------ |
 | id               | String   |     -     | Used as the HTML `id` attribute.                                                                              |
-| type             | String   |     -     | Used as the HTML `type` attribute. Supported values are type: `submit`, `reset`, `button`.                    |
+| type             | String   |     -     | Used as the HTML `type` attribute. Supported values are `submit`, `reset` and `button`.                       |
 | label            | String   |     ✓     | The text that should be used as `aria-label`                                                                  |
 | onClick          | Function |     ✓     | Callback that is called when the button is clicked                                                            |
 | isToggled        | Bool     |     -     | Set this prop to indicate a toggle button and it's state                                                      |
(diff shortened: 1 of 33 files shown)


src/components/buttons/accessible-button/README.md
```diff
@@ -42,6 +42,7 @@ export default Button;
 | Property         | Type     | Required? | Description                                                                                                   |
 | :--------------- | :------- | :-------: | :------------------------------------------------------------------------------------------------------------ |
 | id               | String   |     -     | Used as the HTML `id` attribute.                                                                              |
+| type             | String   |     -     | Used as the HTML `type` attribute. Supported values are type: `submit`, `reset`, `button`.                    |
```
montezume: supported values are `submit`, `reset` and `button`.

src/components/buttons/icon-button/icon-button.percy.js
```diff
@@ -140,5 +140,35 @@ screenshot('IconButton', () => (
         theme="blue"
       />
     </Spec>
+
+    <Spec label="type - when button">
```
montezume: What would be visually different in these cases? Doesn't seem like we need percy tests for this 🤔 . I would remove them.

src/components/buttons/secondary-button/secondary-button.percy.js
```diff
@@ -81,5 +81,17 @@ screenshot('SecondaryButton', () => (
     <Spec label="when used as link">
       <SecondaryButton label="A label text" linkTo="/" />
     </Spec>
+
+    <Spec label='when type is "button"'>
```
montezume: would remove these as well

src/components/buttons/icon-button/icon-button.percy.js
dferber90: I agree that it's unlikely that this breaks. However browsers might have default styles for different button types or somebody might confuse `type` for `tone` while working on the component. This is here to ensure that there are none and the buttons look the same - no matter the type. There's no cost to adding more specs (aside from more to review) and we don't need to worry about it again once it's approved. I'd keep them for peace of mind.

That said, buttons will probably be redone soonish anyways so I we don't need to argue about this a lot. If you feel strongly about it, please go ahead and just push onto this branch :)

dferber90:
Thanks for fixing things @montezume. Some tests broke due to the added warning (because we were passing `type` and `linkTo` in our tests). I pushed another commit removing the `type` in those cases to avoid the warning. All working again 👍

src/components/buttons/secondary-button/secondary-button.spec.js
```diff
@@ -343,7 +343,7 @@ describe('interaction', () => {
     let wrapper;
     describe('when not disabled', () => {
       beforeEach(() => {
-        props = createProps({ linkTo: '/foo/bar' });
+        props = createProps({ type: undefined, linkTo: '/foo/bar' });
```
montezume: nice

PR #350, opened 2018-12-18
Reorganize README & CONTRIBUTING and add testing docs

dferber90:
This PR reorganises the README by moving less relevant sections out to CONTRIBUTING or collapsing them behind a `details` component. 

This PR also adds a new [*Testing*  section](https://github.com/commercetools/ui-kit/blob/f247c2ddf48608f04388dd3bbd1e1a3cbb070ed2/CONTRIBUTING.md#testing) which explains the differences between our testing approaches and the decisions made there.

*I've had this PR sitting for a while but wanted to wait until we've got Visual Regression Testing up and running.*


CONTRIBUTING.md
```diff
@@ -18,7 +18,117 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+  - use visual testing instead (see below)
+- if mocking is required, try to do it at the level the user interacts with the component
+  - Examples
+    - good: mocking network responses
+    - bad: mocking a fetching component
+  - Try to keep mocking minimal. The less we mock, and the later we mock, the more our tests will cover and the more confidence we gain that our components work
+
+#### Visual testing
+
+- we will use a tool called [Percy](https://percy.io/) for visual regression testing
```
emmenko:
```suggestion
- we use a tool called [Percy](https://percy.io/) for visual regression testing
```

CONTRIBUTING.md
```diff
@@ -18,7 +18,117 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+- we test the components from the produced bundle
+  - this gives us confidence that the bundling process works
+  - this tests components under conditions closer to how consumers would use them
+  - this helps us to avoid accidentally not exposing implemented components from the main `index.js` file
+- we test components in multiple states in one snapshot
+  - it is nice to see all states in one snapshot
+  - it uses less snapshots (we have a monthly budget)
+  - we use custom components called `Suite` and `Spec` to surround our component-under-test, which makes the review easier
+  - we set a min-height for every `Spec` to ensure that changes in height of a component's state don't break the visual diff of states following below that
+- we can currently not "interact" with components (like opening a dropdown menu) before taking a screenshot. We might be able to solve this by using Puppeteer in the future, but we wanted to get up and running first.
```
emmenko:
```suggestion
- at the moment we cannot "interact" with components (like opening a dropdown menu) before taking a screenshot. We might be able to solve this by using Puppeteer in the future, but we wanted to get up and running first.
```

CONTRIBUTING.md
```diff
@@ -18,7 +18,117 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+
+#### Bundle testing
+
+We have test files ending with `*.bundlespec.js`. They are used to test the bundle produced by the rollup build. It is necessary to run the build before running these tests. We use these to further ensure that our bundling process works.
```
emmenko:
```suggestion
We have test files ending with `*.bundlespec.js`. They are used to test the bundle produced by the rollup build, to ensure that our bundling process works.

Note that it is necessary to run the build before running these tests.
```

dferber90: This suggestion is missing the part which says that one needs to build before running the test. And I don't like the repeated use of "to ... to ...". So I left this one out. Thanks for the other two suggestions!

emmenko: That line is there is the suggestion, didn’t you see it? 😅

CONTRIBUTING.md
```diff
@@ -18,7 +18,119 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+  - an `it` can describe the flow of a user
+  - it is okay to have many `expect` statements in an `it`
+  - code comments can help explain the `expect`s better
+- avoid testing for visuals (classnames or the DOM structure)
+  - the user does not care about the DOM structure or the classnames themselves
+  - use visual testing instead (see below)
+- if mocking is required, try to do it at the level the user interacts with the component
+  - Examples
+    - good: mocking network responses
+    - bad: mocking a fetching component
```
montezume: maybe we can give a more relevant example to UI Kit, since it's unlikely we will ever have network requests in the lib itself 🤔

CONTRIBUTING.md
```diff
@@ -18,7 +18,119 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+  - Try to keep mocking minimal. The less we mock, and the later we mock, the more our tests will cover and the more confidence we gain that our components work
```
montezume: I'm not sure what this means. Can you explain it in a different way? 😇

CONTRIBUTING.md
```diff
@@ -18,7 +18,119 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+  - it uses less snapshots (we have a monthly budget)
+  - we use custom components called `Suite` and `Spec` to surround our component-under-test, which makes the review easier
+  - we set a min-height for every `Spec` to ensure that changes in height of a component's state don't break the visual diff of states following below that
+- at the moment we cannot "interact" with components (like opening a dropdown menu) before taking a screenshot. We might be able to solve this by using Puppeteer in the future, but we wanted to get up and running first.
+
+#### Bundle testing
+
+We have test files ending with `*.bundlespec.js`. They are used to test the bundle produced by the rollup build, to ensure that our bundling process works.
+
+Note that it is necessary to run the build before running these tests.
```
montezume: to run `build` before ...

CONTRIBUTING.md
```diff
@@ -18,7 +18,119 @@ These are general and loose rules components in UI Kit should strive to fulfill.
...
+  - the user does not care about the DOM structure or the classnames themselves
+  - use visual testing instead (see below)
+- if mocking is required, try to do it at the level the user interacts with the component
+  - Examples
+    - good: mocking network responses
+    - bad: mocking a fetching component
+  - Try to keep mocking minimal. The less we mock, and the later we mock, the more our tests will cover and the more confidence we gain that our components work
+
+#### Visual testing
+
```
montezume: maybe we should add the same as in Bundle Testing and explain that they need to run `build`?

CONTRIBUTING.md
dferber90: I was trying to come up with a better example too, but I couldn't really. Maybe this is better?

Maybe:
- good: mocking `localStorage`
- bad: mocking a whole component which uses local storage

It's unlikely too. It's more about getting readers in the mindset that "the more tests reflect how a user uses an application, the more confidence they can give you" than about providing an actual example. So we should use mocks at the intersection with a user (DOM, network, local-storage, cookies) instead of mocking implementation parts of the app. Exceptions apply where it's must more convenient (like `ApolloMockProvider`) and worth the tradeoff.

CONTRIBUTING.md
dferber90: I tried to reword it in https://github.com/commercetools/ui-kit/pull/350#discussion_r242853430. It boils down to the famous "[The more your tests resemble the way your software is used, the more confidence they can give you.](https://twitter.com/kentcdodds/status/977018512689455106)."

README.md
```diff
@@ -39,6 +39,9 @@ import {
 
 When you are developing your application using the UI Kit components, chances you want to use the same design tokens as our design system. We provide them through both **CSS variables** and **JavaScript variables**.
```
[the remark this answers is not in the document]
dferber90: I'd prefer without "the"

PR #358, opened 2018-12-21
Convert MultilineTextField

dferber90:


src/components/fields/multiline-text-field/multiline-text-field.spec.js
```diff
@@ -1,189 +1,181 @@
...
+
+const renderMultilineTextField = (props, options) =>
+  render(<Story {...props} />, options);
+
+it('should render a text field', () => {
+  const { getByLabelText } = renderMultilineTextField();
+  expect(getByLabelText('MultilineTextField')).toBeInTheDocument();
+});
+
+it('shoult render a title', () => {
```
tdeekens: shault!

src/components/fields/multiline-text-field/multiline-text-field.spec.js
```diff
@@ -1,189 +1,181 @@
...
-    it('should mark the MultilineTextInput as default closed', () => {
-      expect(wrapper.find(MultilineTextInput)).toHaveProp(
-        'isDefaultClosed',
-        true
-      );
-    });
+describe('when required', () => {
+  it('should add `*` to title`', () => {
+    const { getByText } = renderMultilineTextField({ isRequired: true });
+    expect(getByText('*')).toBeInTheDocument();
```
tdeekens: I guess this would pass if I have a title of * but not a isRequired passed. I would guess it is gine 😌

dferber90: Yeah it would, but the Visual Regression Test catches this anyways 🤷‍♂️

PR #361, opened 2018-12-22
Table Exploration

dferber90:
> ⚠️ **This PR is not intended to be merged!**

kapture 2018-12-23 at 11 33 47

> Checkout [the demo](https://deploy-preview-361--mc-uikit.netlify.com/?selectedKind=Table&selectedStory=cool%20table&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)

I briefly looked into using CSS grid for our table as I was curious about it and wanted to share my findings.

## Findings

- the code to create the table is very straightforward
- CSS grid is awesome
- sticky headers are no problem
- columns with different sizes are easy
- no virtualisation, so browser search works
- horizontal scrolling is straightforward
- limited height table is straightforward

## Problems

- CSS grid is limited to 999 rows in Chrome and 9999 rows in Firefox

## Unknowns

I did not look into these things.

- sticky columns (Briefly tried using `position: sticky; left: 0` and seemed okay. Styling was missing though)

## Complexity

To compare, this table consists of roughly 80 lines of code (HTML & CSS) while the original table needs almost 800 lines of code (HTML & CSS), but it also needs all `react-virtualized` itself! Of course the `CoolTable` does a bit less but I doubt it would get as complex.

### DIFF
--- a/src/components/table/cooltable/cooltable.js
+++ b/src/components/table/cooltable/cooltable.js
@@ -17,16 +17,20 @@ export default class CoolTable extends React.Component {
     ).isRequired,
     renderItem: PropTypes.func.isRequired,
     maxHeight: PropTypes.number,
+    maxWidth: PropTypes.number,
   };
   render() {
     return (
       <div
         className={styles.grid}
         style={{
-          gridTemplateColumns: `repeat(${this.props.columns.length}, auto)`,
+          gridTemplateColumns: this.props.columns
+            .map(column => column.width || 'auto')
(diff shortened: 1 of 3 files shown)


emmenko:
I like this!! Given our past experience, the simpler the Table, the better 😇

dferber90:
I'll close this for now, but I really like this approach and thing it might be worth to continue investigating .

PR #362, opened 2018-12-23
Add `isAutofocussed` and `isReadOnly` to `MoneyInput` and `MoneyField`

### LINKED ISSUE #330: `MoneyInput` improvements

montezume:
Closes #330 

#### Summary

* Adds `isAutofocussed` prop to `<MoneyInput>` and `<MoneyField>`
* Adds `isReadOnly` prop  to `<MoneyInput>` and `<MoneyField>`
* `<MoneyInput>`: Wraps `SingleValue` of react-select with a label pointing at the currencyCode input, so that it can be targeted using RTL.

### DIFF
--- a/src/components/fields/money-field/money-field.spec.js
+++ b/src/components/fields/money-field/money-field.spec.js
@@ -46,9 +46,6 @@ class Story extends React.Component {
         <label htmlFor={MoneyInput.getAmountInputId(this.props.id)}>
           Amount
         </label>
-        <label htmlFor={MoneyInput.getCurrencyDropdownId(this.props.id)}>
-          Currency Code
-        </label>
         <MoneyField
           {...this.props}
           value={this.state.value}
@@ -65,7 +62,7 @@ const renderMoneyField = (props, options) =>
 it('should render a money field', () => {
   const { getByLabelText } = renderMoneyField();
   expect(getByLabelText('Amount')).toBeInTheDocument();
-  expect(getByLabelText('Currency Code')).toBeInTheDocument();
+  expect(getByLabelText('EUR')).toBeInTheDocument();
 });
 
 it('should render a title', () => {
@@ -97,8 +94,8 @@ it('should call onFocus when amount input is focused', () => {
 it('should call onFocus when currency select is focused', () => {
   const onFocus = jest.fn();
   const { getByLabelText } = renderMoneyField({ onFocus });
-  getByLabelText('Currency Code').focus();
-  expect(getByLabelText('Currency Code')).toHaveFocus();
+  getByLabelText('EUR').focus();
+  expect(getByLabelText('EUR')).toHaveFocus();
   expect(onFocus).toHaveBeenCalled();
 });
 
@@ -115,10 +112,10 @@ it('should call onBlur when amount input loses focus', () => {
 it('should call onBlur when currency select loses focus', () => {
   const onBlur = jest.fn();
   const { getByLabelText } = renderMoneyField({ onBlur });
-  getByLabelText('Currency Code').focus();
-  expect(getByLabelText('Currency Code')).toHaveFocus();
-  getByLabelText('Currency Code').blur();
-  expect(getByLabelText('Currency Code')).not.toHaveFocus();
+  getByLabelText('EUR').focus();
+  expect(getByLabelText('EUR')).toHaveFocus();
+  getByLabelText('EUR').blur();
+  expect(getByLabelText('EUR')).not.toHaveFocus();
   expect(onBlur).toHaveBeenCalled();
 });
 
@@ -131,8 +128,8 @@ it('should call onChange when changing the value', () => {
   });
   fireEvent.change(getByLabelText('Amount'), { target: { value: '20' } });
 
-  fireEvent.focus(getByLabelText('Currency Code'));
-  fireEvent.keyDown(getByLabelText('Currency Code'), { key: 'ArrowDown' });
+  fireEvent.focus(getByLabelText('EUR'));
+  fireEvent.keyDown(getByLabelText('EUR'), { key: 'ArrowDown' });
 
   expect(onChange).toHaveBeenCalledTimes(1);
   expect(onChange).toHaveBeenCalledWith({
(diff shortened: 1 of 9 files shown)


src/components/inputs/money-input/money-input.js
```diff
@@ -14,12 +15,25 @@ import currencies from './currencies.json';
 import createSelectStyles from '../../internals/create-select-styles';
 import vars from '../../../../materials/custom-properties.json';
 
+const SingleValue = ({ id, ...props }) => (
+  <components.SingleValue {...props}>
+    <label htmlFor={id}>{props.children}</label>
```
montezume: wrap the value with a label pointing at react-select's input.

src/components/fields/money-field/money-field.spec.js
```diff
@@ -65,7 +62,7 @@ const renderMoneyField = (props, options) =>
 it('should render a money field', () => {
   const { getByLabelText } = renderMoneyField();
   expect(getByLabelText('Amount')).toBeInTheDocument();
-  expect(getByLabelText('Currency Code')).toBeInTheDocument();
+  expect(getByLabelText('EUR')).toBeInTheDocument();
```
tdeekens: 💯

src/components/fields/money-field/money-field.spec.js
```diff
@@ -115,13 +112,20 @@ it('should call onBlur when amount input loses focus', () => {
...
-  getByLabelText('Currency Code').blur();
-  expect(getByLabelText('Currency Code')).not.toHaveFocus();
+  getByLabelText('EUR').focus();
+  expect(getByLabelText('EUR')).toHaveFocus();
+  getByLabelText('EUR').blur();
+  expect(getByLabelText('EUR')).not.toHaveFocus();
   expect(onBlur).toHaveBeenCalled();
 });
 
+it('should have focus the amount input automatically when isAutofocussed is passed', () => {
```
tdeekens: should have focussed? or should focus the?

src/components/inputs/money-input/money-input.js
```diff
@@ -328,7 +349,9 @@ class MoneyInput extends React.Component {
     onFocus: PropTypes.func,
     onBlur: PropTypes.func,
     isDisabled: PropTypes.bool,
-    onChange: PropTypes.func.isRequired,
+    isReadOnly: PropTypes.bool,
+    isAutofocussed: PropTypes.bool,
+    onChange: requiredIf(PropTypes.func, props => !props.isReadOnly),
```
tdeekens: Curious: is this a general pattern across other inputs or what we strive for and should change elsewhere to if not already?

src/components/inputs/money-input/money-input.js
```diff
@@ -502,12 +527,18 @@ class MoneyInput extends React.Component {
           }}
         >
           <Select
-            inputId={MoneyInput.getCurrencyDropdownId(this.props.id)}
+            inputId={id}
             name={getCurrencyDropdownName(this.props.name)}
             value={option}
-            isDisabled={this.props.isDisabled || hasNoCurrencies}
+            isDisabled={
+              this.props.isDisabled || hasNoCurrencies || this.props.isReadOnly
```
tdeekens: Should the select rather also be in readonly mode if this component is in it? Feels a bit odd to mix states in that readonly in one becomes disabled on another.

montezume: @tdeekens react-select doesn't seem to have a `readonly` prop [ref](https://react-select.com/props)

src/components/inputs/money-input/money-input.js
montezume: it's the same in `LocalizedTextInput`, and quite a few inputs have quite a bit of copy / paste from that input

src/components/fields/money-field/money-field.spec.js
montezume: 🙈

montezume:
Open question:

I'm considering adding a prop, something like (isCurrencySelectionDisabled), where under this case, instead of rendering a react-select, we will render a normal label with it's for pointing at the amountInputId. That way, the user can click on `EUR`, and focus on the input-able input. Same as with LocalizedTextInput. This would be helpful for #339. 

Any comments @dferber90 @tdeekens

dferber90:
This already happens (except for the label part) when you pass no currencies to `MoneyInput`. It might make sense to also disable the selection it in case only one currency is passed.

I'm undecided about the API. Adding a separate prop opens up new combinations which we'd have to give answers for. What would happen for:
- `isCurrencySelectionDisabled: false` and `currencies: []`
- `isCurrencySelectionDisabled: false` and `currencies: ['EUR']`
- `isCurrencySelectionDisabled: false` and `currencies: undefined`
- `isCurrencySelectionDisabled: true` and `currencies: ['EUR']`

It seems easier to say that "currency selection will be enabled when more than one currency is passed".

I don't see users wanting to dynamically switch between having currency selection enabled and disabled. It seems more like something that's decided while building a feature. So devs can then decide to omit `currencies`. Even if it's needed dynamically they could do `currencies={isCurrencySelectionDisabled ? [] : ['EUR', 'USD']}`.

Let's have a chat about it 📞

montezume:
Okay in [1de572c](https://github.com/commercetools/ui-kit/pull/362/commits/1de572c0b0f772accfdfb181f9e4d2e6dd4fc38d) I removed `isCurrencySelectDisabled`, and reworked the component to display a label with an htmlFor pointing at the amount input id if there are no currencies.

montezume:
I found too that the current docs are better represented by this new change, than by the current state in master 😆 
```
List of possible currencies. When not provided or empty, the component renders a label with the value's currency instead of a dropdown```

PR #370, opened 2018-12-27
Reorganize design tokens and add story

dferber90:
- [Adds a "Basics" section](https://user-images.githubusercontent.com/1765075/50478396-254b0880-09d1-11e9-85b6-6307f63bcfbf.png) to Storybook
- [Adds a Tokens story](https://user-images.githubusercontent.com/1765075/50478408-372cab80-09d1-11e9-8804-113b439c087a.png)
- Removes `--token` prefix from all tokens
- Reorganises how design tokens are created (uses a single validated `definition.yaml`) instead of separate *base*-files. Creates `custom-properties.json` and `custom-properties.css` from that file through `generate-custom-properties.js` script.

## `definition.yaml`

This definition file is read by `generate-custom-properties.js` when the script is called to create `custom-properties.{json,css}`. The generation of `custom-properties.{json,css}` fails in case an invalid design token was provided (e.g. one not matching the naming pattern, or one not using any of the predefined choices).

## Token Story

A new Storybook section shows the design tokens we are now able to use. The tokens can be searched. The choices, component groups and available states are shown as well.

---

BREAKING CHANGE: All custom properties no longer use the `--token` prefix.


materials/internals/definition.yaml
```diff
@@ -0,0 +1,325 @@
...
+  #     size-height-input:
+  #       description: ''
+  #       value: 32px
+  #     size-height-tag:
+  #       description: ''
+  #       value: 26px
+
+# These tokens are deprecated and not convertible to the new format
+# as we don't have a group for heights
+plainTokens:
```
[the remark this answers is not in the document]
dferber90: I would have removed them if possible, but they're still in use for the moment. Ultimately we should not use `plainTokens` at all, but it takes a bit to get there.

The current situation is that we use a mix of directly using choices (like `color-green`) and tokens. The tokens we use are old and don't follow the new conventions. They are also not in line with our vision. And they don't even follow the previously established structure (e.g. `  small-button-height` vs `size-height-input` follow completely different naming schemas).

The tokens in use don't follow the `<property>-for-<component-group>-when-<state>` [Token Structure](https://github.com/commercetools/ui-kit/blob/5b27e9beccb63f9d437b5522cc565127d6e44805/philosophy/design-tokens.md#15-token-structure). So all of our current tokens are deprecated (indicated by the `deprecated: true` flag on each decision in `decision.yaml`).

All tokens need to be converted to the new format (by dropping the deprecated token and introducing a new token) and we need to rethink how much sense some of those tokens even make in the first place. This takes some time and requires coordination with the design team. I've been working on this with Sven and we have an idea of where we want to go, which is becoming more and more concrete.

Right now we offer deprecated tokens but we don't offer the "correct" new way yet. The deprecated flag is used to avoid validation of those token names for now and to indicate the deprecated components.

dferber90:
A complete refactoring of how our design-tokens are built and visual regression testing gives us confidence that nothing was screwed up (it actually caught a regression in between)!

materials/internals/definition.yaml
```diff
@@ -0,0 +1,350 @@
...
+  readonly:
+    description: 'When the element can not be modified'
+  warning:
+    description: 'When the element has slightly wrong input'
+  error:
+    description: 'When the element has invalid input'
+
+componentGroups:
+  input:
+    description: 'Input elements'
```
dferber90: I just used the `input` to get us started. Ultimately we identified the following groups. We want to add them as we go instead of defining them all upfront.

- `button`: Clickable elements which trigger an action
- `panel`: Visual grouping
- `dropdown`: Elements which offer more choices in an appearing menu
- `label`: Elements which give meaning to another element and focus the related element when clicked
- `icon`: Icons
- `input`: Elements which get information from the user
- `stamp`: Indicate additional information about an element
- `checkbox`: Elements which allow users to select a subset from a list of options
- `radio`: Elements which allow users to select one option from a list
- `toggle`: Elements which allow users to select a boolean option
- `tag`: Elements which indicate one item of a selection
- `text-headline`: Headline
- `text-subheadline`: Subheadline
- `text-body`: Body Text
- `text-detail`: Detail Text

materials/internals/definition.yaml
```diff
@@ -0,0 +1,350 @@
...
+  pristine:
+    description: 'The element was not interacted with yet'
+  disabled:
+    description: 'When the element can not be interacted with'
+  readonly:
+    description: 'When the element can not be modified'
+  warning:
+    description: 'When the element has slightly wrong input'
+  error:
+    description: 'When the element has invalid input'
```
dferber90: We identified the following states:

| State       | Maps to CSS pseudo class | Description                                                                                           |
| ----------- | ------------------------ | ----------------------------------------------------------------------------------------------------- |
| `visited`   | `:visited`               |  The user has visited this link before                                                                |
| `hovered`   | `:hover`                 |  The pointer is over this element                                                                     |
| `active`    | `:active`                |  The element is currently being activated by the user                                                 |
| `focused`   | `:focus`                 | The element is currently selected to receive input                                                    |
| `disabled`  | `:disabled`              | The element can not be interacted with                                                                |
| `read-only` | `:read-only`             | The element can not be modified                                                                       |
| `checked`   | `:checked`               | The element is checked by the user (e.g. for checkboxes).                                             |
| `warning`   | -                        | The element contains a warning. Warnings do not block the further flow but should be resolved.        |
| `error`     | -                        | The element contains an error. Errors block the further flow and must be resolved.                    |
| `info`      | -                        | The element contains a hint. Hints do not block the further flow and need do not need to be resolved. |

PR #373, opened 2018-12-28
Rework visual regression testing to use Puppeteer and Percy 🎭👺📸

montezume:
### Current approach

Currently, we use [react-percy](https://github.com/percy/react-percy) for our visual regression tests. We did this by creating files ending in `.percy.js`, and inside of those files, calling `percySnapshot` with a react component. We used our custom [Suite](https://github.com/commercetools/ui-kit/blob/38214e8bb9c6cd9689e34a67a0b3c044ed3387e8/test/percy/suite.js) and [Spec](https://github.com/commercetools/ui-kit/blob/38214e8bb9c6cd9689e34a67a0b3c044ed3387e8/test/percy/spec.js) components for those tests.

Our tests looked like this

```js
// primary-button.percy.js

import { PrimaryButton } from '../../../dist/ui-kit.esm';
import { Suite, Spec, screenshot } from '../../../test/percy';

screenshot('PrimaryButton', () => (
  <Suite>
    <Spec label="in default state">
      <PrimaryButton label="Primary button" />
    </Spec>
    <Spec label="when disabled">
      <PrimaryButton disabled={true} label="Primary button" />
    </Spec>
  </Suite>
));
```

When we ran `yarn percy`, `react-percy` would take all files ending in `.percy.js`, and generate one html file for each screenshot assertion within those files. So for ~50 something files, `react-percy` would generate ~50 something html files. Then, behind the scenes, percy would use puppeteer to take screenshots of each of those html files, and upload them to their service.

#### Downsides with this approach

* `react-percy` has been deprecated.
* we don't have any control over how `react-percy` builds our html, all we can do is pass in webpack configuration.
* we were not able to write tests for some of our components, and couldn't figure out why (`<Table>`, `<CollapsiblePanel>`, etc).
* we were not able to interact with our components before taking screenshots
  * since we do not want to add props just for testing, this meant that for some components, like the DatePicker for example, we were not able to take screenshots of it in an open state.

### Changes introduced 📸

Now, we write visual regression tests in a different way. This PR introduces a new folder to the project called `visual-testing-playground`. As a contributor to `UI-Kit`, you probably won't need to make changes to this playground.

#### Create files ending in `.visualroute.js`

Our `visual-testing-playground` detects all files ending in `.visualroute.js`, and creates a route for each of these files. We can then use these routes in our test files ending in `visualspec.js`

#### Write tests in ending `.visualspec.js`

We use a library called [jest-puppeteer](https://github.com/smooth-code/jest-puppeteer) to run our visual tests. We use puppeteer to navigate to routes generated by our `*.visualroute.js` files, and then we use a tool called [Percy](https://percy.io/) to take snapshots of that route.

#### Visual Routes (`*.visualroute.js`)

Now we create visual route files ending in `.visualroute.js`. When the `visual-testing-playground` is built, it will create a react-router route for each of these routes. An example of a visual route:

```js
// primary-button.visualroute.js

import React from 'react';
import { InformationIcon, PrimaryButton } from 'ui-kit';

export const routePath = '/primary-button';

  <Suite>
    <Spec label="in default state">
      <PrimaryButton label="Primary button" />
    </Spec>
    <Spec label="when disabled">
      <PrimaryButton disabled={true} label="Primary button" />
    </Spec>
  </Suite>
);
```

When we run `yarn visual-testing-playground:build` and `yarn visual-testing-playground:serve` we can then navigate to `localhost:3001/primary-button` to see our route. Generally the route we want to follow is to use one route per snapshot. This however, is just the route. To write a complete visual regression test, we need to also create a file ending in `.visualspec.js`

#### Visual tests (`*.visualspec.js`)

```js
// primary-button.visualspec.js

import { percySnapshot } from '@percy/puppeteer';

describe('PrimaryButton', () => {
  beforeAll(async () => {
    await page.goto(`http://${HOST/primary-button`);
  });

  it('PrimaryButton', async () => {
    await expect(page).toMatch('A label text');
    await percySnapshot(page, 'PrimaryButton');
  });
});
```

We can now run our test by running `yarn test:visual`. Notice how before we call `percySnapshot`, we add an expect assertion on some text on the page. This is to ensure that the tests will fail if your route cannot be found. Otherwise, we could end up uploading a lot of `route not founds` to Percy.

#### Interacting with components before taking their snapshot.

Here is an example of a simple test that clicks an element before taking it's snapshot.

```js
// date-input.visualspec.js
import { percySnapshot } from '@percy/puppeteer';

describe('DateInput', () => {
  it('Open', async () => {
    await page.goto(`${HOST}/date-input-open`);
    await page.click('#date-input');
    await expect(page).toMatch('November');
    await percySnapshot(page, 'DateInput - open');
  });
});
```

As you can see, we simply click the `<DateInput>`, and take it's snapshot. You can see the output on percy [here](https://percy.io/commercetools-GmbH/ui-kit/builds/1286423). Works perfectly! 🎉 

### Next steps

With this setup, and with the current `ApplicationContext` and the `react-testing-library` setup in app kit, we can move on to adding VRT tests to the app-kit in the very near future. That's why I tagged you @emmenko 😄 👍 

#### Add `pptr-testing-library`

With [`pptr-testing-library`](https://github.com/patrickhulce/pptr-testing-library), we can write our visual tests more like we write our react-testing-library tests. Then we can use getByTestId.


package.json
```diff
@@ -55,8 +56,9 @@
     "test": "jest --config jest.test.config.js",
     "test:watch": "jest --config jest.test.config.js --watch",
     "test:bundle": "jest --config jest.bundle.config.js",
-    "test:bundle:watch": "jest --config jest.bundle.config.js --watch",
-    "percy": "react-percy"
+    "test:percy": "percy exec -- yarn test:visual",
+    "test:visual": "jest --config jest.visual.config.js --maxWorkers=1",
```
montezume: using multiple workers got errors from puppeteer about reusing requests... it seems to run relatively quickly like this though.

src/components/inputs/date-time-input/date-time-input.visualspec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('DateTimeIpnut', () => {
+  it('Default', async () => {
+    await page.goto('http://localhost:3001/date-time-input');
+    await expect(page).toMatch('minimal');
+    await percySnapshot(page, 'DateTimeInput', { widths });
+  });
+  it('Open', async () => {
+    await page.goto('http://localhost:3001/date-time-input-open');
+    await page.click('#date-time-input');
```
montezume: here's an example of interacting with a component before taking it's picture

src/components/avatar/avatar.visualspec.js
```diff
@@ -0,0 +1,13 @@
+import { percySnapshot } from '@percy/puppeteer';
+import { widths } from '../../../test/percy/widths';
+
+describe('Avatar', () => {
+  beforeAll(async () => {
+    await page.goto('http://localhost:3001/avatar');
+  });
+
+  it('Avatar', async () => {
+    await expect(page).toMatch('when gravatar hash is known');
```
montezume: I just like having these expect's because if your route isn't loaded, the test fails. Else the tests will pass even if the route doesn't exist... Would be nice actually to have titles or something for each snapshot and display those.

montezume:
Seem to be having a problem where locally it builds properly, but on CI it's missing all the @emotion/styled styles...

package.json
emmenko: For sequential runs, you can use `—runInBand`

vrt-examples/src/App.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
+
+const componentsContext = require.context(
+  '../../src/components',
+  true,
+  /\.example\.js$/
```
montezume: load all files ending in `example.js`

vrt-examples/src/App.js
```diff
@@ -0,0 +1,39 @@
...
+  return components;
+}, {});
+
+class App extends React.Component {
+  render() {
+    return (
+      <Router>
+        <Switch>
+          <Route path="/" exact component={() => <div>Home</div>} />
+          {Object.values(allComponents).map(Component => (
```
montezume: create a route from all `example.js` files, using their exported `routePath` and `component`

.travis.yml
```diff
@@ -16,15 +16,18 @@ script:
...
-  - "! test -f fail-tests-because-there-was-an-unhandled-rejection.lock"
+  - yarn run jest --projects jest.bundle.config.js --maxWorkers=4 --reporters jest-silent-reporter
   # skip percy builds when this is not a pull request or when it's not master to save precious snaps
   # also skip when PR is opened by `renovate`
-  - bash ./scripts/run-percy.sh
   # back to default in case internal travis scripts return non zero response codes.
+  # - bash ./scripts/run-percy.sh
+  - yarn vrt-playground:build
+  - yarn percy
+  - "! test -f fail-tests-because-there-was-an-unhandled-rejection.lock"
```
dferber90: Did you move this line down on purpose?

I had it before percy to prevent us from running percy in case there was an unhandled rejection in earlier tests. I would keep it right after `yarn run jest --projects jest.test.config.js`.

vrt-examples/webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+  } catch (e) {
+    return null;
+  }
+})();
+
+if (!info) {
+  // We can only start ui-kit when it was built first
+  console.info(
+    '\x1b[33m%s\x1b[0m', // log in yellow
+    '⚠️  You need to run "yarn build" or "yarn build:watch" before starting the playground!'
```
dferber90: This message should be adapted as it's not a playground.

vrt-examples/src/index.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
dferber90: Reminder that you still wanted to rename `vrt-examples` to something else.

I wonder whether we could eventually somehow turn this into a reusable package we publish on npm? I don't have too clear of a picture of what it would entail though.

.travis.yml
montezume: ah yeah, good call

vrt-examples/src/index.js
montezume: yeah, let's do some brainstorming on naming. Shopify have a folder called `playground` and a folder called `examples`. We might even just want to merge this with our current playground setup.

.eslintrc.yaml
```diff
@@ -24,18 +24,16 @@ plugins:
...
   - 'react'
-  - 'react-percy'
 overrides:
   - files: ['*.spec.js']
     rules:
       'react/display-name': 0
-  - files: ['*.percy.js', 'test/percy/suite.js']
-    env:
-      'react-percy/globals': true
+  - files: ['*.example.js']
```
dferber90: Maybe using `.visualexample.js` would help see the connection between `visualspec.js` and `visualexample.js` a bit more?

.eslintrc.yaml
```diff
@@ -28,7 +28,7 @@ overrides:
   - files: ['*.spec.js']
     rules:
       'react/display-name': 0
-  - files: ['*.example.js']
+  - files: ['*.visualroute.js']
```
montezume: I think this is more clear now

test/percy/suite.js
```diff
@@ -1,18 +1,15 @@
...
 import PropTypes from 'prop-types';
-import { HashRouter as Router } from 'react-router-dom';
 import { IntlProvider, addLocaleData } from 'react-intl';
 import en from 'react-intl/locale-data/en';
 import { i18n as messages } from '../../dist/ui-kit.esm';
 
 addLocaleData(en);
 
 const Suite = props => (
-  <Router>
```
montezume: instead now we render the Router in App.js

.travis.yml
tdeekens: Is this still valid/open or resolved?

src/components/buttons/primary-button/primary-button.visualspec.js
```diff
@@ -0,0 +1,13 @@
+import { percySnapshot } from '@percy/puppeteer';
+import { widths } from '../../../../test/percy/widths';
```
tdeekens: Why do we mostly import both withs and percySnapshots and not export a pre-filled/wrapped percySnapshot from the percy/utils? Would that shorten and centralise things a bit more?

src/components/fields/async-select-field/async-select-field.visualspec.js
```diff
@@ -0,0 +1,13 @@
+import { percySnapshot } from '@percy/puppeteer';
+import { widths } from '../../../../test/percy/widths';
+
+describe('AsyncSelectField', () => {
+  beforeAll(async () => {
+    await page.goto('http://localhost:3001/async-select-field');
```
tdeekens: 1. Can‘t we just return the promise here instead of having the beforeAll being async?
2. What if we have a gotoComponent exported from the utils somewhere to not repeat localhost and ports here over and over again?

src/components/inputs/date-time-input/date-time-input.visualspec.js
tdeekens: Is this a case for data-testid rathet than an id?

src/components/tag/tag.visualspec.js
```diff
@@ -0,0 +1,13 @@
+import { percySnapshot } from '@percy/puppeteer';
+import { widths } from '../../../test/percy/widths';
+
+describe('Tag', () => {
+  beforeAll(async () => {
+    await page.goto('http://localhost:3001/tag');
+  });
+
+  it('Tag', async () => {
```
tdeekens: Is this „Normal“? Otherwise it is double Tag 😁

src/components/fields/async-select-field/async-select-field.visualspec.js
tdeekens: gotoRoute might be more appropriate as route specs export a route.

src/components/tag/tag.visualspec.js
montezume: you mean the naming? What really matters to use is the name we pass to percySnapshot, as that's what shows up on the dashboard there. The Describe and It blocks are only used by jest puppeteer, which assuming the routes were correctly created, will always pass. The passing and failing of the VRT tests happens in Percy and not here 🤷‍♂️

montezume: Maybe instead of 'Tag', here it would be 'Default'.

src/components/buttons/primary-button/primary-button.visualspec.js
dferber90: We had this in our setup before, where we would preconfigure the snapshot-fn with the widths. I think it makes sense to do the same thing here as well. Nice suggestions! 👍

src/components/tag/tag.visualspec.js
tdeekens: Yes, that‘s what I meant. Describe is component name and it is default.

src/components/buttons/primary-button/primary-button.visualspec.js
montezume: I think instead we can just use percy.yml and specify widths
[ref](https://docs.percy.io/v1/docs/configuration)

src/components/fields/async-select-field/async-select-field.visualspec.js
montezume: I added HOST to the jest globals instead [here](https://github.com/commercetools/ui-kit/pull/373/commits/4abd35969cf95d7076e19c2d229c8bb02ff099c4)

src/components/buttons/primary-button/primary-button.visualspec.js
montezume: I went for that approach instead so that it's easier to understand what's happening with the percySnapshot [ref](https://github.com/commercetools/ui-kit/pull/373/commits/8778d86143c47d93a0ce31a5751943e3543f6f7b)

PR #374, opened 2018-12-29
Add tokens README

dferber90:
Adds an in-depth README to our tokens story.

### DIFF
--- a/materials/internals/TOKENS.md
+++ b/materials/internals/TOKENS.md
@@ -1,0 +1,215 @@
+# Design Tokens
+
+> This document is WIP and subject to change as we learn more about design tokens.
+
+We use design tokens as the smallest form of information in our visual language.
+The tokens give semantic meaning to the plain palette of choices we have.
+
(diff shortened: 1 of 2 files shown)


montezume:
@tdeekens the only thing weird about it is that we have it inside this `internal` directory.

dferber90:
🇨🇦 Thanks for the fixes @montezume! 


> the only thing weird about it is that we have it inside this internal directory.

Yeah, I'm not too happy with the setup either. But it's good enough for now. I'm not too sure what would be better. Maybe organising our folders the way we're going to organise Storybook sections makes sense: See [screenshot](https://user-images.githubusercontent.com/1765075/50559630-5725dd00-0cf9-11e9-8c22-439e07e64c50.png) taken from experimental `df-add-form-philosophy` branch)

```
- Philosophy
- Basics
- Components
- Examples
```

I'm not too sure about the actual sections, but in general it seems like a good idea to use the same hierarchy.

PR #376, opened 2018-12-30
Add storysource

dferber90:
Adds [`@storybook/addon-storysource`](https://www.npmjs.com/package/@storybook/addon-storysource):

image

This allows us to see the source code of our stories which can be useful e.g. in the form examples.


montezume:
Is there a way to rename the tab?

dferber90:
I was also a bit bothered by that, but didn't find a mention of it in the docs. You can check the docs again, maybe I missed it.

PR #381, opened 2019-01-02
chore: improve vrt app and add table vrt test

montezume:
* Add a VRT test for the basic table example
* Improves the VRT app setup a bit by splitting node_modules and ui-kit.
* ~~Use webpack production mode for VRT app build~~

### DIFF
--- a/src/components/table/table.visualspec.js
+++ b/src/components/table/table.visualspec.js
@@ -1,9 +1,7 @@
 import { percySnapshot } from '@percy/puppeteer';
 
 describe('Table', () => {
-  beforeAll(async () => {
-    await page.goto(`${HOST}/table`);
-  });
+  beforeAll(async () => page.goto(`${HOST}/table`));
 
   it('Default', async () => {
     await expect(page).toMatch('Basic example');
--- a/src/components/table/table.visualspec.js
+++ b/src/components/table/table.visualspec.js
@@ -1,0 +1,12 @@
+import { percySnapshot } from '@percy/puppeteer';
+
+describe('Table', () => {
+  beforeAll(async () => {
+    await page.goto(`${HOST}/table`);
+  });
+
+  it('Default', async () => {
+    await expect(page).toMatch('Basic example');
+    await percySnapshot(page, 'Table');
+  });
+});
--- a/visual-testing-app/webpack.config.js
+++ b/visual-testing-app/webpack.config.js
@@ -22,6 +22,7 @@ if (!info) {
 }
 
 module.exports = {
+  target: 'web',
   mode: 'production',
   stats: 'minimal',
   entry: './visual-testing-app/src/index.js',
@@ -31,6 +32,7 @@ module.exports = {
     publicPath: '/',
   },
   optimization: {
+    minimize: false,
     splitChunks: {
       cacheGroups: {
         vendors: {
--- a/visual-testing-app/webpack.config.js
+++ b/visual-testing-app/webpack.config.js
@@ -22,18 +22,38 @@ if (!info) {
 }
 
 module.exports = {
-  mode: 'development',
+  mode: 'production',
   stats: 'minimal',
   entry: './visual-testing-app/src/index.js',
   output: {
-    filename: 'main.js',
+    filename: '[name].js',
     path: path.resolve(__dirname, 'dist'),
+    publicPath: '/',
   },
+  optimization: {
+    splitChunks: {
+      cacheGroups: {
+        vendors: {
+          test: /[\\/]node_modules[\\/]/,
+          name: 'vendors',
+          chunks: 'all',
+          priority: -20,
+        },
+        'ui-kit': {
+          test: /ui-kit.esm/,
+          name: 'ui-kit',
+          chunks: 'all',
+          priority: -15,
+        },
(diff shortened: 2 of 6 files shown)


src/components/table/table.visualspec.js
```diff
@@ -0,0 +1,12 @@
+import { percySnapshot } from '@percy/puppeteer';
+
+describe('Table', () => {
+  beforeAll(async () => {
+    await page.goto(`${HOST}/table`);
```
tdeekens: Nitpick: we can just return the promise. No need for async/await. But it is kind of a 🤷‍♀️

visual-testing-app/webpack.config.js
```diff
@@ -22,18 +22,40 @@ if (!info) {
...
+          test: /[\\/]node_modules[\\/]/,
+          name: 'vendors',
+          chunks: 'all',
+          priority: -20,
+        },
+        'ui-kit': {
+          test: /ui-kit.esm/,
+          name: 'ui-kit',
+          chunks: 'all',
+          priority: -15,
```
tdeekens: Funky numbers?

montezume: copy paste from shopify 😆

tdeekens: Haha. Maybe add a comment then 😂

visual-testing-app/webpack.config.js
```diff
@@ -22,18 +22,40 @@ if (!info) {
 }
 
 module.exports = {
-  mode: 'development',
+  target: 'web',
+  mode: 'production',
```
montezume: this doesn't seem to work with Percy. It might be a bug to do with percy?

When I run `yarn percy` locally after having built the project using mode: 'production', I get unstyled js in css, as shown [here](https://percy.io/commercetools-GmbH/ui-kit/builds/1290053). I even tried adding a wait for 2 seconds in case it was some sort of flash of unstyled content problem, and still got the same problem. When I removed back to development, it worked again as shown [here](https://percy.io/commercetools-GmbH/ui-kit/builds/1290049)

emmenko:
Btw, I had troubles running RTL tests in the app-kit playground to test a view with a Table, I guess because of react-virtualised and the DOM (for instance, the width was always `0`, so the Table was not visible). I'm not sure if you have a similar problem, but I fear that our Table is not really testable with RTL at the moment.

And in case you have encountered this problem before, please let me know what's the trick to fix it 😜

montezume:
@emmenko yeah, we had the same issues with our initial percy implementation. We also had troubles rendering "CollapsiblePanel". However, with the new implementation, _it just works_

visual-testing-app/webpack.config.js
montezume: I got an email back from Percy support. The gist of it is 

>  It looks like emotion uses the CSSOM (CSS Object Model) to inject its styles. Percy currently works by taking a snapshot of the DOM and the external assets linked in that DOM snapshot. So, at the moment, we're missing things in the CSSOM because it's not serialized into the DOM.


So it's not worth trying to solve why we can't use the webpack production build with emotion at this moment.

emmenko: So would it work if we extract the CSS into a file and use that? This would also help with using ui-kit for SSR.

Regarding SSR, it's still not an ideal solution, as we only want to inject the "critical styles" when rendering the page, but it might be a good starting point.
Anyway, this is a bit off topic...

visual-testing-app/webpack.config.js
```diff
@@ -22,18 +22,41 @@ if (!info) {
 }
 
 module.exports = {
+  target: 'web',
+  // note: we have to use 'development' for now, because when we use production
+  // our emotion styles aren't displayed in our percy snapshots.
```
emmenko: Can we also mention some follow up or something, or link to an issue related to this (in case there is one)?

montezume: I emailed support directly so there isn't a github issue. I did create a reproduction repo to show to support [here](https://github.com/montezume/percy-puppeteer-emotion).

visual-testing-app/webpack.config.js
tdeekens: I think for UIKit as a library it's impossible to serve consumers a static `.css` file while fulfilling a potential requirement for only include above the fold css (or critical styles). Also if a static file was offered, and consumers include/use it, UIKit would have to stop inlining styles. In that it would have to know it the static css is/was used. At least in my head it's like a dog catching its tail. 

I think going all in on emotion, also for SRR and static extraction of css, is a good goal but doesn't really relate to Percy or what we're attempting to solve here.

montezume:
Can I get some reviews related to the scope of this PR? (Add VRT test for Table) and split vendors and ui-kit from main bundle in visual-testing-app webpack config? 🙏

visual-testing-app/webpack.config.js
montezume: The numbers are similar in the webpack docs. The best place to learn about splitChunks is [there](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunks-cachegroups-priority) anyways 🤷‍♂️

PR #383, opened 2019-01-03
feat: encapsulate `Typography` styles.

### LINKED ISSUE #382: Q: typography styles
As I was [playing around with ui-kit being used with e.g. Gatsby/Next.js](https://github.com/commercetools/ui-kit/issues/353#issuecomment-449629247), I noticed that the `Text` components use kind of "global styles", even though the styles are defined in the css file.

https://github.com/commercetools/ui-kit/blob/d1acba93d97419f8a85db51471edf602087515fa/src/components/typography/text/text.mod.css#L15-L70

I'm wondering about a couple of things:
- should we really target html tags instead of using specific classes?
- if we do want to have it like this, we should maybe move those selectors to our e.g. `reset.mod.css`

Thoughts?

montezume:
Closes #382

### DIFF
--- a/materials/internals/reset.mod.css
+++ b/materials/internals/reset.mod.css
@@ -20,15 +20,6 @@ html {
   box-sizing: border-box;
 }
 
-h1,
-h2,
-h3,
-h4,
-h5,
-h6 {
-  margin: 0;
-}
-
 button {
   cursor: pointer;
 }
@@ -38,9 +29,4 @@ a {
   cursor: pointer;
   text-decoration: none;
   transition: color 0.2s ease-in;
-}
-
-p {
-  color: var(--color-black);
-  margin: 0;
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -9,7 +9,6 @@ import postcssPresetEnv from 'postcss-preset-env';
 import postcssReporter from 'postcss-reporter';
 import cleanup from 'rollup-plugin-cleanup';
 import replace from 'rollup-plugin-replace';
-import postcssNesting from 'postcss-nesting';
 import postcssCustomProperties from 'postcss-custom-properties';
 import postcssCustomMediaQueries from 'postcss-custom-media';
 import postcssColorModFunction from 'postcss-color-mod-function';
@@ -50,7 +49,6 @@ const postcssPlugins = [
     preserve: false,
     importFrom: { 'custom-properties': customProperties },
   }),
-  postcssNesting(),
   postcssReporter(),
 ];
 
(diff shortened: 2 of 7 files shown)


src/components/typography/typography.story.js
```diff
@@ -66,15 +66,15 @@ storiesOf('Basics|Typography/Text', module)
...
-          'none',
-          'primary',
-          'information',
-          'secondary',
-          'positive',
-          'negative',
-          'inverted',
-        ])}
+        tone={select('Text tone', {
+          none: null,
```
montezume: fixes storybook proptype warning

materials/internals/reset.mod.css
```diff
@@ -39,8 +30,3 @@ a {
   text-decoration: none;
   transition: color 0.2s ease-in;
 }
-
-p {
```
montezume: making reset.mod.css a bit lighter 👍

rollup.config.js
```diff
@@ -49,6 +50,7 @@ const postcssPlugins = [
     preserve: false,
     importFrom: { 'custom-properties': customProperties },
   }),
+  postcssNesting(),
```
tdeekens: Nice. This removes the need to `compose` even more.

src/components/typography/typography.story.js
dferber90: Could it also be fixed by providing a default value to `select`?

src/components/typography/text/text.mod.css
```diff
@@ -12,97 +12,110 @@
...
 }
 
-h4 {
+h4.subheadline {
   font-size: var(--font-h4);
   font-weight: normal;
 }
 
-h5 {
+h5.subheadline {
```
dferber90: Could we remove those `h1` prefixes from the selectors? Our JS code should ensure that they're applied to the right elements. I wouldn't consider it part of the CSS's job.

montezume:
Yay for Percy 🎉

src/components/typography/text/text.mod.css
```diff
@@ -118,3 +131,43 @@ small {
...
+    color: var(--color-gray-60);
+  }
+
+  &.positive {
+    color: var(--color-green-25);
+  }
+
+  &.negative {
+    color: var(--color-red);
+  }
```
dferber90: Could we define separate top-level classes instead? Then we might not need the `postcss-nesting` plugin as well.

So something like

```css
.bold {
  font-weight bold;
}

.positive {
   color: var(--color-green-25);
 }

.subheadline {
   margin: 0;
   color: var(--font-color);
 }
```

We are using the same "modifier" values from what I can tell, so this should reduce the amount of CSS as well.

montezume: done, I removed `postcss-nesting` and updated the PR to simplify the CSS. VRT still passing. I love VRT.

src/components/typography/text/text.mod.css
montezume: We still need it for the font size though right? either that or we need two different classnames.

dferber90: Ah yeah, I missed that part. I'm okay with either approach then 👍

PR #387, opened 2019-01-04
Self contained styles (+pptr)

### LINKED ISSUE #185: Self-contained styles
### Problem

Right now the UI-Kit requires consumers to provide some style, otherwise the ui-kit looks pretty ugly.

For example, when an application only includes the `PrimaryButton` and the `MoneyInput`, then the result looks like [this](https://user-images.githubusercontent.com/1765075/47426533-c4634200-d78d-11e8-80c2-b57a1d51c3bf.png).

The reason the ui-kit looks unstyled is because the UI-Kit currently requires parents to apply these styles:

- `@commercetools-frontend/ui-kit/materials/reset.mod.css`
- `@commercetools-frontend/ui-kit/materials/grid.mod.css`

By doing 

```js
import '@commercetools-frontend/ui-kit/materials/grid.mod.css';
import '@commercetools-frontend/ui-kit/materials/reset.mod.css';
```

For the MC, this is currently done [here](https://github.com/commercetools/merchant-center-application-kit/commit/be9bf41003a21054e363260b18a33d27c06fcd9b#diff-68ba9644f2a46c8f8ae5b2442b149a5dR6).

Notice that we currently also need to import these styles into `storybook`, otherwise the ui-kit would look unstyled. Additionally, the `reset.mod.css` uses features like `@import` and `var(--color-black);`. This is bad because it requires the consumers to have the tools ready to transpile these features. You can see the build-output at [unpkg.com/@commercetools-frontend/ui-kit@2.0.0-rc.11/materials/reset.mod.css](https://unpkg.com/@commercetools-frontend/ui-kit@2.0.0-rc.11/materials/reset.mod.css).

### Ideal solution

Ideally, the ui-kit would be self-contained. All components of the ui-kit would their full styles applied. The ui-kit should not add any global styling to the consumer, nor should it require the consumer to apply any global styling.

This is a bigger step to achieve, as we'd have to add all global overwrites of `reset.mod.css` to each component (e.g.`font-family` to all components).

### Intermediate suggestion

Because it is quite a bit of work to move the internal styles around, we can take an intermediate step. We can include `reset.mod.css` and `grid.mod.css` into `index.js` of ui-kit. The global will then get applied automatically, and consumers no longer need to import styles from ui-kit only to apply them to the body afterwards. This also lifts the inconvenience of ui-kit consumers needing to have a setup which can parse custom-properties (`color: var(--color-black);`) and `@import` statements. This also allows us to remove the import from `app-kit` and from `storybook`.

This intermediate suggestion further already heads in the right direction - as the parent on longer needs to add any styling. We still apply global styles, but moving those to component-styles will be an internal change then.

### Notes

This is a lot of text, but reaching the intermediate solution shouldn't actually be that much work.

montezume:
#### Summary

* Adds [pptr-testing-library)](https://github.com/patrickhulce/pptr-testing-library), which allows us to do dom-testing-library (think RTL) style queries with puppeteer.
* Encapsulates all styles (Closes #185)
* Bug fix: Text.Detail no longer "jumps" down when you set it to inline. (this lead to some tiny vrt diffs)

### DIFF
--- a/materials/internals/reset.mod.css
+++ b/materials/internals/reset.mod.css
@@ -1,5 +1,0 @@
-html,
-body {
-  color: var(--color-black);
-  font-size: 13px;
-}
--- a/playground/src/playground.css
+++ b/playground/src/playground.css
@@ -2,5 +2,4 @@ html,
 body {
   margin: 0;
   padding: 0;
-  height: 100vh;
 }
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,5 @@
 /* This is the entry point of the public package interface */
 
-import '../materials/internals/reset.mod.css';
 import * as i18n from '../i18n';
 
 export {
--- a/src/components/avatar/avatar.js
+++ b/src/components/avatar/avatar.js
@@ -37,7 +37,7 @@ const GravatarImg = props => (
     className={styles['gravatar-img']}
     src={createGravatarImgUrl(props.hash, getSizeInPx(props.size))}
     srcSet={oneLineTrim`
-      ${createGravatarImgUrl(props.hash, getSizeInPx(props.size))} 1x, 
+      ${createGravatarImgUrl(props.hash, getSizeInPx(props.size))} 1x,
       ${createGravatarImgUrl(props.hash, getSizeInPx(props.size) * 2)} 2x
     `}
   />
@@ -50,7 +50,9 @@ GravatarImg.propTypes = {
 };
 
 const Initials = props => (
-  <div className={styles.initials}>
+  <div
+    className={classnames(styles.initials, styles[`initials-${props.size}`])}
+  >
     {getInitialsFromName({
       firstName: props.firstName,
       lastName: props.lastName,
@@ -62,6 +64,7 @@ Initials.displayName = 'Initials';
 Initials.propTypes = {
   firstName: PropTypes.string,
   lastName: PropTypes.string,
+  size: PropTypes.oneOf(['s', 'm', 'l']).isRequired,
 };
 
 const Avatar = props => (
@@ -76,7 +79,11 @@ const Avatar = props => (
     {...filterDataAttributes(props)}
   >
     <GravatarImg hash={props.gravatarHash} size={props.size} />
-    <Initials firstName={props.firstName} lastName={props.lastName} />
+    <Initials
+      size={props.size}
+      firstName={props.firstName}
+      lastName={props.lastName}
+    />
   </div>
 );
 Avatar.displayName = 'Avatar';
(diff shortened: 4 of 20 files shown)


materials/internals/reset.mod.css
```diff
@@ -9,24 +9,9 @@
 html,
 body {
   color: var(--color-black);
-  font-family: 'Open Sans', sans-serif;
   font-size: 13px;
```
emmenko: Unrelated but should we consider using `rem` for sizes instead of hard-coded pixels?

package.json
```diff
@@ -186,6 +186,7 @@
     "postcss-preset-env": "6.5.0",
     "postcss-reporter": "6.0.0",
     "postcss-safe-parser": "4.0.1",
+    "pptr-testing-library": "0.3.1",
```
emmenko: Nice 🎉

src/components/buttons/secondary-button/secondary-button.visualspec.js
```diff
@@ -1,12 +1,17 @@
...
 describe('SecondaryButton', () => {
   beforeAll(async () => {
     await page.goto(`${HOST}/secondary-button`);
   });
 
   it('Default', async () => {
-    await expect(page).toMatch('A label text');
+    const doc = await getDocument(page);
+    const button = await getByLabelText(doc, 'A label text');
+    expect(button).toBeTruthy();
```
montezume: feels a bit more like our style of testing 😆

playground/src/playground.css
```diff
@@ -0,0 +1,6 @@
+html,
+body {
+  margin: 0;
+  padding: 0;
+  height: 100vh;
```
dferber90: Is this `height: 100vh` actually needed? As far as I remember it is something the MC needs which somehow ended up in ui-kit, but our playground should not need it.

montezume:
Going to do more encapsulate in this PR since I'm on a roll.

playground/src/playground.css
montezume: not sure, I just added it so it looked the same as before 🤷‍♂️

src/index.js
```diff
@@ -1,6 +1,5 @@
 /* This is the entry point of the public package interface */
 
-import '../materials/internals/reset.mod.css';
```
tdeekens: Can you elaborate how we can be sure that no app-code implicitly depends on these?

montezume: because both of those files are duplicated in the app shell and are imported there

https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell/src/components/application-shell/reset.mod.css

montezume: 😄

montezume:
Okay this is ready for review again 🗡

PR #385, opened 2019-01-03
Add form philosophy

dferber90:
Adds a [Form Philosophy](https://github.com/commercetools/ui-kit/blob/df-add-form-philosophy/philosophy/FORMS.md) document to storybook.

Also
- Adds a Philosophy section
- Reorganizes the storybook sections
- Fixes outdated usage of `yarn lint:css -- -xy` where the intermediate `--` is deprecated by yarn
- Limits story width for documentation-only stories to make them more readable

Checkout the newly organized Storybook here: https://deploy-preview-385--mc-uikit.netlify.com/

PR #388, opened 2019-01-04
Add inputs and fields philosophy

dferber90:
Check out #385 first. This PR builds on top of that.

Adds more information about our inputs and fields.

### DIFF
--- a/philosophy/FIELDS.md
+++ b/philosophy/FIELDS.md
@@ -1,8 +1,8 @@
 # Fields
 
-Fields combine a label, an input element and valdiation messages.
+Fields combine a label, an input element and validation messages.
 
-Think of Fields as a form of API sugar expressed as additional components.
+Think of Fields as a form of syntactic sugar expressed as additional components.
 They forward most properties to the underlying components while combining some
 APIs of the components for easier usage.
 
@@ -42,6 +42,7 @@ All field components should at least accept these properties
 | `value`      | `any`      |
 | `onChange`   | `function` |
 | `isDisabled` | `boolean`  |
+| `data-*`     | `any`      |
 
 ## Auto-generated ids
 
--- a/philosophy/FIELDS.md
+++ b/philosophy/FIELDS.md
@@ -1,0 +1,104 @@
+# Fields
+
+Fields combine a label, an input element and valdiation messages.
+
+Think of Fields as a form of API sugar expressed as additional components.
+They forward most properties to the underlying components while combining some
+APIs of the components for easier usage.
+
+They are an abstraction which should be used to build forms with.
+In case it's necessary to use slightly different behaviour, it's always possible
+to drop down to the building blocks (label, input element, validation message) to
+achieve customizaion. In most cases this won't be necessary though.
+
+## Forwarding of props
+
+Generally all fields forward most properties to the underlying `FieldLabel`, input component and `FieldErrors` component.
+
+All field components should at least accept these properties
+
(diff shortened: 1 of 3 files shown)


philosophy/FIELDS.md
```diff
@@ -0,0 +1,104 @@
+# Fields
+
+Fields combine a label, an input element and valdiation messages.
```
montezume: typo: validation

philosophy/FIELDS.md
```diff
@@ -0,0 +1,104 @@
...
+
+Think of Fields as a form of API sugar expressed as additional components.
```
montezume: ~~API sugar~~ syntactic sugar?

philosophy/FIELDS.md
```diff
@@ -0,0 +1,104 @@
...
+| `hint`                 | `string`              |
+| `hintIcon`             | `node`                |
+| `badge`                | `node`                |
+| `horizontalConstraint` | `string`              |
+| `errors`               | `object`              |
+| `renderError`          | `function`            |
+| `onInfoButtonClick`    | `function`            |
+| `isRequired`           | `boolean`             |
+| `touched`              | `boolean` or `object` |
+
```
montezume: how about dataTest props?

philosophy/INPUTS.md
```diff
@@ -0,0 +1,188 @@
...
+This allows the parent not to worry about determining valid/invalid dates while still being able to customize validation. Users won't get validation warnings when they enter invalid dates (the date will just disappear on blur), but they will get validation warnings when the date is outside of a certain range. This allows some flexibility in validation without burdening all consumers with date validation.
+
+## Error states
+
+As our inputs leave validation to the form they are used in, they can't determine themselves whether they are currently valid or not.
+This means the parent needs to tell an input about its validation state. The input can use this information to render itself in an appropriate style (e.g. with a red border in case of an error). Do not try to determine the validation state form within the input.
+
+Usually our inputs are told about the validation state by accepting a `hasError` property which contains a boolean.
+They also accept an additional boolean `hasWarning` property for intermediate cases (usually results in an orange border).
+Ask the design team about the difference between `hasError` and `hasWarning`. In most cases, consumers will use `hasError`.
```
montezume: 😆

philosophy/FIELDS.md
dferber90: Ah, right! Yeah we could mention them.

philosophy/FIELDS.md
dferber90: To me "API sugar" is closer to what we mean. Syntactic sugar is something like `foo?.bar` instead of having to write `foo && foo.bar`.

Here it's more about providing a nicer API / set of components to work with. I dunno, it's just an analogy anyways so it's not too important.

montezume: Sure, but syntactic sugar is an expression, API sugar isn't. 
[ref](https://www.google.com/search?q=api+sugar&rlz=1C5CHFA_enDE799DE799&oq=api+sugar&aqs=chrome..69i57j69i60j69i61j69i60j69i61j0.825j0j7&sourceid=chrome&ie=UTF-8)

dferber90: ......but, .........but it could be. Let a boy dream 👦 :P

Let's change it to syntactic sugar then. I really don't mind either way.

dferber90:
I fixed the things, but I'm waiting for #385 before merging this one.

PR #389, opened 2019-01-04
Reorganise props of multiline and localized inputs and fields

dferber90:
Unifies the APIs of multiline and localized inputs and fields.

### Before

This is a list of prop-names each input is using. Notice that the respective field components use the same prop-names, and they have changed as well!

| Concern | `MultilineTextInput`  | `LocalizedTextInput` | `LocalizedMultilineTextInput`  |
| --- | --- | --- | --- |
| Picture | image | image | image |
| Multiline Expansion | `isDefaultClosed` |  | `isMultilineDefaultExpanded` |
| Language Expansion | | `isDefaultExpanded` | `areLanguagesDefaultOpened` |
| Language Controls | | `hideExpansionControls` | `hideLanguageControls` |

### Now

| Concern | `MultilineTextInput`  | `LocalizedTextInput` | `LocalizedMultilineTextInput`  |
| --- | --- | --- | --- |
| Picture | image | image | image |
| Multiline Expansion | `defaultExpandMultilineText` |  | `defaultExpandMultilineText` |
| Language Expansion | | `defaultExpandLanguages` | `defaultExpandLanguages` |
| Language Controls | | `hideLanguageExpansionControls` | `hideLanguageExpansionControls` |

The same property renamings happened for `LocalizedMultilineTextField`, `LocalizedTextField` and `MultilineTextField`.

Notice that `isDefaultClosed` was renamed to `defaultExpandMultilineText` for `MultilineTextInput`, `MultilineTextField`, `LocalizedMultilineTextInput` and `LocalizedMultilineTextField`. This means the default has been flipped! You now need to provide `defaultExpandMultilineText` in cases where you were not passing `isDefaultClosed` to keep the same behaviour! It also means you can remove any places where you were passing `isDefaultClosed={true}` as that is the default now.

### BREAKING

Changed APIs of `LocalizedMultilineTextField`, `LocalizedTextField`, `MultilineTextField`, `LocalizedMultilineTextInput`, `LocalizedTextInput` and `MultilineTextInput`.

Closes #313

#### Flipped defaults
- `MultilineTextInput` and `MultilineTextField`
  - `isDefaultClosed` was dropped. The multiline-text is now closed by default. Pass `defaultExpandMultilineText` to keep the original behaviour.

#### Simple renamings
- `LocalizedTextInput` and `LocalizedTextField`
  - Rename `isDefaultExpanded` to `defaultExpandLanguages`
  - Rename `hideExpansionControls` to `hideLanguageExpansionControls` 
[...]

PR #390, opened 2019-01-05
docs(changelog): for release v6.0.0

emmenko:
Based on this diff: https://github.com/commercetools/ui-kit/compare/v5.0.0...master


montezume:
I would like to get #387 in as well, and I think we are really close with #339

dferber90:
#389 is ready code-wise as well. It's only missing some reviews.

Whether we should wait for #389 mostly depends on whether we want to do one big upgrade of the MC or two successive smaller ones. I'm fine either way, but I'd rather get #389 in as well.

emmenko:
Cool, let's try to get this out in the next days then! 🚀

dferber90:
Added the #389 stuff now. @montezume is still doing some fixes and then we're ready for the next release.

PR #399, opened 2019-01-07
build(travis): publish canary releases

emmenko:
Same idea as in https://github.com/commercetools/merchant-center-application-kit/pull/243

We need to prepare the release manually though as we are not using lerna (which has this built-in).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ script:
   # also skip when PR is opened by `renovate`
   # back to default in case internal travis scripts return non zero response codes.
   - "! test -f fail-tests-because-there-was-an-unhandled-rejection.lock"
-  - bash ./scripts/run-percy.sh
+  - ./scripts/run-percy.sh
   - set +e
 
 before_deploy: yarn build
@@ -39,6 +39,9 @@ deploy:
   on:
     tags: true
   tag: next
+
+after_success:
+  - if [ "$TRAVIS_BRANCH" = "master" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ ! "$TRAVIS_COMMIT_MESSAGE" =~ \[skip\ publish\] ]]; then node ./scripts/release-canary.js; fi
 
 # Keep it disabled until this issue is clarified
 # https://twitter.com/emmenko/status/1040544486101856256
--- a/package.json
+++ b/package.json
@@ -207,6 +207,7 @@
     "rollup-plugin-postcss": "1.6.3",
     "rollup-plugin-replace": "2.1.0",
     "rollup-plugin-url": "2.1.0",
+    "semver": "5.6.0",
     "shelljs": "0.8.3",
     "storybook-readme": "4.0.2",
     "style-loader": "0.23.1",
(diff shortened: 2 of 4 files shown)


montezume:
I would rather not have this in, for the same reason we decided not to go forward with using [semantic-release](https://github.com/semantic-release/semantic-release). It's a small team, getting smaller, working on ui-kit, and releasing doesn't take much effort anyways.

dferber90:
We thought about having different release/branching strategies. We currently face the problem of not being able to easily release bug fixes when breaking changes have been merged to `master` since the last release. Having a canary release setup would complicate this a bit (as we'd need to do canary releases from a `next` branch, while doing regular releases from `master`).

Not having canary releases hasn't been a problem so far (afaik), so I would like to keep it simple for now.

dferber90:
If we were to do canary releases at the moment:
- we can not determine their version automatically as we're not using appropriate commit-messages (as it's to hard to get everyone to follow it properly - it's easier to determine the next version when writing a change log/releasing for projects which have a large set of often-changing contributors)
- consumers would often have a hard time using canary releases as they might have to go through breaking changes first
- I'm not sure whether we could still generate change logs correctly
- we have a ton of releases
- it is more we need to think about when we try to come up with a better releasing strategy
- we need to decide whether to write change logs for canary releases
- we need to figure out how to do canary releases properly (it's not obvious to me from this PRs description. What's the "manual" part that's left?)
- we need to communicate that we still have to write change logs from the last published version to the current one
- we have two competing ways of doing preview releases
  - canary releases with `1.0.0-canary.0+{sha}`
  - the dist-tags where we release to `next`
- teams shouldn't use differing versions of ui-kit, as app-kit enforces a ui-kit version anyways and they'd be shipping two versions of ui-kit then

The code added in this PR is the minor part of the work and this PR leaves us with a lot of unanswered questions. Neither @montezume or me are seeing canary releases as something that's urgent at the moment. That's why were hesitant to go for it.

dferber90:
> We need to prepare the release manually though as we are not using lerna 

I'm not sure what you mean by "prepare the release manually". If we need to do something manually, we might as well release to `next` with our current setup?  We already have a way to do preview-releases. So there is already a way to get some changes out earlier.

tdeekens:
I see this making sense for app-kit. We agreed there to try it out and see how much it actually helps and how sensible it is. I see UIKit having a bit different and lower cadence release cycles for now. Given those two I don't see why we can't test it out in app-kit first and come back to this after. Given how "easy" it is to add it I don't see why we should push for it here now already.

dferber90:
> > The code added in this PR is the minor part of the work and this PR leaves us with a lot of unanswered questions.
>
> What work? There is nothing else to do here.

I meant coordination and communication with other team members (including myself) so we understand how to use this. When to use a canary release vs publishing to `next` etc.

> For me it's important that we have frequent releases, which is obviously not the case at the moment.

I don't think frequent releases are a good indication of anything. We can release a 100 times a day, but if the MC can't upgrade to those releases in a timely manner then they're worthless.

---

Thanks for suggesting canary releases and thanks for letting go again.

PR #401, opened 2019-01-07
chore(collapsible-panel): move onClick in header

montezume:
This has bugged me for awhile. You should be able to toggle the `CollapsiblePanel` by clicking anywhere in the header.

### DIFF
--- a/src/components/panels/collapsible-panel/collapsible-panel.js
+++ b/src/components/panels/collapsible-panel/collapsible-panel.js
@@ -89,6 +89,7 @@ export default class CollapsiblePanel extends React.PureComponent {
             })}
           >
             <div
+              onClick={this.props.isDisabled ? undefined : toggle}
               className={classnames(styles['base-header-container'], {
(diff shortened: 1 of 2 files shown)


src/components/panels/collapsible-panel/collapsible-panel.js
```diff
@@ -89,6 +89,7 @@ export default class CollapsiblePanel extends React.PureComponent {
             })}
           >
             <div
+              onClick={this.props.isDisabled ? undefined : toggle}
```
dferber90: This might result in bugs when there's a button/input in the header and the event propagation is not stopped. Our `headerControls` should be safe though, as we already call `event.stopPropagation()` for them.

montezume: okay, I'll try adding a test then 👍

montezume: ready for re-review

PR #403, opened 2019-01-07
docs(storybook): fix PrimaryActionDropdown story

dferber90:
## Broken

broken

## Working

ok

### DIFF
--- a/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.story.js
+++ b/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.story.js
@@ -1,14 +1,12 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { withKnobs, boolean, text, select } from '@storybook/addon-knobs';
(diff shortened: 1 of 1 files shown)


src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.story.js
```diff
@@ -1,37 +1,35 @@
...
   .addDecorator(withKnobs)
   .addDecorator(withReadme(Readme))
   .add('PrimaryActionDropdown', () => (
     <Section>
       <PrimaryActionDropdown>
         <Option
-          iconLeft={React.createElement(
-            icons[select('icon', iconNames, iconNames[0])]
-          )}
+          iconLeft={<BoxIcon />}
```
dferber90: It was failing because only the first option had an icon. When the first option didn't have an icon, the `React.cloneElement` done inside of `primary-action-dropdown.js` failed.

Back when this component was created, we decided that every option needs to have an icon. This PR ensures there is an icon for every option.

I removed the possibility to select the icon to keep the story simple.

PR #405, opened 2019-01-08
feat(time-input): add default placeholder

montezume:
As requested by design, adds default placeholder to TimeInput if no placeholder is passed, similar to DateInput. Also fixes some proptype warnings.


src/components/inputs/time-input/time-input.js
```diff
@@ -149,7 +151,11 @@ export class TimeInput extends React.Component {
           isDisabled={this.props.isDisabled}
           hasError={this.props.hasError}
           onClear={() => this.emitChange('')}
-          placeholder={this.props.placeholder}
+          placeholder={
+            typeof this.props.placeholder === 'string'
```
montezume: this is pretty much copy paste from DateInput

PR #408, opened 2019-01-08
fix(primary-action-dropdown): use secondary tone when disabled

dferber90:
Use font-color gray-60 when input is disabled (when all options are disabled) as requested by @filippobocik.

| Before | After |
| ---- | --- |
| image | image |


src/components/dropdowns/primary-action-dropdown/primary-actions-dropdown.mod.css
```diff
@@ -43,7 +43,6 @@
 .disabledDropdownHeadChevron {
   box-shadow: none;
   background-color: var(--color-navy-98);
-  color: var(--color-gray);
```
dferber90: This wasn't doing anything as `Text.Detail` defines its own colour anyways.

PR #413, opened 2019-01-08
chore: rename localized-money-input specs and use pptr

montezume:
#### Summary
the files shouldn't have `percy` in the name.
we should use pptr 😄


src/components/inputs/localized-money-input/localized-money-input.visualspec.js
```diff
@@ -1,12 +1,17 @@
...
 describe('LocalizedMoneyInput', () => {
   beforeAll(async () => {
     await page.goto(`${HOST}/localized-money-input`);
   });
 
   it('Default', async () => {
-    await expect(page).toMatch('minimal');
+    const doc = await getDocument(page);
+    const cadInput = await getByLabelText(doc, 'CAD');
+    expect(cadInput).toBeTruthy();
```
tdeekens: What do we test here? That there is a CAD input in the document? Why do we use toBeTruthy?

montezume: > What do we test here

The idea is the same as `expect(page).toMatch('minimal')`. Just to make sure that the route is being rendered, and that our component is being rendered on screen.

> Why do we use toBeTruthy?

 Do you mean instead of using toBeInTheDocument? We can't use that as it's from `jest-dom`

PR #415, opened 2019-01-08
chore: update changelog and bump package.json

montezume:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools-frontend/ui-kit",
-  "version": "6.0.0",
+  "version": "6.1.0",
```
emmenko: You shouldn’t bump this manually

emmenko: Sorry my bad I got confused 🙈

montezume: 😆

PR #421, opened 2019-01-10
feat(CollapsiblePanel): add `hideExpansionControls` (not hidden by default)

tdeekens:
#### Summary

This pull request adds an `hideExpansionControls` to the `<CollapiblePanel />` while (by default) showing the expansion controls in disabled visuals when the panel itself is disabled.

#### Description

The images visualises the change. The change itself is reasoned in not having the content in the header of the panel "shift" or move when disabling or enabling it. So a previously clickable part remains clickable and does not move away underneath a user's mouse.

cleanshot 2019-01-10 at 10 06 44


src/components/panels/collapsible-panel/collapsible-panel.js
```diff
@@ -81,32 +82,40 @@ export default class CollapsiblePanel extends React.PureComponent {
...
           <div
-            className={classnames(this.props.className, {
-              [styles['container-condensed']]: this.props.condensed,
-              [styles['container-open']]: isOpen,
-              [styles['container-theme-light']]: this.props.theme === 'light',
-              [styles['container-theme-dark']]: this.props.theme === 'dark',
-            })}
+            className={classnames(
+              this.props.className,
+              styles[`container-theme-${this.props.theme}`],
```
tdeekens: A similar classname was built up by convention elsewhere in the file. Doing the same here shortens the `classnames` "block".

src/components/panels/collapsible-panel/collapsible-panel.js
```diff
@@ -81,32 +82,40 @@ export default class CollapsiblePanel extends React.PureComponent {
...
+                <div
+                  {...dataProps}
+                  className={classnames(styles.header, {
+                    [styles['header-disabled']]: this.props.isDisabled,
+                  })}
+                >
                   <div className={styles['truncate-header']}>
                     <Spacings.Inline alignItems="center" scale="s">
-                      {!this.props.isDisabled && (
+                      {!this.props.hideExpansionControls && (
```
tdeekens: The actual change. Instead of hiding them when disabled we show them.

montezume:
@tdeekens can you ask @mariabarrena to review the percy diff?

tdeekens:
> @tdeekens can you ask @mariabarrena to review the percy diff?

Yup was about to.

src/components/panels/collapsible-panel/README.md
```diff
@@ -33,21 +33,22 @@ The component can be used in a controlled or uncontrolled manner. Please refer t
...
+| `description`           | `String` |                                                          | -                     |           | If passed will be shown below the title as more information regarding the panel                                                           |
+| `isDisabled`            | `bool`   |                                                          | -                     | false     | Disables the panel and all interactions with it                                                                                           |
+| `children`              | `node`   |                                                          | -                     | -         | The actual content rendered inside the panel                                                                                              |
+| `hideExpansionControls` | `bool`   |                                                          | -                     | -         | Controls the visibility of the expansion controls on the left                                                                             |
```
montezume: I'm not sure about this prop name. @dferber90 any ideas?

tdeekens: It's an attempt to align it with some inputs which also have an option to be expanded. The intend was to have transferable vocabulary also across components of different groups. Let me know if you come up with a better name and the reasoning for it.

mariabarrena:
It looks exactly how it was designed. Thanks!
Before approve it on percy, I want to double check with any of you one thing. In the screenshot I see some of the possible states of the panel being showed on light and dark themes (f.e. "urgent and dark" and "urgent and light") However for disabled I just see one theme being showed. We definitely need the disabled and the hide expansion control states for both themes. Do we need to show both on percy?

src/components/panels/collapsible-panel/README.md
montezume: but we no longer have any other props called `hideExpansionControls`, they were changed [here](https://github.com/commercetools/ui-kit/pull/389).

montezume:
@mariabarrena sure, it makes sense to visually test all the possible iterations 👍 .

src/components/panels/collapsible-panel/README.md
tdeekens: Your call. Gimme a name :)

tdeekens:
> Do we need to show both on percy?

Sure, I'll add them. I just build on what was there. I'll add all variations coming to mind.

src/components/panels/collapsible-panel/README.md
qmateub: Well in this case you are expanding or collapsing... maybe the name is correct after all 🤷‍♂️ 

Just a note this is just in case the component is disabled right? So we could assume that you set it whenever is disabled meaning that the panel is collapsed so now you can only expand 🤔

src/components/panels/collapsible-panel/README.md
```diff
@@ -33,21 +33,22 @@ The component can be used in a controlled or uncontrolled manner. Please refer t
...
+| `headerControls`        | `node`   |                                                          | -                     | -         | Controls at the top right part of the panel                                                                                               |
+| `tone`                  | `oneOf`  |                                                          | ['urgent', 'primary'] | 'primary' | -                                                                                                                                         |
+| `className`             | `bool`   |                                                          | -                     | -         | -                                                                                                                                         |
```
qmateub: Note in the scope of this PR obviously but this should go away so no custom CSS leaks into the component 😋

tdeekens:
Alright @mariabarrena I added more visual specs. Adding all is close to bit silly as it's quite a bit of combinations. I added at least one per case and "combination variation". So condensed is disabled and urgent too yadada.

src/components/panels/collapsible-panel/README.md
montezume: I don't really have strong opinions on it. hideExpansionControls is fine by me. 🤷‍♂️

tdeekens: I will wait for the @dferber90 to give his feedback for this.

src/components/panels/collapsible-panel/README.md
montezume: Yeah, it would be more appropriate to create an issue for that.

src/components/panels/collapsible-panel/README.md
dferber90: I'm fine with this name 👍 

We used more specific names (`defaultExpandMultilineText`, `defaultExpandLanguages`) in `LocalizedMultilineTextInput` as there are two things which can be expanded/collapsed. Here it's only one thing, so `hideExpansionControls` seems fine. On a sidenote, `LocalizedMultilineTextInput` only accepts `hideLanguageExpansionControls` as the multiline expansion controls can't be hidden (yet).

<details>
<sumary>Sidenote</summary>

`LocalizedMultilineTextInput` always acts as an uncontrolled component regarding its expansion states (it doesn't accept the state from the parent). Therefore it doesn't accept props like `expandMultilineText` or `expandLanguages` (or `isMultilineTextExpanded` and `areLanguagesExpanded`).

For the component in this PR, I slightly dislike that `hideExpansionControls` and `isClosed` don't use the same terminology (open/closed vs expanded/collapsed). We could do `hideExpansionControls` and `isExpanded` so that they use the same terminology for the same concept which shows that the props belong together. But that's probably not worth the refactoring as we'd also need to flip the default. So let's just keep `isClosed` and add `hideExpansionControls`.

</details>

montezume:
@mariabarrena can you approve the percy diff so we can get this merged then? 😇

mariabarrena:
"Condensed light", "Condensed light - hideExpansionControls" and "Condensed light is Disabled" are showing the dark theme and not the light one. Once that is fixed, we are good to go. I'll wait for the change to approve it.

PR #423, opened 2019-01-10
fix(radio-option, checkbox, switch): move data-attributes to input

montezume:
#### Summary

If you run `yarn lint` on the master branch, we get a bunch of errors.

This PR fixes those errors.

It also moves the `filter-data-attributes` on `Switch`, `RadioOption` and `Checkbox` to the underlying input.

### DIFF
--- a/src/components/inputs/async-creatable-select-input/async-creatable-select-input.spec.js
+++ b/src/components/inputs/async-creatable-select-input/async-creatable-select-input.spec.js
@@ -18,6 +18,9 @@ class TestComponent extends React.Component {
             props,
             ...rest
           ),
+    loadOptions: PropTypes.func,
+    defaultOptions: PropTypes.bool,
+    isSearchable: PropTypes.bool,
     onChange: PropTypes.func,
   };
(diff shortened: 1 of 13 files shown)


src/components/inputs/async-creatable-select-input/async-creatable-select-input.spec.js
```diff
@@ -18,6 +18,9 @@ class TestComponent extends React.Component {
             props,
             ...rest
           ),
+    loadOptions: PropTypes.func,
+    defaultOptions: PropTypes.bool,
```
tdeekens: Is this a bool or object?

montezume: it was set as a bool in the existing defaultProps

dferber90: It can either be an array of objects (options) or a boolean, see [`react-select` docs](https://react-select.com/props#async-props).

PR #424, opened 2019-01-10
fix(error-message): support data-attributes

montezume:
#### Summary

@amine-benselim reported that `ErrorMessage` didn't support data-foo attributes.

#### Approach

Make it support data-attributes.

### DIFF
--- a/src/components/messages/error-message/error-message.js
+++ b/src/components/messages/error-message/error-message.js
@@ -1,9 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Text from '../../typography/text';
+import filterDataAttributes from '../../../utils/filter-data-attributes';
 
 const ErrorMessage = props => (
-  <Text.Detail tone="negative">{props.children}</Text.Detail>
+  <Text.Detail tone="negative" {...filterDataAttributes(props)}>
+    {props.children}
+  </Text.Detail>
 );
 ErrorMessage.displayName = 'ErrorMessage';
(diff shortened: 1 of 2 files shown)


dferber90:
For react-testing-library tests `getByText` should be preferred over `getByTestId`. Adding support for `data-`  attributes opens up the possibility to find `ErrorMessage` components using `getByTestId` whereas until now users would have had to use the more preferable `getByText`. By making it inconvenient to attach data-props, we were guiding users into the right direction of using `getByText`. If we add data-props to `ErrorMessage` we make it easier to do the wrong thing (for react-testing-library tests).

I liked that we made/impossible to do the wrong thing so far.

However we might need `data-test-id` (not `data-testid`)  for E2E tests. Since both are `data` props we can't really have one without the other. So I'm okay with this, even though it sucks a bit that we lose this guidance now.

montezume:
@dferber90 but we are using `data-testid` in the MC-FE for cypress, and in their [docs](https://docs.cypress.io/guides/references/best-practices.html) it suggests to use one of 

```
data-cy
data-test
data-testid
```

🤔

montezume:
I agree with you though that for `react-testing-library`, we should be using `getByText` or `queryByText`

dferber90:
@montezume That was my point. If we only used react-testing-library I would not have wanted the `data` props on `ErrorMessage`, but since cypress also wants to pass `data` props, we need to support them for the sake of cypress.

PR #425, opened 2019-01-10
feat(storybook): expose svg icons as static assets

emmenko:
This is an attempt/experiment for solving the issue about dynamically rendering the icons by name:

https://github.com/commercetools/merchant-center-application-kit/blob/ffe96d2d9922e98d8d4d5c588070548e64db791c/packages/application-shell/src/components/navbar/navbar.js#L63-L68

The idea here is that those icons are available at the uikit website https://uikit.commercetools.com/icons so that we can load them as normal ``.

**I'm playing around with this locally first, so let's not merge this yet.**

If we decide to proceed with this approach we need to put those static icons into something more stable (e.g. google storage). Most likely we would also need to namespace those static assets by release version.


dferber90:
Exposing the primitive svgs is a nice idea! Instead of deploying them to a URL, we could also expose them from the package. Then consumers could read them from `node_modules`. Using webpack contexts this would probably work.

However consumers would then need to have loaders in place for those SVGs. Also not sure how colour and size props would work then.

If this is only about the icons of custom apps, then we could make apps export/contain an SVG icon instead? Maybe we shouldn't mix that specific use-case into our overall icons approach.

emmenko:
> If this is only about the icons of custom apps

No it will also apply to our apps, since I'm currently moving the menu items (including icons) out of the AppShell (https://github.com/commercetools/merchant-center-application-kit/issues/37#issuecomment-444500693). It's not super crucial but it still feels wrong to have this coupling between NavBar and app icons when they belong in different places.

> Instead of deploying them to a URL, we could also expose them from the package.

Yeah that could also work. 

> However consumers would then need to have loaders in place for those SVGs

We already have that: https://github.com/commercetools/merchant-center-application-kit/blob/ffe96d2d9922e98d8d4d5c588070548e64db791c/packages/mc-scripts/config/create-webpack-config-for-production.js#L273-L289

> Also not sure how colour and size props would work then.

This is the point that I'm looking into at the moment. In theory we can't fiddle with the svg anymore since we load it as an image. I think we have a couple of options here:
- hack our way through it (e.g. found this after a quick google search, which (mis)uses CSS `filter` https://blog.union.io/code/2017/08/10/img-svg-fill/)
- duplicate each svg file into a separate coloured version (although we might end up with a lot of files, depending on the combinations)

emmenko:
This is the code diff require to make it work: https://github.com/commercetools/merchant-center-application-kit/compare/nm-navbar-remote-svg-icon?expand=1

montezume:
I also like the idea of just exposing the raw SVGs in `node_modules`. What do you think @emmenko ?

emmenko:
Yep, I wanted to try that next.

Besides that, what do you think about this approach?

package.json
```diff
@@ -36,13 +36,14 @@
   ],
   "scripts": {
     "changelog": "conventional-changelog -p angular -u",
-    "prebuild": "node ./scripts/generate-icon-exports.js && node ./scripts/generate-custom-properties.js",
-    "prestorybook:start": "node ./scripts/generate-icon-exports.js && node ./scripts/generate-custom-properties.js",
+    "prestorybook:build": "yarn run prebuild && node ./scripts/copy-svg-icons-to-assets.js",
```
montezume: I'm confused as to why this needs to be associated with storybook? Seems more appropriate to do it as a prebuild step.

.gitignore
```diff
@@ -72,6 +72,7 @@ typings/
 
 # Storybook build output
 .public
+docs/assets/icons
```
montezume: I think I would rather have a new `assets/icons` directory

package.json
emmenko: Don't worry much about this stuff, it was to just to quickly get it to work

montezume:
The approach from the `app-kit` side looks good to me. From the `ui-kit` side, we can just copy the icons into a `assets/icons` directory at build time, and add thoses to the `files` array in our `package.json`.

emmenko:
> The approach from the app-kit side looks good to me

Alright, that's the important part for me.

> From the ui-kit side, we can just copy the icons into a assets/icons directory at build time, and add thoses to the files array in our package.json.

Sure, I'll probably open a new PR and close this one. I'll test first to load the files from `node_modules`. 

Regarding the `SvgIcon` component, should we implement it in the ui-kit?

dferber90:
I like that we're able to do this, but I'm still not sure if (or why) we actually need to?

We might still be missing two different concerns:
- having icons in ui-kit
- allowing applications to define their own icons in the MC

I think we can solve the latter without changing anything in the ui-kit icons. And then there's no need anymore to expose them statically, since the app icons are the only place requiring that so far.

tdeekens:
I like the approach and the possibility but I have some thoughts and concerns

1. We currently "relate" icons by name and convention
   - This is a loose coupling and can break silently with a breaking change in UIKit (removing and icon) while not updating it here
   - It's a simple approach, easy to read and "performant" as icons are statically imported (moving apart potential issues with tree-shaking)
2. This adds another "export" to UIKit for a very specific use case (currently only existing in the menu
3. It adds quite a bit of code complexity (two refs, state management etc)
   - On the plus side we can easily support any icon without having to change code in app-kit (which doesn't strictly relate). I have no idea if this is a product requirement.
4. It adds network requests on app boot
   - Each icon fetch will result in a request. Given that browser limit the amount of requests this can block (as we can specify priorities yet) other more important requests. Especially during boot of the application.

dferber90:
Closing after speaking to @emmenko since this probably isn't the right solution to our problem.

We'll talk about it more tomorrow 👍

PR #426, opened 2019-01-11
chore: update deps

montezume:
Just updating some deps now that the new version of rollup is [out](https://github.com/rollup/rollup/releases)

As usual, if tests and VRT pass, then we should be fine.


montezume:
@dferber90 it really makes it simple 💯

PR #428, opened 2019-01-11
chore: move Toggle, radio and checkbox to input, emit event for toggle

montezume:


src/components/inputs/toggle/toggle.js
```diff
@@ -22,6 +22,17 @@ export class Toggle extends React.PureComponent {
     size: 'big',
   };
 
+  emitChange = () => {
+    const event = {
+      target: {
```
montezume: this is the breaking change

src/components/inputs/time-input/time-input.spec.js
```diff
@@ -72,7 +72,7 @@ describe('TimeInput', () => {
     expect(container.querySelector('input')).toHaveAttribute('value', 'foo');
   });
 
-  it('should call onChange when chaning the value', () => {
+  it('should call onChange when changing the value', () => {
```
montezume: typo fix

src/components/inputs/toggle/toggle.js
```diff
@@ -22,6 +22,17 @@ export class Toggle extends React.PureComponent {
...
+        id: this.props.id,
+        name: this.props.name,
+        value: !this.props.isChecked,
+      },
+    };
+    this.props.onChange(event);
```
emmenko: Why are we constructing the event ourselves? 🤔

src/index.js
```diff
@@ -91,7 +91,9 @@ export { default as PasswordInput } from './components/inputs/password-input';
 export { default as SelectInput } from './components/inputs/select-input';
 export { default as TextInput } from './components/inputs/text-input';
 export { default as TimeInput } from './components/inputs/time-input';
-
+export { default as Toggle } from './components/inputs/toggle';
+export { default as Checkbox } from './components/inputs/checkbox';
+export { default as Radio } from './components/inputs/radio';
```
emmenko: I think we should name them consistently with the other component, so with `Input` suffix.

montezume:
It's being used with formik with this workaround in the MC

```
// FIXME: the Toggle component should send the `event` in the `onChange` handler
function patchToggleChange(isChecked) {
  // This is bound to the function
  const props = this;
  const event = {
    target: {
      type: 'checkbox',
      name: `${props.fieldName}.isActive`,
      checked: isChecked,
    },
  };
  props.formikProps.handleChange(event);
}

```

src/index.js
tdeekens: Very valid!

emmenko:
> It's being used with formik with this workaround in the MC

I know, I wrote that 😄

That's why I'm asking: why now simply `onChange={this.props.onChange}`?

emmenko:
The `input` sends the correct `event`, the workaround was to make it work with formic because the component didn't return an `event`.

montezume:
In fact, if I try it in storybook, what is returned from the event, doesn't contain what we need (at least in the console log...)

screen shot 2019-01-11 at 4 32 41 pm

src/components/inputs/toggle/toggle.js
montezume: I can't get it to work with the native event. You are welcome to try it, but I spent about an hour on it and got nowhere with RTL. The shape is the same now as the native event though.

emmenko: Oh really? Sorry about that...

Hmm that's weird, I would have expected this to work. I can have a look over the weekend.

montezume: switching to `defaultChecked` fixed the probelm

src/index.js
montezume: I renamed `Toggle` -> `ToggleInput`. If we are happy with this, I will rename the rest of them on Monday

src/components/inputs/toggle-input/README.md
```diff
@@ -0,0 +1,37 @@
...
+| `ToggleInput` | Two               | Single    |
+
+#### Properties
+
+| Props        | Type     | Required | Values                  | Default | Description                            |
+| ------------ | -------- | :------: | ----------------------- | ------- | -------------------------------------- |
+| `isDisabled` | `bool`   |    -     | -                       | `false` | Disables the ToggleInput               |
+| `isChecked`  | `bool`   |    -     | -                       | `false` | Checks the ToggleInput                 |
+| `onChange`   | `func`   |    ✅    | -                       | -       | called with the new value.             |
+| `size`       | `string` |    ✅    | oneOf(['small', 'big']) | `big`   | The size of the ToggleInput component. |
```
emmenko: Some props are missing here

src/components/inputs/toggle-input/toggle-input.js
```diff
@@ -38,7 +38,7 @@ export class Toggle extends React.PureComponent {
           className={styles.inputWrapper}
           id={this.props.id}
           name={this.props.name}
-          onChange={() => this.props.onChange(!this.props.isChecked)}
```
emmenko: 💯

montezume:
I updated the PR with the renaming of the rest of the switches

PR #431, opened 2019-01-14
Fix proptypes

### LINKED ISSUE #319: [Docs] Inconsistent / missing props in the component READMEs
**Describe the bug**
I noticed why using the components that the READMEs are not up-to-date with the components. For example, some props are not documented, are wrong or inconsistent.

**Expected behavior**
The READMEs should always reflect and document the components APIs.

dferber90:
- I went through most components to compare the prop-types to the ones described in the README (I only compared the existence of the props, not the value/description/defaults in most cases)
- I applied the same order of props to the README as in the prop-types. We often use the same order for the tests as well and having the same order everywhere makes it easier to compare prop types across files.

Closes #319


montezume:
Can you rebase and merge once the conflict is fixed @dferber90 ? I'd like to limit my merge conflicts with #325

PR #442, opened 2019-01-16
feat: add Grid component

### LINKED ISSUE #73: Grid
Right now we only use  `Constraints` and `Spacings` to create a sort-of grid. We should add a proper grid to the UI-Kit.

emmenko:
Closes #73 

#### Summary

Implements a `<Grid>` component that accepts all the supported [CSS Grid properties](https://css-tricks.com/snippets/css/complete-guide-grid).

**`<Grid>` component**

**Example for fixed columns**

**Example for auto-sizing columns**

#### Design

_NA_

##### Technical

- Documentation

  - [x] Has `README.md` file
  - [x] Explains use-cases
  - [x] Explains props
  - [x] Has story for Storybook

- Props

  - [ ] Has default props for labels (where applicable)

- Design

  - [x] Uses design-tokens (instead of e.g. color variables)

- Testing

  - [x] Uses [react-testing-library](https://github.com/kentcdodds/react-testing-library) tests
  - [x] Has visual regression tests

- Accessibility (optional)
  - [ ] Supports `aria` attributes


emmenko:
This component can obviously be extended and made more configurable as we go, for now I would consider this the bare minimum to make it usable.

src/components/grid/grid.js
```diff
@@ -0,0 +1,69 @@
...
+import styled from '@emotion/styled';
+import customProperties from '../../../materials/custom-properties.json';
+import filterDataAttributes from '../../utils/filter-data-attributes';
+
+// TODO: check with design if those "breakpoints" are ok
+const minColumnSizes = {
+  1: '600px',
+  2: '450px',
+  3: '350px',
+  4: '275px',
```
dferber90: Should these be the same as our horizontal constraints?

dferber90: Our possible values are all over the place right now. We sometimes use numbers, words or shirt sizes.

I'd use shirt-sizes for now, so `s`, `m`, `l`, `xl`

src/components/grid/grid.js
```diff
@@ -0,0 +1,69 @@
...
+  >
+    {props.children}
+    {insertGridPlaceholders(
+      React.Children.count(props.children),
+      props.columns
+    )}
+  </GridContainer>
+);
+Grid.displayName = 'Grid';
+Grid.propTypes = {
```
dferber90: I wonder about the API. The current one is pretty restrictive and it would suck if we had to upgrade the whole MC while we learn more about how to provide a flexible API. Exploring more possible use-cases now can probably avoid migration pain afterwards.

Maybe we can explore using Grid for more advanced use-cases before publishing it?

For example, we could try to build the [`CollapsiblePanel` header](https://user-images.githubusercontent.com/1765075/51261947-94ed5b80-19b1-11e9-8723-a6145dd47230.png) (if that's even something we intend it to be used for).

src/components/grid/grid.js
```diff
@@ -0,0 +1,69 @@
...
+  3: '350px',
+  4: '275px',
+};
+
+const GridContainer = styled.div`
+  display: grid;
+  grid-gap: ${customProperties['--spacing-16']};
+  grid-auto-columns: 1fr;
+  grid-template-columns: ${props => {
+    switch (props.mode) {
```
tdeekens: Wondering if a `Grid.Fixed` and `Grid.AutoFit` would be nicer than a `prop`. It feels a bit like a `type` on a button. I know it would likely complicate the implementation a bit cause we'd have the `GridContainer` and compose it with more specific `grid-template-columns` depending if it's `Grid.Fixed` and `Grid.Autofit`. Not sure if the suggestion makes sense or if others would agree.

src/components/grid/grid.js
emmenko: Sure, we don't need to merge this in right away. We can keep it open for now.

> The current one is pretty restrictive

Definitely. There is a lot of room for improvements, this one is the bare minimum.

src/components/grid/grid.js
emmenko: Yeah, that's one of the things we need to check with the design team.

emmenko: Hmm good point. I can check in the places where we use this already if the sizes make sense.

src/components/grid/grid.js
emmenko: Sure, we can do that. It would avoid "throwing" errors because of unsupported modes.

dferber90: We could also make one mode the default instead of throwing.

emmenko: https://github.com/commercetools/ui-kit/pull/442/commits/f9ac92b71af797caad5f13954bbefce89f75531d

qmateub:
Nice! Looking forward to use this! Thanks a lot @emmenko 🎉

emmenko:
Alright, I've rewrote the `<Grid>` component after talking to @dferber90. The component does not do anything per-se now, it just forwards the given css grid props to the styled component.

There are a couple of examples in the storybook to show the setup using fixed columns or auto-sizing columns.

dferber90:
I approved the new snaps as they're not really design-related ✅

PR #446, opened 2019-01-18
fix(money-input): avoid messing with floating point numbers

dferber90:
### Problem

Certain values get rounded on blur:

moneymoneymoney

### Problem

We run into JavaScript floating point number problems with our current calculations.

We are executing `amountAsNumber * 10 ** fractionDigitsOfAmount` which boils down to e.g. `8.066652 * 10 ** 6` for a high-precision price. However this is where we're running into floating point problems:

image

The function above converts `8.066652` into `8066652`

### Solution

We avoid floating point math problems by using string manipulation to get the `preciseAmount` instead.

---

Affects: `MoneyInput`, `MoneyField`, `LocalizedMoneyInput`, `LocalizedMoneyField`

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -221,7 +221,18 @@ export const createMoneyValue = (currencyCode, rawAmount) => {
       currencyCode,
       centAmount,
       preciseAmount: parseInt(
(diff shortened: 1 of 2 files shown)


montezume:
Awesome 🎉

tdeekens:
Nifty.

PR #447, opened 2019-01-19
WIP - TypeScript setup

emmenko:
**Not ready yet**

> This is an exploratory branch to evaluate using TypeScript.

---

TODOs:
- [x] Basic setup (`tsconfig.json`, etc)
- [x] Compile when bundling with Rollup 
- [x] Convert `<Constraints.Horizontal>` component (using VSCode plugin: https://github.com/lyft/react-javascript-to-typescript-transform)
- [x] Jest setup and convert tests of `<Constraints.Horizontal>` component
- [x] Setup `tslint`
- [x] Check that Storybook works
- [ ] Convert component using emotion library
- [ ] Try out the docs generator (see #320)

PR #449, opened 2019-01-20
WIP - ReasonML/BuckleScript setup

emmenko:
**Not ready yet**

> This is an exploratory branch to evaluate using ReasonML instead of TypeScript (#447).

---
TL;DR;

Please do not jump to conclusions here! Bear with me for a second.

In #447 we have a basic TS setup, so far this is my feedback:
- it took me about 3h to get there, where 2h 1/2 were spent googling and looking around to get the bare minimum setup working (and I haven't finished yet)
- I had to make some workarounds to get things working
- the tooling feels a bit "heavy", as in speed
- good vscode integration (obviously)
- type definitions are exported with the library (great for consumers)

I was more or less happy with the result. Then I thought, "_wait, why can't we use Reason instead and generate the TypeScript types/definitions instead?_"

So here's the idea:
- we write our components in Reason (we leverage a real type system)
- we can use typed CSS!!! ❤️ (https://www.npmjs.com/package/bs-css, which are bindings for Emotion)
-  we use `gentype` to generate the bindings!!!
  - if we target TypeScript, we can then export the type definitions in the bundle, so it's a win-win
- BuckleScript compiles extremely fast
- good editor integration
- we can optionally publish ui-kit as a Reason package, so that consumers can write Reason code without bindings

I'm not saying we need to do that, but from what I can see this way it will bring more benefits over a TypeScript approach only. As for the learning curve, I suppose there is not much difference, we will need to "learn" the types from both sides.

Let's think about this, experiment a bit more with both and then reach a decision.


tdeekens:
What is your expected response from us here?

tdeekens:
I appreciate the effort, suggestion and novelty of this. I also see the point in this having advantages over going to TypeScript. What confuses me is the fact that we settled on "attempting" to adopt TypeScript two days ago and now have to balance a discussion between TypeScript and Reason. 

Generally, I would not disagree that Reason has strengths over TypeScript itself. From e.g. generating TypeScript definitions or the type system itself.

However I think, at least I understood it in our discussions, that TypeScript would give us anything we'd need now:

1. Strong editor integration
2. Trust in refactoring due to a good type system
3. Ability to generate docs
4. Good interop with JavaScript
5. A not to steep learning curve while allowing incrementally typing things

Some of the above Reason can certainly also provide (plus as mentioned other things).

My main goal with UIKit is to keep it as approachable as possible. Given that it already, as any Design System, can (and is) a bottle neck for teams to get through, I'd like to keep the fear of contributing as low as possible. Part of our discussion in adopting TypeScript was that we at first won't require contributions to be typed (at least from what I recall) why we can then guide people to do so.

Taking the good arguments made:

1. we write our components in Reason (we leverage a real type system)
   - So TypeScript is not a "real" (whatever that means without context) type systems or put another way: can not provide the level of expressiveness we currently need for UIKit?
2. we can use typed CSS!!! ❤️ (npmjs.com/package/bs-css, which are bindings for Emotion)
   - Emotion has to my knowledge TypeScript bindings [here](https://github.com/emotion-js/emotion/blob/master/docs/typescript.md) which give some sense of security. What exactly does Reason provide we fear of running into bugs with?
3. BuckleScript compiles extremely fast
   - So TypeScript compiles too slow for the project size of UIKit and it is a drag during development?
4. good editor integration
   - Holds for both TS and Reason, right?
5. we can optionally publish ui-kit as a Reason package, so that consumers can write Reason code without bindings
   - Do we have anyone who would like to use UIKit in Reason? Do we have anyone using UIKit apart from us?

I am by no mean strictly against it (even if it may read as such). I just want the solution to balance the needs for UIKit while maintaining a low barrier to entry (not saying Reason has a frustratingly high barrier but I sense it would be higher than TS to others) and whatever we do not go after quenching our thirst for novel (and good do not get me wrong) technology. 

My personal gut feeling would be to be on the lookout for Reason and take that step after having proven (which I think it will be) and educated ourselves and others with TypeScript. Even if that may be slightly more boring or even technologically not strictly the best decision.

emmenko:
Sure thing, I didn't say that we should take this path. I just want us to evaluate different options before proceeding further.

src/components/constraints/horizontal/Horizontal.re
```diff
@@ -0,0 +1,43 @@
...
+module Styles = {
+  open Css;
+
+  let container = style([width(`percent(100.)), position(`relative)]);
+
+  let style_of_constraint = c => {
+    let def =
+      switch (c) {
+      | Scale => width(`percent(100.))
+      | Xs => minWidth(CustomProperties.constraintXs)
```
emmenko: This turned out to be pretty nice 🤓

PR #457, opened 2019-01-22
Refactor to use `@babel/preset-typescript` for TypeScript transpilation

tdeekens:
#### Summary

This pull request suggests/evaluates using the `@babel/preset-typescript` to compile out TypeScript.

#### Description

`@babel/preset-typescript` comes with most of (not all) TypeScript features to be compiled away. Generally, "How does Babel handle TypeScript code? It removes it". It supports most features. I only know that it does not not support namespaces or const enums because those require type information to transpile. Also does not support export = and import =, because those cannot be transpiled to ESM. I would argue that this could be fine as the initial step. Furthermore, babel is known to be faster than tsc (totally polemic argument though given the size of this code base).

If we would want to go this way we should add an `yarn type-check` script that invokes the TypeScript compiler. As this changes the workflow a bit - allowing easier experimentation - as you write code, you save, and it compiles (very quickly) without checking for type safety just yet.

1. Generally I like that we say within babel for both webpack and rollup. 
   - It's not like we sent our files through TS and then Babel. 
2. We can continue to use any JavaScript proposal and syntax we like (requires some thought)
   - Do we use `private` or `#instanceVar`?
   - do expressions could be a thing (not that they should)


package.json
```diff
@@ -219,7 +219,6 @@
     "tslint-config-prettier": "1.17.0",
     "tslint-eslint-rules": "5.4.0",
     "tslint-react": "3.6.0",
-    "typescript": "3.2.4",
```
emmenko: Don't we still need typescript? 🤔

tdeekens:
> Good idea, if it works it's fine by me, it seems to make the configuration a bit easier 👌

Yes, I saw that as the main downside with the TypeScript adoption that it bloats our build setup a bit. I wanted to minimise the impact here a bit.

> Don't we still need typescript? 🤔

Yes, for the type checking eventually to my understanding. Will look into that now.

tdeekens:
The last two commits add `tsc` for TypeScript while the compiler now runs in `noEmit`-mode for it. Additionally the `type-check` run script now runs on CI.

tsconfig.json
```diff
@@ -16,53 +16,53 @@
   "compilerOptions": {
     /* Basic Options */
     "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
-    "module": "esnext",                    /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
-    "lib": ["es2015", "dom"],              /* Specify library files to be included in the compilation. */
-    "allowJs": true,                       /* Allow javascript files to be compiled. */
+    "module": "esnext",                       /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
```
emmenko: PS: there was a reason why I left the comments indented, to mark the options that have changed from the default values.

Eventually we can clean up this file, as we're testing things out I found it useful to be honest 😇

tsconfig.json
```diff
@@ -16,53 +16,53 @@
...
     // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
     // "sourceMap": true,                     /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    "outDir": "dist",                      /* Redirect output structure to the directory. */
+    "outDir": "dist",                         /* Redirect output structure to the directory. */
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     "pretty": true,
     // "composite": true,                     /* Enable project compilation */
-    "removeComments": true,                /* Do not emit comments to output. */
-    // "noEmit": true,                        /* Do not emit outputs. */
```
emmenko: I think this broke storybook...or something else I don't remember exactly. There was a problem with this option as far as I could remember.

We should check that everything works, maybe with the babel setup it's not a problem anymore. Just wanted to raise awareness about this.

tdeekens: Ah, yeah thänks. `noEmit` now can be used as babel removes the types and actually creates the bundle. While before when set to `true` the compiler would not yield anything to webpack hence breaking storybook.

I fixed the option as suggested in the `tsconfig.declarations.json` so we emit there.

tdeekens:
Let me know if you're in on this and want this in your base branch.

PR #458, opened 2019-01-22
Refactor `Stamp` to use emotion

montezume:
#### Summary

Refactors `stamp` to use emotion for styling.


src/components/stamp/stamp.js
```diff
@@ -13,8 +13,61 @@ export const availableTones = [
...
+    }
+    default:
+      return css``;
+  }
+};
+
 const Label = props => (
-  <div className={classnames(styles.container, styles[`tone-${props.tone}`])}>
+  <div
+    css={[
```
emmenko: Do you want to do it like this now instead of concatenating strings?

emmenko:
What is wrong with the percy snapshots?

tdeekens:
> What is wrong with the percy snapshots?

To me it looks like font rendering issues. At least judging from where the red marks appear.

src/components/stamp/stamp.js
tdeekens: I kind of like the array form for composition more to be fair. It feels a bit less bulky than template literals.

emmenko: Sure, I just need to know so that I can also follow that approach 😇

I just saw this now --> https://emotion.sh/docs/composition 😅

montezume:
Maybe the font color is wrong. i'll check it out later before merging.

src/components/stamp/stamp.js
tdeekens: Yeah it's kind of cool what they all offer also in relation to object styles. I guess we can just try out some approaches and then decide on one. Should all be internal to the library.

PR #459, opened 2019-01-22
refactor(buttons): to use emotion styles

emmenko:
This turned out to be a bit more work than expected. For once, because our buttons are really weird both in implementation and styling, and also because the tests were like 🤦‍♂️

I believe I also fixed some errors (see inline comments).

And tests have been simplified using RTL, for the visual stuff we have VRT, so we don't need to write unit tests for it.

There would be some things to check/confirm on Percy I would imagine...take your time to review.


src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -1,73 +1,74 @@
...
+        padding: 0;
+        margin: 0;
+        white-space: nowrap;
+        cursor: pointer;
+        color: inherit;
+        font: inherit;
+        font-size: ${vars['--font-size-default']};
+        font-family: ${vars['--font-family-default']};
+      `,
+      props.isDisabled &&
```
montezume: Nice 💯

src/components/buttons/accessible-button/README.md
```diff
@@ -39,19 +39,19 @@ export default Button;
...
+| type             | String    |     -     | Used as the HTML `type` attribute. Supported values are `submit`, `reset` and `button`.                                                                                                      |
+| label            | String    |     ✓     | The text that should be used as `aria-label`                                                                                                                                                 |
+| children         | Node      |     ✓     | Button content                                                                                                                                                                               |
+| isToggled        | Bool      |     -     | Set this prop to indicate a toggle button and it's state                                                                                                                                     |
+| ref              | React ref |     -     | Used as button reference                                                                                                                                                                     |
```
emmenko: I changed that to use `React.createRef` and `React.forwardRef`. We actually never use this prop, so I'm not sure if we still need it or not...

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -1,73 +1,74 @@
...
+        css`
+          cursor: not-allowed;
+        `,
+    ]}
+    // Allow to override the styles by passing a `className` prop.
+    // Custom styles can also be passed using the `css` prop from emotion.
+    // https://emotion.sh/docs/css-prop#style-precedence
+    className={props.className}
```
montezume: 👍

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -1,73 +1,74 @@
...
+        color: inherit;
+        font: inherit;
+        font-size: ${vars['--font-size-default']};
+        font-family: ${vars['--font-family-default']};
+      `,
+      props.isDisabled &&
+        css`
+          cursor: not-allowed;
+        `,
+    ]}
```
tdeekens: So is there a rule of thumb for us the extract these things?

src/components/buttons/icon-button/icon-button.js
```diff
@@ -1,26 +1,25 @@
...
+import {
+  getStateStyles,
+  getShapeStyles,
+  getSizeStyles,
+  getThemeStyles,
+} from './icon-button.styles';
 
 // Gets the color which the icon should have based on context of button's state/cursor behavior
 const getIconThemeColor = props => {
+  const isActive = props.isToggleButton && props.isToggled;
```
emmenko: Based on the prop types validation, the active state is only set when both props are true.

Please correct me if I'm wrong...

src/components/buttons/icon-button/icon-button.js
```diff
@@ -33,56 +32,55 @@ const getIconThemeColor = props => {
...
+        getThemeStyles(props.theme),
+      ]}
     >
       <AccessibleButton
-        buttonAttributes={dataProps}
+        buttonAttributes={buttonAttributes}
         type={props.type}
         label={props.label}
         onClick={props.onClick}
+        isToggleButton={props.isToggleButton}
```
emmenko: This was not passed

src/components/buttons/link-button/link-button.spec.js
```diff
@@ -1,141 +1,43 @@
...
-        expect(wrapper.find('Link')).toHaveProp('onClick');
-      });
-      it('when `onClick` is fired, event default is prevented', () => {
-        expect(testEvent.preventDefault).toHaveBeenCalled();
-      });
-    });
-
-    it('renders the icon with "grey" theme', () => {
-      expect(wrapper.find(PlusThinIcon)).toHaveProp('theme', 'grey');
-    });
```
tdeekens: Nice that we move this to rtl. Could this have been another PR though or was it required in the refactor to emotion?

src/components/buttons/icon-button/icon-button.js
```diff
@@ -33,56 +32,55 @@ const getIconThemeColor = props => {
...
-  return themeClassName;
-};
-
 export const IconButton = props => {
-  const dataProps = {
+  const buttonAttributes = {
     'data-track-component': 'IconButton',
     ...filterDataAttributes(props),
   };
+  const isActive = props.isToggleButton && props.isToggled;
```
emmenko: Same here

src/components/buttons/secondary-button/secondary-button.styles.js
```diff
@@ -0,0 +1,97 @@
...
+      box-shadow: ${vars['--shadow-9']};
+      background-color: ${vars['--color-white']};
+    }
+  `;
+};
+
+const getThemeStyles = theme => {
+  if (!theme) return css``;
+
+  if (theme === 'default') return css``;
```
tdeekens: Maybe move into the switch?

src/components/buttons/primary-button/primary-button.js
```diff
@@ -1,48 +1,51 @@
...
       >
-        {Boolean(props.iconLeft) && (
-          <span className={styles['icon-container']}>
-            {React.cloneElement(props.iconLeft, {
-              theme: props.isDisabled ? 'grey' : 'white',
-            })}
-          </span>
-        )}
-        <span>{props.label}</span>
+        <Spacings.Inline alignItems="center" scale="xs">
```
emmenko: I saw that this was used in the SecondaryButton and figured it should be used here as well

src/components/buttons/flat-button/flat-button.js
```diff
@@ -1,48 +1,83 @@
...
-      size: 'medium',
-      theme: iconTheme,
-    });
+  return React.cloneElement(props.icon, {
+    size: 'medium',
+    theme: iconTheme,
+  });
+};
+
+const getColorTone = (tone, isHover = false) => {
```
montezume: shouldn't this be getColor or getPColor or getTextColor?

src/components/buttons/icon-button/icon-button.js
tdeekens: I us previously fiddling a lot with this but it looks still good.

src/components/buttons/flat-button/flat-button.js
```diff
@@ -1,48 +1,83 @@
...
+          font-size: 1rem;
+          border: none;
+          background: none;
+          padding: 0;
+          min-height: initial;
+
+          > * + * {
+            margin: 0 0 0 ${vars['--spacing-4']};
+          }
+          p {
```
montezume: This is because we are using Text.Body right? Maybe it makes more sense to allow Text.Body to accept a className and use `css` to pass the className for this.

src/components/buttons/secondary-button/secondary-button.js
```diff
@@ -38,65 +29,94 @@ export const getIconThemeColor = props => {
...
+            })}
+          </span>
+        )}
+        <span>{props.label}</span>
+      </Spacings.Inline>
+    </AccessibleButton>
+  );
 
-  const shouldLink = !props.isDisabled && Boolean(props.linkTo);
-  const WrapperComponent = shouldLink ? Link : Div;
```
emmenko: I found it weird to have this component. Instead we should simply render the `<Link>` (with the `to` prop) or a normal `<div>`.

src/components/buttons/icon-button/icon-button.js
```diff
@@ -33,56 +32,55 @@ const getIconThemeColor = props => {
...
+          box-shadow: ${vars['--shadow-7']};
+          color: ${vars['--color-black']};
+          transition: background-color ${vars['--transition-linear-80ms']},
+            box-shadow 150ms ease-in-out;
+        `,
+        getStateStyles(props.isDisabled, isActive, props.theme),
+        getShapeStyles(props.shape, props.size),
+        getSizeStyles(props.size),
+        getThemeStyles(props.theme),
+      ]}
```
montezume: I like this style 🎉

src/components/buttons/icon-button/icon-button.js
```diff
@@ -33,56 +32,55 @@ const getIconThemeColor = props => {
...
-          className={classnames(
-            styles['button-layout'],
-            styles[`container-${props.size}`]
-          )}
+          css={[
+            css`
+              display: flex;
+              align-items: center;
+              justify-content: center;
+              > * + * {
```
montezume: I find these kinds of styles super weird.

src/components/buttons/accessible-button/accessible-button.js
emmenko: You mean extract into a separate file?

src/components/buttons/link-button/link-button.spec.js
emmenko: Yes and no. But since I changed some stuff internal to the components, I would have had to fix all the tests. This was way quicker...

src/components/buttons/accessible-button/accessible-button.js
tdeekens: Yeah. I guess it's just gut feeling.

src/components/buttons/link-button/link-button.spec.js
tdeekens: Makes total sense.

src/components/buttons/flat-button/flat-button.js
emmenko: Yeah, I think you were trying to do the same right?

src/components/buttons/icon-button/icon-button.js
emmenko: Actually there are super useful, that's what the spacings are based for. However, I can check if we need to have to manually apply them in the button components...

src/components/buttons/secondary-button/secondary-button.styles.js
emmenko: The switch default is used for the invariant check, that's way it's defined before.

src/components/buttons/accessible-button/accessible-button.js
emmenko: I find it useful to do once you start having more than a couple of long `getXxxStyles` functions, at which point it only pollutes the component file. Otherwise, for "small" styles I think it's better to have it within the component, for readability.

emmenko:
I'll inspect percy and fix up the styles.

src/components/buttons/primary-button/primary-button.js
emmenko: I guess this relates to this VRT diff

image

I think it should be fine to accept this. /cc @montezume

emmenko: Same for this one:

image

emmenko:
I had to convert also the `PrimaryActionDropdown` because of specificity when passing `className` to the `AccessibleButton` (I suppose the style passed as `className` gets applied before the `className` transformed by the `css` prop.

src/components/buttons/icon-button/icon-button.js
montezume: Sure, but since we don't allow users to pass in children to the buttons, why not just add the spacing to the text and the icon?

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -54,7 +96,33 @@ const DropdownChevron = React.forwardRef((props, ref) => (
...
+      css`
+        display: flex;
+        align-items: center;
+        padding: 0 ${vars['--spacing-4']};
+        border-radius: 0 ${vars['--border-radius-6']}
+          ${vars['--border-radius-6']} 0;
+        border-left: 1px solid ${vars['--color-gray']};
+        height: ${vars['--big-button-height']};
+      `,
+      props.isDisabled
```
tdeekens: Sorry to be annoying but can we extract this maybe into a little helper. I find it easier to follow with e.g. an if/else there as this gets quite bulky.

src/components/buttons/accessible-button/README.md
montezume: 💯

src/components/buttons/accessible-button/README.md
```diff
@@ -39,19 +39,19 @@ export default Button;
...
+| isToggleButton   | Bool      |     -     | Used to set `aria-pressed˚ state                                                                                                                                                             |
+| isDisabled       | Bool      |     -     | Set this prop to indicate a disabled button                                                                                                                                                  |
+| onClick          | Function  |     ✓     | Callback that is called when the button is clicked                                                                                                                                           |
+| className        | String    |     -     | Will be applied to the `<button/>` to allow for custom styling. Alternatively, you can pass a `css` prop if you're [styling your components with Emotion](https://emotion.sh/docs/css-prop). |
```
montezume: I'm not sure we need to include this doc because it's pretty much just explaining how emotion `css` prop works 🤔

src/components/buttons/flat-button/flat-button.js
montezume: I guess for now we leave it like it is and we make another PR to allow Text to accept classNames?

src/components/buttons/accessible-button/README.md
emmenko: Yeah I'll remove it, also because passing a `className` doesn't really work. See https://github.com/commercetools/ui-kit/pull/459#issuecomment-456497484

src/components/buttons/secondary-button/secondary-button.styles.js
tdeekens: I saw that. Nevermind.

emmenko:
Alright, this is good to go now. VRT changes have been approved after checking them with Filip.

src/components/buttons/flat-button/flat-button.js
emmenko: Yep, makes sense

montezume:
@emmenko it looks like the border radius of the PrimaryActionDropdown is wrong

emmenko:
Fixed, all good now!

montezume:
@emmenko I just pushed fixes...

montezume:
Can you hold off on this?

montezume:
NEvermind, looks good 👍

PR #462, opened 2019-01-23
refactor(content-notification): use emotion

montezume:
#### Summary

Refactors `ContentNotification` to be styled with `emotion`.


src/components/notifications/content-notification/content-notification.js
```diff
@@ -1,9 +1,24 @@
...
+const getIconContainerBackgroundColour = props => {
+  switch (props.type) {
+    case 'error':
+      return vars['--color-red'];
+    case 'info':
+      return vars['--color-blue'];
+    case 'warning':
+      return vars['--color-orange'];
+    case 'success':
+      return vars['--color-green'];
```
tdeekens: Wait. This is not to way to go now?

montezume: Why not?

montezume: I missed that we made a breaking change 🤦‍♂️

PR #463, opened 2019-01-23
refactor: to import custom props as js object

emmenko:
Follow up of #460 and #461

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ You can also access the JavaScript variables like this
 ```js
 import { customProperties } from '@commercetools-frontend/ui-kit';
 
-const green = customProperties['--color-green'];
+const green = customProperties.colorGreen;
 ```
 
-> Please look at the [file](materials/custom-properties.json) itself to inspect which variables are available (_documentation will be provided in the future_).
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -91,10 +91,10 @@ You can also access the JavaScript variables like this
 ```js
 import { customProperties } from '@commercetools-frontend/ui-kit';
```
emmenko: I think this export should point to the js object, no the json one. I also changed the export in `src/index`

package.json
```diff
@@ -20,6 +20,7 @@
     "i18n",
     "materials/custom-properties.css",
     "materials/custom-properties.json",
+    "materials/custom-properties.js",
```
emmenko: Not sure it's necessary since we are exporting it now also as a named export from `src/index`

tdeekens: Yeah I guess it's not really needed but we can keep it given anybody wants to copy it into their projects I suppose.

emmenko:
I spoke to @montezume and we decided to mark this as a breaking change, because of the named export now returning a JS object.

emmenko:
Given that I got two approvals and that all builds are green, I'm going ahead and merge this, so that we can unblock other PRs

PR #466, opened 2019-01-23
refactor(field-label): to use emotion styles

emmenko:
That's right, we basically do not need to pass custom styles anymore for the FieldLabel.


emmenko:
I'll check with Filip if the new spacings are ok

montezume:
Looks like there's a problem here

screen shot 2019-01-23 at 3 20 04 pm

montezume:
A few problems with things stacking when they should not.

emmenko:
Yes I noticed that, I'm on it

emmenko:
FYI: I'll check with Filip about the visual diffs and approve if it's ok like this.

emmenko:
Snapshots have been approved. @montezume anything else left from your side?

PR #467, opened 2019-01-23
chore: speed up storybook build

montezume:
I noticed that when we build storybook, it's including all the moment locales and (I think), slowing it down a bit. So I added the plugin to strip out unused locales.

screen shot 2019-01-23 at 3 17 52 pm

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
 const postcssImport = require('postcss-import');
 const postcssPresetEnv = require('postcss-preset-env');
 const postcssReporter = require('postcss-reporter');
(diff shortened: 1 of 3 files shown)


.storybook/webpack.config.js
```diff
@@ -21,6 +22,9 @@ const sourceFolders = [
 ];
 
 module.exports = (storybookBaseConfig, configType) => {
+  storybookBaseConfig.plugins.push[
```
tdeekens: This looks very odd. You sure `push[1]`?

montezume: What do you mean `push[1]`? Seems like the same here

https://github.com/storybooks/storybook/issues/2115

tdeekens: Do we invoke a function here or dereference an array?

montezume: Sorry 🤦‍♂️

PR #468, opened 2019-01-23
refactor(icons): to use emotion styles

emmenko:
This is exiting!

The HOCs are now gone, maybe tree shaking works again now 🤓


src/components/icons/index.js
```diff
@@ -106,323 +105,225 @@ import OrigVerifiedIcon from './svg/verified.react.svg';
...
-export { TagStackedIcon };
-export { TagIcon };
-export { TreeStructureIcon };
-export { TruckIcon };
-export { UserFilledIcon };
-export { UserLinearIcon };
-export { VerifiedIcon };
-export { WarningIcon };
-export { WorldIcon };
+export const AngleDownIcon = createStyledIcon(
```
emmenko: We can export this directly now 🎉

src/components/icons/create-styled-icon.js
```diff
@@ -0,0 +1,91 @@
...
+      invariant(
+        theme,
+        `ui-kit/Icon: the specified theme '${theme}' is not supported.`
+      );
+      return 'inherit';
+    }
+  }
+};
+
+export default function createStyledIcon(Component, displayName) {
```
tdeekens: Why do we need a factory function for this? We have quite a few of these in the app and they cause wasteful renders if I recall. Would it be an option to just have an `StyledIcon` component accepting an `iconComponent` prop and `iconDisplayName`?

emmenko: The function is only used to create the component. The exported component is already styled, there are wasteful rerenders

montezume:
How come the MoneyField tests are failing here?

emmenko:
The scale size seems broken. I'll look into it

src/components/icons/create-styled-icon.js
```diff
@@ -13,7 +13,7 @@ const getSizeStyle = size => {
   switch (size) {
     case 'scale':
       return `
-        svg.scale:not(:root) {
+        &:not(:root) {
```
montezume: 👍

montezume:
Excited to see if this helps with tree shaking 😇

PR #469, opened 2019-01-23
refactor(inputs/checkbox): to use emotion styles

emmenko:


src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -1,12 +1,80 @@
...
+  return css`
+    cursor: pointer;
+    &:hover svg [id$='borderAndContent'] > [id$='border'] {
+      stroke: ${vars.borderColorInputFocus};
+    }
+  `;
+};
+
+const getCheckboxWrapperStyles = (isDisabled, hasError, isHovered) => {
+  const baseStyles = css`
```
tdeekens: Can we hoist this out of the function and name it checkboxWrapperBaseStyles? Would avoid the variable to kind of regenerate the css on each invocation. Not sure it makes a difference in the grand scheme of things but my inner urge still sees this as a bit wasteful.

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -1,12 +1,80 @@
...
+      baseStyles,
+      css`
+        svg [id$='borderAndContent'] > [id$='border'] {
+          stroke: ${vars.borderColorInputFocus};
+        }
+      `,
+    ];
+  }
+  return baseStyles;
+};
```
tdeekens: So before we moved them into a .styles. Then we inline them. Now we have them in fns above. What do we prefer. Should we settle in a way - at least for fns in .styles or above?

emmenko: From my point of view so far:
- simple styles can be "inlined" so it's easier to read them
- if the style requires some "complex" conditionals, move the style into a function
- if you start having a lot of functions that pollutes the component, move them into a `styles` file

So for me it's more about readability. As long as the component is readable, it's fine to have the styles within the same file, otherwise move it to a separate file.

In this particular case I guess we can move the functions into a separate file

tdeekens: Yes, I agree with the loose rule of thumb and also that in this case they should be moved.

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -49,21 +52,15 @@ class CheckboxInput extends React.PureComponent {
...
             <div
-              className={classnames(styles.checkboxWrapper, {
-                [styles.isDisabled]: this.props.isDisabled,
-                [styles.isHovered]:
-                  this.props.isHovered &&
-                  !this.props.isDisabled &&
-                  !this.props.hasError,
-                [styles.hasError]: this.props.hasError,
-              })}
+              css={getCheckboxWrapperStyles(
```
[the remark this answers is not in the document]
emmenko: Yeah, I started doing this in the new components I'm migrating. I'll fix up those functions after that

emmenko:
Closing in favour of #473

PR #470, opened 2019-01-24
Migrate `Tag` to emotion

montezume:


src/components/tag/tag.js
```diff
@@ -1,51 +1,96 @@
...
+
+const getClickableContentWrapperStyles = ({ type, isRemoveable }) =>
   type === 'warning'
-    ? classnames(styles.clickableContentWrapperWarning, {
-        [styles.clickableRemovableContentWrapperWarning]: isRemovable,
-      })
-    : classnames(styles.clickableContentWrapperNormal, {
-        [styles.clickableRemovableContentWrapperNormal]: isRemovable,
-      });
+    ? [
```
tdeekens: Just wondering if we could even do something like:

```js
const getClickableContentWrapperStyles = ({ type, isRemoveable }) => [
  type === 'warning' && css`
    &:hover {
      border-color: ${vars.borderColorTagWarningHover};
      border-right: 1px solid ${vars.borderColorTagPristine};
    }
  `,
  type === 'warning' && isRemoveable && css`
    border-right: 1px solid var(--border-color-tag-warning);
  `

  type =!== 'warning' && css`
    &:hover {
      border-color: ${vars.borderColorTagFocus};
    }
  `
  type =!== 'warning' && isRemoveable && css`
    &:hover {
      border-right: 1px solid ${vars.borderColorTagPristine};
    }
  `
].filter(Boolean)
```

Not sure maybe that's easier to follow. Just what I thought when reading this.

src/components/tag/tag.js
```diff
@@ -1,51 +1,96 @@
...
+        css`
+          &:hover {
+            border-color: ${vars.borderColorTagWarningHover};
+            border-right: 1px solid ${vars.borderColorTagPristine};
+          }
+        `,
+        isRemoveable &&
+          css`
+            border-right: 1px solid var(--border-color-tag-warning);
```
tdeekens: Is this a `vars.$` too?

src/components/tag/tag.js
montezume: I find the whole component hard to follow, I would rather refactor it not to be duplicated with the Link and the Regular, and I find this even more confusing

src/components/tag/tag.js
montezume: ya thanks 👍

src/components/tag/tag.js
tdeekens: Sure, it's subjective in the end just wanted to make a suggestion.

PR #473, opened 2019-01-24
refactor(inputs): to use emotion styles

emmenko:
You can review on a commit basis


emmenko:
I'll see if I can extract a basic input styled component, as most of the inputs have the same basic styles

src/components/inputs/localized-multiline-text-input/translation-input.styles.js
```diff
@@ -0,0 +1,112 @@
...
+
+/* we need this line-height to achieve 32px height when the component has only one row */
+const sizeInputLineHeight = '22px';
+
+// NOTE: order is important here
+// * a disabled-field currently does not display warning/error-states so it takes precedence
+// * a readonly-field cannot be changed, but it might be relevant for validation, so error and warning are checked first
+// how you can interact with the field is controlled separately by the props, this only influences visuals
+const getTextareaStyles = props => {
+  const baseInputStyles = [
```
tdeekens: I know it doesn't make much of a difference but these could be hoisted on top of the function? Which would also imply we would invoke `css` once?

src/components/inputs/text-input/text-input.styles.js
```diff
@@ -0,0 +1,85 @@
...
+        border-color: ${vars.borderColorInputPristine};
+        background: ${vars.backgroundColorInputPristine};
+        color: ${vars.fontColorForInput};
+      }
+    `,
+  ];
+};
+
+// eslint-disable-next-line import/prefer-default-export
+export { getInputStyles };
```
tdeekens: I like these. All super easy to follow!

src/components/inputs/toggle-input/toggle-input.styles.js
```diff
@@ -0,0 +1,185 @@
+import { css } from '@emotion/core';
+import vars from '../../../../materials/custom-properties';
+
+const toggleSwitchSizes = {
+  small: {
+    height: '16px',
```
tdeekens: Can't we use spacings here?

emmenko: Those are all custom numbers, I found it totally confusing how it was before. In this case, I prefer to have them explicitly readable.

src/components/inputs/localized-multiline-text-input/translation-input.styles.js
```diff
@@ -0,0 +1,58 @@
...
+/* we need this line-height to achieve 32px height when the component has only one row */
+const sizeInputLineHeight = '22px';
+
+// NOTE: order is important here
+// * a disabled-field currently does not display warning/error-states so it takes precedence
+// * a readonly-field cannot be changed, but it might be relevant for validation, so error and warning are checked first
+// how you can interact with the field is controlled separately by the props, this only influences visuals
+const getTextareaStyles = props => {
+  const baseStyles = [
+    getInputStyles(props),
```
emmenko: The "shared" input styled now can simply be spread, and overriden depending on the component

src/components/inputs/number-input/number-input.spec.js
```diff
@@ -75,21 +75,6 @@ describe('NumberInput', () => {
     expect(container.querySelector('input')).toHaveAttribute('type', 'number');
   });
 
-  it('should have pristine style when untouched', () => {
-    const { container } = render(<NumberInput {...baseProps} />);
-    expect(container.querySelector('input')).toHaveClass('pristine');
```
emmenko: Those tests are not necessary, since we have VRT.
Also, this doesn't work anymore as-is, because the class names from emotion are different.

emmenko:
Some small issues reported by percy, I'm going to fix those now

emmenko:
The errors reported by percy are actually ok, I went through them with Filip and we will approve them (basically there were some inconsistencies, so this refactoring fixed them as well) 😉

emmenko:
Just to make it more clear:

- text color for error state should be red

image

- disabled multi line text inputs should still have the "resize" control

image

emmenko:
Oh I forgot about the calendar picker styles (they are in the `internals` folder). Will push that in this PR as well

src/components/inputs/toggle-input/toggle-input.js
```diff
@@ -25,17 +24,25 @@ class ToggleInput extends React.PureComponent {
   render() {
     return (
       <label
-        className={classnames(styles.labelWrapper, {
-          [styles.labelWrapperDisabled]: this.props.isDisabled,
-        })}
+        css={
+          this.props.isDisabled
+            ? css`
+                cursor: not-allowed;
```
montezume:
```suggestion
                cursor: ${this.props.isDisabled ? 'not-allowed': 'pointer'};
```

src/components/inputs/toggle-input/toggle-switch.js
```diff
@@ -1,61 +1,42 @@
...
-      props.isChecked
-    )}
-  />
+  <Icons.ButtonDefault css={getToggleButtonStyles(props)} />
 );
-
 ToggleButton.displayName = 'ToggleButton';
 ToggleButton.propTypes = {
-  size: PropTypes.string.isRequired,
+  size: PropTypes.oneOf(['small', 'big']).isRequired,
```
montezume: unrelated: weird how we use 'small' and 'big' here and yet 's' and 'l' in LoadingSpinner.

src/components/internals/calendar-content/calendar-content.js
```diff
@@ -0,0 +1,8 @@
+import styled from '@emotion/styled';
+
+const CalendarCalendar = styled.ul`
```
montezume: nitpick: the file is renamed so this should be renamed as well?

src/components/inputs/date-time-input/date-time-input.js
```diff
@@ -6,7 +6,7 @@ import { parseTime } from '../../../utils/parse-time';
 import CalendarBody from '../../internals/calendar-body';
 import CalendarMenu from '../../internals/calendar-menu';
 import CalendarHeader from '../../internals/calendar-header';
-import CalendarCalendar from '../../internals/calendar-calendar';
+import CalendarContent from '../../internals/calendar-content';
```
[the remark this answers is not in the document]
emmenko: Suggestion? I find `CalendarCalendar` very confusing to be honest...

PR #474, opened 2019-01-24
chore: migirate loading spinner to emotion

montezume:


src/components/loading-spinner/loading-spinner.js
```diff
@@ -1,34 +1,69 @@
...
+  transform: rotate(0deg);
+}
+
+to {
+  transform: rotate(360deg);
+}
+`;
+
+const LoadingSpinner = props => {
+  const size = props.scale === 's' ? sizeSmallSvg : sizeLargeSvg;
```
tdeekens: Needless but could also be something like:

```js
const sizePerScale = {
   s: '18px',
   l: '32px',
};

const size = sizePerScale[props.scale] && '32px'
```

src/components/loading-spinner/loading-spinner.js
```diff
@@ -1,34 +1,72 @@
...
-const LoadingSpinner = props => (
-  <Spacings.Inline alignItems="center">
-    <div className={styles[`container-${scaleClassMap[props.scale]}`]}>
-      <svg
-        className={styles[`svg-${scaleClassMap[props.scale]}`]}
-        viewBox="0 0 40 40"
+const positionOrigin = '20px';
+
+const LoadingSpinner = props => {
+  const size = sizePerScale[props.scale] && '32px';
```
emmenko: Do we need the `&& '32px'`?

montezume: Ah I guess not since we have a default prop.

montezume: I guess it means that if someone puts in an invalid value, we will get '32' as a default? otherwise if I pass in 'm' I'll get undefined without the && '32px'

PR #475, opened 2019-01-24
Refactor to remove `recompose`

tdeekens:
#### Summary

This pull request removes the dependency on `recompose` and refactors all HoCs to work without while using `flowRight` from lodash instead of `compose`.

#### Description

While playing with TypeScript I noticed that `recompose` HoCs are a bit hard to type. Then I saw that we only use recompose for two rather simple HoCs. I think it's not really worth to keep recompose just for them. This should reduce bundle size a bit (if not at least simplify the code).

### DIFF
--- a/src/hocs/with-mouse-down-state/with-mouse-down-state.js
+++ b/src/hocs/with-mouse-down-state/with-mouse-down-state.js
@@ -1,5 +1,5 @@
+import flowRight from 'lodash.flowright';
 import {
-  compose,
   withState,
   withHandlers,
   wrapDisplayName,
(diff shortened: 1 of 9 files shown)


src/hocs/with-mouse-down-state/with-mouse-down-state.js
```diff
@@ -1,19 +1,36 @@
...
+    };
+
+    setMouseDown = nextMouseDown =>
+      this.setState({ isMouseDown: nextMouseDown });
+
+    handleMouseDown = () => this.setMouseDown(true);
+    handleMouseUp = () => this.setMouseDown(false);
+
+    render() {
+      return React.createElement(BaseComponent, {
```
emmenko: Any reason why using `createElement` over `<BaseComponent {...props} />`?

tdeekens: I like `createElement`, I have no idea why. I can change it.

emmenko: I don't mind, I was just curious if there was a specific reason 😇

PR #476, opened 2019-01-24
refactor(label): to use emotion styles

emmenko:


emmenko:
Who cares, as long as we get it done 😉

tdeekens:
We can all have a shared call with a drink when postcss is out of the library.

PR #478, opened 2019-01-25
refactor(table): to use emotion styles

emmenko:
I survived the Table, with some scratches though... 💪🤕

I had to work around some things, at least the output should be the same as before. Given that we plan to replace the table with a new implementation, I wouldn't care much about those things. Same for the tests, I didn't dare re-writing them with RTL.


src/components/table/base-table/base-table.js
```diff
@@ -55,15 +56,23 @@ export default class BaseTable extends React.Component {
          * Custom class that is added to the header's cell container.
          */
         headerClassName: PropTypes.string,
+        headerStyle: PropTypes.object,
```
emmenko: Same as `headerClassName`, but for CSS-in-JS usage

src/components/table/base-table/base-table.js
```diff
@@ -55,15 +56,23 @@ export default class BaseTable extends React.Component {
...
          * Function to return a custom class that is added to the cell's
          * container.
          */
         classNameGetter: PropTypes.func,
+        cellClassNameGetter: PropTypes.func, // alias
+        /**
+         * Function to return a custom class that is added to the cell's
+         * container.
+         */
+        cellStyleGetter: PropTypes.func,
```
emmenko: Same as `classNameGetter`, but for CSS-in-JS

src/components/table/base-table/base-table.js
```diff
@@ -90,6 +99,7 @@ export default class BaseTable extends React.Component {
     sortDirection: PropTypes.oneOf(['ASC', 'DESC']),
     sortBy: PropTypes.string,
     tableClassName: PropTypes.string,
+    tableStyle: PropTypes.object,
```
emmenko: Same as `tableClassName`, but for CSS-in-JS

src/components/table/base-table/base-table.js
```diff
@@ -422,41 +462,79 @@ export default class BaseTable extends React.Component {
...
+
+  itemRenderer = renderParams => (
+    <CellMeasurer
+      cache={this.cellMeasurerCache}
+      columnIndex={renderParams.columnIndex}
+      key={renderParams.key}
+      parent={renderParams.parent}
+      rowIndex={renderParams.rowIndex}
+    >
+      {this.renderItemCell(renderParams)}
```
emmenko: Just to make the tests setup a bit more clean

src/components/table/base-table/base-table.js
```diff
@@ -329,51 +339,81 @@ export default class BaseTable extends React.Component {
...
 
-  itemRenderer = renderParams => {
+  renderItemCell = renderParams => {
     // `columnDefinition` is the original column-definition from where you defined the table
     // `renderParams` holds similar looking-information but is the data provided by "react-virtualized"
     const columnDefinition = this.columns[renderParams.columnIndex];
+    if (renderParams.rowIndex === 0) {
+      return (
+        <ClassNames>
+          {({ css: makeClassName }) => (
```
emmenko: Given that we need to merge the class names provided to the component with some default ones, we need to use this workaround to generate a classname from emotion, and merge that with the user defined ones.

montezume:
Given don't have VRT's for Table (I tried but something to do with the virtualization led to constant false positives), we should do some manual testing 😫

src/components/table/base-table/base-table.js
```diff
@@ -329,51 +339,81 @@ export default class BaseTable extends React.Component {
...
-              styles['header-cell'],
-              columnDefinition.headerClassName,
-              styles[`align-${columnDefinition.align || 'left'}`]
-            )}
-          >
-            {this.headerRenderer(columnDefinition)}
-          </div>
-        ) : (
+      <ClassNames>
+        {({ css: makeClassName }) => (
```
emmenko: Same here

src/components/table/base-table/base-table.js
```diff
@@ -422,41 +462,79 @@ export default class BaseTable extends React.Component {
...
+              style={{
+                ...(this.props.tableStyle || {}),
+                width: this.state.width,
+              }}
+              {...filterDataAttributes(this.props)}
+            >
+              <MultiGrid
+                ref={this.registerMultiGrid}
+                styleTopLeftGrid={{ background: vars.colorGray90 }}
+                styleBottomLeftGrid={{
```
emmenko: The MultiGrid already accepted both "className" and "style" props

src/components/table/base-table/base-table.js
```diff
@@ -329,51 +339,81 @@ export default class BaseTable extends React.Component {
...
+              columnDefinition.cellClassName,
               columnDefinition.classNameGetter
                 ? columnDefinition.classNameGetter({
                     rowIndex: this.getBodyRowIndex(renderParams.rowIndex),
                     columnKey: columnDefinition.key,
                     height: renderParams.height,
                     width: renderParams.width,
                   })
+                : undefined,
+              columnDefinition.cellClassNameGetter
```
tdeekens: Just suggestions:

1. Defaulting it could avoid the check
2. `getCellClassName` Maybe? Then the `*Getter` is not rly needed

src/components/table/base-table/base-table.js
tdeekens: 💯

src/components/table/base-table/base-table.js
emmenko: Hmm, I wouldn't change the name, otherwise it's a breaking change as this is already used (actually the `classNameGetter` is, the `cellClassNameGetter` is just an alias, but they should have a consistent name).

emmenko:
The only thing I noticed so far that needs to be fixed is the arrow for the sortable header.

PS: I'm using the ui-kit bundle in the mc apps to test it out.

src/components/table/base-table/base-table.js
tdeekens: Ah didn't notice that. Makes total sense. Sorry. Yay for non-breaking changes.

emmenko:
Alright, tested all views in the MC and couldn't spot anything "broken". I think this is good to go now 🚀

src/components/table/base-table/base-table.js
```diff
@@ -329,51 +339,81 @@ export default class BaseTable extends React.Component {
...
+                ? columnDefinition.cellClassNameGetter({
+                    rowIndex: this.getBodyRowIndex(renderParams.rowIndex),
+                    columnKey: columnDefinition.key,
+                    height: renderParams.height,
+                    width: renderParams.width,
+                  })
                 : undefined
             )}
+            style={{
```
tdeekens: This is passed to a `style` prop. Would it make sense to use object styles and pass it to a `css(columnDef.cellStyleGetter`?

emmenko: I thought about it but honestly I would keep it as is (it was like this before), I don't want to mess up with possible specificity problems 😅

PR #482, opened 2019-01-26
[EXPERIMENT] Use REM instead of pixels

montezume:
Thought it would be cool to try this out.


.storybook/preview-head.html
```diff
@@ -4,6 +4,7 @@
   body {
     margin: 0;
     padding: 0;
+    font-size: 13px;
```
montezume: rem inherits from the root elements from size, so 1rem === 13px.

src/components/typography/text/text.styles.js
```diff
@@ -48,15 +48,15 @@ const getTone = tone => {
 const getElementFontSize = elementType => {
   switch (elementType) {
     case 'h1':
-      return '32px';
+      return '2.462rem';
```
montezume: used https://offroadcode.com/rem-calculator/

tdeekens: Can you even calc the thing maybe? Like calc(var(—root-font-size * x))em? I don‘t even know. Was just wondering.

montezume: I wouldn't put --root-font-size here though, because the whole advantage of using REM is that a user could use a different base size... 

Instead maybe we could decide that h1 should be X percentage bigger than 1, and use that for the calculation?

montezume: For example, if I go to the netlify branch deploy and I change the root font size to 16, then the computed value for our H1 becomes 39.3846px, while at the current root font size of 13 it's ~32px.

montezume:
Percy passed 😄

src/components/typography/text/text.styles.js
montezume: Yeah I guess we could have something like 

`--h1-scale` = 32 / 13 = `2.46153846154`

tdeekens: That's what I meant, cause the numbers are otherwise kind of hard to reconstruct or understand and hence to adjust. Given we just have the scales it also quickly shows if they actually make sense or if they're kind of random (not that we should change that on this PR).

.storybook/preview-head.html
emmenko: Does it mean we need to define the root font size in each app?

montezume: Yes, you need to set the font-size on the `html` root element.

montezume: So if someone wanted to use `ui-kit` for a static site or a documentation website, they could set the root font-size to something more appropriate, like 16px.

tdeekens: That's the cool part though :)

emmenko: Yes of course, just wanted to make sure 😄

I guess we need to document it and make it clear. What is the default font size of the browser btw?

montezume: [This](https://webdesign.tutsplus.com/tutorials/comprehensive-guide-when-to-use-em-vs-rem--cms-23984) is a decent article that explains it well.

Generally the default is 16px

montezume: FYI we are already setting the font size to the 13px for each app.

[ref](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell/src/components/application-shell/reset.mod.css)

tdeekens: I still think this would be a breaking change, right? I mean, to be frank, nobody uses UIKit apart from us but I guess we want to do semver.

montezume: Yep, I would say breaking. 

Interesting [articlee](https://medium.com/eightshapes-llc/visual-breaking-change-in-design-systems-1e9109fac9c4) on this

emmenko: 7a93fec5-e115-4cda-b6c3-6bae147cb5da

tdeekens: I agree and see the point. However, I think we just enable supporting REMs here which is cool. Whatever we then do on documentation sites and in apps is another thing? I would also argue, even though I agree with the point the article makes, that there may be a difference between web app and website. I know the lines are incredibly blurry these days but I think an app (e.g the MC) has to decide how much it can support variable (browser set) font sizes. In a perfect world anything should just let a browser and user decide. Totally agreed. Thanks for sharing. Really interesting.

emmenko: Yeah I agree, I also think for apps is a bit different.

I'm totally fine btw for going with `rem` for ui-kit, in fact I even proposed that some time ago 😜 https://github.com/commercetools/ui-kit/pull/387#discussion_r245276573

montezume: @emmenko yeah I remember, that's what gave me the idea to make this PR 😆

montezume:
I don't see why we can't merge this. It doesn't require any migration on the side of the MC. Any objections @qmateub @emmenko ?

src/components/typography/text/text.styles.js
```diff
@@ -48,15 +48,15 @@ const getTone = tone => {
...
+      return '1.8461538461rem';
     case 'h3':
-      return '20px';
+      return '1.5384615384rem';
     case 'h4':
-      return '16px';
+      return '1.2307692307rem';
     case 'h5':
-      return '14px';
+      return '1.0769230769rem';
```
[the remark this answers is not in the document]
montezume: Yeah, I mostly put them like this so Percy would pass. We could just as easily set them now to max 3 digits and then just approve the failing tests?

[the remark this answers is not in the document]
tdeekens: That why I thought of adding them as calc. It kind of gives a better insight into the ratios. Can obv we a follow up or we round.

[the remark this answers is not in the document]
emmenko: As long as we let the design team know about it and that they're fine with it.

We can also do that as a follow up, I don't mind.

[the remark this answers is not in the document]
montezume: I think it's sort of weird to do the calc because that's sort of what REM is, calculated font sizes. Feels weird to send a calc to the calc so it can calc? 🤔 

yo dawg

montezume:
Ah I already rounded them. It's really not that big of a diff though, I think when I looked at it, instead of getting 32px, we get 31.99px. I can discuss it with the designers too.

src/components/typography/text/text.styles.js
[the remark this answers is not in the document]
tdeekens: Ok, I smiled you funny, lovely... coffee nerd.

PR #490, opened 2019-01-28
Update deps

montezume:
Update deps

As always, since we use the bundled code in our visual specs, if our tests pass and visual tests pass, then we should be good to 🚀 

* Closes #489 
* Closes #488 
* Closes #487 
* Closes #486 
* Closes #485 
* Closes #484 

** Removed jest update to 24 as it broke visual tests, had trouble having our setup play nicely with "jest-puppeteer".


package.json
```diff
@@ -86,56 +86,55 @@
...
+    "@storybook/addon-actions": "4.1.11",
+    "@storybook/addon-info": "4.1.11",
+    "@storybook/addon-knobs": "4.1.11",
+    "@storybook/addon-links": "4.1.11",
+    "@storybook/addon-options": "4.1.11",
+    "@storybook/addon-storysource": "4.1.11",
+    "@storybook/react": "4.1.11",
     "@svgr/rollup": "4.1.0",
     "@svgr/webpack": "4.1.0",
-    "babel-core": "7.0.0-bridge.0",
```
montezume: don't need this anymore as of jest 24.

montezume:
I'm looking into the failing visualspec config.

src/components/table/table.story.js
```diff
@@ -528,6 +530,7 @@ class Wrapper extends React.PureComponent {
               'shouldFillRemainingVerticalSpace',
               false
             ),
+            // eslint-disable-next-line no-alert
```
tdeekens: Nitpick. Just use action(‚click‘) over alert?

PR #494, opened 2019-01-29
Adds `Card` Component

montezume:
### Summary

Adds `Card` component from core.

### DIFF
--- a/src/components/card/README.md
+++ b/src/components/card/README.md
@@ -8,7 +8,7 @@ import { Card } from '@commercetools-frontend/ui-kit';
 
 #### Description
 
-Cards are areas that display content and actions on a single topic.
+Cards are used to display content and actions on a single topic.
 
 #### Usage
 
@@ -24,5 +24,5 @@ Cards are areas that display content and actions on a single topic.
 | ----------- | -------- | :------: | ---------------- | -------- | --------------------------------- |
 | `theme`     | `string` |    -     | `light`, `dark`  | `light`  | background color of the card      |
 | `type`      | `string` |    -     | `raised`, `flat` | `raised` | The type of card                  |
-| `classname` | `string` |    -     | -                | `raised` | className passed to the container |
+| `classname` | `string` |    -     | -                | -        | className passed to the container |
 | `children`  | `node`   |    ✅    | -                | -        | Content rendered within the card  |
--- a/src/components/card/README.md
+++ b/src/components/card/README.md
@@ -1,0 +1,27 @@
+# Card
+
+## Usage
+
+```js
+import { Card } from '@commercetools-frontend/ui-kit';
+```
+
+#### Description
+
+Cards are areas that display content and actions on a single topic.
+
+#### Usage
+
+```js
+<Card theme="light" type="raised">
+  <p>You'll never guess where I've been!</p>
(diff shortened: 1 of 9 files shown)


src/components/card/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### Description
+
+Cards are areas that display content and actions on a single topic.
+
+#### Usage
+
+```js
+<Card theme="light" type="raised">
+  <p>You'll never guess where I've been!</p>
```
montezume: 🦞

src/components/card/README.md
```diff
@@ -0,0 +1,28 @@
...
+</Card>
+```
+
+#### Properties
+
+| Props       | Type     | Required | Values           | Default  | Description                       |
+| ----------- | -------- | :------: | ---------------- | -------- | --------------------------------- |
+| `theme`     | `string` |    -     | `light`, `dark`  | `light`  | background color of the card      |
+| `type`      | `string` |    -     | `raised`, `flat` | `raised` | The type of card                  |
+| `classname` | `string` |    -     | -                | `raised` | className passed to the container |
```
emmenko: The default is wrong

montezume: Ah yeah should be empty

montezume:
Hah, Percy is failing because I forgot to add the export 😆 🤦‍♂️

emmenko:
Should we add some padding in the section container, so that percy can snapshot the border shadow of the cards?

src/components/card/README.md
```diff
@@ -0,0 +1,28 @@
...
+</Card>
+```
+
+#### Properties
+
+| Props       | Type     | Required | Values           | Default  | Description                       |
+| ----------- | -------- | :------: | ---------------- | -------- | --------------------------------- |
+| `theme`     | `string` |    -     | `light`, `dark`  | `light`  | background color of the card      |
+| `type`      | `string` |    -     | `raised`, `flat` | `raised` | The type of card                  |
+| `classname` | `string` |    -     | -                | -        | className passed to the container |
```
qmateub: Are we sure that we want that custom styles can be passed to this? Last time I spoke with Dom regarding this (classname in Collapsible Panel) we agreed that we should aim at some point to not have this in any UiKit component. Is really needed? 🤔 or what do you think?

montezume: Two points

1) I don't think it's feasible for the `ui-kit` team to refactor every single usage of the Card component. 
2) Accepting a classname is actually really useful when you are using CSS in JS because otherwise it can be impossible to override styles without !important.

qmateub: > I don't think it's feasible for the ui-kit team to refactor every single usage of the Card component.

Fair enough 👌 

> Accepting a classname is actually really useful when you are using CSS in JS because otherwise it can be impossible to override styles without !important.

Yeah, I didn't say that it is not useful, I was just asking because of the last information about this I had 😅

montezume: Yeah, in a perfect world we would have props that would deal with every scenario and we wouldn't need to accept a classname. But this requires doing some research into how `Card` is being used.

Maybe it's possible in the next iteration to drop support for `classname` in Card. 🤷‍♂️

PR #498, opened 2019-01-31
chore: update changelog and bump version

montezume:


materials/custom-properties.css
```diff
@@ -154,7 +154,7 @@
   --shadow-15-second: 0 1px 4.75px rgba(0, 0, 0, 0.12);
   --font-family-default: 'Open Sans', sans-serif;
   --font-size-default: 1rem;
-  --font-size-small: 12px;
+  --font-size-small: 0.9231rem;
```
montezume: accidental regression here https://github.com/commercetools/ui-kit/blame/00fa71c65f6089f229f216c88eb6671f4f9cb98c/materials/custom-properties.js#L156

montezume: for some reason when the card component was added, this goes back to how it was after the REM was merged.

PR #501, opened 2019-02-01
Adds `Tooltip` component

montezume:
# Tooltip

#### Description

Tooltips display informative text when users hover over or focus on an element.

#### Usage

```js
<Tooltip
  position="left"
  label="If you buy a pizza, you will also get a free ice cream :)"
>
  <button onClick={orderPizza({ freeIceCream: 'yes please' })}>Submit</button>
</Tooltip>
```

#### Working with disabled child elements

When you use a tooltip with a disabled element, you should the style `pointer-events: none` to the disabled element to stop it from capturing events.

```js
<Tooltip
  position="left"
  label="You do not have permission to delete the database"
>
  <button
    disabled
    onClick={deleteDb()}
    style={{
      pointerEvents: 'none',
    }}
  >
    Delete production database
  </button>
</Tooltip>
```

#### Customizing the wrapper

The tooltip applies event listeners (`onMouseOver`, `onMouseLeave`, `onFocus`, and `onBlur`) to a wrapping component around the passed in component. By default, this wrapper is displayed with style `inline-block`. If you want to customize this behaviour, then you can pass in a custom wrapping element.

```js
const Wrapper = styled.div`
  display: block;
`;

const FullWidthButton = styled.button`
  display: block;
  width: 100%;
`;

<Tooltip title="Delete" components={{ WrapperComponent: Wrapper }}>
  <FullWidthButton>Submit</FullWidthButton>
</Tooltip>;
```

#### Properties

| Props                  | Type     | Required | Values                                                                                                                                       | Default | Description                                                                                     |
| ---------------------- | -------- | :------: | -------------------------------------------------------------------------------------------------------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------- |
| `isOpen`               | `bool`   |    -     | -                                                                                                                                            | -       | If passed, the tooltip's open and closed states are controlled by this prop                     |
| `leaveDelay`           | `number` |    -     | -                                                                                                                                            | 0       | Delay (in milliseconds) between the end of the user interaction, and the closing of the tooltip |
| `onOpen`               | `func`   |    -     | -                                                                                                                                            | -       | Called when the tooltip is opened                                                               |
| `onClose`              | `func`   |    -     | -                                                                                                                                            | -       | Called after the tooltip is closed                                                              |
| `position`             | `object` |    -     | `top`, `top-start`, `top-end`, `right`, `right-start`, `right-end`, `bottom`, `bottom-end`, `bottom-start`, `left`, `left-start`, `left-end` | `top`   | How the tooltip is positioned relative to the child element                                     |
| `horizontalConstraint` | `object` |    -     | `xs`, `s`, `m`, `l`, `xl`, `scale`                                                                                                           | `scale` | Horizontal size limit of the tooltip                                                            |
| `children`             | `node`   |    ✅    | -                                                                                                                                            | -       | Content rendered within the tooltip                                                             |
| `components`           | `object` |    -     | `WrapperComponent`                                                                                                                           | -       | If passed, the tooltip will wrap your component with this element                               |


src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -28,6 +28,7 @@ const AccessibleButton = React.forwardRef((props, ref) => (
       props.isDisabled &&
         css`
           cursor: not-allowed;
+          pointer-events: none;
```
montezume: otherwise it swallows up events

montezume:
Percy is not playing at all nicely with the positioning library I'm using. It might not be possible to make useful VRTs for these tooltips.

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,191 @@
...
+    );
+  }
+}
+
+Tooltip.propTypes = {
+  children: PropTypes.node.isRequired,
+  horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale'])
+    .isRequired,
+  leaveDelay: PropTypes.number.isRequired,
+  open: PropTypes.bool,
```
tdeekens: Nitpick: how is UIKit's naming convention. Is it `isOpen` in this case maybe?

src/components/tooltip/tooltip.spec.js
```diff
@@ -0,0 +1,237 @@
...
+      ];
+
+      constructor() {
+        return {
+          destroy: () => {},
+          scheduleUpdate: () => {},
+        };
+      }
+    }
+);
```
tdeekens: Maybe put into a `ui-kit/mocks/applyPopperMocks.js` cause then it could be reused.

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,191 @@
...
+      this.setState({
+        open: false,
+      });
+    }
+    if (this.props.onClose) {
+      this.props.onClose(event);
+    }
+  };
+
+  onRootRef = ref => {
```
tdeekens: Is this more a `setRootRef`? I am just trying to follow along.

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,191 @@
...
+    const childrenProps = {
+      'aria-describedby': open ? this.state.id : null,
+      // for seo and accessibility, we add the tooltip's title
+      // as a native title when the title is hidden
+      title:
+        !open && typeof this.props.title === 'string' ? this.props.title : null,
+      ...this.props.children.props,
+    };
+
+    childrenProps.onMouseOver = this.handleEnter;
```
tdeekens: Why do we do this here and not in L117? Just `onMouseLeave: this.handleLeave` etc?

src/components/buttons/primary-button/primary-button.js
```diff
@@ -5,23 +5,27 @@ import { css } from '@emotion/core';
...
+import filterEventAttributes from '../../../utils/filter-event-attributes';
 import Spacings from '../../spacings';
 import AccessibleButton from '../accessible-button';
 import {
   getButtonLayoutStyles,
   getButtonStyles,
 } from './primary-button.styles';
 
-const PrimaryButton = props => {
+const PrimaryButton = React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
montezume: Warning: Function components cannot be given refs. Attempts to access this ref will fail.

Yeah, good call. Should add this to the docs.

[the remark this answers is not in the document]
montezume: Reworked it to not need this. All it requires is that the child spreads the event listener props

src/components/buttons/flat-button/flat-button.js
```diff
@@ -37,9 +38,10 @@ const getTextColor = (tone, isHover = false) => {
 
 export const FlatButton = props => {
   const dataProps = {
-    'data-track-component': 'LinkButton',
+    'data-track-component': 'FlatButton',
     ...filterAriaAttributes(props),
     ...filterDataAttributes(props),
+    ...filterEventAttributes(props),
```
tdeekens: Is this function exported from UI-Kit then for app level components to spread event props? I still wonder if we should rather export a

```js
import { Tooltip } from 'ui-kit';

const Component = Tooltip.forwardProps(props => <div {...props} /> 
```

which would internally can use the `filterEeventAttributes` but I like the api more as the origin of the props is clear and the Tooltip gets the change to potentially spead any other props or even refs (2nd arg).

montezume: I don't really see the point of having this functionality when most people will just follow the readme and do something like 

js```
const Component = props => (
<div {...props}>Foo</div>
)```

This forwarding props only making sense because we are so against prop forwarding, but that's a technical decision we don't need to force on others...

tdeekens: Ok.

src/components/tooltip/tooltip.spec.js
montezume: Yeah good call 👍

src/components/buttons/accessible-button/accessible-button.js
emmenko: We should leave a comment to explain the reasons

src/components/internals/root-ref/root-ref.js
```diff
@@ -0,0 +1,48 @@
...
+    this.ref = null;
+    setRef(this.props.rootRef, null);
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+RootRef.propTypes = {
```
emmenko: Nitpick: make it a `static` prop as well within the class?

src/components/internals/root-ref/root-ref.js
```diff
@@ -0,0 +1,48 @@
+// inspired from https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/RootRef/RootRef.js
```
emmenko: Can we also copy over the comments, just in case they get deleted from the source? Then we have a bit more context about the scope of this component. Thanks

src/components/tooltip/README.md
```diff
@@ -0,0 +1,67 @@
...
+    >
+      Delete production database
+    </button>
+  </div>
+</Tooltip>
+```
+
+#### Working with custom child elements
+
+The tooltip needs to apply DOM event listeners to its child element. If the child is a custom React element, you need to make sure that it spreads its properties to the underlying DOM element.
```
emmenko: It would be helpful to list the DOM event listeners that we are referring to (`onMouseOver`, etc)

src/components/tooltip/README.md
```diff
@@ -0,0 +1,67 @@
...
+<Tooltip
+  position="left"
+  label="You do not have permission to delete the database"
+>
+  <div style={{ cursor: 'not-allowed' }}>
+    <button
+      disabled
+      onClick={deleteDb()}
+      style={{
+        pointerEvents: 'none',
```
emmenko: Should we make it clear that this needs to be set?

src/components/tooltip/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+When you use a tooltip with a disabled element, you need to wrap element in a div, as
+disabled elements do not fire events.
+
+```js
+<Tooltip
+  position="left"
+  label="You do not have permission to delete the database"
+>
+  <div style={{ cursor: 'not-allowed' }}>
```
emmenko: How about we let the tooltip do the wrapping and the consumer can pass custom styles for this wrapper div, in case it's necessary. For example:

```js
<Tooltip
  disabledWrapperStyles={{...}}
  disabledWrapperClassName=""
>
  <button />
</Tooltip>
```

tdeekens: Not sure. I'd be as restrictive with UIKit as one can be. Passing `classNames` or `styles` we should maybe avoid for as long as we ever can? Maybe if needed than rather a `renderWrapper` or something to give full control.

montezume: Either way it's a trade off. I tried with creating a wrapper div (like we do in `core` Tooltip), and I found it's a lot easier for styling purposes if we don't create a wrapper div.

src/components/internals/root-ref/root-ref.js
montezume: I don't understand 🤔 Can you explain?

src/components/buttons/accessible-button/accessible-button.js
montezume: Maybe instead more examples of uses with different elements would be useful>?

src/components/tooltip/README.md
emmenko: The styles/classname prop names are on purpose very specific to the "wrapped element", so it's not that we spread them everywhere.

I just find it a bit weird to have to wrap a div, only when in disabled state. It's going to be a "mess" from a consumer perspective

```js
<Tooltip
>
  {isDisabled ? (
    <div style={{ cursor: 'not-allowed' }}>
      <MyComponent isDisabled={isDisabled} />
    </div>
  ) : (
    <MyComponent />
  )}
</Tooltip>
```

src/components/internals/root-ref/root-ref.js
emmenko:
```js
class RootRef extends React.Component {
  static propTypes = {}
}
```

src/components/buttons/accessible-button/accessible-button.js
emmenko: Sure, but also I mean for future references when someone looks at this code and it might not be clear what the reason for disabling pointer events is.

src/components/tooltip/README.md
montezume: @emmenko but then we are just complaining about how the dom / html works. It's not unique to tooltips that disabled elements don't fire events 🤔

montezume: 🤷‍♂️ . If we wrapped in a div, we could have like a "display: block" vs "display: inline-block" option.

emmenko: Whatever works, as long as the API is simple to use.

qmateub: Just leaving my humble opinion here after @montezume explained me both approaches.

I think I'm more with the mindset of Tobi, of not allowing the consumer "insert" custom styles in the component (eventhough this custom style for the wrapping div are not intended for customizing but for consistency I would say).

The reason is because allowing the user to add custom styles is not how the rest of components of UIKit works (maybe just collapsible panel which is quite old). So, for the sake of consistency (API wise) I would vote for just rely on the user to add proper children elements to the `Tooltip`.

If it's well documented, explaining the use case and what you should do to make it consistent in a specific scenario, for me it's more than enough as a consumer 🤷‍♂️ .

montezume: I would like to keep the API like it is for now, and if we run into problems with the migration, I'm 100% open to reworking it.

src/components/internals/root-ref/root-ref.js
montezume: oh yeah 🤦‍♂️ . Thanks

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+Tooltip.propTypes = {
+  children: PropTypes.node.isRequired,
+  horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale'])
+    .isRequired,
+  leaveDelay: PropTypes.number.isRequired,
+  isOpen: PropTypes.bool,
+  onClose: PropTypes.func,
```
tdeekens: would a `onClose: requiredIf(PropTypes.func, props => !props.isOpen)` make sense?

montezume: Just to understand, by `!props.isOpen`, do you mean that the component is controlled by it's internal state, or it's controlled by `isOpen` as a prop?

tdeekens: Maybe it doesn't make sense true. I've just seen this elsewhere. If you pass e.g. `isOpen` you mean to control the component. You however can't if don't use the event handlers (which is not true in this case).

montezume:
I think it makes the most sense to do a call tomorrow afternoon about this because there is some controversy surrounding the API

src/components/tooltip/tooltip.js
```diff
@@ -22,6 +24,49 @@ const requiredProps = [
...
+      'left',
+      'left-start',
+      'left-end',
+    ]),
+    title: PropTypes.string.isRequired,
+    components: PropTypes.shape({
+      WrapperComponent: (props, propName) => {
+        if (props[propName] && !isValidElementType(props[propName])) {
+          return new Error(
+            `Invalid prop 'component' supplied to 'Route': the prop is not a valid React component`
```
emmenko: `Route` 🤔

src/components/tooltip/wrapper.js
```diff
@@ -0,0 +1,22 @@
...
+  </div>
+);
+
+Wrapper.displayName = 'Wrapper';
+
+Wrapper.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default Wrapper;
```
emmenko: I think this all thing can be

```js
export default styled.div`display: inline-block;`
```

src/components/tooltip/tooltip.js
montezume: oh 😆

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,212 @@
...
+      }, {});
+
+  componentDidMount() {
+    // eslint-disable-next-line no-console
+    const childrenProps = this.props.children.props;
+
+    invariant(
+      !(childrenProps.disabled || childrenProps.isDisabled),
+      [
+        'ui-kit: you are providing a disabled `button` child to the Tooltip component.',
```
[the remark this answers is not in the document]
montezume: true

montezume:
Okay I updated the Tooltip to now wrap the passed in component with a div. This way, if your element is disabled, the consumer doesn't need to wrap it. By default our wrapper uses display: block, but a consumer can pass in their own custom Wrapping component like this.

```js
import { Tooltip } from 'ui-kit;

const Button = styled.button`
   width: 100%;
   display: block;
`;

const Wrapper = styled.div`
   display: block;
`;

<Tooltip title="Hi there" components={{WrapperComponent: Wrapper>
   <Button>Submit</Button>
</Tooltip>

```

montezume:
Could use some more reviews now that @emmenko's concerns have been addressed by adding the Wrapping component / cc @tdeekens

src/components/internals/root-ref/root-ref.spec.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { render } from '../../../test-utils';
+import RootRef from './root-ref';
+
+class TestComponent extends React.Component {
+  state = {
+    show: this.props.show,
```
tdeekens: Nitpick: even though internal we should follow our own conventions maybe and name is `isOpen` or `shouldShow`. I don't mind too much though.

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,198 @@
...
+      'right-end',
+      'bottom',
+      'bottom-start',
+      'bottom-end',
+      'left',
+      'left-start',
+      'left-end',
+    ]),
+    title: PropTypes.string.isRequired,
+    components: PropTypes.shape({
```
tdeekens: Nice and extensible!

src/components/tooltip/tooltip.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    fireEvent.blur(button);
+    // should call callbacks
+    expect(onBlur).toHaveBeenCalled();
+    expect(onClose).toHaveBeenCalled();
+    // should hide tooltip
+    expect(queryByText('What kind of bear is best?')).not.toBeInTheDocument();
+    // should add the title again
+    expect(button).toHaveProperty('title', 'What kind of bear is best?');
+  });
+});
```
tdeekens: Really nice tests 💯

src/components/tooltip/README.md
```diff
@@ -0,0 +1,69 @@
...
+```js
+<Tooltip
+  position="left"
+  label="If you buy a pizza, you will also get a free ice cream :)"
+>
+  <button onClick={orderPizza({ freeIceCream: 'yes please' })}>Submit</button>
+</Tooltip>
+```
+
+#### Working with disabled child elements
```
emmenko: We should mention that some of our components already support that (e.g. buttons, etc)

src/components/tooltip/README.md
```diff
@@ -0,0 +1,69 @@
...
+    }}
+  >
+    Delete production database
+  </button>
+</Tooltip>
+```
+
+#### Customizing the wrapper
+
+The tooltip applies event listeners (`onMouseOver`, `onMouseLeave`, `onFocus`, and `onBlur`) to a wrapping component around the passed in component. By default, this wrapper is displayed with style `inline-block`. If you want to customize this behaviour, then you can pass in a custom wrapping element.
```
emmenko:
```suggestion
The tooltip applies event listeners (`onMouseOver`, `onMouseLeave`, `onFocus`, and `onBlur`) to a wrapping `div` component around the children element. By default, this wrapper is displayed with style `inline-block`. If you want to customize this behaviour, then you can pass in a custom element.
```

src/components/tooltip/README.md
```diff
@@ -0,0 +1,69 @@
...
+| ---------------------- | -------- | :------: | -------------------------------------------------------------------------------------------------------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------- |
+| `isOpen`               | `bool`   |    -     | -                                                                                                                                            | -       | If passed, the tooltip's open and closed states are controlled by this prop                     |
+| `leaveDelay`           | `number` |    -     | -                                                                                                                                            | 0       | Delay (in milliseconds) between the end of the user interaction, and the closing of the tooltip |
```
emmenko: Nitpick: should we maybe name this `closeAfter` or something similar? I feel that it's less confusing. I don't mind much though...

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,198 @@
...
+      WrapperComponent: (props, propName) => {
+        if (props[propName] && !isValidElementType(props[propName])) {
+          return new Error(
+            `Invalid prop 'components.WrapperComponent' supplied to 'WrapperComponent': the prop is not a valid React component`
```
emmenko:
```suggestion
            `Invalid prop 'components.WrapperComponent' supplied to 'Tooltip': the prop is not a valid React component`
```

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,198 @@
...
+    if (this.props.onClose) {
+      this.props.onClose(event);
+    }
+  };
+
+  setChildrenRef = ref => {
+    this.childrenRef = ref;
+  };
+
+  isControlled = this.props.isOpen != null;
```
emmenko:
```suggestion
  isControlled = Boolean(this.props.isOpen);
```

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,198 @@
...
+              <RootRef rootRef={ref}>
+                {React.cloneElement(this.props.children, { ...childrenProps })}
+              </RootRef>
+            </WrapperComponent>
+          )}
+        </Reference>
+        {open && (
+          <Popper placement={this.props.placement}>
+            {({ ref, style, placement }) => (
+              <div
```
emmenko: We wanted to extract this into a `<TooltipContent>` or something, in order to run VRT on it, right?

src/components/tooltip/tooltip.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    leaveDelay: PropTypes.number,
+    onMouseLeave: PropTypes.func,
+    onMouseOver: PropTypes.func,
+    components: PropTypes.shape({
+      WrapperComponent: PropTypes.any,
+    }),
+  };
+
+  static defaultProps = {
+    title: 'What kind of bear is best?',
```
emmenko: Is this some kind of Canadian thing to say? 😅

src/components/tooltip/tooltip.visualroute.js
```diff
@@ -0,0 +1,29 @@
...
+
+const title = 'What kind of bear is best';
+const noop = () => {};
+
+export const routePath = '/tooltip';
+
+export const component = () => (
+  <Suite>
+    <Spec label="Closed" omitPropsList>
+      <Tooltip title={title} eventsEnabled={false}>
```
emmenko: Where is this prop `eventsEnabled` defined? 🤔

src/components/tooltip/wrapper.js
```diff
@@ -0,0 +1,5 @@
+import styled from '@emotion/styled';
+
+export default styled.div`
```
emmenko: Nitpick: we can define this component in the `tooltip.js` file directly I suppose, right?

src/components/tooltip/tooltip.spec.js
montezume: @emmenko [here](https://www.youtube.com/watch?v=WaaANll8h18)

src/components/tooltip/tooltip.visualroute.js
montezume: ah yeah that's a leftover from trying to get poppers to work with percy. I'll remove it.

src/components/tooltip/tooltip.js
montezume: I thought about that but it seems a bit weird since then we would need to export that, since the VRT uses the bundled ui-kit...

src/components/tooltip/tooltip.js
montezume: @emmenko but that isn't what we want?

If the user passed in `isOpen`, then they have to control it's open and closed state. 

So if I pass in false, I want to control it myself.

```

this.props.isOpen != null true
tooltip.js:141 Boolean(this.props.isOpen)) false
```

emmenko: Oh nvm, now I see the `!=` loose comparator. 🤦‍♂️
All good, I was wondering why checking only for `null` and e.g. not for `undefined`.

src/components/tooltip/tooltip.js
emmenko: Hmm right...that's unfortunate 😑

src/components/tooltip/README.md
```diff
@@ -0,0 +1,71 @@
...
+  label="You do not have permission to delete the database"
+>
+  <button
+    disabled
+    onClick={deleteDb()}
+    style={{
+      pointerEvents: 'none',
+    }}
+  >
+    Delete production database
```
emmenko: 😅

src/components/tooltip/tooltip.js
```diff
@@ -0,0 +1,202 @@
+// inspired from https://github.com/mui-org/material-ui/blob/9ecc8db8abbfb829111d3b5c0678267827984024/packages/material-ui/src/RootRef/RootRef.js#L7-L36
```
emmenko: Wrong file?

src/components/tooltip/README.md
emmenko: What about this one? 👆

src/components/tooltip/tooltip.visualroute.js
emmenko: Do you want to remove it?

PR #509, opened 2019-02-04
feat(localized-money-input): LocalizedMoneyInput is collapsed by default

gnerkus:
#### Summary
- Define `isTouched` prop to determine if the input was touched
- If there are errors or warnings, expand input only if it has been
touched

localized-money-input-default-collapsed


montezume:
Are we sure this is the behaviour we want? If so, can you add more percy tests to show that the component will be open when it has `hasError` and `hasTouched` and update the wording to reflect these two cases?

gnerkus:
@montezume This is the behaviour we want. To summarize the steps:

**1**. In the discount view, the type is switch from _relative_ to _absolute_ via a radio button. This triggers `Formik`'s validation.
**2**. The `money` field in `Formik` is populated with empty values i.e
```
{
  EUR: {currencyCode: 'EUR', amount: ''},
  USD: {currencyCode: 'USD', amount: ''}
}
```
**3**. Prior to this change, the `LocalizedMoneyInput` would be expanded, displaying inputs for each of the currencies in the empty values in **_2_**. above. This is because each of the empty values produces an error (once the validation in **_1_**. is completed).

emmenko:
Are you sure? It looks like it does to me 🤔

https://github.com/commercetools/ui-kit/blob/04632e0c60135220080aebf8134d5c306ba4e358/src/components/inputs/localized-money-input/localized-money-input.js#L216-L226

```js
const hasErrorOnRemainingCurrencies =
  props.hasError
```

gnerkus:
@montezume 
A workaround for the issue was found with @emmenko 's help.

```
errors={this.props.touched.value?.money && this.props.errors.value?.money}
```

We only send the errors when the `LocalizedMoneyInput` has been touched (any of its constituent `MoneyField`s).

This PR will be closed as a result.

PR #511, opened 2019-02-06
fix(tag): wrong css is applied in case linkTo prop is provided

pa3:
#### Summary

Fix <Tag /> component appearance.

### DIFF
--- a/src/components/tag/tag.story.js
+++ b/src/components/tag/tag.story.js
@@ -12,7 +12,7 @@ const Story = () => (
   <Section>
     <Tag
       type={select('type', ['normal', 'warning'], 'normal')}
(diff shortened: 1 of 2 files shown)


src/components/tag/tag.story.js
```diff
@@ -12,7 +12,7 @@ const Story = () => (
   <Section>
     <Tag
       type={select('type', ['normal', 'warning'], 'normal')}
-      linkTo={select('linkTo', ['/foo', undefined], undefined)}
+      linkTo={select('linkTo', ['/foo', undefined], '/foo')}
```
montezume: ah thanks for fixing this as well!

PR #517, opened 2019-02-07
UIKIT-55: select inputs line between items

vidurar:
#### Summary

add a `shouldDivideOptionGroups` prop to `SelectInput`,`CreatableSelectInput`, `AsyncSelectInput` and `AsyncCreatableSelectInput`, to display a divider between option groups


materials/custom-properties.js
```diff
@@ -105,6 +105,7 @@ export default {
   borderRadiusInput: '6px',
   borderRadiusTag: '2px',
   borderRadiusForInput: '6px',
+  borderTopForOptionGroupSeparator: '1px solid #ccc',
```
montezume: this file is custom generated, so you can't add this here as it will get overridden. Maybe just use the custom property ${colorGray} and don't bother with adding a token

vidurar: 👍

src/components/inputs/async-select-input/async-select-input.js
```diff
@@ -102,6 +102,7 @@ export class AsyncSelectInput extends React.Component {
     ]),
     loadOptions: PropTypes.func.isRequired,
     cacheOptions: PropTypes.any,
+    shouldDivideOptionGroups: PropTypes.bool,
```
montezume: I'm not sure about this prop maybe. Maybe `showOptionGroupDivider` would be more appropriate 🤔

src/components/internals/create-select-styles.js
```diff
@@ -157,22 +157,27 @@ const multiValueRemoveStyles = () => base => ({
...
-    borderTop: `1px ${vars.borderColorSeparator} solid`,
+  '&:not(:first-child)': {
+    borderTop: props.shouldDivideOptionGroups
+      ? `1px solid ${vars.colorGray}`
+      : base.borderTop,
   },
 });
 
-const groupHeadingStyles = () => base => ({
+export const groupHeadingStyles = () => base => ({
```
montezume: Why this change?

vidurar: this was something I added while figuring out the implementation and forgot to remove. Will get rid of this

src/components/inputs/async-select-input/async-select-input.js
qmateub: To me `divide` is already an action, so maybe directly `divideOptionGroups` is enough 😅

vidurar: i think `showOptionGroupDivider` is a little clearer

montezume:
Just need to modify those VRT routes so that we have 3 snapshot-ed states for the open select input and we should be good to go 👍 

- Open without option groups
- Open with option groups
- Open with option groups and `showOptionGroupDivider` enabled

montezume:
Okay I tagged @mariabarrena to review the Percy VRTs.

src/components/internals/create-select-styles.js
```diff
@@ -157,11 +157,13 @@ const multiValueRemoveStyles = () => base => ({
...
-const groupStyles = () => base => ({
+const groupStyles = props => base => ({
   ...base,
   padding: 0,
-  '&:not(:firstChild)': {
-    borderTop: `1px ${vars.borderColorSeparator} solid`,
+  '&:not(:first-child)': {
+    borderTop: props.showOptionGroupDivider
+      ? `1px solid ${vars.colorGray}`
+      : base.borderTop,
```
emmenko: Is this correct? Where does it come from? 🤔

montezume: base.borderTop is `undefined`. I think this is what we want?

vidurar: This value is undefined , but used `base.borderTop` instead of hardcoding `undefined` to stick to the patterns found elsewhere in this file

tdeekens: This seems closed over from [here](https://github.com/commercetools/ui-kit/blob/47d5de1e53f5dca1a569bcd8f1158bbe10ace967/src/components/internals/create-select-styles.js#L195) with the `props` [here](https://github.com/commercetools/ui-kit/blob/7c53c7d3b301a3930917c8456eea2f87d7c446e0/src/components/inputs/select-input/select-input.js#L175). Then I fail to follow from react-select what `base is. Should that be our `vars`?

vidurar: From when i looked into it, `base` are the default styles for the component provided by `react-select` itself.

montezume: Yeah, I think it's fine to use base.borderTop. We are doing similar things in the rest of these functions.

montezume:
@emmenko do you have any more concerns about this or is it okay to merge from your side?

PR #518, opened 2019-02-08
[SPIKE] - Bundling multiple outputs

montezume:
#### Summary

I wanted to experiment with bundling multiple outputs to facilitate proper tree shaking. For an example, I used the `Stamp` component. 

If you run `yarn build`, the output contains some interesting things.

- All of the customProperties are bundled.
- All of the Spacings components are bundled, not just `Inset`

#### How / what / wtf / why

 If we were to go the approach of exporting multiple bundles, I think it would make sense to look into writing some sort of babel plugin that converts imports like 

```
import { Stamp, Spacings } from 'ui-kit';

```

to 
```
import Stamp from 'ui-kit/stamp`
import Spacings from 'ui-kit/spacings`
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -57,20 +57,33 @@ const defaultExternal = deps.concat(peerDeps);
 
 // We need to define 2 separate configs (`esm` and `cjs`) so that each can be
 // further customized.
-const config = {
-  input: 'src/index.js',
-  external: defaultExternal,
-  output: [
-    {
-      file: pkg.module,
-      format: 'esm',
-    },
-    {
-      file: pkg.main,
-      format: 'cjs',
-    },
-  ],
-  plugins,
-};
+const config = [
+  {
+    input: 'src/index.js',
+    external: defaultExternal,
+    output: [
+      {
+        file: pkg.module,
+        format: 'esm',
+      },
+      {
+        file: pkg.main,
+        format: 'cjs',
+      },
+    ],
+    plugins,
+  },
+  {
+    input: 'src/components/stamp/index.js',
+    external: defaultExternal,
+    output: [
+      {
+        file: 'dist/stamp.esm.js',
+        format: 'esm',
+      },
+    ],
+    plugins,
+  },
+];
 
 export default config;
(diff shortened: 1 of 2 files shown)


rollup.config.js
```diff
@@ -57,20 +57,33 @@ const defaultExternal = deps.concat(peerDeps);
...
+    input: 'src/components/stamp/index.js',
+    external: defaultExternal,
+    output: [
+      {
+        file: 'dist/stamp.esm.js',
+        format: 'esm',
+      },
+    ],
+    plugins,
+  },
```
tdeekens: How would we guide consumers to these entries?

montezume: We could offer both the current [`big bundle`](https://www.youtube.com/watch?v=VDLDVPV8kDE), as well as the specific bundles. I posted above about writing a possible babel transform or something 🤔

tdeekens: Yeah I get that. But when we publish small bundles to npm. How would consumers import them as the `module`, `main` etc all point to the bug bundle.

montezume: They would need to either use the babel plugin or manually import the small bundles directly. Other libraries follow this pattern, I'll find an example soon

tdeekens: Thanks.

emmenko: I tried to programmatically listing the exported components paths but I don't think it's possible. I guess we need to manually list the paths here?
We should probably find a better solution, for example we generate the main export file from a list, which we can use to define the exports bundle here as well.

emmenko:
> * All of the customProperties are bundled.

That's expected, as we export it as a default (#461, https://github.com/commercetools/ui-kit/pull/460#discussion_r249953661).

> * All of the Spacings components are bundled, not just `Inset`

I guess because we also export it as a default? 🤔 If I import one of the spacings components directly then it works as expected.

montezume:
Even if we can't write a babel plugin to reword all of the imports, it still might be useful for new projects. This way the Gatsby projects wouldn't end up with things like `react-select`, `react-virtualized`, etc.

emmenko:
We could avoid using `default` exports in proxy exports such as spacings, etc and use named exports. It's an internal change only anyway.

tdeekens:
I am all for default exports dying 😄. I only find it useful when exported HoC'ed un non-HoC'ed components in apps.

montezume:
I'll come back to this later.

PR #537, opened 2019-02-13
feat: add RadioField component, rework RadioInput

emmenko:
This PR aims to rework the `RadioInput` component API to be consistent with the other input components, as well as to add the `RadioField` component.

There are a couple of extra changes though, as we decided for example to remove some default styling on the `Text` components in order to make it easier to apply e.g. some colors on the text.
This became particularly obvious in the radio input options.


src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### Description
+
+A controlled text input component for passwords with validation
+states.
+
+## Usage
+
+```js
+import { PasswordField } from '@commercetools-frontend/ui-kit';
```
tdeekens: RadioField

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,64 @@
...
+| `isReadOnly`           | `bool`             |    -     | -                                                     | `false` | Indicates that the field is displaying read-only content                                                                                                                                                                                                              |
+| `placeholder`          | `string`           |    -     | -                                                     | -       | Placeholder text for the input                                                                                                                                                                                                                                        |
+| `autoComplete`         | `string`           |    -     | `on` \| `off` \| `current-password` \| `new-password` | -       | Password autocomplete mode                                                                                                                                                                                                                                            |  |
```
tdeekens: I think this doesn't apply for a radio-field.

src/components/styles-provider/styles-provider.js
```diff
@@ -0,0 +1,50 @@
...
+        h4,
+        h5,
+        h6,
+        p {
+          margin: 0;
+        }
+      `}
+    />
+    {props.children}
+  </React.Fragment>
```
tdeekens: To by using the RadioField you "import" the side effect of a global style?

src/components/fields/radio-field/README.md
emmenko: I forgot to update the readme 🤦‍♂️

src/components/styles-provider/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+```js
+import { StylesProvider } from '@commercetools-frontend/ui-kit';
+```
+
+## Properties
+
+| Props          | Type               | Required | Values | Default | Description                                                                                      |
+| -------------- | ------------------ | :------: | ------ | ------- | ------------------------------------------------------------------------------------------------ |
+| `baseFontSize` | `PropTypes.string` |    -     | -      | `1rem`  | Defines the font size baseline from which the `rem` font sizes of the application are based from |
```
montezume: It doesn't make sense to set baseFontSize to `1rem`. It should be a pixel value. Maybe name it something like `bodyFontSize` or `htmlFontSize` to imply that it's being set on the body...

montezume:
I labelled it a breaking change because removing margin, color and font-family from the text components is a breaking visual change.

emmenko:
Yes thanks, I didn't finish updating the PR description etc. It should be a breaking change of course.

src/components/styles-provider/README.md
emmenko: Sure, I wanted to discuss this with you tomorrow.

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
+# RadioField
+
+#### Description
+
+A controlled radio input component with validation states and a label.
```
tdeekens: This still feels off?

src/components/fields/radio-field/radio-field.spec.js
```diff
@@ -0,0 +1,193 @@
...
+    expect(getByLabelText(/RadioField Option 1/)).toHaveAttribute('disabled');
+    expect(getByLabelText(/RadioField Option 2/)).toHaveAttribute('disabled');
+  });
+});
+
+describe('when readOnly', () => {
+  it('should disable the input', () => {
+    const { getByLabelText } = renderRadioField({ isReadOnly: true });
+    expect(getByLabelText(/RadioField Option 1/)).toHaveAttribute('readonly');
+    expect(getByLabelText(/RadioField Option 2/)).toHaveAttribute('readonly');
```
tdeekens: Are you sure we need this fuzzy matching everywhere? I understand the intend to have tests potentially more stable/flexible but I don't feel case for it really exists?

src/components/inputs/radio-input/radio-group.js
```diff
@@ -36,25 +52,49 @@ class Group extends React.PureComponent {
...
+            onBlur: this.props.onBlur,
+          });
+        }
+        return child;
+      }
+    );
+    if (this.props.direction === 'inline') {
+      return (
+        <div id={this.props.id}>
+          <Spacings.Inline
```
tdeekens: Not here but maybe in the future. I like the notion of `components: {}` we recently adopted far more. So it would be `components: { spacing: <Spacings.Inline scale='' /> }` maybe. Is that worth of something to track as an issue?

src/components/fields/radio-field/README.md
emmenko: Hmm, that's the same as other field components

tdeekens: Okay then. Just read "radio input component" and was like but it's a "radio field component".

src/components/inputs/radio-input/radio-group.js
emmenko: I see your point. In this case though, you really only have 2 choices: stack or inline. I'm not sure having the freedom of passing any component would make sense here 🤔

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
...
+| ---------------------- | ------------------ | :------: | ----------------------- | ---------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`                   | `string`           |    -     | -                       | -                | Used as HTML `id` property. An `id` is auto-generated when it is not specified.                                                                                                                                                                                       |
+| `horizontalConstraint` | `object`           |          | `m`, `l`, `xl`, `scale` | `scale`          | Horizontal size limit of the input fields.                                                                                                                                                                                                                            |
```
montezume: isn't this `horizontalConstraint` of the RadioGroup?

src/components/inputs/radio-input/radio-group.js
tdeekens: Yeah, me neither fully. Just wanted to share the thought. One could argue (not saying has to) that having an consistent/expected API is also nice. You should go "How can I customise this?" "Ah, just like everywhere, maybe pass in the `components` thing.".

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `onBlur`               | `func`             |    -     | -                       | -                | Called when input is blurred                                                                                                                                                                                                                                          |
+| `onFocus`              | `func`             |    -     | -                       | -                | Called when input is focused                                                                                                                                                                                                                                          |
+| `isDisabled`           | `bool`             |    -     | -                       | `false`          | Indicates that the input cannot be modified (e.g not authorised, or changes currently saving).                                                                                                                                                                        |
```
montezume: does this make sense to have isDisabled for an entire radiogroup?

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `onChange`             | `func`             |    -     | -                       | -                | Called with an event containing the new value. Required when input is not read only. Parent should pass it back as `value`.                                                                                                                                           |
+| `onBlur`               | `func`             |    -     | -                       | -                | Called when input is blurred                                                                                                                                                                                                                                          |
+| `onFocus`              | `func`             |    -     | -                       | -                | Called when input is focused                                                                                                                                                                                                                                          |
```
montezume: Is this right? Or it's called when any of the RadioOptions are focused?

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `value`                | `string`           |    ✅    | -                       | -                | Value of the input component.                                                                                                                                                                                                                                         |
+| `onChange`             | `func`             |    -     | -                       | -                | Called with an event containing the new value. Required when input is not read only. Parent should pass it back as `value`.                                                                                                                                           |
+| `onBlur`               | `func`             |    -     | -                       | -                | Called when input is blurred                                                                                                                                                                                                                                          |
```
montezume: ditto

src/components/fields/radio-field/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `name`                 | `string`           |    -     | -                       | -                | Used as HTML `name` of the input component. property                                                                                                                                                                                                                  |
+| `value`                | `string`           |    ✅    | -                       | -                | Value of the input component.                                                                                                                                                                                                                                         |
+| `onChange`             | `func`             |    -     | -                       | -                | Called with an event containing the new value. Required when input is not read only. Parent should pass it back as `value`.                                                                                                                                           |
```
montezume: can a RadioGroup be read only?

src/components/fields/radio-field/radio-field.spec.js
```diff
@@ -0,0 +1,188 @@
...
+
+describe('when `badge` is passed', () => {
+  it('should render a badge', () => {
+    const { getByText } = renderRadioField({ badge: 'foo badge' });
+    expect(getByText('foo badge')).toBeInTheDocument();
+  });
+});
+
+describe('when disabled', () => {
+  it('should disable the input', () => {
```
montezume: `inputs`?

montezume: Feels weird though. to me it makes more sense to pass isDisabled to the individual RadioOptions rather than the field.

src/components/inputs/radio-input/radio-group.js
```diff
@@ -36,25 +52,49 @@ class Group extends React.PureComponent {
...
+      this.props.children,
+      (child, index) => {
+        // NOTE: Allowing to intersperse other elements than `Option`.
+        if (child && child.type.displayName === Option.displayName) {
+          return React.cloneElement(child, {
+            id: `${this.props.id}-${index}`,
+            name: this.props.name,
+            isChecked: this.props.value === child.props.value,
+            isDisabled: child.props.isDisabled || this.props.isDisabled,
+            isReadOnly: this.props.isReadOnly,
```
montezume: I find this functionality confusing. Maybe the prop names are wrong.

isDisabled makes every single radio option disabled?

src/components/fields/radio-field/README.md
emmenko: I've put it on the each `RadioInput.Option`. Do you think we should leave them out (onBlur/onFocus) or should we keep them like this?

src/components/fields/radio-field/README.md
emmenko: Maybe, maybe not. I doesn't hurt to have it I suppose, as it's inline with the other inputs/fields.

src/components/fields/radio-field/radio-field.spec.js
emmenko: You usually pass `isDisabled` to the entire group. In some special cases, you can then disable only specific options by passing `isDisabled` directly to them.

src/components/fields/radio-field/radio-field.spec.js
emmenko: I simply like it more like this 🙃 Do you see any problems with this approach?

src/components/inputs/radio-input/README.md
```diff
@@ -8,46 +8,65 @@ import { RadioInput } from '@commercetools-frontend/ui-kit';
...
 #### Properties
 
 ##### RadioInput.Group
 
-| Props      | Type     | Required | Values                      | Default | Description                                                                       |
-| ---------- | -------- | :------: | --------------------------- | ------- | --------------------------------------------------------------------------------- |
-| `children` | `node`   |    ✅    | -                           | -       | The `RadioInput.Option` or another node (mixed children are allowed)              |
-| `value`    | `string` |    -     | -                           | -       | The selected value of the of a `RadioInput.Option` determining its checked status |
-| `onChange` | `func`   |    ✅    | -                           | -       | What will trigger whenever an `RadioInput.Option` is clicked                      |
-| `scale`    | `string` |    -     | ['xs', 's', 'm', 'l', 'xl'] | `m`     | Spacing between options                                                           |
```
emmenko: Technically, this is what causes the breaking change, as it's been renamed to `directionProps: { scale }`

src/components/inputs/radio-input/radio-group.js
```diff
@@ -36,25 +52,49 @@ class Group extends React.PureComponent {
...
-              return React.cloneElement(child, {
-                isChecked: this.props.value === child.props.value,
-                name: this.props.name,
-                onChange: this.props.onChange,
-              });
-            return child;
-          })}
-        </DirectionWrapper>
+      <div id={this.props.id}>
+        <Constraints.Horizontal constraint={this.props.horizontalConstraint}>
```
emmenko: Only the stack mode has the horizontal constraint, as it does not make sense to have it in the inline mode.

src/components/inputs/radio-input/radio-option.js
```diff
@@ -7,68 +7,73 @@ import Text from '../../typography/text';
...
-            />
-          </Spacings.Inline>
-        </label>
+const Option = props => (
+  <label
+    css={getLabelStyles(props)}
+    role="radio"
+    aria-checked={props.isChecked}
+    onFocus={props.onFocus}
+    onBlur={props.onBlur}
```
emmenko: I decided to put the onBlur/onFocus event handlers here, as it's what makes more sense to me, because the input is hidden and the entire label is clickable.

Not sure if we really need them though...

src/components/inputs/radio-input/radio-option.js
```diff
@@ -7,68 +7,73 @@ import Text from '../../typography/text';
...
-              type="radio"
-              {...filterDataAttributes(this.props)}
-            />
-          </Spacings.Inline>
-        </label>
+const Option = props => (
+  <label
+    css={getLabelStyles(props)}
+    role="radio"
+    aria-checked={props.isChecked}
```
emmenko: The `role` and `aria-checked` are for a11y purposes.
https://www.w3.org/TR/wai-aria/#aria-checked

src/components/inputs/radio-input/radio-option.js
```diff
@@ -1,74 +1,82 @@
...
+      </div>
+      <div
+        css={css`
+          width: 100%;
+          font-size: 1rem;
+          font-family: ${vars.fontFamilyDefault};
+          color: ${props.isDisabled
+            ? vars.fontColorDisabled
+            : vars.fontColorDefault};
+        `}
```
emmenko: So I ended up removing the `Text.Body` wrapper, as it was just causing problems with the HTML markup (e.g. `<div>` nested inside a `<p>`).
The styles to apply are actually just a few, I think it's better this way.

/cc @montezume

emmenko:
Alright, VRTs have been approved by design, I'd be happy to get another round of review here. Thanks!

tdeekens:
I lost track of this with the other PRs in app-kit. Where are we here? Looks like some sort of walkthrough was supposed to happen?

emmenko:
What do you mean "walkthrough"?

tdeekens:
I just read somewhere that Malcolm and you wanted a call. Maybe that happened. Sorry then. I was just wondering how we wanted to proceed...

emmenko:
Ah yeah, I guess we talked about it last week when I was in Berlin

emmenko:
I'll check it with Filip tomorrow (the percy diffs), otherwise there is nothing else to block this.

emmenko:
@tdeekens @montezume this is now ready to go (VRTs have been approved by design).

Is there anything left from your side?

PR #539, opened 2019-02-15
feat(tooltip): Add `off` prop

montezume:
#### Summary

Adds the prop `isEnabled` to the tooltip. It sort of mirrors the old "display" prop from the core tooltip.

#### Motivation

While refactoring the MC and adding the new tooltip, I found a lot of cases when I had to make code like this 

```js
const Button = props =>
  props.isDisabled ? (
    <Tooltip title="Disabled">
      <button disabled>Submit</button>
    </Tooltip>
  ) : (
    <button>Submit</button>
  );
```

This prop should make it a bit cleaner.

```js
const Button = props => (
    <Tooltip title="Disabled" off={!props.isDisabled}>
      <button disabled={props.isDisabled}>Submit</button>
    </Tooltip>
  );
```

### DIFF
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -73,13 +73,13 @@ const Body = styled.div`
 
 ### Conditionally displaying tooltips
 
-There may be cases when you only want to enable the display of a tooltip under a certain condition. In these cases, you may want to use the `isDisabled`.
+There may be cases when you only want to enable the display of a tooltip under a certain condition. In these cases, you may want to use the `off` prop.
 
 In the following example, the tooltip text only appears on hover when the button is disabled.
 
 ```js
 <Tooltip
-  isDisabled={!props.isDisabled}
+  off={props.isDisabled}
   title="You do not have permission to perform this action"
 >
   <button disabled={props.isDisabled}>Submit</button>
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -71,9 +71,11 @@ const Body = styled.div`
 </Tooltip>;
 ```
 
-### Disabling the tooltip functionality
+### Only displaying the tooltip under certain conditions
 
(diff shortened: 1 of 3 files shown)


src/components/tooltip/README.md
```diff
@@ -71,10 +71,26 @@ const Body = styled.div`
 </Tooltip>;
 ```
 
+### Only displaying the tooltip under certain conditions
+
+There may be cases when you only want to enable the display of a tooltip under a certain condition. In these cases, you may want to set the `isEnabled` prop to false.
+
+In the following example, the tooltip text only appears on hover when the button is disabled.
```
tdeekens: Reading this I like this more than `isVisible` cause the latter may spark the understanding that it is forced to be visible even without hovering. Maybe `isSurpressed` is another name for the prop?

montezume: @emmenko do you have any opinions on the prop naming?

* isDisabled
* isEnabled
* isSurpressed

tdeekens: I liked isSurpressed but am leaning towards isDisabled for API consistency.

emmenko: Hmm, if you end up doing `isEnabled={!props.isDisabled}` maybe `isDisabled` makes more sense?
I'm also thinking about something like `off`, which maybe better indicates that the event listeners are turned off.

montezume: I renamed it to `isDisabled`. The other option could be isOff 🤔.

emmenko: Or just `off` 🤔

montezume: all of our other props start with `is` though

montezume: I'm okay with `off` if @tdeekens is as well. I just want to :shipit:

tdeekens: `off` is fine. I kind of don't like to break with patterns unless there is a good reason. So my choice would have been `isOff` even if that is more to write.

PR #550, opened 2019-02-18
Fix money parsing

### LINKED ISSUE #538: MoneyInput incorrectly parse value when the currency has no fraction digits
**Describe the bug**
When having currency that doesn't have fraction digits, and amount greater than thousand.
the entered value is formatted with separator
e.g `100000.00` to `100,000`
but then this separator is treated as fraction symbol (`100.000`), and the value will be considered just `100`

**Expected behavior**
Should parse values like 
```js
{
  amount: "100,000",
  currencyCode: "JPY"
}
```

as 
```js
{
  "type":"centPrecision",
  "currencyCode":"JPY",
  "centAmount": 100000,
  "fractionDigits":0
}
```

**Screenshots**
kapture 2019-02-14 at 14 40 07

islam3zzat:
#### Summary
Fixes #538 
The problem is explained thoroughly in the linked issue
This PR fixes the issue, with the price of removing the following "feature" that 
if the user calls `MoneyInput.convertToMoneyValue` with the amount `10,000.00` or `10.000,00` it'll parse it.
After this PR user have to explicitly provide the locale that the conversion based on.
Now the user has to enter the separator according to the current locale

| Before | After|
|-|-|
| lost-feature | after |

this was necessary for solving 

| Before | After |
|-|-|
| kapture 2019-02-18 at 14 20 36 |  kapture 2019-02-18 at 14 21 51 |


src/components/fields/money-field/money-field.form.story.js
```diff
@@ -14,129 +15,137 @@ import Readme from './README.md';
...
-            price: MoneyInput.parseMoneyValue(),
-            pricePerTon: MoneyInput.parseMoneyValue(),
-            discountedPrice: MoneyInput.parseMoneyValue(),
-          }}
-          validate={values => {
-            const errors = { price: {}, pricePerTon: {}, discountedPrice: {} };
-            if (MoneyInput.isEmpty(values.price)) errors.price.missing = true;
-            else if (MoneyInput.isHighPrecision(values.price))
-              errors.price.unsupportedHighPrecision = true;
+const Story = injectIntl(props => {
```
islam3zzat: I moved the story to be a variable to injectItl, will point out the real changes

src/components/fields/money-field/money-field.form.story.js
```diff
@@ -14,129 +15,137 @@ import Readme from './README.md';
...
+          // is not required.
+          // Supplying either only currency or only an amount will be treated
+          // as leaving the field blank.
+          // UX could want to show an error in this case, but then we'd also
+          // need a way to unset the currency, which is not possible now.
+          if (
+            !MoneyInput.isEmpty(values.discountedPrice) &&
+            MoneyInput.isHighPrecision(
+              values.discountedPrice,
+              props.intl.locale
```
islam3zzat: added the locale here

src/components/fields/money-field/money-field.form.story.js
```diff
@@ -14,129 +15,137 @@ import Readme from './README.md';
...
-                    default:
-                      return null;
-                  }
-                }}
+          return omitEmpty(errors);
+        }}
+        onSubmit={(values, formik) => {
+          action('onSubmit')(values, formik);
+          // eslint-disable-next-line no-console
+          console.log('doc', formToDoc(values, props.intl.locale));
```
islam3zzat: added locale here too

islam3zzat: no other changes in the file, u can move a long :D

src/components/inputs/money-input/money-input.form.story.js
```diff
@@ -14,79 +15,79 @@ import Spacings from '../../spacings';
...
   if (MoneyInput.isEmpty(formValues.price)) {
     errors.price.missing = true;
-  } else if (MoneyInput.isHighPrecision(formValues.price)) {
+  } else if (MoneyInput.isHighPrecision(formValues.price, locale)) {
     errors.price.unsupportedHighPrecision = true;
   }
   return omitEmpty(errors);
 };
-
+const Story = injectIntl(props => (
```
islam3zzat: I moved the story to be a variable to injectItl, will point out the real changes

src/components/inputs/money-input/money-input.form.story.js
```diff
@@ -14,79 +15,79 @@ import Spacings from '../../spacings';
...
+  <Section>
+    <Formik
+      initialValues={{ price: { currencyCode: '', amount: '' } }}
+      validate={formValues => validate(formValues, props.intl.locale)}
```
islam3zzat: added locale here

src/components/inputs/money-input/money-input.form.story.js
```diff
@@ -14,79 +15,79 @@ import Spacings from '../../spacings';
...
+      onSubmit={(values, formik) => {
+        // eslint-disable-next-line no-console
+        console.log(
+          'money value',
+          MoneyInput.convertToMoneyValue(values.price, props.intl.locale)
```
islam3zzat: added locale here too

emmenko:
@islam3zzat I see that there are a lot of changes in `yarn.lock`. I think you're using an old version of `yarn`. Can you check and upgrade?

src/components/inputs/money-input/money-input.js
```diff
@@ -185,15 +182,15 @@ export const parseRawAmountToNumber = rawAmount => {
...
+export const createMoneyValue = (currencyCode, rawAmount, locale) => {
   if (!currencyCode) return null;
 
   const currency = currencies[currencyCode];
   if (!currency) return null;
 
   if (rawAmount.length === 0 || !isNumberish(rawAmount)) return null;
 
-  const amountAsNumber = parseRawAmountToNumber(rawAmount);
+  const amountAsNumber = parseRawAmountToNumber(rawAmount, locale);
```
tdeekens: This would be a breaking change as it stands? I may also not be following along adequately. But Can't the passed `locale` be considered a hint. So it works without but if passed, needed in certain circumstances, it ensures that parsing covers the last x% which otherwise fail?

tdeekens:
> Can you check and upgrade?

We should be on 1.13.0.

src/components/inputs/money-input/money-input.js
islam3zzat: Yes this would be a breaking change.
That's a good point. We can use the previous logic if locale is not passed, but this can cause incorrect value for fractionless currencies.
given that we can:
Add invariant that locale is required if the currency has no fractions,
and if locale isn't passed and currency has fraction digits use previous logic.
WDUT?

src/components/inputs/localized-money-input/localized-money-input.visualroute.js
```diff
@@ -15,6 +15,10 @@ const value = {
     currencyCode: 'CAD',
     amount: '19.82',
   },
+  JPY: {
```
montezume: why did you add this here if it's not actually showing up in any visual tests?

src/components/inputs/money-input/money-input.js
tdeekens: To we know the cases where it won't work beforehand or can it happen any time with MoneyInput and just depends on what the user has defined. Depending on that this might have to be breaking and we should always pass the locale.

islam3zzat: We know that the functions `MoneyInput.convertToMoneyValue` and `MoneyInput.isHighPrecision` will always fail if the currency has no fraction digits.

islam3zzat: Added an `invariant` that will throw if the currency has no fraction digits, and no local is provided

tdeekens: Cool. Makes sense.

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -84,25 +85,42 @@ describe('MoneyInput.getAmountInputId', () => {
...
       expect(
-        MoneyInput.convertToMoneyValue({ currencyCode: 'foo', amount: '1' })
+        MoneyInput.convertToMoneyValue(
+          { currencyCode: 'foo', amount: '1' },
+          'en'
+        )
       ).toEqual(null);
     });
   });
+  describe('when an unknown currency is used2', () => {
```
montezume: shouldn't this be `when a currency with no fractions digits is used with a locale being passed`?

src/components/inputs/money-input/money-input.js
islam3zzat: Will this now be considered breaking change?

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -84,25 +85,42 @@ describe('MoneyInput.getAmountInputId', () => {
...
+    });
+  });
 
   describe('when no amount is present', () => {
     it('should return an invalid object', () => {
       expect(
-        MoneyInput.convertToMoneyValue({ currencyCode: 'EUR', amount: '' })
+        MoneyInput.convertToMoneyValue(
+          { currencyCode: 'EUR', amount: '' },
+          'en'
```
montezume: is this right?

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -120,7 +138,10 @@ describe('MoneyInput.convertToMoneyValue', () => {
       // tests here.
 
       expect(
-        MoneyInput.convertToMoneyValue({ currencyCode: 'EUR', amount: 'foo' })
+        MoneyInput.convertToMoneyValue(
+          { currencyCode: 'EUR', amount: 'foo' },
+          'en'
```
montezume: do we still need all these default locales?

src/components/inputs/money-input/money-input.js
tdeekens: Not to what I can tell.

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,42 @@ const createCurrencySelectStyles = ({
...
+// Since most users are not careful about how they enter values, we will parse
+// both `.` and `,` as decimal separators.
 // This means the highest amount always wins. We do this by comparing the last
 // position of `.` and `,`. Whatever occurs later is used as the decimal
 // separator.
-export const parseRawAmountToNumber = rawAmount => {
+export const getLocalsSperatorAndThrowaway = (rawAmount, locale) => {
+  if (locale) {
+    const [throwaway, separator] = (9999.999)
+      .toLocaleString(locale)
```
tdeekens: I dont think I can follow each step here. Can we mapbe add inline comments as to why we do these things and what the outcomes are?

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,42 @@ const createCurrencySelectStyles = ({
...
 // When a value is `1.000,00` we parse it as `1000`.
 // When a value is `1,000.00` we also parse it as `1000`.
-//
+// Since most users are not careful about how they enter values, we will parse
+// both `.` and `,` as decimal separators.
 // This means the highest amount always wins. We do this by comparing the last
 // position of `.` and `,`. Whatever occurs later is used as the decimal
 // separator.
-export const parseRawAmountToNumber = rawAmount => {
+export const getLocalsSperatorAndThrowaway = (rawAmount, locale) => {
```
montezume: `Locals` or Locale? or Locales?
`Sperator` -> Seperator?

I don't think this function name really makes sense.

src/components/inputs/money-input/money-input.spec.js
islam3zzat: yes, but the default locales are not needed though

src/components/inputs/money-input/money-input.js
```diff
@@ -160,13 +160,13 @@ const createCurrencySelectStyles = ({
...
+export const getLocalesSperatorAndThrowaway = (rawAmount, locale) => {
   if (locale) {
     const [throwaway, separator] = (9999.999)
-      .toLocaleString(locale)
-      .replace(/9/g, '')
+      .toLocaleString(locale) // after this step it'll be either 9,999.999 or 9.999,999 based in the locale
+      .replace(/9/g, '') // then we remove the number `9` to endup with `,.` or `.,`
       .split('')
-      .map(symbol => (symbol === '.' ? '\\.' : symbol));
+      .map(symbol => (symbol === '.' ? '\\.' : symbol)); // here we escape the '.' to use it as regex
```
tdeekens: What is the goal behind the `map`. I still can't follow 100%.

islam3zzat: it is exactly like https://github.com/commercetools/ui-kit/blob/e65a1ec06baca13aecf5f64673c2bf1b9b7e7f0f/src/components/inputs/money-input/money-input.js#L176

when we remove the throwaway 
https://github.com/commercetools/ui-kit/blob/e65a1ec06baca13aecf5f64673c2bf1b9b7e7f0f/src/components/inputs/money-input/money-input.js#L190

if we didn't do this map, the `.` will be the regex any character, so we escape it

src/components/fields/money-field/money-field.js
```diff
@@ -107,7 +111,10 @@ class MoneyField extends React.Component {
             badge={
               this.props.hasHighPrecisionBadge &&
               !MoneyInput.isEmpty(this.props.value) &&
-              MoneyInput.isHighPrecision(this.props.value) ? (
+              MoneyInput.isHighPrecision(
+                this.props.value,
+                this.props.intl.locale
```
montezume: I'm still confused about this. Why do we use this locale to determine this? I don't need to have my computer set to Japanese to enter prices in Yen

montezume: couldn't we just determine with the currency code whether it's capable of being high precision?

package.json
```diff
@@ -147,6 +147,7 @@
     "html-webpack-plugin": "3.2.0",
     "husky": "1.3.1",
     "identity-obj-proxy": "3.0.0",
+    "intl": "^1.2.5",
```
emmenko: Why do we need this dependency? 🤔

src/components/fields/money-field/money-field.js
emmenko: As far as I know, the `intl.locale` is what you pass to the `IntlProvider`, which is usually what we set.

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,42 @@ const createCurrencySelectStyles = ({
...
 // }
 // which equals 0.0123456 €
-
-// Parsing
-// Since most users are not careful about how they enter values, we will parse
-// both `.` and `,` as decimal separators.
+// when locale is provided we use the number 9999.999 becauause it'll either be formated to
+// A: 9,999.999
+// B: 9.999,999
+// otherwise, we infere the throwaway, separator from the amount
```
emmenko: 👆Can you document the part above a bit better? It's a bit hard to understand...

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,42 @@ const createCurrencySelectStyles = ({
...
 // When a value is `1.000,00` we parse it as `1000`.
 // When a value is `1,000.00` we also parse it as `1000`.
-//
+// Since most users are not careful about how they enter values, we will parse
+// both `.` and `,` as decimal separators.
 // This means the highest amount always wins. We do this by comparing the last
 // position of `.` and `,`. Whatever occurs later is used as the decimal
 // separator.
-export const parseRawAmountToNumber = rawAmount => {
+export const getLocalesSperatorAndThrowaway = (rawAmount, locale) => {
```
emmenko: What is a `throwaway`? 🤔

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,42 @@ const createCurrencySelectStyles = ({
...
+  if (locale) {
+    const [throwaway, separator] = (9999.999)
```
emmenko: I'm very confused now about this part. Why is the number `9999.999` hardcoded?

src/components/inputs/money-input/money-input.spec.js
```diff
@@ -633,9 +655,7 @@ describe('MoneyInput', () => {
         language,
         options
       ) {
-        return `(${this}).toLocaleString(${language}, ${JSON.stringify(
-          options
-        )})`;
+        return new Intl.NumberFormat(language, options).format(this);
```
emmenko: Why not using react-intl to format the number?

package.json
islam3zzat: It's a devDependency
in tests JSDOM only have `en` 
we use Intl here to have the mock the behaviour of `Number.prototype.toLocaleString`
https://github.com/commercetools/ui-kit/blob/e65a1ec06baca13aecf5f64673c2bf1b9b7e7f0f/src/components/inputs/money-input/money-input.spec.js#L649
https://github.com/commercetools/ui-kit/blob/e65a1ec06baca13aecf5f64673c2bf1b9b7e7f0f/src/components/inputs/money-input/money-input.spec.js#L650

src/components/inputs/money-input/money-input.js
islam3zzat: I've added step by step inline comments in the function itself

src/components/inputs/money-input/money-input.js
islam3zzat: the thousands separator symbol

src/components/inputs/money-input/money-input.js
islam3zzat: it is explained in the following lines

src/components/inputs/money-input/money-input.spec.js
islam3zzat: I don't know if they provide a similar function, that we can use here

src/components/fields/money-field/money-field.js
islam3zzat: > Why do we use this locale to determine this?

In currencies like JPY the number `12.555` is high precision according to the `en` local, but cent precision according to the `de` locale.

> I don't need to have my computer set to Japanese to enter prices in Yen

You don't, you can enter the number in all cases, but whether it is  high precision or not should be based on the locale

> couldn't we just determine with the currency code whether it's capable of being high precision

All currencies can be high precision if the entered value has more fractions than the currency's `fractionDigits`

> As far as I know, the `intl.locale` is what you pass to the `IntlProvider`, which is usually what we set

Yes it is, would it be better if we get a prob from the user instead?

montezume: I thought in currencies like JPY there is no such thing as cent precision or high precision in any locales. 

There's no such thing as cents, so how could they have high precision prices? 😕

islam3zzat: When the price of example was 7 JPY and there is 50% off then it's 3.5 JPY.
there's no cents or smaller units, but there are the cases when u have a fractions of the smallest unit.

islam3zzat: it's like Euro, it has 2 fraction digits `.01` is cent, but u can have a balance of `1.0000003` euro

montezume: Yen doesn't support fractional fractional units though. You can't have 3.5 JPY 😕 Can you find a source supporting this fact? AFAIK they removed the "sen" value from their currency in 1950s.

src/components/inputs/money-input/money-input.js
emmenko: So do we need this comment then? I find it confusing to be honest. I suppose it's clear for you that are familiar with the topic but we should try to make it understandable for everyone

src/components/inputs/money-input/money-input.js
emmenko: Can we find a better name? I find it confusing to understand what it does/is by the name

src/components/inputs/money-input/money-input.js
emmenko: I still don't understand the purpose of this 😕

Can you try to explain it a bit more?

src/components/fields/money-field/money-field.js
islam3zzat: https://www.xe.com/currencyconverter/convert/?Amount=1&From=EUR&To=JPY
they removed the "sen", but you can still have .3 JPY

src/components/inputs/money-input/money-input.spec.js
emmenko: I would expect that to work, because react-intl builds on top of intl as far as I know. Can you look into that?

src/components/fields/money-field/money-field.js
montezume: Can we involve someone from the API in this discussion and find out whether or not the API actually supports fractional YEN amounts?

montezume: I believe that fractional digits are only used for exchange and stock trading

filippobocik: Any currency can be high precision and it should not depend on any locale from my understanding. The use case are product variants with very low unit prices e.g print items like business cards, envelops etc. The price for one single business card could be EUR 0,00001 ... For EUR any price with more than 2 fraction digits is high precision. For JPY, any price with => 1 fraction digit is high precision. Lets do a short meeting tomorrow all together to get rid of the confusion

src/components/inputs/money-input/money-input.spec.js
islam3zzat: https://github.com/yahoo/react-intl/wiki#the-react-intl-module
none of the named exports help in mocking the `toLocaleString` here
as they are targeted more to rendering the localized values into react component, not used as util functions as we need in this tests

package.json
montezume: I would like to look into doing this without this new dep

src/components/inputs/money-input/money-input.js
```diff
@@ -149,23 +149,31 @@ const createCurrencySelectStyles = ({
...
 // }
 // which equals 0.0123456 €
-
-// Parsing
-// Since most users are not careful about how they enter values, we will parse
-// both `.` and `,` as decimal separators.
 // When a value is `1.000,00` we parse it as `1000`.
 // When a value is `1,000.00` we also parse it as `1000`.
-//
+// Since most users are not careful about how they enter values, we will parse
```
montezume: Can you revert the order of these comments? They don't make sense anymore.

montezume:
@emmenko do you want to take another look at this or is it good to go from your side as well?

PR #556, opened 2019-02-19
fix(card): do not apply padding to outer flex container

### LINKED ISSUE #554: Fix flex padding/margin overflow in Card component
**Describe the bug**
The ~~Spacing~~ Card element (which are `display: flex`) should not have any `padding`. Instead there should be an inner container that implements the spacing `padding` or `margin`.

**To Reproduce**
I noticed that when implementing the `{Info,Confirmation}Dialog` (https://github.com/commercetools/merchant-center-application-kit/pull/367, https://github.com/commercetools/merchant-center-application-kit/pull/367).

screenshot 2019-02-19 at 11 07 46

According to this SO answer (https://stackoverflow.com/a/37785605) this is because of how the calculation of the flex container is done in regards to padding/margin.

Setting the padding/margin to `0` will obviously remove the "overflow".

screenshot 2019-02-19 at 11 11 27

**Expected behavior**
The ~~Spacing~~ Card component should not cause overflow on the parent container element.

I would suggest to have a wrapping `<div>` inside the Card container, where the `<div>` will set the padding/margin

screenshot 2019-02-19 at 11 14 26

I can spin up a PR to better illustrate the fix.

emmenko:
Fixes #554

### DIFF
--- a/src/components/card/card.js
+++ b/src/components/card/card.js
@@ -16,7 +16,6 @@ const Card = props => (
         width: 100%;
         box-shadow: ${props.type === 'raised' ? vars.shadow1Second : 'none'};
         border-radius: ${vars.borderRadius6};
-        padding: ${vars.spacing16};
         background: ${props.theme === 'dark'
           ? vars.colorGray95
           : vars.colorWhite};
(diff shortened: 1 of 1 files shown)


src/components/card/card.js
```diff
@@ -16,15 +16,20 @@ const Card = props => (
...
-        padding: ${vars.spacing16};
         background: ${props.theme === 'dark'
           ? vars.colorGray95
           : vars.colorWhite};
       `,
     ]}
     className={props.className}
   >
-    {props.children}
+    <div
```
tdeekens: Curious: `styled.div`?

PR #558, opened 2019-02-19
fix(spacings): specificity of margin for children elements

### LINKED ISSUE #542: Fix specificity problem of margin for Spacings -> Text components
The `margin: 0` set on the typography components is really annoying when using them with e.g. `<Spacings.{Inline,Stack}>` components, because the margin from the spacing components doesn't get applied (lower specificity).
To work around that, you would wrap the `<Text>` components in a `<div>`.

Should we maybe simply force the spacing components to apply their margin by using `!important`?
It feels like a little hack but maybe it's more of a trade-off.

Thoughts?

emmenko:
Fixes #542

### DIFF
--- a/src/components/buttons/flat-button/flat-button.js
+++ b/src/components/buttons/flat-button/flat-button.js
@@ -74,9 +74,7 @@ export const FlatButton = props => {
       >
         <Spacings.Inline scale="xs" alignItems="center">
           {props.iconPosition === 'left' && getIconElement(props)}
-          <div>
-            <Text.Body>{props.label}</Text.Body>
-          </div>
+          <Text.Body>{props.label}</Text.Body>
           {props.iconPosition === 'right' && getIconElement(props)}
         </Spacings.Inline>
       </AccessibleButton>
--- a/src/components/buttons/link-button/link-button.js
+++ b/src/components/buttons/link-button/link-button.js
@@ -50,9 +50,7 @@ const LinkButton = props => (
           size: 'medium',
           theme: props.isDisabled ? 'grey' : 'green',
         })}
-      <div>
-        <Text.Body>{props.label}</Text.Body>
-      </div>
+      <Text.Body>{props.label}</Text.Body>
     </Spacings.Inline>
   </Link>
 );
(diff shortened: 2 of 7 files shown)


src/components/spacings/stack/stack.js
```diff
@@ -5,7 +5,15 @@ import getStyles from './stack.styles';
...
   <div css={getStyles(props)} {...filterDataAttributes(props)}>
-    {props.children}
+    {/* We need to wrap the children into a <div> to ensure that
+    the margins defined by the spacing component are properly applied.
+    Some children might have `margin: 0` defined, which results in a
+    higher CSS specificity, causing the margings of the spacing
+    components to be ignored.
+    See https://github.com/commercetools/ui-kit/issues/542 */}
+    {React.Children.map(props.children, child => (
+      <div>{child}</div>
```
tdeekens: Can we maybe additionally remove the lobotomized owls selector on the parent and just have the margin applied here? Would make CSS a bit more atomic.

emmenko: We can't, because the first child doesn't have the margin.

emmenko:
Should we go with the `!important` approach then?

montezume:
Yeah, !important seems like a cleaner approach to me.

qmateub:
I think I'm also more in the approach from Malcolm to use !important.

don't you have the feeling (with the wrapping div) of step in the consumer's toes 😅 in terms of we doing something "else" with the content that he/she defines in the Spacings? Apart from the fact that we add more stuff to the DOM tree 🤷‍♂️

emmenko:
Sure, that was my initial approach as well (read related github issue).

Anyway, I’ll change it to `!important` then as we seem to agree on that

emmenko:
Alright, changed it to `!important`

montezume:
Can you also remove the extra divs we added in various components to make up for this?

montezume:
Here's one such [example](https://github.com/commercetools/ui-kit/blob/a60b6ab333c983e3346bb5ff848dc2cca9050bb8/src/components/buttons/flat-button/flat-button.js#L78)

PR #560, opened 2019-02-21
feat(tooltip): add option for tooltip wrapper

montezume:
#### Summary

Adds another `component` prop to the `Tooltip` component, `TooltipWrapperComponent`. This can be used along with `ReactDOM.renderPortal` to render the Tooltip to another place inside of the document. This is useful when dealing with virtualized items, like our table.

The following example uses a simple `TooltipWrapperComponent` that renders the portal as a descendent of `document.body`. 

#### Before

before

#### After

after

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -15,11 +15,11 @@ const Wrapper = styled.div`
   display: inline-block;
 `;
 
-const PopperWrapperComponent = props => (
+const TooltipWrapperComponent = props => (
   <React.Fragment>{props.children}</React.Fragment>
 );
-PopperWrapperComponent.displayName = 'PopperWrapperComponent';
-PopperWrapperComponent.propTypes = {
+TooltipWrapperComponent.displayName = 'TooltipWrapperComponent';
+TooltipWrapperComponent.propTypes = {
   children: PropTypes.node.isRequired,
 };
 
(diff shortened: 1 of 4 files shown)


src/components/tooltip/tooltip.spec.js
```diff
@@ -319,6 +333,59 @@ describe('when used with a custom wrapper component', () => {
...
+    const button = getByText('Submit');
+    fireEvent.focus(button);
+    // should call callbacks
+    expect(onFocus).toHaveBeenCalled();
+    expect(onOpen).toHaveBeenCalled();
+
+    // should not render the tooltip inside of the main div\
+    const mainContainer = container.querySelector('#main');
+    expect(
+      mainContainer.querySelector("[data-testid='tooltip-custom-body']")
```
montezume: this is a cool part about react-testing-library

src/components/tooltip/tooltip.spec.js
```diff
@@ -319,6 +333,59 @@ describe('when used with a custom wrapper component', () => {
...
+    ).not.toBeInTheDocument();
+
+    // should render the tooltip inside of the portal
+    const portalContainer = container.querySelector('#portal-id');
+    expect(
+      portalContainer.querySelector("[data-testid='tooltip-custom-body']")
```
montezume: 😄

src/components/tooltip/tooltip.js
```diff
@@ -51,6 +59,14 @@ class Tooltip extends React.Component {
         }
         return null;
       },
+      PopperWrapperComponent: (props, propName) => {
```
tdeekens: 🤔 aren't we leaking implementation details with this name? The fact that we use popper?

montezume: 🤔 yeah makes sense.

How about `TooltipWrapperComponent`?

src/components/tooltip/tooltip.js
```diff
@@ -180,7 +196,8 @@ class Tooltip extends React.Component {
 
     const WrapperComponent = this.props.components.WrapperComponent || Wrapper;
     const BodyComponent = this.props.components.BodyComponent || Body;
-
+    const TooltipWrapper =
```
emmenko: Nitpick: to keep it consistent with the other variable names

- rename variable `TooltipWrapper` --> `TooltipWrapperComponent`
- rename component `TooltipWrapperComponent` --> `TooltipWrapper`

montezume: Done 👍

PR #563, opened 2019-02-21
feat: make text-input `theme-able`

montezume:
# Making UI-Kit `Themeable` 

This PR contains two major changes.

## New choices

The first change introduces *new* **choices**. These choices have more semantically meaningful names that the old choices. For example, `color-green`, becomes `color-primary`. For now, this is a *non breaking change*. The choices are duplicated, and the old choices remain. At some point in the future, we will remove these now **deprecated** choices. See the table below for the new (and old) names.

| (deprecated) choice 	| new choice           	| decision (value)                               	|
|---------------------	|----------------------	|------------------------------------------------	|
| color-green         	| color-primary        	| `#00b39e`                                      	|
| color-green-25      	| color-primary-25     	| `hsl(172.9608938547486, 100%, 25%)`            	|
| color-green-40      	| color-primary-40     	| `hsl(172.9608938547486, 100%, 40%)`            	|
| color-green-85      	| color-primary-85     	| `hsl(172.9608938547486, 100%, 85%)`            	|
| color-green-95      	| color-primary-95     	| `hsl(172.9608938547486, 100%, 95%)`            	|
| color-navy          	| color-accent    	| `#213c45`                                      	|
| color-navy-30       	| color-accent-30 	| `hsl(195, 35.2941176471%, 30)`                 	|
| color-navy-40       	| color-accent-40 	| `hsl(195, 35.2941176471%, 40)`                 	|
| color-navy-95       	| color-accent-95 	| `hsl(195, 35.2941176471%, 95)`                 	|
| color-navy-98       	| color-accent-98 	| `hsl(195, 35.2941176471%, 98)`                 	|
| color-gray          	| color-neutral        	| `#cccccc`                                         	|
| color-gray-60       	| color-neutral-60     	| `hsl(0, 0%, 60%)`                              	|
| color-gray-90       	| color-neutral-90     	| `hsl(0, 0%, 90%)`                              	|
| color-gray-95       	| color-neutral-95     	| `hsl(0, 0%, 95%)`                              	|
| color-blue          	| color-info           	| `#078cdf`                                      	|
| color-blue-85       	| color-info-85        	| `hsl(203.05555555555554, 93.9130434783%, 85%)` 	|
| color-blue-95       	| color-info-95        	| `hsl(203.05555555555554, 93.9130434783%, 95%)` 	|
| color-orange        	| color-warning        	| `#f16d0e`                                      	|
| color-orange-95     	| color-warning-95     	| `hsl(25.110132158590307, 89.0196078431%, 95%)` 	|
| color-red           	| color-error          	| `#e60050`                                      	|
| color-red-95        	| color-error-95       	| `hsl(339.1304347826087, 100%, 95%)`            	|
| color-black         	| color-solid           	| `#1a1a1a`                                      	|
| color-white         	| color-surface          	| `#ffffff`                                         	|

Internally we now use these new choices for our design tokens. 

## Introducing the concept of a `theme`

By default, we can think about UI-Kit's theme as being the relationship between a `choice` and a `decision`. For example, in our (default) theme, we set the `color-surface` choice to the decision `#FFF`. 

## Overriding the default theme for a specific component.

In this PR, the ability to override the default theme for the component `text-input` has been added. A user can set a custom theme to the component as follows.

```js

import { ThemeProvider } from 'emotion-theming';
import { TextInput } from '@commercetools-frontend/ui-kit';

const darkTheme = {
  colorSurface: 'black',
  colorSolid: 'white',
  colorDarkAccent98: 'white',
  colorError: 'darkred',
};

const DarkTextInput = ({value, onChange}) => (
    <ThemeProvider theme={darkTheme}>
        <TextInput name="dark-input" value={value} onChange={onChange} />
    </ThemeProvider>
);
```

## Overriding the default theme for an entire application.

`ThemeProvider` supports nested themes, so once we have added theming to more components, users will be able to override the theme once in their app, and have it propagated everywhere UI-Kit is used. 

## What's next?

Once this PR has been merged, we can go about supporting themeing in the other components.


materials/design-tokens.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  backgroundColorForInput: 'colorWhite',
```
montezume: this file would need to be autogenerated

src/components/inputs/styles.js
```diff
@@ -1,42 +1,54 @@
 import { css } from '@emotion/core';
 import vars from '../../../materials/custom-properties';
+import designTokens from '../../../materials/design-tokens';
+
+const getInputStyles = (props, theme) => {
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
```
montezume: basically how it works, is in your theme, you overwrite one of our choices. So for "colorGreen", you can specify "pink".

src/components/inputs/styles.js
```diff
@@ -1,42 +1,54 @@
...
+  };
 
-const getInputStyles = props => {
   const baseStyles = css`
     appearance: none;
-    background-color: ${vars.backgroundColorForInput};
-    border: 1px solid ${vars.borderColorForInput};
-    border-radius: ${vars.borderRadiusForInput};
+    background-color: ${overwrittenVars[designTokens.backgroundColorForInput]};
```
montezume: we use design tokens, but the design tokens variables contain "choices"

tdeekens:
Nice. Can you maybe, for us on the outside, give a set of use cases we need theming fir?

emmenko:
> Can you maybe, for us on the outside, give a set of use cases we need theming fir?

We wanted to look into it to see if/how we can use the ui-kit components/design tokens for other use cases other than the MC (e.g. documentation websites).

However, I feel like having a "theme-like" structure, will help/force us to keep the design system/tokens more consistent. For example, the color names as we have them now don't make much sense. Instead of `--color-green` we should probably use `--color-primary`.

tdeekens:
> the color names as we have them now don't make much sense

Yes, that's my main concern. We would need more generic/semantic colour names. Like `primary`, `accent`, `warning` etc pp.

montezume:
We can support theming without actually coming up with other colour palettes. As you can see from this PR, the code changes are very minimal 😇. 

The blocker right now is renaming the choices.

tdeekens:
I like the names!

Shecki:
in the design-team, we had a closer look at the naming and changed three proposals to make it more generic. (bold=new)
 
green = Primary
*navy = accent*
Blue = Information
Orange = Warning
Red = Error
Black = solid
*Gray = Neutral*
*White = surface*

sorry for mention it so late.. could these one be updated?
This would be awesome! Thanks!

tdeekens:
Thanks. I am a bit worried about these

```
Black = solid
Gray = Neutral
White = surface
```

White is not the surface whenever a theme is inverted. Just saying.

emmenko:
I agree with @tdeekens . Think about if you would have a dark theme, what would the color names map to.

Why not keeping black/white as dark/light?

Shecki:
maybe this graphic will help a bit.
if we would use (light) + (dark) all colors have to be at least light or dark.. we already setting limitations
image

this is why we went for (surface) (solid)

Hope it helps?

emmenko:
Yeah dark/light names for the inverted colors is a bit weird.

I think we all want the same thing eventually, solid/surface for me also feel a bit weird but maybe we just need to get used to it. At least I can't think of any better name...

Thanks

tdeekens:
Thanks. Agreed. Maybe it's just we're not used to it but eventually the smallest eveil name :)

montezume:
Okay I updated the color names 👍

montezume:
Sven gave the okay from the design side. Is there anything from the code review side or can I merge? @tdeekens @emmenko ?

materials/custom-properties.css
```diff
@@ -30,6 +30,29 @@
   --color-purple: #b866ff;
   --color-black: #1a1a1a;
   --color-white: #fff;
+  --color-primary: #00b39e;
+  --color-primary-25: hsl(172.9608938547486, 100%, 25%);
+  --color-primary-40: hsl(172.9608938547486, 100%, 40%);
+  --color-primary-85: hsl(172.9608938547486, 100%, 85%);
+  --color-primary-95: hsl(172.9608938547486, 100%, 95%);
```
tdeekens: `172.9608938547486` 👀

montezume: AFAIK these values haven't changed 😆. They are just copy pasted of `color-green-25`, etc. But yeah, the weird `172.9608938547486` value comes because before we were using some postcss plugin to generate the value. Might make sense to round these 🤔

src/components/inputs/text-input/text-input.story.js
```diff
@@ -8,35 +10,54 @@ import Section from '../../../../.storybook/decorators/section';
 import TextInputReadme from './README.md';
 import TextInput from './text-input';
 
+const darkTheme = {
+  colorLight: 'black',
+  colorDark: 'white',
+  colorDarkAccent98: 'white',
```
emmenko: Shouldn't those be solid/surface?

materials/custom-properties.css
tdeekens: I guess rounding them would be a separate task.

emmenko: Or maybe use rgba instead of hsl?

src/components/inputs/text-input/text-input.story.js
montezume: oh yeah true, thanks!

montezume: 🤦‍♂️

PR #569, opened 2019-02-25
chore: update deps

montezume:
* Closes #568 
* Closes #567 
* Closes #566 
* Closes #565


package.json
```diff
@@ -175,13 +175,13 @@
     "react-dom": "16.8.3",
     "react-intl": "2.8.0",
     "react-router-dom": "4.3.1",
-    "react-testing-library": "5.9.0",
+    "react-testing-library": "6.0.0",
```
montezume: the only breaking change is moving `testHooks` to it's own lib. we don't even use hooks so 🤷‍♂️ 

[ref](https://github.com/kentcdodds/react-testing-library/issues/302)

PR #571, opened 2019-02-28
chore: remove double emotion plugin

montezume:
I was investigating tree shaking issues with `ui-kit`, and I came across an (unrelated) problem. We don't need `emotion-plugin-babel` because it's already included in the preset.

[ref](https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop)

```
@emotion/babel-preset-css-prop includes the emotion plugin. The babel-plugin-emotion entry should be be removed from your config and any options moved to the preset.

```


rollup.config.js
```diff
@@ -17,8 +17,6 @@ const plugins = [
   }),
   // To use the nodejs `resolve` algorithm
   resolve(),
-  // To automatically externalize `peerDependencies` and `dependencies`
```
montezume: this comment is outdated because we removed the associated plugin.

montezume:
yay for VRT 😄

package.json
```diff
@@ -178,7 +177,7 @@
     "react-testing-library": "6.0.0",
     "react-value": "0.2.0",
     "rimraf": "2.6.3",
-    "rollup": "1.2.3",
+    "rollup": "1.3.2",
```
montezume: why not?

PR #572, opened 2019-02-28
feat: remove unnecessary lodash map values dep

montezume:
#### Summary

While looking into our bundle sizes, I noticed that `ui-kit` was making apps bundle `lodash.mapvalues`, but it's only used in one component for a pretty trivial function. I replaced it with ES6 reduce. 

For apps that aren't already using this dep, this shaves ~14 kbs (gzipped) from the vendor bundle.

screen shot 2019-02-28 at 4 21 27 pm


package.json
```diff
@@ -159,6 +158,7 @@
     "json-loader": "0.5.7",
     "lint-staged": "8.1.4",
     "lodash.camelcase": "4.3.0",
+    "lodash.mapvalues": "4.6.0",
```
emmenko: Do we still need it or not? 🤔

montezume: I moved it to dev deps, as we are using it in a story 🤷‍♂️

src/components/inputs/localized-multiline-text-input/localized-multiline-text-input.js
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import requiredIf from 'react-required-if';
-import mapValues from 'lodash.mapvalues';
-import oneLine from 'common-tags/lib/oneLine';
```
montezume: we can use the es6 version

package.json
montezume: It's in the story...

montezume: https://github.com/commercetools/ui-kit/blob/2861a1b289f2e5c05f57d7bbd2f1d7a950b95dd6/examples/form-inputs.example.story.js#L10

src/components/inputs/localized-multiline-text-input/localized-multiline-text-input.js
```diff
@@ -115,9 +117,14 @@ export class LocalizedMultilineTextInput extends React.Component {
 
   expandAllTranslations = () =>
     this.setState(prevState => ({
-      expandedTranslations: mapValues(
-        prevState.expandedTranslations,
-        () => true
+      expandedTranslations: Object.keys(prevState.expandedTranslations).reduce(
```
tdeekens: You could also just dump this into a `utils` in UIKit itself as a `mapValues` then the code would be even shared.

src/components/inputs/localized-multiline-text-input/localized-multiline-text-input.js
```diff
@@ -95,9 +94,12 @@ export class LocalizedMultilineTextInput extends React.Component {
     // This state is used to show/hide the remaining translations
     areLanguagesOpened: this.props.defaultExpandLanguages,
     // This state is to manage the expand/collapse of multiline text inputs
-    expandedTranslations: mapValues(this.props.value, () =>
-      Boolean(this.props.defaultExpandMultilineText)
-    ),
+    expandedTranslations: Object.keys(this.props.value).reduce((acc, key) => {
+      return {
+        [key]: Boolean(this.props.defaultExpandMultilineText),
+        ...acc,
```
qmateub: Can we maybe be more precise on naming the variables of the `reduce` ?

`acc` -> translations
`key` -> translation

🤔 

Also, this took me by surprised because I'm used to have code first the `...acc` and then the assignment 😅

montezume: It's not translations though... The object looks like this

```
en: true,
es: true,
```
It's key value of locale and whether or not it's expanded...

qmateub: Then  `allLocales` maybe?

package.json
montezume: okay I removed it here as well

montezume:
netlify is weird sometimes

PR #576, opened 2019-03-02
chore(rollup): output size snapshot of esm build

emmenko:
To provide useful information about the bundle size, treeshaking, etc.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,0 +1,16 @@
+{
+  "dist/ui-kit.esm.js": {
+    "bundled": 619262,
+    "minified": 449296,
+    "gzipped": 83470,
+    "treeshaked": {
+      "rollup": {
+        "code": 321329,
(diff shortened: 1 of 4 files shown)


emmenko:
This should help us improving treeshaking support, by looking at those numbers (current values below). I still don't know exactly what they mean by that what we should be looking at.

```json
"treeshaked": {
  "rollup": {
    "code": 321329,
    "import_statements": 1038
  },
  "webpack": {
    "code": 343714
  }
}
```

emmenko:
Also, interesting point here: https://github.com/ReactTraining/react-router/issues/6464#issuecomment-438373423

Basically, avoid using classes and use hooks instead 😅

emmenko:
Closing this for now, as it currently fails because of this problem: https://github.com/TrySound/rollup-plugin-size-snapshot/issues/24

PR #597, opened 2019-03-14
Fix parse money value in MoneyInput

islam3zzat:
#### Summary
While fixing previouse issue https://github.com/commercetools/ui-kit/issues/538 another issue was added :( 
Which makes incorrect value rendered in the german locale German.
Will walk you through the issue

| Before | After |
|-|-|
| if-fix-parse-money-value-before | money-input after |

### The issue
Let's say that the output from `getAmountAsNumberFromMoneyValue` was `14.55`

1. When we get the formatted amount in `parseMoneyValue`,  we stringify the amount first using `String` function. (which doesn't consider the locale ) so the output of `String(getAmountAsNumberFromMoneyValue(moneyValue))` in our case will be `"14.55"` no matter what the locale was
https://github.com/commercetools/ui-kit/blob/e124de8154c84a8cc20769cd2cd744a034e06862/src/components/inputs/money-input/money-input.js#L343 
2. And in the `formatAmount` function we create the money value using this string (rawAmount is `"14.55"`)
https://github.com/commercetools/ui-kit/blob/e124de8154c84a8cc20769cd2cd744a034e06862/src/components/inputs/money-input/money-input.js#L277
3. And in `createMoneyValue` function 
we parse  (rawAmount is `"14.55"`)  https://github.com/commercetools/ui-kit/blob/e124de8154c84a8cc20769cd2cd744a034e06862/src/components/inputs/money-input/money-input.js#L210

4. And in `parseRawAmountToNumber` function 
we remove the thousand separator (`.` in `de` locale, `,` in `en` locale) from the rawAmount
https://github.com/commercetools/ui-kit/blob/e124de8154c84a8cc20769cd2cd744a034e06862/src/components/inputs/money-input/money-input.js#L180

## Proposed fix
In the beginning, when we stringify the value, use a function that stringify with respect to the locale, because we'll latter remove the thousands spirators based on the provided locale

## Why did we miss this issue during development
- The issue happens when we convert to formValues in `de` locale ( or `es` or any locale that uses `.` as a thousands separator).
And in our story, we initialise the form with empty values.
- Also the test env doesn't have `de` locale, so we can't reproduce it in tests

## Repreduce on MC
- Switch your profile to `de` locale
- in [Merchant Center cart discount list](https://mc.escemo.com/sunrise-data-alex/discounts/carts?query=eyJzZWFyY2hUZXh0IjoidGVzdC1pbmNvbnNpc3RlbnQtdmFsaWRpdHkiLCJzb3J0aW5nIjp7ImtleSI6InNvcnRPcmRlciIsIm9yZGVyIjoiYXNjIn0sImZpbHRlcnMiOnt9LCJwYWdlIjoxLCJwZXJQYWdlIjoyMH0%3D) notice the value of a discount, then
- open the details, you'll find that the value is not accurate.

### DIFF
--- a/src/components/inputs/money-input/money-input.form.story.js
+++ b/src/components/inputs/money-input/money-input.form.story.js
@@ -3,7 +3,7 @@ import { Formik } from 'formik';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import { injectIntl } from 'react-intl';
-import { withKnobs } from '@storybook/addon-knobs';
+import { withKnobs, number } from '@storybook/addon-knobs';
 import withReadme from 'storybook-readme/with-readme';
 import omitEmpty from 'omit-empty';
 import ErrorMessage from '../../messages/error-message';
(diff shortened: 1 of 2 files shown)


islam3zzat:
@tdeekens @emmenko currently `Merging is blocked`. Who can continue with this from here (merge and release)

PR #612, opened 2019-03-25
Update dependencies

emmenko:
Closes #606 
Closes #607 
Closes #608 
Closes #609 
Closes #610 
Closes #611


emmenko:
Right, I see that now in the logs. Thanks for the reminder!

scripts/get-babel-preset.js
```diff
@@ -38,6 +38,7 @@ module.exports = function getBabelPreset() {
           targets: {
             browsers: ['last 1 versions'],
           },
+          corejs: 2,
```
tdeekens: We also need to make the dependency explicit and have it in our pkg.json.

emmenko: Hmm are you sure? It seems to work like this..anyway, I can add it

tdeekens: Yes, pretty much.

1. https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpolyfill

> Right now @babel/polyfill is mostly just an alias of core-js v2

I would maybe remove @babel/polyfill and make core-js@2 explicit?

emmenko: We don't use `@babel/polyfill`. I added the `core-js` as a dependency. Is that ok now?

.travis.yml
```diff
@@ -11,6 +11,11 @@ cache:
   directories:
     - node_modules
 
+before_install:
+  # Install yarn
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
+  - export PATH=$HOME/.yarn/bin:$PATH
```
tdeekens: Will add this to app-kit.

emmenko: We have it already there, you can bump the yarn version though 😉

tdeekens: Did that.

PR #613, opened 2019-03-28
chore: update deps

montezume:
Updates deps.


visual-testing-app/webpack.config.js
```diff
@@ -75,7 +75,7 @@ module.exports = {
     },
   },
   plugins: [
-    new CleanWebpackPlugin(['dist']),
+    new CleanWebpackPlugin(),
```
montezume: [ref](https://github.com/johnagan/clean-webpack-plugin/issues/106)

tdeekens: Yeah should work for UIKit in our apps (and hence app-kit) the change is a bit more involved.

package.json
```diff
@@ -94,38 +94,39 @@
...
-    "@storybook/react": "5.0.3",
-    "@storybook/theming": "5.0.3",
+    "@storybook/addon-actions": "5.0.5",
+    "@storybook/addon-info": "5.0.5",
+    "@storybook/addon-knobs": "5.0.5",
+    "@storybook/addon-links": "5.0.5",
+    "@storybook/addon-options": "5.0.5",
+    "@storybook/addon-storysource": "5.0.5",
+    "@storybook/react": "5.0.5",
+    "@storybook/theming": "5.0.5",
```
montezume: seems to be a breaking change here as it fails without having `emotion-theming` installed 🤦‍♂️

package.json
```diff
@@ -193,7 +194,7 @@
     "react": "16.x",
     "react-dom": "16.x",
     "react-intl": "2.x",
-    "react-router-dom": "4.x"
+    "react-router-dom": "4.x || 5.x"
```
montezume: we are already using 5 in our apps

tdeekens: In a perfectly fine semver world this would be a breaking change too. As we change our peer deps and the consumer does have to change something (even though it's not a breaking change in itself). However, we also didn't bump to breaking in app-kit.

emmenko: We should just go with `5.x`

montezume:
Not sure why these visual tests are failing here and in my theming branch 🤔

PR #615, opened 2019-03-28
Refactor to use CircleCI instead of TravisCI

tdeekens:
#### Summary

This moves our CI config to CircleCI. Hoping for faster builds and to have all projects in one tool.


.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+      - bundle_test:
+          requires:
+            - build
+      - vrt_test:
+          requires:
+            - build
+      - publish:
+          requires:
+            - bundle_test
+          filters:
```
tdeekens: Only on `master` and only with `tags` present on the commit we want to publish.

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+      - checkout
+      - restore_cache: *restore_yarn_cache
+      - install_dependencies
+      - build_artifacts
+      - run:
+          name: Authenticating with npm registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > << parameters.working_directory >>/.npmrc
+      - run:
+          name: Publishing to next dist-tag
+          command: yarn release:next
```
tdeekens: Need to define the run script now.

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+        type: string
+        default: *working_directory
+    steps:
+      - checkout
+      - restore_cache: *restore_yarn_cache
+      - install_dependencies
+      - build_artifacts
+      - run:
+          name: Authenticating with npm registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > << parameters.working_directory >>/.npmrc
```
tdeekens: Both `NPM_TOKEN` and `PERCY_TOKEN` are setup in CircleCI.

scripts/run-percy.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
 
 # skip percy builds when this is not a pull request or when it's not master
-if [[ $TRAVIS_PULL_REQUEST != 'false' || ${TRAVIS_BRANCH} == 'master' ]]
+if [[ -n $CIRCLE_PR_NUMBER || ${CIRCLE_BRANCH} == 'master' ]]
```
tdeekens: Hope this makes sense. `-n` to check that it is there (contrary to `-z`.

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+    steps:
+      - run: yarn run jest --projects jest.test.config.js --maxWorkers=3 --reporters jest-silent-reporter
+  bundle_test:
+    description: 'Running bundle tests'
+    steps:
+      - run: yarn run jest --projects jest.bundle.config.js --maxWorkers=3 --reporters jest-silent-reporter
+  vrt_test:
+    description: 'Running Visual Regression Tests'
+    steps:
+      - run: ./scripts/run-percy-.sh
```
montezume: I think there's a typo here? Shouldn't it be `run-percy.sh`?

tdeekens: Snap.

tdeekens:
For me installing the deps like this manually until we have a docker image is fine. It takes only 8 secs. When we then have our own public docker image we can use that here as an executor.

.circleci/config.yml
```diff
@@ -0,0 +1,156 @@
...
+          name: Install Chrome headless dependencies
+          working_directory: *working_directory
+          command: |
+            sudo apt-get install -yq gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 \
+            libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 \
+            libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 \
+            libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 \
+            ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget --fix-missing
+      - run: ./scripts/run-percy.sh
+  publish_to_next:
```
emmenko: I'm a bit surprised that there is no `npm` orb yet 😕

.circleci/config.yml
```diff
@@ -0,0 +1,156 @@
...
+jobs:
+  lint:
+    executor: node_10
+    steps:
+      - install_dependencies
+      - lint
+  unit_test:
+    executor: node_10
+    steps:
+      - install_dependencies
```
emmenko: What I don't like much here is that we potentially install the dependencies twice, if lint + tests run in parallel.

I think I would prefer to have only one job so that we only install the deps once.

.circleci/config.yml
```diff
@@ -0,0 +1,156 @@
...
+            - build
+      - vrt_test:
+          requires:
+            - build
+      - publish:
+          requires:
+            - bundle_test
+          filters:
+            tags:
+              only: /^v.*/
```
emmenko: 👌

jest-puppeteer.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  launch: {
+    args: ['--no-sandbox', '--disable-setuid-sandbox'],
```
emmenko: Can we add a comment explaining why we need this?

E.g.: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#setting-up-chrome-linux-sandbox

scripts/publish-docs.js
```diff
@@ -3,7 +3,7 @@ const shelljs = require('shelljs');
 const docsOrigin = 'origin-docs';
 const docsBranch = 'latest/docs';
 
-if (process.env.TRAVIS === 'true') {
+if (process.env.CI === 'true') {
```
emmenko: We can remove this file, we don't use it

.circleci/config.yml
```diff
@@ -0,0 +1,156 @@
...
+    name: 'Restoring yarn cache'
+    keys:
+      - v1-yarn-cache-{{ checksum "yarn.lock" }}
+      - v1-yarn-cache
+
+  - &save_yarn_cache
+    name: 'Saving yarn cache'
+    key: v1-yarn-cache-{{ checksum "yarn.lock" }}
+    paths:
+      - ~/.cache/yarn
```
emmenko:
```suggestion
      - ~/.cache/yarn
      - node_modules
```

.circleci/config.yml
tdeekens: We can always improve. I would keep it like this for now and later we can perist it to the workspace. Given that we have a yarn cache in place the only thing which may take a bit longer is building native bindings. Again, I hope to keep it like this and given we have numbers fine tune as a follow up.

.circleci/config.yml
tdeekens: Won't add. Isn't mentioned [here](https://circleci.com/docs/2.0/yarn/). Would prefer to just store the yarn cache and not installed files.

emmenko: We do this in all our circleci repos!

https://github.com/commercetools/merchant-center-frontend/pull/4303

.circleci/config.yml
emmenko: > I would keep it like this for now and later we can perist it to the workspace

What do you want to persist?

> Again, I hope to keep it like this and given we have numbers fine tune as a follow up.

Here you go, same workflow, double installation of dependencies.

Screenshot 2019-03-28 at 18 15 44

---

I don't understand why you are against this. We do the same in our frontend repo:

image

tdeekens: In 1d83a94 I added the `node_modules` to the cache. From what I understand you, you want to persist everything installed in a `install` job to disc and restore it in every subsequent step?

emmenko: Join into one job (as you did https://github.com/commercetools/ui-kit/pull/615/commits/059fe0dc6d725cc29542dfc0f72b6a3fcada7ed8)

tdeekens:
This is ready from my point of view. The installation of Puppeteer deps can remain as is as it's really fast and pulling down a Docker image is not cached.

Percy may fail as it now runs in another Docker container which slightly changes font rendering. Not sure.

montezume:
@tdeekens Percy doesn't render fonts inside of our container though. They zip up the DOM / CSSDOM and send it to their servers where they render it 🤔

.circleci/config.yml
```diff
@@ -0,0 +1,151 @@
+version: 2.1
+
+aliases:
+  - &working_directory ~/ui-kit
+
+  - &restore_yarn_cache
+    name: 'Restoring yarn cache'
+    keys:
+      - v1-yarn-cache-{{ checksum "yarn.lock" }}
```
montezume: is this something we will have to manually bump in the future?

.circleci/config.yml
```diff
@@ -0,0 +1,151 @@
...
+      - persist_to_workspace:
+          root: *working_directory
+          paths:
+            - dist
+  restore_workspace:
+    steps:
+      - attach_workspace:
+          at: *working_directory
+  lint:
+    description: 'Running linters'
```
montezume: we only have one linter 😁 😉

montezume:
Can you try triggering the job again to see if percy passes this time?

.circleci/config.yml
tdeekens: Generally not. However it has prooven to be helpful to not only rely on the checksum or the yarn.lock. Whenever CircleCI's cache needs to be pruned we can just bump that v{x} thing.

tdeekens:
> Can you try triggering the job again to see if percy passes this time?

Triggered again. Note

montezume:
Okay yeah, let's just 🚀

tdeekens:
You're saying I should merge and re-configure the branch requirements on GitHub?

montezume:
Yeah unless @emmenko has some final thoughts? Looks good to me.

tdeekens:
It's the same as in app-kit so I assume it's good. I am happy to take any further improvements noted as follow-ups too.

montezume:
Thanks @tdeekens 🎉

PR #616, opened 2019-03-28
refactor(collapsible-panel): add `min-height`

adnasa:
#### Summary

- adds `min-height` of `32px` on CollapsiblePanel's header section
This is done to anticipate use cases that renders `{Secondary|Primary}Button` on the header

With this, we can experience that the header's height "jumps" when `{Secondary|Primary}Button` are conditionally rendered

### DIFF
--- a/src/components/panels/collapsible-panel/collapsible-panel.js
+++ b/src/components/panels/collapsible-panel/collapsible-panel.js
@@ -106,7 +106,10 @@ export default class CollapsiblePanel extends React.PureComponent {
               <Spacings.InsetSquish scale={scale}>
                 <div
                   {...dataProps}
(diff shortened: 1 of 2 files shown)


adnasa:
:trollface:

Shecki:
Normal Panel is fine with the 8px more. ✅
**condensed-version** should stay like it was ❌. (hope this is possible?)

montezume:
@Shecki yep, we can pass in `isCondensed` and set this conditionally. Nice catch 👏

montezume:
Maybe need to rebase to master for circle to pick up the build?

adnasa:
this `unit_test` step is taking ages..

PR #617, opened 2019-03-29
fix(select-inputs): always show scrollbar

montezume:
#### Summary

The design team requested for us to always show the scroll bar in the `SelectInputs`. Right now it's not very obvious to users when they can scroll inside the select inputs.

This PR introduces that change.

#### Before

beforeS

#### After

afterS

### DIFF
--- a/src/components/internals/create-select-styles.js
+++ b/src/components/internals/create-select-styles.js
@@ -96,7 +96,6 @@ const menuListStyles = () => base => ({
   '&::-webkit-scrollbar-track': {
     '-webkit-border-radius': '5px',
     borderRadius: '5px',
-    // background: 'rgba(0,0,0,0.1)',
   },
 
   '&::-webkit-scrollbar-thumb': {
--- a/src/components/internals/create-select-styles.js
+++ b/src/components/internals/create-select-styles.js
@@ -90,13 +90,27 @@ const menuListStyles = () => base => ({
   backgroundColor: vars.backgroundColorInputPristine,
 
   '&::-webkit-scrollbar': {
-    '-webkit-appearance': 'none',
-    width: '7px',
+    width: '9px',
+  },
+
+  '&::-webkit-scrollbar-track': {
+    '-webkit-border-radius': '5px',
+    borderRadius: '5px',
+    background: 'rgba(0,0,0,0.1)',
   },
 
   '&::-webkit-scrollbar-thumb': {
-    borderRadius: '4px',
-    backgroundColor: 'rgba(0, 0, 0, .5)',
+    '-webkit-border-radius': '5px',
+    borderRadius: '5px',
+    background: 'rgba(0,0,0,0.2)',
+  },
+
+  '&::-webkit-scrollbar-thumb:hover': {
+    background: 'rgba(0,0,0,0.4)',
+  },
+
+  '&::-webkit-scrollbar-thumb:window-inactive': {
+    // background: 'rgba(0,0,0,0.05)',
   },
 });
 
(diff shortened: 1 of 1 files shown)


emmenko:
You know you can open the PR as a draft, right? 😉

montezume:
@emmenko no need, ready for review 😄 😉 :trollface:

src/components/internals/create-select-styles.js
```diff
@@ -89,6 +88,29 @@ const menuListStyles = () => base => ({
   padding: '0',
   borderRadius: vars.borderRadiusInput,
   backgroundColor: vars.backgroundColorInputPristine,
+
+  '&::-webkit-scrollbar': {
```
emmenko: Where does all of this come from? 👀

Can you at least add some comments or something? Otherwise it looks like a huge hack (to me).

montezume: The default setting for scrolling in Mac OSX is to only show scrollbars on scroll.

Screen Shot 2019-03-29 at 2 55 19 PM

The design team wants us to override this an always show the scrollbar. So we use CSS to "hack it" so that we can see the scrollbar.

src/components/internals/create-select-styles.js
```diff
@@ -89,6 +88,33 @@ const menuListStyles = () => base => ({
   padding: '0',
   borderRadius: vars.borderRadiusInput,
   backgroundColor: vars.backgroundColorInputPristine,
+
+  // this CSS is here to override the default scroll behaviour in Mac OSX
```
emmenko:
```suggestion
  // this CSS is here to override the default scroll behaviour in MacOs
```

src/components/internals/create-select-styles.js
montezume: I added a comment, but it is a huge hack 😄

emmenko: Ah ok, so this is OS specific instead of browser specific?

tdeekens: I appriciate the effort but am unsure what to think of this. It's almost like styling the scrollbar in the 90s ain't it 😄. Why do we overwrite a default which a user likely make on purpose on his OS?

montezume: Maybe @Shecki can chime in and explain the reasoning from the design side.

Shecki:
the issue appears when we were testing some functionality.
it fails when they should something from the drop field.
they were not aware that you can scroll it.

even when there is the cursor inside, it won't show the scroll bar
image
this is why we have to  show the scroll bar when its scrollable

emmenko:
> this is why we have to show the scroll bar when its scrollable

Given that this is an OS specific setting, you would have this problem is ALL scrollable views, right?

montezume:
@emmenko I guess this one is more important because it's an input. We don't have any other scrollable inputs do we?

Shecki:
in tables for example, you can see that the content is cutted:
image

so here it would be clear,that the content is scrollable.

the Dropfields instead have no cutted content from the hight, so it looks quite like "perfect-fit".
this is where the trouble happens

emmenko:
Would it make sense to have a small inset shadown at the bottom to give the impression that the content overflows?

src/components/internals/create-select-styles.js
montezume: To answer the OS specific questions, yes this only effects Mac OS. On Windows, nothing changes.

islam3zzat:
there might be a simpler solution to this, than always show the scroll bar (which will make the behaviour inconsistent with the rest of the OS :D )
e.g adjust the default dropdown height so that it shows part of an option, this will show that the options are cutted, exactly like the tables

PR #618, opened 2019-03-29
chore: update changelog, bump version to v9.1.2

adnasa:


tdeekens:
Lets see if the CirleCI publishing setup works.

PR #627, opened 2019-04-02
fix: LinkButton to use inline TextBody

### LINKED ISSUE #588: LinkButton should render the label in a <span> element
**Describe the bug**
The `<LinkButton>` renders the label in a `<Text.Body>` element, which defaults to `<p>`.
This becomes problematic when the `<LinkButton>` is rendered within a paragraph text, as nested `<p>` are not valid.

**To Reproduce**
image

https://codesandbox.io/s/k06ln8z2xr

**Expected behavior**
The `<LinkButton>` should render the label within a `<span>` element

jonnybel:
#### Summary

This PR:
- Updates documentation of `Text.Body` and `Text.Detail` to include the `isInline` prop.
- Fixes #588 by using the aforementioned prop.

#### Description

There is still a problem with the `LinkButton` rendering a `div` (from the `Spacings.Inline` wrapper), which inside a `p` is also an invalid DOM nesting. I shall address this in a follow-up PR.

### DIFF
--- a/src/components/buttons/link-button/link-button.js
+++ b/src/components/buttons/link-button/link-button.js
@@ -22,12 +22,12 @@ const LinkButton = props => (
         min-height: initial;
         cursor: pointer;
         text-decoration: none;
-        p {
+        span {
           color: ${vars.colorGreen};
           ${props.isDisabled ? `color: ${vars.colorGray};` : ''}
         }
         &:hover {
-          p {
+          span {
             color: ${vars.colorGreen25};
             ${props.isDisabled ? `color: ${vars.colorGray});` : ''}
           }
@@ -50,7 +50,7 @@ const LinkButton = props => (
           size: 'medium',
           theme: props.isDisabled ? 'grey' : 'green',
         })}
-      <Text.Body>{props.label}</Text.Body>
+      <Text.Body isInline={true}>{props.label}</Text.Body>
     </Spacings.Inline>
   </Link>
 );
--- a/src/components/typography/text/README.md
+++ b/src/components/typography/text/README.md
@@ -94,7 +94,7 @@ Wraps the given text in a `<p>` element, for normal content.
 | `children` | `PropTypes.node` |    ✅    | -                                                                             | -       |
 | `title`    | `String`         |    -     | -                                                                             | -       |
 | `truncate` | `Bool`           |    -     | -                                                                             | `false` |
-| `inline`   | `Bool`           |    -     | -                                                                             | `false` |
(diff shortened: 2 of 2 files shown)


src/components/buttons/link-button/link-button.js
```diff
@@ -50,7 +50,7 @@ const LinkButton = props => (
           size: 'medium',
           theme: props.isDisabled ? 'grey' : 'green',
         })}
-      <Text.Body>{props.label}</Text.Body>
+      <Text.Body isInline>{props.label}</Text.Body>
```
tdeekens: Nit: `isInline={true}`. I didn't see us abbreviating that here. If we would maybe renaming to `inline` would be apt but I don't see that breaking change happening.

src/components/typography/text/README.md
```diff
@@ -94,6 +94,7 @@ Wraps the given text in a `<p>` element, for normal content.
 | `children` | `PropTypes.node` |    ✅    | -                                                                             | -       |
 | `title`    | `String`         |    -     | -                                                                             | -       |
 | `truncate` | `Bool`           |    -     | -                                                                             | `false` |
+| `inline`   | `Bool`           |    -     | -                                                                             | `false` |
```
tdeekens: `inline` or `isInline`? I am confused by the prop naming inconsistencies. When did it change and way if so from one convention to the other?

jonnybel: Sorry, I copied the wrong thing and didn't even notice 🤦‍♂️

src/components/typography/text/README.md
```diff
@@ -94,6 +94,7 @@ Wraps the given text in a `<p>` element, for normal content.
 | `children` | `PropTypes.node` |    ✅    | -                                                                             | -       |
 | `title`    | `String`         |    -     | -                                                                             | -       |
 | `truncate` | `Bool`           |    -     | -                                                                             | `false` |
+| `isInline` | `Bool`           |    -     | -                                                                             | `false` |
```
montezume: is this used anywhere?

src/components/typography/text/README.md
montezume: I don't see this prop actually being used though? Shouldn't the default be "true" then?

src/components/typography/text/README.md
emmenko: This is the README of the Text component, I think the props was missing from the documentation...right?

jonnybel: Yes, it was already being used, just not documented.

montezume: oooh 🤦‍♂️

PR #630, opened 2019-04-03
feat(localized-text-input): support custom theming, use design tokens

montezume:
#### Summary

Uses design tokens in `LocalizedTextInput`, and supports custom theming.

### DIFF
--- a/src/components/inputs/localized-text-input/localized-text-input.styles.js
+++ b/src/components/inputs/localized-text-input/localized-text-input.styles.js
@@ -1,13 +1,14 @@
 import { css } from '@emotion/core';
 import vars from '../../../../materials/custom-properties';
+import designTokens from '../../../../materials/design-tokens';
 import { getInputStyles } from '../styles';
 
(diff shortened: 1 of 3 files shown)


src/components/inputs/localized-text-input/localized-text-input.styles.js
```diff
@@ -16,22 +17,36 @@ const getLocalizedInputStyles = props => [
...
-  border-top-left-radius: ${vars.borderRadiusInput};
-  border-bottom-left-radius: ${vars.borderRadiusInput};
-  border: 1px ${vars.borderColorInputDisabled} solid;
-  padding: 0 ${vars.spacing8};
-  transition: ${vars.transitionStandard};
-  border-right: 0;
-  box-shadow: none;
-  appearance: none;
-`;
+const getLanguageLabelStyles = (props, theme) => {
```
montezume: even those we aren't using `props` here, I added it for consistency sake with the other methods.

PR #631, opened 2019-04-03
feat(text): Add `as` prop while deprecating `elementType`

jonnybel:
#### Summary

This PR adds a `as` prop to the Text components, as suggested by @tdeekens in https://github.com/commercetools/ui-kit/issues/588#issuecomment-469685493.

Deprecates `elementType` prop of `Text.Headline` and `Text.Subheadline` components.
Deprecates `isInline` prop of `Text.Body` component.

#### Questions

- **Which element types should we allow to render for the `Body` component?** (currently just `span` and `p`)

### DIFF
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -48,23 +48,10 @@ const Headline = props => {
 
 Headline.displayName = 'TextHeadline';
 Headline.propTypes = {
-  as(props, propName, componentName, ...rest) {
-    // is required if the prop 'elementType' is not defined
-    if (!props.elementType) {
-      return PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired(
-        props,
-        propName,
-        componentName,
-        ...rest
-      );
-    }
-    return PropTypes.oneOf(['h1', 'h2', 'h3'])(
-      props,
-      propName,
-      componentName,
-      ...rest
-    );
-  },
+  as: requiredIf(
+    PropTypes.oneOf(['h1', 'h2', 'h3']),
+    props => !props.elementType
+  ),
   elementType(props, propName, componentName, ...rest) {
     if (props[propName] != null) {
       const message = oneLine`
@@ -93,7 +80,6 @@ Headline.propTypes = {
       ...rest
     );
   },
-  children: PropTypes.node.isRequired,
   title: nonEmptyString,
   truncate: PropTypes.bool,
   intlMessage: requiredIf(intlMessageShape, props => !props.children),
@@ -119,23 +105,7 @@ const Subheadline = props => {
 
 Subheadline.displayName = 'TextSubheadline';
 Subheadline.propTypes = {
-  as(props, propName, componentName, ...rest) {
-    // is required only if the prop 'elementType' is not defined
-    if (!props.elementType) {
-      return PropTypes.oneOf(['h4', 'h5']).isRequired(
-        props,
-        propName,
-        componentName,
-        ...rest
-      );
-    }
-    return PropTypes.oneOf(['h4', 'h5'])(
-      props,
-      propName,
-      componentName,
-      ...rest
-    );
-  },
+  as: requiredIf(PropTypes.oneOf(['h4', 'h5']), props => !props.elementType),
   elementType(props, propName, componentName, ...rest) {
     if (props[propName] != null) {
       const message = oneLine`
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -4,8 +4,9 @@ import { FormattedMessage } from 'react-intl';
 import requiredIf from 'react-required-if';
 import isNil from 'lodash/isNil';
 import { oneLine } from 'common-tags';
-import warning from 'warning';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
+import throwDeprecationWarning from '../../../utils/warn-deprecated-prop';
+
 import {
   bodyStyles,
   detailStyles,
@@ -54,11 +55,11 @@ Headline.propTypes = {
   ),
   elementType(props, propName, componentName, ...rest) {
     if (props[propName] != null) {
-      const message = oneLine`
-            "${propName}" property of "${componentName}" has been deprecated
-            and will be removed in the next major version.\n
-            Please use "as" prop instead.`;
-      warning(false, message);
+      throwDeprecationWarning(
+        propName,
+        componentName,
+        `\n Please use "as" prop instead.`
+      );
 
       if (props.as) {
         return new Error(oneLine`
@@ -108,11 +109,11 @@ Subheadline.propTypes = {
   as: requiredIf(PropTypes.oneOf(['h4', 'h5']), props => !props.elementType),
   elementType(props, propName, componentName, ...rest) {
     if (props[propName] != null) {
-      const message = oneLine`
-            "${propName}" property of "${componentName}" has been deprecated
-            and will be removed in the next major version.\n
-            Please use "as" prop instead.`;
-      warning(false, message);
+      throwDeprecationWarning(
+        propName,
+        componentName,
+        `\n Please use "as" prop instead.`
+      );
 
       if (props.as) {
         return new Error(oneLine`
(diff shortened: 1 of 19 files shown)


src/components/typography/text/text.js
```diff
@@ -33,14 +36,58 @@ const Headline = props => {
...
+      ...rest
+    );
+  },
+  elementType(props, propName, componentName, ...rest) {
+    if (props[propName] != null) {
+      const message = oneLine`
+            "${propName}" property of "${componentName}" has been deprecated
+            and will be removed in the next major version.\n
+            Please use "as" prop instead.`;
+      warning(false, message);
```
emmenko: This will cause a massive amount of warnings.

I would proceed as following instead:
- introduce the `as` prop without touching the other props
- wait for a new major bump
- deprecate the `elementType` prop and add the warning
- wait for a new major bump
- remove the `elementType` prop

Ideally, we would also provide a codemod to run the migration.

src/components/typography/text/text.js
```diff
@@ -33,14 +36,58 @@ const Headline = props => {
...
+    // is required if the prop 'elementType' is not defined
+    if (!props.elementType) {
+      return PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired(
+        props,
+        propName,
+        componentName,
+        ...rest
+      );
+    }
+    return PropTypes.oneOf(['h1', 'h2', 'h3'])(
```
emmenko: I'm a bit confused now. What's the purpose of changing the prop `elementType` to `as` if you have the same exact restrictions as before? 🤔

tdeekens: ...to settle on a convention most other component libraries also have in offeing a prop called `as` not `elementType`. Other components, before we started with #628 assuming there too, would then offer an `as` prop too if needed and not `elementType`.

src/components/typography/text/text.js
tdeekens: I think this is what was aimed for here and also agreed upon before. Generally, I see the intend here to warn through prop-types given the prop is used. Then fix up the warnings (while still supporting the prop) by migrating to `as` and then in the next major dropping the `elementType`.

How did you read us wanting a quick breaking change from this code?

src/components/typography/text/text.js
tdeekens: > To be honest, I think there are several implications that need to be discussed first instead of jumping into the implementation.

We have an issue where the prop name was suggested. So lets move the discussion there. Can you lay out the specific bigger implication you have for text components there?

src/components/typography/text/text.js
montezume: I think providing a codemod is a bit overkill considering we could make the changes in the MC-FE in about 2 minutes with find and replace. 

I think the path described by @tdeekens seems quite reasonable.

tdeekens: Yeah, would be a nice learning experience to write a codemod during tech time if anybody wants but seems faster without. Assuming that not too many people use ui-kit yet (whom the find and replace is also quick an easy).

I am just lagging behind to how the suggested implementation here seemed to suggest another path than outlined above.

src/components/typography/text/text.js
```diff
@@ -33,14 +36,58 @@ const Headline = props => {
 
 Headline.displayName = 'TextHeadline';
 Headline.propTypes = {
-  elementType: PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired,
+  as(props, propName, componentName, ...rest) {
```
tdeekens: We can use `requiredIf` to achieve this a bit more dense. For reference look [here](https://github.com/commercetools/ui-kit/blob/c85074af2ce933a89d62791068f6066568f4db31/src/components/fields/time-field/time-field.js#L34). I like the idea however to only make the prop required if (and only if) elementType is not given (during the migration phase).

src/components/typography/text/text.js
tdeekens: Starting a new thread on using `warning` here. We can just return an `Error` to my knowledge.

Example from the official React docs:

```js
customProp: function(props, propName, componentName) {
    if (!/matchme/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  },
```

src/components/typography/text/text.js
```diff
@@ -98,12 +203,30 @@ const Body = props =>
...
   isBold: PropTypes.bool,
   isItalic: PropTypes.bool,
-  isInline: PropTypes.bool,
+  isInline(props, propName, componentName, ...rest) {
+    if (!isNil(props[propName])) {
+      const message = oneLine`
+            "${propName}" property of "${componentName}" has been deprecated
+            and will be removed in the next major version.\n
+            Please use "as" prop instead.`;
+      warning(false, message);
```
tdeekens: Same here. Just returning an error I think would do.

src/components/typography/text/text.js
```diff
@@ -98,12 +203,30 @@ const Body = props =>
...
+
+      if (props.as) {
+        return new Error(oneLine`
+        Invalid prop "${propName}" supplied to "${componentName}".
+        "${propName}" does not have any effect when "as" is defined`);
+      }
+    }
+
+    return PropTypes.bool(props, propName, componentName, ...rest);
```
tdeekens: Idea. Maybe we can share these prop-types in `typography/prop-types`. I know that they will be removed but it makes the actual code less noisy.

src/components/typography/text/text.js
jonnybel: I'd prefer to use a Warning rather than an Error here because it's **not invalid** to use the prop, but the developer should still be _warned_ about the future deprecation to be encouraged to migrate to the new prop.

I also ~stole it~ took inspiration from the official React docs:
https://reactjs.org/warnings/dont-call-proptypes.html#fixing-the-false-positive-in-third-party-proptypes

tdeekens: Ah cool. Thanks. Will resolve then.

jonnybel:
Thinking about merging this into 10.0.0, since it already implies breaking changes.

I'm updating the branch and we can discuss this further.

jonnybel:
I have a few things I'd like to discuss about the `Text.Detail` component and some positive implications of changing it.

* The `isInline` prop isn't doing anything 🤔  We can just remove it.
* It has a `className` prop. I searched on both the `ui-kit` and the `merchant-center-frontend` but couldn't find any usage of it...

As such, given these changes, almost all Text components (except for `Wrap`) could compose on top of the same "BaseText" component to reduce repetition of code, as they have almost the same API and mostly just accept different values as props.
For instance, the `Text.Detail` component is simply a `Text` _as_ a `<small>` element, and the rest is all the same as the `Text.Body`.

Thoughts on this?

montezume:
@jonnybel 

1) Yeah, seems to me like the usage of `isInline` could just be replaced with a wrapping div. This makes more sense to me 🤔. I would remove it.
2) Yes we should remove the className prop.

montezume:
The diff from the icon refactoring is showing in this PR even though this is aimed at `10.0.0`. Can you rebase it maybe so it fixes that?

jonnybel:
@montezume Done!

jonnybel:
@emmenko anything else you'd like to have changed?

We can use the same codemod I've added to #747 to automate the refactoring of these props!
This PR could be added to our v10 project, since it also implies breaking changes.

src/components/typography/text/text.js
```diff
@@ -45,15 +48,46 @@ const Headline = props => {
...
+  as: requiredIf(
+    PropTypes.oneOf(['h1', 'h2', 'h3']),
+    props => !props.elementType
+  ),
+  elementType(props, propName, componentName, ...rest) {
+    if (props[propName] != null) {
+      const message = oneLine`
+            "${propName}" property of "${componentName}" has been deprecated
+            and will be removed in the next major version.\n
+            Please use "as" prop instead.`;
```
emmenko: We can add a note also to use the codemod for the migration. WDYT?

jonnybel: Sure! Although I was thinking of adding that after #747 is merged, because then I can also refactor this to use the `util` that it adds [(this one)](https://github.com/commercetools/ui-kit/blob/4495d051e55a531118f00e5d929cb3a8c75682e6/src/utils/warn-deprecated-prop.js).

jonnybel:
Ok, the VRT detected changes now are only the prop name changes, in addition to the ones inherited from `10.0.0`.

PR #635, opened 2019-04-04
feat(password-input): support theming

montezume:
#### Summary

This is the last easy one.

### DIFF
--- a/src/components/inputs/password-input/password-input.visualroute.js
+++ b/src/components/inputs/password-input/password-input.visualroute.js
@@ -7,15 +7,7 @@ const value = 'hello world how are you?';
 
 export const routePath = '/password-input';
 
-const darkTheme = {
-  colorSurface: 'black',
-  colorSolid: 'white',
-  colorNeutral60: 'rgba(255,255,255,0.60)',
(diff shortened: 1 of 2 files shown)


src/components/inputs/password-input/password-input.visualroute.js
```diff
@@ -1,11 +1,20 @@
 import React from 'react';
+import { ThemeProvider } from 'emotion-theming';
 import { PasswordInput } from 'ui-kit';
 import { Suite, Spec } from '../../../../test/percy';
 
 const value = 'hello world how are you?';
 
 export const routePath = '/password-input';
 
+const darkTheme = {
```
emmenko: Instead of repeating this in ever VRT test, we should define this once and we can define the theme provider at the app level directly.

I can spin up a separate PR to make it more clear

emmenko: https://github.com/commercetools/ui-kit/pull/636

montezume: Yeah good idea 🎉

PR #636, opened 2019-04-04
refactor(vrt): define dark theme once and pass it as prop to tests

emmenko:
As suggested here: https://github.com/commercetools/ui-kit/pull/635#discussion_r272116681

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -30,6 +30,7 @@ overrides:
       'react/display-name': 0
   - files: ['*.visualroute.js']
     rules:
+      'react/prop-types': 0
       'react/display-name': 0
   - files: ['*.visualspec.js']
     globals:
--- a/test/percy/spec.js
+++ b/test/percy/spec.js
@@ -48,7 +48,7 @@ const PropValue = styled.span`
 `;
 
 const Box = styled.div`
-  background-color: ${props => (props.inverted ? '#111' : '#eee')};
+  background-color: ${props => props.theme.colorSurface};
 `;
 
 const Pill = props => {
@@ -102,14 +102,13 @@ const Spec = props => (
   <SpecContainer>
     <Label>{props.label}</Label>
     {!props.omitPropsList && <Props>{props.children}</Props>}
-    <Box inverted={props.inverted}>{props.children}</Box>
+    <Box>{props.children}</Box>
   </SpecContainer>
 );
 
 Spec.propTypes = {
   label: PropTypes.string.isRequired,
   children: PropTypes.node,
-  inverted: PropTypes.bool,
   omitPropsList: PropTypes.bool,
 };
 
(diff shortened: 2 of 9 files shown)


visual-testing-app/src/App.js
```diff
@@ -44,7 +52,7 @@ class App extends React.Component {
             <Route
               key={Component.routePath}
               path={Component.routePath}
-              component={Component.component}
+              render={() => <Component.component theme={darkTheme} />}
```
montezume:
```suggestion
              render={() => <Component.component themes={{darkTheme: darkTheme}} />}
```
?

emmenko: Do you plan to test multiple themes?

montezume: Maybe. Either way is fine for me. I also have an idea to use the themes in `Spec` instead of inverted

montezume: So in that case it would be nice to have a theme around the whole App, and then the specific ones around the inverted tests.

emmenko: I'm not sure I follow you... 🤔

.eslintrc.yaml
```diff
@@ -30,6 +30,7 @@ overrides:
       'react/display-name': 0
   - files: ['*.visualroute.js']
     rules:
+      'react/prop-types': 0
```
montezume: makes sense to do this here instead of in each component.

test/percy/spec.js
```diff
@@ -48,7 +48,7 @@ const PropValue = styled.span`
 `;
 
 const Box = styled.div`
-  background-color: ${props => (props.inverted ? '#111' : '#eee')};
+  background-color: ${props => props.theme.colorSurface};
```
montezume: here we can use the background colour of the app instead.

emmenko: Q: is `inverted` supposed to be the same as "dark". Or we just happen to use the dark theme as the "inverted" one?

visual-testing-app/src/App.js
```diff
@@ -22,41 +32,43 @@ const allComponents = componentsContext.keys().reduce((components, file) => {
...
-                    <li key={Component.routePath}>
-                      <a href={Component.routePath}>{Component.routePath}</a>
-                    </li>
-                  ))}
-                </ul>
-              </div>
-            )}
-          />
-          {Object.values(allComponents).map(Component => (
+      <ThemeProvider theme={customProperties}>
```
emmenko: Nitpick: maybe instead of spreading the custom properties, should we create and export a "default theme". Then creating new themes would be easier, as you can start by copying the "default theme".

Anyway, just a side thought. It might make more sense to look into this once we've implemented theming to all components.

montezume: Yeah a default theme makes a lot of sense.

PR #639, opened 2019-04-04
feat(i18n): support new locales fr-FR and zh-CN

emmenko:
Besides the obvious changes, there is one important (breaking) change: the static `i18n` export has been removed and instead an async export function is now exported to load the intl messages

```js
import { loadMessages } from '@commercetools-frontend/ui-kit';
```

This is now working the same way as we do it in the application messages.

The reason for this is that now that we have more locales, it doesn't make sense anymore to bundle all of them together in the main bundle.

**Update**
One important reason though (that I forgot to mention) is that we can’t export variables like `fr-FR`, as it’s not a valid syntax.

```js
import { i18n } from 'ui-kit';
const messages = i18n[locale];
```

If `locale` is e.g. `fr-FR`, it won't work.


emmenko:
image

emmenko:
> Starting to wonder how much % of the bundle size translations in UIKit actually are.

If you bundle them together is like 15Kb or something

rollup.config.js
```diff
@@ -66,17 +66,21 @@ const config = [
     input: 'src/index.js',
     external: defaultExternal,
     output: {
-      file: pkg.main,
       format: 'cjs',
+      dir: 'dist',
+      chunkFileNames: 'ui-kit-[name]-[hash].cjs.js',
```
montezume: What's this change for?

emmenko: For the split chunks

montezume:
I'm not really convinced the chunking is necessary considering how we only have 30 strings in 5 languages. It's pretty inconsequential.

.storybook/decorators/intl/intl.js
```diff
@@ -1,18 +1,23 @@
 import React from 'react';
 import { IntlProvider, addLocaleData } from 'react-intl';
-import { withKnobs, select } from '@storybook/addon-knobs/react';
+import { select } from '@storybook/addon-knobs/react';
 import en from 'react-intl/locale-data/en';
 import de from 'react-intl/locale-data/de';
 import es from 'react-intl/locale-data/es';
-import * as messages from '../../../i18n';
+import frFR from 'react-intl/locale-data/fr';
+import zhCN from 'react-intl/locale-data/zh';
```
montezume: this file is (minified) 11kb. If anything, I would chunk these before chunking our translations.

emmenko: I don't follow. The react-intl data does not get bundled, we do that only in app-kit

emmenko:
> Can you remove the chunk splitting and do it in another PR?

What's the reason?

montezume:
If I change the locales in the knob, it no longer works, even for locales we have translations for (DE and ES).

montezume:
Screen Shot 2019-04-04 at 4 47 47 PM
in prod

in this branch
Screen Shot 2019-04-04 at 4 46 16 PM

tdeekens:
If we do a breaking change with this, can we add the `as` prop renaming into that breaking change?

emmenko:
> If we do a breaking change with this, can we add the as prop renaming into that breaking change?

I thought we decided to keep it on hold for now (based on the meeting we had this morning) 🤔

tdeekens:
> I thought we decided to keep it on hold for now (based on the meeting we had this morning) 🤔

Back then nobody knew a breaking change was around the corner we could jump onto I suppose.

montezume:
Another approach to splitting the locales is to do as follows in the rollup config


in rollup.config.js
```js

    input: {
      index: 'src/index.js',
      messages: 'i18n/index.js',
    },
```

in src/index.js

remove the import and export of i18n

in your app

```js
import { en  } from '../../dist/ui-kit.esm/messages';
```

montezume:
1. I want to go towards having multiple entry points in this library.

2. I'm not sure how multiple entry points plays together with code splits.

If we are going to introduce code splits, I would like to investigate 2.

Does that make sense? It's not about bashing people over the head about breaking changes.

tdeekens:
In hindsight I appreciate @montezume point. Given the size (even though there is one) of the new translations we can split concerns and add translations in one and discuss code splitting in another. Also given that we might not even release inbetween.

emmenko:
Closing in favor of https://github.com/commercetools/ui-kit/pull/643

Thanks a lot for the feedback so far!

PR #640, opened 2019-04-04
idea: use multiple entry points instead of code splitting

montezume:
Here is an example of what I'm talking about in #639


montezume:
I would like to play around more with this, especially after reading [this](https://levelup.gitconnected.com/code-splitting-for-libraries-bundling-for-npm-with-rollup-1-0-2522c7437697)

emmenko:
This should not be relevant anymore, closing. See #643

PR #641, opened 2019-04-04
feat(tags): use design tokens, add theming

montezume:
#### Summary

Create new tokens to use for `Tags`. These tokens will also be used in the `SelectInputs` (for when isMulti is enabled).

Enable theming as for Tags.


materials/internals/definition.yaml
```diff
@@ -171,6 +171,10 @@ states:
 componentGroups:
   input:
     description: 'Input elements'
+  tag:
+    description: 'Tag elements'
+  tag-warning:
```
montezume: wasn't 100% sure on whether to allow `multiple states` or to put it as a separate component group;

montezume:
I think I made a mistake making the visual spec. I'll fix it tomorrow 😥

src/components/tag/tag.js
```diff
@@ -3,62 +3,88 @@ import { Link } from 'react-router-dom';
...
-  border-style: solid;
-  border-width: 1px;
-  border-color: ${props.type === 'warning'
-    ? vars.borderColorTagWarning
-    : vars.borderColorTagPristine};
+const getContentWrapperStyles = (props, theme) => {
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
+  };
```
emmenko: Q: do we allow all custom properties to be overwritten or should we only allow design tokens?

src/components/tag/tag.visualroute.js
```diff
@@ -63,10 +64,10 @@ export const component = () => (
         {longText}
       </Tag>
     </Spec>
-    <Spec label="Normal - onRemove (disabled)">
```
emmenko: Did you mean to remove this? 🤔

montezume: No, made a mistake 🙈

src/components/tag/tag.js
montezume: That's something I've thought about. The idea is that (for now), users can override `choices`. And since our designTokens use `choices`, that's kind of how it's working together for now.

However, there's some other variables that we could allow users to override as well, but I think it would make sense to rename them first.

#### UI Kit wide variables

example:
```
spacing8
```

These are often named in a way that would make overriding them weird. Spacing8 is set to 16? 😕 . We would probably want to rename these.

#### Component specific variables
These are sort of like "magic" numbers. If we allow users to override them, it should be documented in the individual components.

``
sizeHeightTag
standardInputHeight
```

Comment (montezume) [reply src/components/tag/tag.js]: #644 to track this

Comment (emmenko) [reply src/components/tag/tag.js]: Right, the spacings tokens should probably be something like `spacingsXs`, `spacingsS`, etc

File: package.json [hunk 4]
```diff
@@ -180,7 +180,7 @@
     "rollup-plugin-json": "4.0.0",
     "rollup-plugin-node-resolve": "4.0.1",
     "rollup-plugin-replace": "2.1.1",
-    "serve": "10.1.2",
+    "serve": "11.0.0",
```
montezume: the warning was annoying me

montezume:
I would like to merge this so I can get started on the select inputs. Can I get a review pretty please @emmenko @tdeekens

PR #649, opened 2019-04-05
Experiments using multiple entry points

emmenko:
## Experiment 1: multiple entry points

Inspired by https://levelup.gitconnected.com/code-splitting-for-libraries-bundling-for-npm-with-rollup-1-0-2522c7437697

We talked a couple of times about this. The idea is to provide multiple entry points, one for each component. This would help with treeshaking.

image

The entry points are dynamically generated, based on glob patterns.

## Experiment 2: use parcel bundler for VRT and use real imports

This relates a bit to the 1st experiment, more specifically to use and test the "single" entry points imports (you can see that all `*.visualroute.js` tests use that).
To do so we are packaging the ui-kit as a normal dependency in `node_modules`, so that we can import it as a normal dependency!


src/components/avatar/avatar.visualroute.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Avatar } from 'ui-kit';
+import Avatar from '@commercetools-frontend/ui-kit/dist/esm/Avatar';
```
emmenko: Consumers of ui-kit could do an import rewrite, like we do for lodash

```js
// Normal import
import Avatar from '@commercetools-frontend/ui-kit/Avatar'

// Transformed to
import Avatar from '@commercetools-frontend/ui-kit/dist/esm/Avatar'
```

tdeekens: Would we still over the combined entry?

emmenko: Actually the rewriting won't work, we should simply use the "full import path"

montezume:
Going to close this for now. We can reopen it later if we decide on this approach over mono repoeing.

PR #650, opened 2019-04-05
fix: Right border of removable tag

### LINKED ISSUE #638: Tag component hover fix
It's been requested from design that when you hover on a tag link component that is removeable, the entire border of the rectangle highlights. Now it only highlights the left, bottom and right.

jonnybel:
#### Summary

Fixes #638 

Kapture 2019-04-05 at 14 23 46
(GIF compression messing those colours 🧐)

#### Description

Using a regular `border` doesn't fit the design because the border takes space and would add a little edge or a flicker between the tag and the removal button.

The easiest alternative would be using `box-shadow` to emulate the border, just on the right side, but when the tag has an onClick prop, it has a box-shadow of its own when hovered, so this new box-shadow border would have to be merged with the current box-shadow and added to the design tokens.
I didn't like where this was going so I tried another approach:

So I used an absolutely positioned `:after` pseudo-element, and it does the job.

### DIFF
--- a/src/components/tag/tag.js
+++ b/src/components/tag/tag.js
@@ -266,7 +266,8 @@ const Tag = props => (
                 background: inherit;
                 border-style: solid;
                 border-width: 1px 1px 1px 1px;
-                &:hover {
+                &:hover,
+                &:focus {
                   border-color: ${overwrittenVars[
                     designTokens.borderColorForTagWarning
                   ]};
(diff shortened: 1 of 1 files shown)


jonnybel:
I've added a fix to the focus styles for the [x] button to improve its a11y.

src/components/tag/tag.js
```diff
@@ -247,7 +266,8 @@ const Tag = props => (
                 background: inherit;
                 border-style: solid;
                 border-width: 1px 1px 1px 1px;
-                &:hover {
+                &:hover,
```
montezume: 👍

src/components/tag/tag.js
```diff
@@ -283,10 +284,6 @@ const Tag = props => (
               `,
               props.isDisabled &&
                 css`
-                  &:hover {
```
jonnybel: This isn't needed because when a button is disabled in the HTML, the `hover` and `focus` states no longer apply.

PR #651, opened 2019-04-05
feat(text): allow to pass intl message instead of children

### LINKED ISSUE #637: [Idea] Allow Text components to accept an intl message
At the moment, we usually wrap the intl messages (e.g. `FormattedMessage`) with the `Text` components.

```js
<Text.Body>
  <FormattedMessage {...messages.title} />
</Text.Body>
```

Allow to pass the intl message directly to the `Text` component to avoid the wrapping

```js
<Text.Body intlMessage={messages.title} />

// the prop can also be named differently, e.g. `message` but we should probably make it explicit that it relates to react-intl
```

emmenko:
Fixes #637

### DIFF
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import requiredIf from 'react-required-if';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import {
   bodyStyles,
@@ -8,6 +10,12 @@ import {
   subheadlineStyles,
   wrapStyles,
 } from './text.styles';
+
+const intlMessageShape = PropTypes.shape({
+  id: PropTypes.string.isRequired,
+  description: PropTypes.string,
+  defaultMessage: PropTypes.string.isRequired,
+});
 
 const nonEmptyString = (props, propName, componentName) => {
   const value = props[propName];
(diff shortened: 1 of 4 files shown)


src/components/typography/text/text.js
```diff
@@ -63,9 +80,10 @@ Subheadline.propTypes = {
     'positive',
     'negative',
   ]),
-  children: PropTypes.node.isRequired,
   title: nonEmptyString,
   truncate: PropTypes.bool,
+  intlMessage: intlMessageShape,
```
tdeekens: Should this be flipped to the case below. Given there are no children then this?

montezume:
Can you add the new VRT's at the end to make reviewing easier 🙏

src/components/typography/text/text.js
emmenko: Ah thanks, good catch!

emmenko:
@montezume do you want to review/approve the VRT?

montezume:
@emmenko 👍 done

src/components/typography/text/text.js
```diff
@@ -26,17 +34,22 @@ const Headline = props => {
...
 };
 
 Headline.displayName = 'TextHeadline';
 Headline.propTypes = {
   elementType: PropTypes.oneOf(['h1', 'h2', 'h3']).isRequired,
-  children: PropTypes.node.isRequired,
   title: nonEmptyString,
   truncate: PropTypes.bool,
+  intlMessage: requiredIf(intlMessageShape, props => !props.children),
+  children: requiredIf(PropTypes.node, props => !props.intlMessage),
```
tdeekens: Thanks!

PR #661, opened 2019-04-09
fix: change high precision money display

montezume:
#### Summary

Currently, when a price is `highPrecision`, the MoneyField "jumps", as it is rendering a label above the field. This is not desired. Instead, we replace it with the use of an icon, and an accompanying tooltip in the MoneyInput.

#### Before

beforemf

#### After

aftermf

Also adds new `FractionDigitsIcon`


src/components/tooltip/README.md
```diff
@@ -111,3 +111,4 @@ In the following example, the tooltip text only appears on hover when the button
...
 | `children`             | `node`   |    ✅    | -                                                                                                                                            | -       | Content rendered within the tooltip                                                             |
 | `components`           | `object` |    -     | `WrapperComponent`, `BodyComponent`                                                                                                          | -       | If passed, the tooltip will wrap your component with this element                               |
+| `customBodyStyles`     | `object` |    -     | -                                                                                                                                            | -       | If passed, these styles will be spread onto the div surrounding the tooltip body                |
```
tdeekens: Not sure if I am overly picky with the name again. But given we tend to like the `components={{header, footer}}` pattern. How about going with `styles={ body }` here?

src/components/inputs/money-input/money-input.js
```diff
@@ -593,31 +607,81 @@ class MoneyInput extends React.Component {
...
+                    },
+                  });
+                this.setState({ amountHasFocus: true });
+              }}
+              value={this.props.value.amount}
+              css={[
+                getAmountInputStyles({ ...this.props, hasFocus }),
+                // accounts for size of icon
+                css`
+                  padding-right: 24px;
```
islam3zzat:
```suggestion
                  padding-right: ${vars.spacing24};
```

src/components/tooltip/README.md
```diff
@@ -111,3 +111,4 @@ In the following example, the tooltip text only appears on hover when the button
...
 | `children`             | `node`   |    ✅    | -                                                                                                                                            | -       | Content rendered within the tooltip                                                             |
 | `components`           | `object` |    -     | `WrapperComponent`, `BodyComponent`                                                                                                          | -       | If passed, the tooltip will wrap your component with this element                               |
+| `styles.body`          | `object` |    -     | -                                                                                                                                            | -       | If passed, these styles will be spread onto the div surrounding the tooltip body                |
```
islam3zzat: the proptypes says it's required.
Should we update the propTypes, or the readme?

src/components/inputs/money-input/money-input.js
montezume: 👍

src/components/tooltip/README.md
montezume: it's required because it's set with a default though, so for the user it's not required.

islam3zzat: thanks for the explanation

src/components/inputs/money-input/money-input.styles.js
```diff
@@ -29,4 +29,20 @@ const getAmountInputStyles = props => [
   `,
 ];
 
-export { getCurrencyLabelStyles, getAmountInputStyles };
+const getHighPrecisionWrapperStyles = props => css`
```
tdeekens: Nit just pass `isDisabled`? to be more specific? Not sure what the pattern is on UIKit.

montezume: 👍

PR #662, opened 2019-04-09
fix: disable sourcemaps when building storybook

montezume:
To try to help with netlify build issues.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,3 +1,4 @@
+const mri = require('mri');
 const path = require('path');
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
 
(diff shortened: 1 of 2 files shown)


tdeekens:
Update: sourcemaps add 30 seconds. Yay, guess was correct. 63 vs 35 secs.

.storybook/webpack.config.js
```diff
@@ -95,5 +106,7 @@ module.exports = ({ config }) => {
     },
   ];
 
+  config.optimization.minimizer[0].options.sourceMap = disableSourceMaps;
```
montezume: hm,  but we just set minimizer to empty array. How can we access this property now?

tdeekens: No. We keep the minimizer in my commit. We just disable source maps emission.

tdeekens: I intended to actually use terser but it to not emit any source maps. Fixed up my commit.

montezume:
Starting to think that any improvements we make is just rearranging deck chairs on the 🚢 . The plan is the problem, not our build 🤔

tdeekens:
Yeah feels like an angle not working or yielding enough. So either up the plan to see if that helps or maybe truely build on circleci. will have to think about it. maybe it can even help having a version switcher. it is a bigger leap obviously though.

PR #663, opened 2019-04-10
chore: internally use non deprecated decisions

montezume:
Just moves to using our non deprecating vars internally. If percy passes, it should be fine.

### DIFF
--- a/src/components/buttons/primary-button/primary-button.styles.js
+++ b/src/components/buttons/primary-button/primary-button.styles.js
@@ -5,7 +5,7 @@ const getButtonLayoutStyles = size => {
   const baseLayoutStyles = css`
     display: inline-flex;
     align-items: center;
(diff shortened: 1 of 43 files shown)


src/components/buttons/primary-button/primary-button.styles.js
```diff
@@ -15,7 +15,7 @@ const getButtonLayoutStyles = size => {
         css`
           border-radius: ${vars.borderRadius4};
           > button {
-            padding: 0 ${vars.spacing8} 0 ${vars.spacing8};
+            padding: 0 ${vars.spacingS} 0 ${vars.spacingS};
```
tdeekens: Yeeees!

montezume: Well I was hoping the hard part of the review could be done by Percy. 😄

PR #664, opened 2019-04-10
refactor: use tag in select input

montezume:
#### Summary

Instead of using custom styles to make React-Select's MultiValue components look like our Tag component, we can directly use our `Tag` component and remove the associated custom styling.


src/components/internals/create-select-styles.js
```diff
@@ -130,45 +130,6 @@ const valueContainerStyles = () => base => ({
   overflow: 'hidden',
 });
 
-const multiValueStyles = () => base => ({
```
montezume: can remove all of these now that we are directly using our tag component.

src/components/internals/multivalue/multivalue.js
```diff
@@ -0,0 +1,41 @@
...
+    <div
+      css={css`
+        display: inline-block;
+        margin: 2px;
+      `}
+    >
+      <Tag
+        styles={{
+          body: css`
+            padding: 4px 8px 3px 6px;
```
montezume: I'm not sure of a better way to do this. Basically our tag needs to be a little bit "slimmer", so that it doesn't make the select input 1 pixel taller 🤔

src/components/internals/multivalue/multivalue.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { css } from '@emotion/core';
+import PropTypes from 'prop-types';
+import Tag from '../../tag';
+
+const MultiValue = props => {
+  return (
```
tdeekens: This can be shortened I think to not return explicitely.

src/components/internals/multivalue/multivalue.js
jonnybel: Is it only because of this that you had to add that `styles.body` prop to the Tag?

tdeekens: Hm. Maybe do math on the vars instead so it's at least able to adjust?

montezume: @jonnybel yeah 😄

jonnybel: Is it possible that we can add a wrapper to the Tag, center the Tag inside it, and with `overflow: hidden` trim the excess?

montezume: wouldn't we lose the border in that case though? 🤔

montezume: another approach would just be to use the big Tag and let design approve it 🤔

src/components/internals/multivalue/multivalue.js
```diff
@@ -0,0 +1,39 @@
...
+  <div
+    css={css`
+      display: inline-block;
+      margin: 2px;
+    `}
+  >
+    <Tag
+      styles={{
+        body: css`
+          padding: 4px 8px 3px 6px;
```
tdeekens: If we don't want to do math based on current sizing vars we should maybe add a comment to say why we need this?

src/components/internals/multivalue/multivalue.js
jonnybel: Yeah you're right.

One question: is there a reason to strictly set the vertical paddings of the body of the Tag, instead of setting a fixed height and let the flexbox align the text to the center?

src/components/internals/multivalue/multivalue.js
montezume: I was going to come back to the math and or comment writing tomorrow. Sorry.

PR #667, opened 2019-04-11
fix: storybook build times

montezume:
I emailed netlify support about our build time issues.

They noticed that our builds have a lot of console logs. They suggested removing them.

`ProgressPlugin` spits out a lot of console logs. So let's disable it.


montezume:
Yay! It builds a lot faster now 😄

montezume:
Screen Shot 2019-04-11 at 1 25 51 PM
:

PR #669, opened 2019-04-11
fix: hint icon in field-label

### LINKED ISSUE #660: hintIcon doesn't change to whatever `size` I apply for `*Field`
**Describe the bug**

If I apply a hintIcon with my own `size` on an `*Field`, the `*Field` doesn't display what I want. This becomes apparent when I apply a multiline `hint`

e.g

```js
<SelectField
  hint={ /* ... multi line hint */}
  hintIcon={<WarningIcon size="big" theme="black" />}
/>
```

**To Reproduce**

For example:

1. Go to https://uikit.commercetools.com/?path=/story/components-fields--selectfield
2. Add a multiline hint (e.g _Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam_)

**Expected behaviour**

The `hintIcon` displays render specifically what I want

**Screenshots**

tiny ass warning-icon

**Additional context**

UI-kit imposes a constraint via `cloneElement` (ref: https://github.com/commercetools/ui-kit/blob/master/src/components/field-label/field-label.js#L40)

Since `hintIcon` is a component prop, we should remove this "fixed" constraint entirely. On top of removing the fixed constraint, let Design decide that there could a "limit" to the constraint (e.g no bigger than `size=big`)

cc @filippobocik

montezume:
Check the Percy diff for the fix.

Closes #660

### DIFF
--- a/src/components/field-label/field-label.js
+++ b/src/components/field-label/field-label.js
@@ -37,7 +37,7 @@ export const FieldLabel = props => (
           {props.hintIcon && (
             <div
               css={css`
-                display: inline-block;
+                display: flex;
               `}
             >
               {React.cloneElement(props.hintIcon, {
--- a/src/components/field-label/field-label.js
+++ b/src/components/field-label/field-label.js
@@ -33,13 +33,16 @@ export const FieldLabel = props => (
       </Spacings.Inline>
 
       {props.hint && (
-        <Spacings.Inline alignItems="center" scale="xs">
-          {props.hintIcon &&
-            // FIXME: add proper tone when tones are refactored
-            React.cloneElement(props.hintIcon, {
-              size: 'medium',
-              theme: props.hintIcon.props.theme || 'orange',
-            })}
+        <Spacings.Inline alignItems="center" scale="s">
+          {props.hintIcon && (
+            <div>
+              {React.cloneElement(props.hintIcon, {
+                // FIXME: add proper tone when tones are refactored
+                size: props.hintIcon.props.size,
+                theme: props.hintIcon.props.theme || 'orange',
+              })}
+            </div>
+          )}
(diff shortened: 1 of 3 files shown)


src/components/field-label/field-label.js
```diff
@@ -33,13 +33,16 @@ export const FieldLabel = props => (
...
-            React.cloneElement(props.hintIcon, {
-              size: 'medium',
-              theme: props.hintIcon.props.theme || 'orange',
-            })}
+        <Spacings.Inline alignItems="center" scale="s">
+          {props.hintIcon && (
+            <div>
+              {React.cloneElement(props.hintIcon, {
+                // FIXME: add proper tone when tones are refactored
+                size: props.hintIcon.props.size,
```
tdeekens: Default size to be `medium`? Otherwise it's a breaking change - ain't it?

montezume: Oh yeah, and Percy caught that as well 😄

montezume:
Visual diff is showing a diff, I guess because of the addition of the div. I'll check out a fix tomorrow

src/components/field-label/field-label.js
```diff
@@ -33,13 +33,20 @@ export const FieldLabel = props => (
...
+        <Spacings.Inline alignItems="center" scale="s">
+          {props.hintIcon && (
+            <div
+              css={css`
+                display: flex;
+              `}
+            >
+              {React.cloneElement(props.hintIcon, {
+                // FIXME: add proper tone when tones are refactored
+                size: props.hintIcon.props.size || 'medium',
```
adnasa: I was supposed to talk to you first about this...
@filippobocik decided against allowing consumers to allow this.

PR #670, opened 2019-04-11
chore(circleci): use manual env variable checks for releae

tdeekens:
#### Summary

This pull request aims to fix the release issue we're experiencing.

We currently can't manage to get the general release config suggested by CircleCI to work:

```yaml
filters:
  tags:
  	only: /^v.*/
  branches:
    ignore: /.*/
```

Never runs. While

```yaml
filters:
  tags:
  	only: /^v.*/
```

Always runs - even without a tag. Leaving us in a state of confusion. 

We can however attempt to replicate the filters with the new CircleCI 2.1 config without much bloat.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,11 +96,11 @@ commands:
           command: |
             if [ ${<< parameters.git_branch >>} != "master" ]; then
               echo "Git branch is ${<< parameters.git_branch >>}, not publishing."
-              exit 0
+              circleci-agent step halt
             fi
             else if [ -z ${<< parameters.git_tag >>}]; then
               echo "Git tag is not set, not publishing."
-              exit 0
+              circleci-agent step halt
             fi
             else
               echo "Git branch is ${<< parameters.git_branch >>} and tag is ${<< parameters.git_tag >>}, publishing."
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,30 +91,27 @@ commands:
         type: string
         default: *working_directory
     steps:
-      - when:
-          condition: << parameters.git_tag >>
-          steps:
-            - run:
-                name: Checking if on master branch
-                command: |
-                  if [ "<< parameters.git_branch >>" != "master" ]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -81,20 +81,40 @@ commands:
       - run: ./scripts/run-percy.sh
   publish_to_next:
     parameters:
+      git_branch:
+        type: env_var_name
+        default: CIRCLE_BRANCH
+      git_tag:
+        type: env_var_name
```
tdeekens: It's a CirclCI type so that the name as to be some POSIX compliant thing.

.circleci/config.yml
```diff
@@ -81,20 +81,40 @@ commands:
...
-          name: Authenticating with npm registry
-          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > << parameters.working_directory >>/.npmrc
-      - run:
-          name: Publishing to next dist-tag
-          command: yarn release:next
+      - when:
+          condition: << parameters.git_tag >>
+          steps:
+            - run:
+                name: Checking if on master branch
```
tdeekens: We exit early when we're not on master as a branch should always be set.

.circleci/config.yml
```diff
@@ -81,20 +81,40 @@ commands:
...
-      - install_dependencies
-      - build_artifacts
-      - run:
-          name: Authenticating with npm registry
-          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > << parameters.working_directory >>/.npmrc
-      - run:
-          name: Publishing to next dist-tag
-          command: yarn release:next
+      - when:
+          condition: << parameters.git_tag >>
```
tdeekens: The tag is optional so we can actuall just use it checking it's existence.

tdeekens: That didn't work.

tdeekens:
Yeeees!

CleanShot 2019-04-11 at 18 20 30@2x

This should work now.

montezume:
Awesome!

PR #671, opened 2019-04-12
feat: allow theming for text

montezume:
#### Summary

Makes Text components themeable.

### DIFF
--- a/src/components/label/label.visualroute.js
+++ b/src/components/label/label.visualroute.js
@@ -18,9 +18,11 @@ export const component = ({ themes }) => (
         Hello
       </Label>
     </Spec>
-    <ThemeProvider theme={themes.darkTheme}>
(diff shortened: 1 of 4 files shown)


src/components/label/label.visualroute.js
```diff
@@ -18,9 +18,11 @@ export const component = ({ themes }) => (
         Hello
       </Label>
     </Spec>
-    <ThemeProvider theme={themes.darkTheme}>
-      <Spec label="when inverted">
-        <Label tone="inverted">Hello</Label>
+    <ThemeProvider theme={{ ...themes.darkTheme, colorSurface: 'purple' }}>
```
montezume: because the spec uses `colorSurface` as the background color, we need to use two themeproviders so we don't render black text on black background.

PR #680, opened 2019-04-15
chore: revert changes in field label pr

montezume:
@adnasa informed me that the desired behaviour is to force the 'medium' save for field labels. As such, I'm reverting all the changes made in #669

### DIFF
--- a/src/components/field-label/field-label.js
+++ b/src/components/field-label/field-label.js
@@ -33,20 +33,13 @@ export const FieldLabel = props => (
       </Spacings.Inline>
 
       {props.hint && (
-        <Spacings.Inline alignItems="center" scale="s">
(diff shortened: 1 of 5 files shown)


src/components/field-label/field-label.js
```diff
@@ -9,66 +10,67 @@ import Label from '../label';
...
-
-      {props.hint && (
-        <Spacings.Inline alignItems="center" scale="s">
-          {props.hintIcon && (
-            <div
-              css={css`
-                display: flex;
-              `}
+export const FieldLabel = props => {
+  if (props.hintIcon && props.hintIcon.props && props.hintIcon.props.size) {
```
montezume: if this is the desired behaviour, we should warn when users try to set the size.

PR #688, opened 2019-04-16
chore: re-add tag filter for circleci to run publish job on tag pushes

tdeekens:
#### Summary

To trigger running on all tags while the release script is smart enough to bail if needed.


tdeekens:
Netlify seems to be using node v8.15.1 (npm v6.4.1) while conventional-changelog-conventionalcommits@1.1.1 requires ">=10.0.0". Did you update that dep @montezume?

tdeekens:
CleanShot 2019-04-16 at 16 49 38@2x

Added an `.nvmrc` with 11.

tdeekens:
The `.nvmrc` seems to do the trick for Netlify and I think is generally a good idea. It's now using node v11.14.0 (npm v6.7.0).

PR #689, opened 2019-04-16
fix: releasing from master and release/*

tdeekens:
#### Summary

This pull request aims to experiment a bit more with the release process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,10 @@ workflows:
       - publish:
           filters:
             tags:
-              only: /^v.*/
+              only: /v[\.0-9]+.*/
+            branches:
+              only:
+                - master
+                - /release\/.*/
           requires:
             - bundle_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,17 +146,28 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - lint_and_test
+      - lint_and_test:
+          filters:
+            tags:
+              only: /v[\.0-9]+.*/
       - build:
           requires:
             - lint_and_test
+          filters:
+            tags:
+              only: /v[\.0-9]+.*/
       - bundle_test:
           requires:
             - build
+          filters:
+            tags:
+              only: /v[\.0-9]+.*/
       - vrt_test:
           requires:
             - build
       - publish:
+          requires:
+            - bundle_test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -163,6 +163,10 @@ workflows:
       - publish:
           filters:
             tags:
-              only: /^v.*/
+              only: /v[\.0-9]+.*/
```
tdeekens: A bit more explicit regex matching semver more.

.circleci/config.yml
```diff
@@ -163,6 +163,10 @@ workflows:
...
+            branches:
+              only:
+                - master
+                - /release\/.*/
```
tdeekens: A little escape hatch and also to have better releasing from branches if needed.

tdeekens:
Okay, I just pushed all variations and tags (and deleted them again) to figure it out (while having publishing halted by the circleci-agent).

The docs are really freakish about this issue and everybody is in utter state of confusion. Essentially you:

1. Need to mark the tag filters all the way up as otherwise the jobs dependencies are not satisfied (even if you push a tag) and it won't run
2. You need to specify the branch dependency on the job too and check the tag again as otherwise it runs too often

Furthermore tags are located at the root of the project so [here](https://circleci.com/gh/commercetools/ui-kit) and can be easily missed. There you see a build for 10.0.0.

Screenshot 2019-04-16 at 20 36 12

Lastly, we can not check for the branch again as _either_ branch of tag are set as CircleCI variables (this I picked up on another discussion in their forum).

Screenshot 2019-04-16 at 20 37 35

In all I think we can give this another try. I am pretty sure this makes more sense then what we had before.

.circleci/config.yml
```diff
@@ -150,19 +146,32 @@ workflows:
...
+              only: /v[\.0-9]+.*/
       - vrt_test:
           requires:
             - build
       - publish:
-          filters:
-            tags:
-              only: /^v.*/
           requires:
             - bundle_test
```
montezume: ah so bundle test needs filters because publish requires bundle test

tdeekens: Yes, and that game bubbles all the way up.

tdeekens: Given you do that you get a run with only those jobs when pushing a tag. The outcome it quite nice.

PR #692, opened 2019-04-17
fix(select-input): fix for prop type

montezume:
#### Summary

We can't use `PropTypes.element` because that defines a react element.

We can't use `Element` because it's undefined in node and crashes the MC at compile time.

More info [here](https://github.com/reactjs/react-modal/issues/15)

### DIFF
--- a/src/utils/get-element.js
+++ b/src/utils/get-element.js
@@ -1,0 +1,3 @@
+const getElement = () => (typeof Element === 'undefined' ? () => {} : Element);
+
+export default getElement;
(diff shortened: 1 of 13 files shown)


src/utils/get-element.js
```diff
@@ -0,0 +1,3 @@
+const getElement = () => (typeof Element === 'undefined' ? () => {} : Element);
+
+export default getElement;
```
emmenko: react-modal uses this: https://github.com/reactjs/react-modal/blob/master/src/helpers/safeHTMLElement.js

We can do the same maybe

montezume: thanks, [4e261485b9ffe6ffcf75e49f60441c2a86363816](https://github.com/commercetools/ui-kit/pull/692/commits/4e261485b9ffe6ffcf75e49f60441c2a86363816)

PR #696, opened 2019-04-18
feat: add `autoComplete` to most inputs / fields.

montezume:
#### Summary

Adds prop `autoComplete` to fields and inputs.


montezume:
@tdeekens but modern browsers accept a whole whack of values for autocomplete...
https://stackoverflow.com/questions/34712926/how-browsers-store-data-for-autocomplete-and-where/34776932#34776932

src/components/fields/select-field/select-field.form.story.js
```diff
@@ -96,7 +96,8 @@ storiesOf('Examples|Forms/Fields', module)
                 errors={formik.errors.animal}
                 isRequired={true}
                 touched={formik.touched.animal}
-                name="animal"
+                name="name"
+                isSearchable={true}
```
jonnybel: Is this supposed to be here?

jonnybel:
This is the type of prop that I'd personally prefer to spread directly to the HTML element in a `remainingProps` fashion, because we're basically just adding more code, tests and documentation for stuff that is already native to the HTML, and seems very unnecessary. (Other examples are props like isDisabled, readOnly, autoFocus, etc, as well as mouse events and whatnot)

But I know this would require a complete refactor of all inputs/fields (and even other components in our UI-kit), so I just wanted to state this for consideration.

That being said, nothing wrong with this PR in particular considering how we're doing things so far 😄

src/components/fields/select-field/select-field.form.story.js
montezume: Nice catch

montezume:
@jonnybel really good point. It just adds noise to the component and especially to the readme the way it is now...

montezume:
@jonnybel do you want to open an issue and list the props that should be whitelisted for this spreading behaviour? I guess aria- props, etc...

jonnybel:
Yes, I can do that.

PR #708, opened 2019-04-25
refactoring(toggle-input): change to new design

montezume:
#### Summary

Refactoring to the Toggles to match the new designs. Still waiting on some design tokens, so the css isn't exactly 100% set, but it's ready for review.

#### Approach

The old Tooltip was built with two SVGs. One SVG for the tooltip "track", and one for the "button". I couldn't really figure out why, it made it harder to style, harder to reason about, and harder to apply things like specific border radiuses without doing a bunch of ... math. I replaced the SVGs with divs instead.


montezume:
Waiting on new shadows from design

src/components/inputs/toggle-input/toggle-input.styles.js
```diff
@@ -1,185 +1,138 @@
...
+      filter: drop-shadow(${vars.shadow15First})
+        drop-shadow(${vars.shadow15Second});
+    }
+    `}
+`;
+
+export const Track = styled.div`
+  position: absolute;
+  left: 0;
+  border-radius: 16px;
```
montezume: Not sure if we should make this a customProperty or not.

src/components/inputs/toggle-input/toggle-input.styles.js
```diff
@@ -1,185 +1,138 @@
...
+  box-shadow: ${vars.shadow9};
```
montezume: really not sure about the box shadows

src/components/inputs/toggle-input/toggle-input.styles.js
```diff
@@ -1,185 +1,138 @@
...
-      position: absolute;
-      left: 0;
-      rect {
-        fill: ${vars.colorNeutral60};
-      }
-    `,
+  ${props =>
+    !props.isDisabled &&
+    `
+    &:hover ${Button} {
```
montezume: this is a cool feature

src/components/inputs/toggle-input/toggle-input.styles.js
```diff
@@ -1,185 +1,138 @@
...
-      }
-    `,
+  transform: translateY(-50%);
+  top: 50%;
+
+  background-color: ${vars.colorSurface};
+  box-shadow: ${vars.shadow7};
+  border-radius: 50%;
+  z-index: 1;
+  transition: transform ${vars.transitionStandard};
```
jonnybel: This doesn't work, because `transitionStandard` already has a property name (`all`), resulting in an invalid rule. So it should just be `transform`.

Not related to this PR, but `transitionStandard` shouldn't have the property specified, and especially not `all`, as it's a very bad practice because of how it can hurt performance.

jonnybel:
The input/button is currently not accessible through the keyboard, but since this is also a problem with a bunch of other UI-kit components, I'm ok with merging this as-is without a11y for now and work on that as a follow-up or as part of a (hopefully upcoming) effort to improve the overall a11y of the UI-kit.

montezume:
Update: based on @jonnybel feedback, I'm refactoring this to work with keyboard control 🎉

src/components/inputs/toggle-input/toggle-input.js
```diff
@@ -1,8 +1,100 @@
...
+    background-color: ${vars.colorSurface};
+    box-shadow: ${vars.shadow7};
+    border-radius: 50%;
+    z-index: 1;
+    transition: transform 0.2s ease, background 0.2s ease;
+  }
+`;
+
+const Input = styled.input`
+  pointer-events: none;
```
montezume: this is just a way of making the input hidden, but still accessible

montezume:
Okay Toggle is now accessible 🎉 . I sort of would prefer if we could make it look like this though under focus, would make it a lot easier to use with the keyboard...

Screen Shot 2019-04-26 at 5 25 00 PM

jonnybel:
Yes, it's very important for a11y to have visual feedback of keyboard focus, only then it is _officially_ accessible...

Usually, that's done with the `outline` property, which we can even colourize, but not make it round... So instead we can use a `box-shadow` on the Track, styled to look like the outline, but rounded.

montezume:
I guess it's also something that could be done as a followup, since it's not a breaking change, and we don't currently support accessibility 🤔 . WDYT @emmenko ?

emmenko:
Yeah we should discuss with design how we want to visualize the "focus" state. That should be a follow up and we should probably open an issue to track it.

montezume:
Okay so, from a design point of view we are just waiting on "new shadows". That means we can get some code reviews. Based on the discussion Friday, I'm rolling the dice and tagging @vidurar @amine-benselim

montezume:
I opened an issue to track accessibility #717

src/components/inputs/toggle-input/toggle-input.js
```diff
@@ -1,8 +1,100 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { css } from '@emotion/core';
-import ToggleSwitch from './toggle-switch';
+import styled from '@emotion/styled';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
+import vars from '../../../../materials/custom-properties';
+
+const thumbSmallSize = '13px';
+const thumbBigSize = '26px;';
```
amine-benselim: are we fine with hardcoded values ?

tdeekens: The might have been before. The track and thumb sizes are not bit tricky.

src/components/inputs/toggle-input/toggle-input.js
```diff
@@ -1,8 +1,100 @@
...
+  ${props =>
+    props.size === 'small'
+      ? `
+        height: 16px;
+        width: 32px;
+      `
+      : `
+
+        height: 32px;
+        width: 64px;
```
tdeekens: Can we maybe "correlate" those sizes? Like in L7 have `thumbSmallSize = 13. Then `thumbBigSize = thumbSmallSize * 2` and then move this up as `labelSize = 16` while multiplying width and height again while everywhere appending the `px`. So that it's clearer how sizes relate.

src/components/inputs/toggle-input/toggle-input.js
montezume: Yeah, I guess we could also use something derived from 

```
   standardInputHeight: 32px;
```

🤔

src/components/inputs/toggle-input/toggle-input.js
montezume: Yeah, that's a good improvement. Thanks 😄

franzi-t:
> Yeah we should discuss with design how we want to visualize the "focus" state. That should be a follow up and we should probably open an issue to track it.
> I opened an issue to track accessibility #717

@montezume thanks! I agree, it makes sense to have this accessibility issue as a separate task to find a uniform way of showing the focus for all inputs and fields.

PR #709, opened 2019-04-25
fix: select-input cursor when disabled

montezume:
#### Summary

Currently, `cursor: not-allowed` doesn't work with SelectInputs. 

This fixes that.

### DIFF
--- a/src/components/internals/create-select-styles.js
+++ b/src/components/internals/create-select-styles.js
@@ -31,19 +31,24 @@ const controlStyles = props => (base, state) => ({
   boxShadow: state.isFocused ? 'none' : base.boxShadow,
 
   '&:hover': {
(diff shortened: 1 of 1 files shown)


src/components/internals/create-select-styles.js
```diff
@@ -31,19 +31,24 @@ const controlStyles = props => (base, state) => ({
...
   },
   '&:focus': {
-    borderColor: vars.borderColorForInputWhenFocused,
+    borderColor: state.isDisabled
+      ? vars.borderColorForInputWhenDisabled
+      : vars.borderColorForInputWhenFocused,
     boxShadow: 'none',
   },
-
-  pointerEvents: state.isDisabled ? 'none' : base.pointerEvents,
```
montezume: no pointer events means we can't have cursor: not-allowed.

PR #718, opened 2019-04-29
refactor(checkbox-input): accessibility concerns

montezume:
#### Summary

This PR makes two changes to the `CheckboxInput`. Neither change affects the design of the component.

1. When the `Input` is focused, we treat it the same as the hovered state. This matches the behaviour in most of our `Input` components, including (but not limited to) `TextInput`, `LocalizedTextInput`, `MoneyInput`, etc, etc.

2. When the `CheckboxInput` is in an indeterminate state, we render it with this HTML property, to help screenreaders.

### DIFF
--- a/src/components/inputs/checkbox-input/checkbox-input.js
+++ b/src/components/inputs/checkbox-input/checkbox-input.js
@@ -18,9 +18,11 @@ const Label = styled.label`
   cursor: ${props => (props.isDisabled ? 'not-allowed' : 'pointer')};
   position: relative;
 
-  &:hover svg [id$='borderAndContent'] > [id$='border'] {
+  ${props =>
+    !props.hasError &&
+    `  &:hover svg [id$='borderAndContent'] > [id$='border'] {
     stroke: ${vars.borderColorInputFocus};
-  }
+  }`}
 `;
 
 class CheckboxInput extends React.PureComponent {
@@ -61,7 +63,7 @@ class CheckboxInput extends React.PureComponent {
 
   render() {
     return (
-      <Label htmlFor={this.state.id}>
+      <Label htmlFor={this.state.id} hasError={this.props.hasError}>
         <Spacings.Inline alignItems="center">
           <Checkbox
             type="checkbox"
@@ -72,6 +74,7 @@ class CheckboxInput extends React.PureComponent {
             disabled={this.props.isDisabled}
             checked={this.props.isChecked && !this.props.isIndeterminate}
             isIndeterminate={this.props.isIndeterminate}
+            hasError={this.props.hasError}
             {...filterDataAttributes(this.props)}
           />
 
--- a/src/components/inputs/checkbox-input/checkbox-input.js
+++ b/src/components/inputs/checkbox-input/checkbox-input.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
+import { css } from '@emotion/core';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import getFieldId from '../../../utils/get-field-id';
 import createSequentialId from '../../../utils/create-sequential-id';
 import Text from '../../typography/text';
-import Spacings from '../../spacings';
 import Icons from './icons';
 import { getCheckboxWrapperStyles } from './checkbox-input.styles';
 import Checkbox from './checkbox';
@@ -14,7 +14,8 @@ import vars from '../../../../materials/custom-properties';
 const sequentialId = createSequentialId('checkbox-input-');
 
 const Label = styled.label`
-  display: block;
+  display: flex;
+  align-items: center;
   cursor: ${props => (props.isDisabled ? 'not-allowed' : 'pointer')};
   position: relative;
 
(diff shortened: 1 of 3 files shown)


src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -52,40 +63,38 @@ class CheckboxInput extends React.PureComponent {
...
+        <Spacings.Inline alignItems="center">
+          <Checkbox
+            type="checkbox"
+            id={this.state.id}
+            name={this.props.name}
+            value={this.props.value}
+            onChange={this.props.onChange}
+            disabled={this.props.isDisabled}
+            checked={this.props.isChecked && !this.props.isIndeterminate}
+            isIndeterminate={this.props.isIndeterminate}
```
adnasa: 👍

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -52,40 +63,38 @@ class CheckboxInput extends React.PureComponent {
...
+            hasError={this.props.hasError}
+            {...filterDataAttributes(this.props)}
+          />
+
+          <div css={getCheckboxWrapperStyles(this.props)}>
+            {(() => {
```
adnasa: why not `do` expression?

montezume: We decided not to use `do` expressions in UI Kit at an earlier date. There's probably a discussion about it somewhere.

emmenko: We don't use those features for uikit and appkit

montezume: I would like to remove the use of our Icons here and use PURE css. But it's a bit more work. So maybe a followup.

src/components/inputs/checkbox-input/checkbox.js
```diff
@@ -0,0 +1,50 @@
...
+
+  componentDidMount() {
+    if (this.props.isIndeterminate) {
+      this.ref.current.indeterminate = true;
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.isIndeterminate !== this.props.isIndeterminate) {
+      this.ref.current.indeterminate = this.props.isIndeterminate;
```
montezume: for some reason in dev tools you don't actually see the property in the dom, but if you change the opacity of the input itself, you can see that it is treated properly 👍

src/components/inputs/checkbox-input/checkbox-input.js
adnasa: > There's probably a discussion about it somewhere.

oh my bad. carry on :)

src/components/inputs/checkbox-input/checkbox.js
```diff
@@ -0,0 +1,50 @@
...
+  &:focus + div > svg [id$='borderAndContent'] > [id$='border'] {
+    stroke: ${vars.borderColorInputFocus};
+  }`}
+`;
+
+class Checkbox extends React.Component {
+  static displayName = 'Checkbox';
+
+  static propTypes = {
+    isIndeterminate: PropTypes.bool,
```
tdeekens: The forwarded prop-types are not noted here. Is that on purpose?

src/components/inputs/checkbox-input/checkbox.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
tdeekens: Why is the module called `checkbox.js` and not `checkbox-input.js`?

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -52,40 +63,38 @@ class CheckboxInput extends React.PureComponent {
...
+      <Label htmlFor={this.state.id} hasError={this.props.hasError}>
+        <Spacings.Inline alignItems="center">
+          <Checkbox
+            type="checkbox"
+            id={this.state.id}
+            name={this.props.name}
+            value={this.props.value}
+            onChange={this.props.onChange}
+            disabled={this.props.isDisabled}
+            checked={this.props.isChecked && !this.props.isIndeterminate}
```
tdeekens: Why do we remap the props here already to "native" DOM input props? Can't we pass them to our checkbox, maintaining our naming convention and then explicitely pass them on to the input?

src/components/inputs/checkbox-input/checkbox.js
montezume: 🤣 because the other file is called 'checkbox-input.js'

src/components/inputs/checkbox-input/checkbox-input.js
montezume: yeah sure 👍

src/components/inputs/checkbox-input/checkbox.js
montezume: Well the idea at the beginning was just to make a super simple component that was just an `<input>` that passed indeterminate state. But yeah, I can refactor it

src/components/inputs/checkbox-input/checkbox.js
tdeekens: Ah right. Thanks

src/components/inputs/checkbox-input/checkbox.js
tdeekens: Yeah I see that. But I think it's two things to 1. Have components with inconsistent API and 2. Components which forward any native DOM props.

montezume: yeah that makes sense, [here](https://github.com/commercetools/ui-kit/pull/718/commits/a6d0a9cd081f35e3364b6286ab97a471c280a015)

PR #719, opened 2019-04-29
refactoring(css): move `accessible-input-styles` to internals

montezume:
#### Summary

As we will be using these styles in `RadioOptions`, `Checkbox` and `Toggle`, it makes sense not to reinvent the wheel 3 times.


src/components/inputs/checkbox-input/checkbox.js
```diff
@@ -52,6 +45,7 @@ class Checkbox extends React.Component {
   render() {
     return (
       <Input
+        css={accessibleHiddenInputStyles}
```
montezume: i kind of like how you can do this 😆

PR #722, opened 2019-04-30
refactor(checkbox-input): accessibility fix

montezume:
Follow up to #721. Based on discussion with @mariabarrena 

Now when a checkbox has error state, we still give it the green hover, and same for "focus".


montezume:
Sure. Let's say we have a simple webform, where you enter your `name`, `email`, and you a checkbox to determine if you read the terms and agreements.

If you submit the form empty, we should show 3 errors right? name, email and the checkbox. And you should be able to navigate to the fields with your keyboard, and fill them in. So the checkbox should be like the rest, and should give you a **visual indication** that is it the input under focus.

montezume:
Hm, maybe it was premature to make the change for hover as well thought

tdeekens:
> Hm, maybe it was premature to make the change for hover as well thought

I meant hover yes.

PR #724, opened 2019-05-02
fix(time-input): readonly fix

### LINKED ISSUE #723: TimeInput doesn't support `isReadOnly`
Related to #695.

montezume:
This PRd does two things

* Properly support the `isReadOnly` prop for the `TimeInput` component.
* Refactor `TimeInput` component to use design tokens and to be themeable. 

NOTE: Until we merge `10.0.0`, a themed TimeInput looks a bit weird because the icons aren't themable. 

Closes #723

### DIFF
--- a/src/components/inputs/time-input/time-input-body.js
+++ b/src/components/inputs/time-input/time-input-body.js
@@ -75,7 +75,7 @@ export default class TimeInputBody extends React.Component {
             id={this.props.id}
             name={this.props.name}
             autoComplete={this.props.autoComplete}
-            css={getTimeInputStyles(this.props)}
+            css={theme => getTimeInputStyles(this.props, theme)}
             placeholder={this.props.placeholder}
             autoFocus={this.props.isAutofocussed}
             disabled={this.props.isDisabled}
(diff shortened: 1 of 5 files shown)


montezume:
@tdeekens I opened it as a draft so I could see the Percy diff and finish it up. 😢

src/components/inputs/time-input/time-input-body.js
```diff
@@ -78,17 +79,21 @@ export default class TimeInputBody extends React.Component {
...
             disabled={this.props.isDisabled}
+            readOnly={this.props.isReadOnly}
             value={this.props.value}
             onChange={this.props.onChange}
             onFocus={this.props.onFocus}
             onBlur={this.props.onBlur}
             {...filterDataAttributes(this.props)}
             /* ARIA */
             role="textbox"
+            aria-readonly={this.props.isReadOnly}
```
montezume: same as we do in the TextInput

src/components/inputs/time-input/time-input-body.js
```diff
@@ -20,7 +20,7 @@ const getIconTheme = (isDisabled, isMouseOver) => {
 export const ClearSection = props => (
   <div
     onClick={props.isDisabled || props.isReadOnly ? undefined : props.onClear}
-    css={getClearSectionStyles(props)}
+    css={theme => getClearSectionStyles(props, theme)}
```
tdeekens: If only if was curried :)

montezume: at least I have VRT to catch my mistakes 🙈

montezume:
Just need to refactor the icons to use colors from css instead of Icon Themes to make theming work

montezume:
Actually a better way would just be to support Icon Theming.

PR #726, opened 2019-05-02
feat(icons): make icons themeable (breaking change)

montezume:
## Summary

As we discussed on slack, in order to make the icons themeable, we need to do something about the `theme` prop. When you use a styled component in emotion, it injects its `theme` prop, which we were overwriting since we had an identically named prop. 

### 💣 BREAKING CHANGES 💣

* Renames the prop `theme` to `color`. Renames it's possible values from being our old colour decisions (white, green, etc) to our new colour decisions (surface, primary, etc).

### Why do we need themeable icons? 

Components that use icons, like for example, the `TimeInput` need this to maintain consistency. 

#### Why are you spelling colour wrong? 

😭 

This will be merged into the `10.0.0` branch. The PR is set to master now so that Percy runs.

### DIFF
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -22,13 +22,13 @@ const getIconThemeColor = props => {
   // if button has a theme, icon should be white when hovering/clicking
   if (props.theme !== 'default' && (isActive || props.isMouseOver)) {
     if (props.isDisabled) {
-      return 'grey';
+      return 'neutral60';
     }
-    return 'white';
+    return 'surface';
   }
 
   // if button is disabled, icon should be grey
-  if (props.isDisabled) return 'grey';
+  if (props.isDisabled) return 'neutral60';
   // if button is not disabled nor has a theme, return icon's default color
   return props.icon.props.theme;
 };
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -19,7 +19,7 @@ import {
 // Gets the color which the icon should have based on context of button's state/cursor behavior
 const getIconThemeColor = props => {
   const isActive = props.isToggleButton && props.isToggled;
-  // if button has a theme, icon should be white when hovering/clicking
+  // if button has a theme, icon should be surface when hovering/clicking
   if (props.theme !== 'default' && (isActive || props.isMouseOver)) {
     if (props.isDisabled) {
       return 'neutral60';
@@ -27,7 +27,7 @@ const getIconThemeColor = props => {
     return 'surface';
   }
 
-  // if button is disabled, icon should be grey
+  // if button is disabled, icon should be neutral60
   if (props.isDisabled) return 'neutral60';
   // if button is not disabled nor has a theme, return icon's default color
   return props.icon.props.theme;
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -84,7 +84,7 @@ export const IconButton = props => {
           {props.icon &&
             React.cloneElement(props.icon, {
               size: props.size,
-              theme: getIconThemeColor(props),
+              color: getIconThemeColor(props),
             })}
         </div>
       </AccessibleButton>
--- a/src/components/field-label/README.md
+++ b/src/components/field-label/README.md
@@ -39,7 +39,7 @@ The `hintIcon` also accepts a custom `theme` while defaulting to `orange` in the
   onInfoButtonClick={() => {}} />}
   hint={<FormattedMessage {...messages.hint} />}
 - hintIcon={<WarningIcon />}
-+ hintIcon={<WarningIcon theme="green" />}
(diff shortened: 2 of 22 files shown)


src/components/field-label/README.md
```diff
@@ -39,7 +39,7 @@ The `hintIcon` also accepts a custom `theme` while defaulting to `orange` in the
   onInfoButtonClick={() => {}} />}
   hint={<FormattedMessage {...messages.hint} />}
 - hintIcon={<WarningIcon />}
-+ hintIcon={<WarningIcon theme="green" />}
++ hintIcon={<WarningIcon color="green" />}
```
emmenko:
```suggestion
+ hintIcon={<WarningIcon color="primary" />}
```

src/components/icons/README.md
```diff
@@ -19,10 +19,10 @@ You can find a list of all available icons in the UIKit.
...
-| `theme` | `string` |          | 'black', 'grey', 'blue', 'green', 'orange', 'red' | 'black' | Specifies the icon theme                                                                               |
+| Props   | Type     | Required | Values                                                        | Default | Description                                                                                            |
+| ------- | -------- | :------: | ------------------------------------------------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| `size`  | `string` |          | 'small', 'medium', 'big', 'scale'                             | 'big'   | Specifies the icon size (if `scale` is selected, the dimensions will scale according with the parents) |
+| `color` | `string` |          | 'solid', 'neutral60', 'info', 'primary40', 'warning', 'error' | 'solid' | Specifies the icon color                                                                               |
```
emmenko:
```suggestion
| `color` | `string` |          | 'solid', 'neutral60', 'info', 'primary', 'primary40', 'warning', 'error' | 'solid' | Specifies the icon color                                                                               |
```

src/components/icons/create-styled-icon.js
```diff
@@ -33,29 +33,33 @@ const getSizeStyle = size => {
...
+    case 'info':
+      return overwrittenVars.colorInfo;
+    case 'primary':
+      return overwrittenVars.colorPrimary;
+    case 'primary40':
+      return overwrittenVars.colorPrimary40;
+    case 'warning':
+      return overwrittenVars.colorWarning;
+    case 'error':
+      return overwrittenVars.colorError;
```
tdeekens: Wondering if this could be more of a convention approach instead of switching?

src/components/panels/collapsible-panel/header-icon.js
```diff
@@ -7,10 +7,10 @@ import vars from '../../../../materials/custom-properties';
 const sizeIconContainer = '24px';
 const sizeIconContainerSmall = '16px';
 
-const getArrowTheme = ({ tone, isDisabled }) => {
-  if (isDisabled) return 'grey';
-  if (tone === 'urgent') return 'white';
-  return 'black';
+const getArrowColor = ({ tone, isDisabled }) => {
```
tdeekens: Good rename!

montezume:
I think the Percy diff it's a combination of breaking things and of the icon VRTs (7 I think) have different labels so therefore have a diff. But yeah, must be some broken things. I'll fix up the broken things tomorrow 😄

montezume:
@tdeekens I think that a v10 branch is a really good idea 👍

montezume:
Ah Percy doesn't run if I have it aimed against a different branch. I guess that makes sense.

src/components/icons/create-styled-icon.js
```diff
@@ -33,55 +33,48 @@ const getSizeStyle = size => {
...
+const capitalize = s => s[0].toUpperCase() + s.slice(1);
+
+const getColor = (color, theme) => {
+  if (!color) return 'inherit';
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
+  };
+
+  const iconColor = overwrittenVars[`color${capitalize(color)}`];
```
montezume: is this more like what you were thinking @tdeekens ?

src/components/icons/create-styled-icon.js
tdeekens: Still valid?

montezume: can you check the latest version?

src/components/buttons/icon-button/icon-button.js
```diff
@@ -22,13 +22,13 @@ const getIconThemeColor = props => {
...
   if (props.theme !== 'default' && (isActive || props.isMouseOver)) {
     if (props.isDisabled) {
-      return 'grey';
+      return 'neutral60';
     }
-    return 'white';
+    return 'surface';
   }
 
   // if button is disabled, icon should be grey
```
amine-benselim: should we update the comments too ?

src/components/icons/create-styled-icon.js
tdeekens: Ah yeah. Thanks!

montezume:
Okay, I'll merge this into `10.0.0` then. Thanks 👍

PR #727, opened 2019-05-03
feat(selects): make selects inputs themable

montezume:
#### Summary

Use the `withTheme` HOC from `emotion-theming` to pass our theme to the createSelectStyles functions to theme our `SelectInputs`.


src/components/internals/create-select-styles.js
```diff
@@ -136,41 +202,79 @@ const valueContainerStyles = () => base => ({
...
-const groupStyles = props => base => ({
-  ...base,
-  padding: 0,
-  '&:not(:first-child)': {
-    borderTop: props.showOptionGroupDivider
-      ? `1px solid ${vars.colorNeutral}`
-      : base.borderTop,
-  },
-});
+const singleValueStyles = (props, theme) => (base, state) => {
```
montezume: this was the styles I added, because we weren't setting the input font color

src/components/inputs/async-select-input/async-select-input.js
```diff
@@ -123,12 +126,15 @@ export class AsyncSelectInput extends React.Component {
               ...customizedComponents,
               ...this.props.components,
             }}
-            styles={createSelectStyles({
-              hasWarning: this.props.hasWarning,
-              hasError: this.props.hasError,
-              showOptionGroupDivider: this.props.showOptionGroupDivider,
-              menuPortalZIndex: this.props.menuPortalZIndex,
-            })}
+            styles={createSelectStyles(
```
montezume: don't have access to emotion here, so don't have the `theme` object without injecting it

PR #728, opened 2019-05-03
feat(checkbox-input): use design tokens / add theming

montezume:
#### Summary

* Use design tokens for `CheckboxInput`
* Make `CheckboxInput` theme-able

### DIFF
--- a/src/components/inputs/checkbox-input/checkbox-input.styles.js
+++ b/src/components/inputs/checkbox-input/checkbox-input.styles.js
@@ -1,15 +1,21 @@
 import { css } from '@emotion/core';
 import vars from '../../../../materials/custom-properties';
+import designTokens from '../../../../materials/design-tokens';
 
-const getCheckboxWrapperStyles = props => {
+const getCheckboxWrapperStyles = (props, theme) => {
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
+  };
   const baseStyles = css`
     display: flex;
     align-items: center;
     svg [id$='borderAndContent'] > [id$='border'] {
-      stroke: 1px ${vars.borderColorInputPristine} solid;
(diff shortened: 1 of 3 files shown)


src/components/inputs/checkbox-input/checkbox-input.styles.js
```diff
@@ -1,26 +1,34 @@
...
+    ...vars,
+    ...theme,
+  };
   const baseStyles = css`
     display: flex;
     align-items: center;
     svg [id$='borderAndContent'] > [id$='border'] {
-      stroke: 1px ${vars.borderColorInputPristine} solid;
+      stroke: ${overwrittenVars[designTokens.borderColorForInput]};
+      fill: ${overwrittenVars[designTokens.backgroundColorForInput]};
```
montezume: this was an oversight, we should have had this before

src/components/inputs/checkbox-input/checkbox-input.styles.js
```diff
@@ -30,11 +38,11 @@ const getCheckboxWrapperStyles = props => {
       baseStyles,
       css`
         svg [id$='borderAndContent'] > [id$='content'] {
-          fill: ${vars.borderColorInputError};
+          fill: ${overwrittenVars[designTokens.borderColorForInputWhenError]};
```
amine-benselim: do we want borderColor for the fill here ?

montezume: Yeah it might make more sense semantically to use the design token fontColorForInputWhenError because here are targeting the id "content", which is the check mark (when checked and error). But that token is not red 🤔. It might make more sense just to use `colorError`

It's pretty confusing, I wanted to rewrite this with CSS and get rid of the SVG usage but haven't got around to it.

PR #740, opened 2019-05-07
refactor: remove deprecated custom props

montezume:
#### Summary

Removes deprecated tokens and custom properties.

## NOTE -> base set first as master for CI run, and then changed to `10.0.0`


montezume:
v diff probably fixed in https://github.com/commercetools/ui-kit/pull/741

montezume:
other VRT diff fixed in #743

Issue #706, opened 2019-04-23
Spread remaining props to HTML elements and remove excess props


jonnybel:
**Is your feature request related to a problem? Please describe.**
We've been adding props to our React components which do little or nothing more than being forwarded to the HTML elements as attributes. This keeps increasing the number of props for these components, giving a false idea of complexity and making our docs longer, not to mention the excess code, and tests for functionality that is already native to the HTML.

The most recent example was [this PR](https://github.com/commercetools/ui-kit/pull/696#issuecomment-484568757) for adding `autocomplete`.

**Describe the solution you'd like**
The vast majority of these props should simply be spread to the underlying HTML elements as `remainingProps`.

Some props, such as `isDisabled`, because they require applying specific styles before adding the `disabled`, shall not be spread directly like the rest but should retain the name of the HTML attribute which they correspond to (`isDisabled` => `disabled`), keeping the API more similar to directly using the HTML attributes.

**Additional context**
Props that could be removed to be passed as `remainingProps`, or renamed to be equal to their HTML attribute correspondents and eventually dismiss their documentation:

Standard Attributes:
`id`, `name`, `value`, `placeholder`, `isAutoFocussed`, `isDisabled`, isRequire `tabIndex`, `data-*`, `aria-*`, ...
Event attributes: 
`onClick`, `onChange`, `onFocus`, `onBlur`, ...

Other like `buttonAttributes` can also be removed altogether.

PR #747, opened 2019-05-08
refactor: Text Inputs to spread attributes and remove wrapper

jonnybel:
#### Summary

This PR is an approach to tackle #706, starting with Text Inputs. If this approach well received, the same can be done for all the other components.

1. Rename input props to use the same names as the native root element attributes, and spread all remaining props
2. Remove the Horizontal Constraint wrapper and apply the sizing style directly on the input.

#### Description

1. The reasoning for this is detailed in #706.
This is obviously a breaking change and if we proceed, we should make the same changes to all the other components that have props which are the same as native html properties/attributes. All the changes would be merged into one single branch for this new major version before merging the full refactor to master.
*Alternatively*, we could keep the naming of some of our props like `isDisabled` and still spread the rest of the props, preventing breaking changes, but this would be less consistent since only _some_ of the properties would have the same namings as the html properties, which partly defeats the purpose of the refactor. For me this would only make sense as a transitional phase, throwing warnings when using the old props.
2. The Horizontal Constraint wrapper is being applied in all inputs, but after researching a bit on our usage of the `horizontalConstraint` prop, I have found that this is rarely useful and mostly only adds noise.
_2.1._ The default value is `scale` which ultimately adds nothing of value and only nests the input in an unnecessary div that scales to 100% of its parent, because the input is already scaled to 100% from its base style 🤷‍♂ .
_2.2._ In practice, most times Inputs are used, they are grouped together, or placed in a container with other elements that share the same horizontal constraint (such as labels), so it usually makes more sense to apply a constraint to the whole group of elements, and so it's irrelevant to repeat the constraint on each individual input.
_2.3._ I've even found usage cases where the developer actually wanted to scale the input individually, but instead of using the `horizontalConstraint` prop, he wrapped the input on an `HorizontalConstraint` component, resulting in two duplicate wrappers around the input.
_2.4._ Some inputs, such as the Checkbox, don't even have this wrapper, so there is already inconsistency in the UI-kit that should be addressed.
_2.5._  Ultimately, I would even prefer to remove this prop from the component, as it is very rarely useful in practice, and if the developer really needs to size an input individually, he can use the `HorizontalConstraint` for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "emotion-theming": "10.0.10",
     "lodash": "4.17.11",
     "lodash-es": "4.17.11",
+    "lodash.omit": "4.5.0",
     "prop-types": "15.7.2",
     "react-is": "16.8.6",
     "react-popper": "1.3.3",
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "emotion-theming": "10.0.10",
     "lodash": "4.17.11",
     "lodash-es": "4.17.11",
-    "lodash.omit": "4.5.0",
     "prop-types": "15.7.2",
     "react-is": "16.8.6",
     "react-popper": "1.3.3",
--- a/src/components/constraints/horizontal/horizontal.js
+++ b/src/components/constraints/horizontal/horizontal.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import vars from '../../../../materials/custom-properties';
 
-const getConstraintSyles = constraint => {
+export const getConstraintSyles = constraint => {
   switch (constraint) {
     case 'xs':
       return css`
--- a/src/components/inputs/styles.js
+++ b/src/components/inputs/styles.js
@@ -71,12 +71,11 @@ const getInputStyles = (props, theme) => {
     return [
       baseStyles,
       css`
-      ${baseStyles}
-      border-color: ${
-        overwrittenVars[designTokens.borderColorForInputWhenWarning]
-      };
-      color: ${overwrittenVars[designTokens.fontColorForInputWhenWarning]};
-    `,
+        border-color: ${overwrittenVars[
+          designTokens.borderColorForInputWhenWarning
+        ]};
+        color: ${overwrittenVars[designTokens.fontColorForInputWhenWarning]};
+      `,
     ];
   }
   if (props.readOnly) {
--- a/src/components/inputs/styles.js
+++ b/src/components/inputs/styles.js
@@ -40,7 +40,7 @@ const getInputStyles = (props, theme) => {
       ]};
     }
   `;
-  if (props.isDisabled) {
+  if (props.disabled) {
     return [
       baseStyles,
       css`
@@ -79,7 +79,7 @@ const getInputStyles = (props, theme) => {
     `,
     ];
   }
-  if (props.isReadOnly) {
+  if (props.readOnly) {
     return [
       baseStyles,
       css`
--- a/src/components/inputs/styles.js
+++ b/src/components/inputs/styles.js
@@ -40,7 +40,7 @@ const getInputStyles = (props, theme) => {
       ]};
     }
   `;
-  if (props.disabled) {
+  if (props.disabled || props.isDisabled) {
     return [
       baseStyles,
       css`
@@ -78,7 +78,7 @@ const getInputStyles = (props, theme) => {
       `,
     ];
   }
-  if (props.readOnly) {
+  if (props.readOnly || props.isReadOnly) {
     return [
       baseStyles,
       css`
--- a/src/components/inputs/text-input/README.md
+++ b/src/components/inputs/text-input/README.md
@@ -22,17 +22,15 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
 | `value`                | `string` |    ✅    | -                            | -       | Value of the input                                                                                     |
 | `autoComplete`         | `string` |    -     | -                            | -       | Used as HTML `autocomplete` property                                                                   |
 | `onChange`             | `func`   |    -     | -                            | -       | Called with the new value. Required when input is not read only. Parent should pass it back as `value` |
-| `onBlur`               | `func`   |    -     | -                            | -       | Called when field is blurred                                                                           |
-| `onFocus`              | `func`   |    -     | -                            | -       | Called when field is focused                                                                           |
-| `isAutofocussed`       | `bool`   |    -     | -                            | -       | Focus the input field on initial render                                                                |
-| `isDisabled`           | `bool`   |    -     | -                            | `false` | Indicates that the field cannot be used (e.g not authorised, or changes not saved)                     |
-| `isReadOnly`           | `bool`   |    -     | -                            | `false` | Indicates that the field is displaying read-only content                                               |
+| `autoFocus`            | `bool`   |    -     | -                            | -       | Focus the input field on initial render                                                                |
+| `disabled`             | `bool`   |    -     | -                            | `false` | Indicates that the field cannot be used (e.g not authorised, or changes not saved)                     |
+| `readOnly`             | `bool`   |    -     | -                            | `false` | Indicates that the field is displaying read-only content                                               |
 | `hasError`             | `bool`   |    -     | -                            | -       | Indicates the input field has an error                                                                 |
 | `hasWarning`           | `bool`   |    -     | -                            | -       | Indicates the input field has a warning                                                                |
 | `placeholder`          | `string` |    -     | -                            | -       | Placeholder text for the input                                                                         |
-| `horizontalConstraint` | `object` |          | `s`, `m`, `l`, `xl`, `scale` | `scale` | Horizontal size limit of the input fields.                                                             |
+| `horizontalConstraint` | `object` |          | `s`, `m`, `l`, `xl`, `scale` | `scale` | Horizontal size limit of the wrapper for input fields.                                                 |
 
-The component further forwards all `data-` attributes to the underlying `input` component.
+Any other properties will be spread to the underlying `input` element.
 
 ### Static methods
 
--- a/src/components/inputs/text-input/text-input.js
+++ b/src/components/inputs/text-input/text-input.js
@@ -7,20 +7,20 @@ import { getInputStyles } from '../styles';
 import { getConstraintSyles } from '../../constraints/horizontal';
 import throwDeprecationWarning from '../../../utils/warn-deprecated-prop';
 
-class TextInput extends React.PureComponent {
-  getInputProps = (props = {}) => {
-    const readOnly = props.readOnly || props.isReadOnly;
+const TextInput = props => {
+  function getInputProps(componentProps = {}) {
+    const readOnly = componentProps.readOnly || componentProps.isReadOnly;
 
     return {
       type: 'text',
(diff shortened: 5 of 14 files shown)


src/components/constraints/horizontal/horizontal.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import vars from '../../../../materials/custom-properties';
 
-const getConstraintSyles = constraint => {
+export const getConstraintSyles = constraint => {
```
jonnybel: I'm exporting this only for internal usage.
Perhaps it might make sense to extract this function into a separate module.

src/components/inputs/styles.js
```diff
@@ -69,15 +69,14 @@ const getInputStyles = (props, theme) => {
     return [
       baseStyles,
       css`
-      ${baseStyles}
```
jonnybel: This is already added to the `css` two lines above.

src/components/inputs/text-input/README.md
```diff
@@ -22,17 +22,15 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
...
 | `hasError`             | `bool`   |    -     | -                            | -       | Indicates the input field has an error                                                                 |
 | `hasWarning`           | `bool`   |    -     | -                            | -       | Indicates the input field has a warning                                                                |
 | `placeholder`          | `string` |    -     | -                            | -       | Placeholder text for the input                                                                         |
-| `horizontalConstraint` | `object` |          | `s`, `m`, `l`, `xl`, `scale` | `scale` | Horizontal size limit of the input fields.                                                             |
+| `horizontalConstraint` | `object` |          | `s`, `m`, `l`, `xl`, `scale` | `scale` | Horizontal size limit of the wrapper for input fields.                                                 |
 
-The component further forwards all `data-` attributes to the underlying `input` component.
+Any other properties will be spread to the underlying `input` element.
```
jonnybel: I'm not sure if we should keep the documentation for the props we already had, but I'm fine with it.

src/components/inputs/text-input/text-input.js
```diff
@@ -1,58 +1,44 @@
...
-  isDisabled: PropTypes.bool,
-  isReadOnly: PropTypes.bool,
   hasError: PropTypes.bool,
   hasWarning: PropTypes.bool,
   placeholder: PropTypes.string,
+  autoComplete: PropTypes.string,
   horizontalConstraint: PropTypes.oneOf(['s', 'm', 'l', 'xl', 'scale']),
 };
 
-TextInput.defaultProps = {
```
jonnybel: I removed this default because the `scale` value in practice is the same as nothing, since the input is already scaled to 100%.

src/components/inputs/text-input/text-input.spec.js
```diff
@@ -82,28 +87,8 @@ describe('TextInput', () => {
     expect(onChange).toHaveBeenCalled();
   });
 
-  it('should call onFocus when the input is focused', () => {
```
jonnybel: No need for testing the behaviour of native HTML properties.

src/components/inputs/text-input/README.md
```diff
@@ -22,17 +22,15 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
...
-| `onFocus`              | `func`   |    -     | -                            | -       | Called when field is focused                                                                           |
-| `isAutofocussed`       | `bool`   |    -     | -                            | -       | Focus the input field on initial render                                                                |
-| `isDisabled`           | `bool`   |    -     | -                            | `false` | Indicates that the field cannot be used (e.g not authorised, or changes not saved)                     |
-| `isReadOnly`           | `bool`   |    -     | -                            | `false` | Indicates that the field is displaying read-only content                                               |
+| `autoFocus`            | `bool`   |    -     | -                            | -       | Focus the input field on initial render                                                                |
+| `disabled`             | `bool`   |    -     | -                            | `false` | Indicates that the field cannot be used (e.g not authorised, or changes not saved)                     |
```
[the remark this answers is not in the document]
montezume: I'm also not sure how difficult / annoying the refactoring will be, since even if we change all of our components to use `disabled` instead of `isDisabled`, there are probably components inside the MC that use that prop and find and replace probably won't work well...

src/components/inputs/text-input/text-input.js
```diff
@@ -1,58 +1,44 @@
...
+        getInputStyles(props, theme),
+        getConstraintSyles(props.horizontalConstraint),
+      ]}
       /* ARIA */
-      aria-readonly={props.isReadOnly}
       role="textbox"
-      contentEditable={!props.isReadOnly}
+      aria-readonly={props.readOnly}
+      contentEditable={!props.readOnly}
+      {...inputProps}
```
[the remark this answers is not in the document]
jonnybel: I haven't, I'll look into this.

src/components/inputs/text-input/README.md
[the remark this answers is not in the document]
jonnybel: OK, so I'm thinking of mapping the old props to the new ones and throw a warning when the old ones are used, the same way I did in #631.

I'll also investigate how to make the code mod.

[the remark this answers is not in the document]
montezume: Yeah that sounds like a good idea to me, then we can slowly do the refactor in the MC

src/components/inputs/text-input/text-input.js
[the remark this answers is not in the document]
jonnybel: Done.

src/components/inputs/text-input/README.md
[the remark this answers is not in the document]
jonnybel: I've added the deprecation warnings and mapped the old props to the new.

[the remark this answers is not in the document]
emmenko: We have a couple of old codemods in the `transforms` folder of the mc-fe repo, in case it helps.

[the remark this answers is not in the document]
jonnybel: I've added a codemod and tested it on UI-kit (refactored TextField to use the new props).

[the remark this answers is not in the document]
jonnybel: @emmenko @montezume 
If we're happy with this, we can start doing the same for _all_ other inputs (and buttons, and other components that reimplement native HTML attributes)

package.json
```diff
@@ -71,6 +71,7 @@
     "emotion-theming": "10.0.10",
     "lodash": "4.17.11",
     "lodash-es": "4.17.11",
+    "lodash.omit": "4.5.0",
```
emmenko: We don't need this

src/components/inputs/text-input/text-input.js
```diff
@@ -1,56 +1,97 @@
...
-      onBlur={props.onBlur}
-      onFocus={props.onFocus}
-      disabled={props.isDisabled}
-      placeholder={props.placeholder}
-      readOnly={props.isReadOnly}
-      autoFocus={props.isAutofocussed}
-      autoComplete={props.autoComplete}
-      css={theme => getInputStyles(props, theme)}
-      {...filterDataAttributes(props)}
+class TextInput extends React.PureComponent {
```
emmenko: Can we keep it a function?

package.json
jonnybel: Removed

src/components/inputs/text-input/text-input.js
jonnybel: Yes we can! Done

src/components/inputs/text-input/text-input.js
```diff
@@ -7,20 +7,20 @@ import { getInputStyles } from '../styles';
 import { getConstraintSyles } from '../../constraints/horizontal';
 import throwDeprecationWarning from '../../../utils/warn-deprecated-prop';
 
-class TextInput extends React.PureComponent {
-  getInputProps = (props = {}) => {
-    const readOnly = props.readOnly || props.isReadOnly;
+const TextInput = props => {
+  function getInputProps(componentProps = {}) {
```
emmenko: Can we move this function outside of the component?

transforms/react-component-rename-props.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable guard-for-in */
+
+// Components whose props shall be renamed
+const componentNamesToRefactor = [
+  // 'TextInput'
```
tdeekens: Should this maybe be a CLI argument to the codemod. Is that possible somehow?

tdeekens: Maybe like:

```txt
/codemods
   /migrations
      /v10
	      text-input.js
          button.js
```

to then `jscodeshift -- --migration v10/text-input.js

jonnybel: Yes, I'd to that as a follow-up, as this PR moves forward and we proceed to refactor all the other components, then list them all those components and props mappings in files that are then imported by this codemod.

src/components/inputs/text-input/text-input.js
jonnybel: Sure, I was thinking it could be useful to be able to call it from the TextInput, like the `isEmpty` fn, but tbh I'm not sure if it would actually ever be used.

transforms/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+Useful options:
+
+- `-d`: dry-run
+- `-p`: print the output for comparison (use it together with `-d`)
+
+## Included scripts
+
+#### `react-rename-component-props`
```
montezume: this is awesome 🎈

src/components/inputs/text-input/text-input.spec.js
```diff
@@ -25,7 +25,12 @@ describe('TextInput', () => {
     expect(container.querySelector('input')).toHaveAttribute('data-foo', 'bar');
   });
 
-  it('should render a number input', () => {
+  it('should forward any attributes', () => {
```
montezume: 👍

src/components/inputs/text-input/text-input.js
```diff
@@ -1,56 +1,95 @@
...
+        'isDisabled',
+        'isAutofocussed',
+      ]),
+    };
+  }
+
+  return (
+    <input
+      css={theme => [
+        getInputStyles(props, theme),
```
montezume: couldn't we use the result of getInputProps here and then we wouldn't need to do the or checks?

jonnybel: You mean the `or` checks inside the `getInputStyles`?
The problem is: other components, which haven't been refactored yet, also use `getInputStyles`, so it still needs to support the old names.

montezume: ah yeah 🤔 makes sense

jonnybel: Also, props such as `hasWarning` would be filtered out but they are needed by `getInputStyles`.

src/components/inputs/text-input/text-input.js
emmenko: In that case you need to assign is as a static property

```js
TextInput.getInputProps = getInputProps
```

But I don't see any usage of it. I would just move it outside.

jonnybel: Ah, right, I forgot about that!

Anyway, it's moved now.

PR #755, opened 2019-05-10
feat: add new `<Link>` component.

montezume:
#### Summary

Adds new `<Link />` component.

This component is a very thin wrapper over the [react-router link component](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md)

However, if you pass in the `isExternal` prop as true, instead it will render a regular <a> tag.

### DIFF
--- a/src/components/links/link/link.js
+++ b/src/components/links/link/link.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import requiredIf from 'react-required-if';
+import { Link as ReactRouterLink } from 'react-router-dom';
+import Styled from '@emotion/styled';
+import vars from '../../../../materials/custom-properties';
+
+const createStyledComponent = component => Styled(component)`
+  font-family: ${vars.fontFamily};
+  color: ${vars.colorPrimary};
+  font-size: ${vars.fontSizeDefault};
+
+  &:hover {
+    color: ${vars.colorPrimary25};
+  }
+`;
+
+const StyledReactRouterLink = createStyledComponent(ReactRouterLink);
+const StyledExternalLink = createStyledComponent('a');
+
+const componentProps = ['isExternal', 'to'];
+
+const Link = props => {
+  // instead of object destructing
+  const remainingProps = Object.keys(props)
+    .filter(key => !componentProps.includes(key))
+    .reduce((obj, key) => {
+      // eslint-disable-next-line no-param-reassign
+      obj[key] = props[key];
+      return obj;
+    }, {});
+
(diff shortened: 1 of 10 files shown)


src/components/links/link/link.js
```diff
@@ -0,0 +1,62 @@
...
+
+const Link = props => {
+  // instead of object destructing
+  const remainingProps = Object.keys(props)
+    .filter(key => !componentProps.includes(key))
+    .reduce((obj, key) => {
+      // eslint-disable-next-line no-param-reassign
+      obj[key] = props[key];
+      return obj;
+    }, {});
```
tdeekens: Should this instead maybe be a general helper? Sounds a bit like a pattern which will start to be repeated. Like `getPassThroughProps`.

src/utils/get-pass-through-props.js
```diff
@@ -0,0 +1,10 @@
+const getPassThroughProps = (props, componentProps) =>
+  Object.keys(props)
+    .filter(key => !componentProps.includes(key))
+    .reduce((obj, key) => {
+      // eslint-disable-next-line no-param-reassign
+      obj[key] = props[key];
+      return obj;
```
tdeekens: Nit: `obj` => `passThroughProps`

src/utils/get-pass-through-props.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        a: 1,
+        b: 2,
+      },
+      'aria-label': 'help us obi wan',
+      href: 'asdf',
+    };
+    const componentProps = ['isDisabled', 'to'];
+
+    expect(getPassThroughProps(props, componentProps)).toEqual({
+      'aria-label': 'help us obi wan',
```
jonnybel: 🤣

src/components/links/link/link.js
```diff
@@ -0,0 +1,56 @@
...
+import Styled from '@emotion/styled';
+import getPassThroughProps from '../../../utils/get-pass-through-props';
+import vars from '../../../../materials/custom-properties';
+
+const createStyledComponent = component => Styled(component)`
+  font-family: ${vars.fontFamily};
+  color: ${vars.colorPrimary};
+  font-size: ${vars.fontSizeDefault};
+
+  &:hover, &:focus, &:active {
```
jonnybel: Just curious: no specific colour for the `:visited` state?

montezume: yeah we will need to ask design 🤔

montezume: Design said for the current / ongoing work we don't need visited state.

PR #767, opened 2019-05-13
fix(collapsible-panel): height of panel header while disabled

montezume:
#### Summary

When `CollapsiblePanel` is disabled, the header should look the same as when it's not disabled.

Thanks to @adnasa for catching this one 👍


montezume:
I'm interested to see how the VRT react

src/components/panels/collapsible-panel/collapsible-panel.styles.js
```diff
@@ -98,6 +98,13 @@ const getHeaderStyles = ({ isDisabled, isCondensed }) => {
       css`
         cursor: default;
       `,
+      !isCondensed &&
```
adnasa: good catch!
I forgot that we had this prop.

PR #769, opened 2019-05-14
refactor(flat-button): simplify code, remove mouseover

montezume:
#### Summary

I believe that the use of the `withMouseOverState` HOC is unnecessary. We can achieve the same effect with pure css, with better performance, and (marginally) more readable code.

If this is acceptable to yall, I'll make PRs to remove from the other components as well.


src/components/buttons/flat-button/flat-button.js
```diff
@@ -62,6 +53,16 @@ export const FlatButton = props => {
               props.isDisabled ? vars.colorNeutral : getTextColor(props.tone)
             };
           }
+
+          svg * {
```
montezume: basically the same code as we have for the hover.

PR #771, opened 2019-05-14
refactor: Checkboxes to spread attributes and improve a11y

jonnybel:
#### Summary

Following #747, this PR is a continuation to tackle #706, this time refactoring the `CheckboxInput` component by following the same approach.

Also, fixes a bug where the input would still visually respond to hovering even when disabled.

Also, improves overall a11y by programmatically setting a few aria attributes.


src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -1,26 +1,41 @@
...
+
 const Label = styled.label`
   display: flex;
   align-items: center;
-  cursor: ${props => (props.isDisabled ? 'not-allowed' : 'pointer')};
+  cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
   position: relative;
   ${props =>
     !props.hasError &&
+    !props.disabled &&
```
jonnybel: This is for fixing the bug.

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -63,41 +98,36 @@ class CheckboxInput extends React.PureComponent {
...
-        </div>
-        {this.props.children && (
-          <div
-            css={css`
-              margin-left: ${vars.spacingS};
-            `}
-          >
+      <Label
+        htmlFor={this.state.id}
+        disabled={disabled}
```
jonnybel: This was also causing the bug: the `disabled` prop wasn't being passed.

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -63,41 +98,36 @@ class CheckboxInput extends React.PureComponent {
...
+        hasError={this.props.hasError}
+      >
+        <Checkbox id={this.state.id} {...getCheckboxProps(this.props)} />
+        <Spacings.Inline alignItems="center" scale="s">
```
jonnybel: Removed the styled `div` in L88 and added this wrapper here for the same effect.

src/components/inputs/checkbox-input/checkbox-input.js
```diff
@@ -63,41 +98,36 @@ class CheckboxInput extends React.PureComponent {
   });
 
   render() {
+    const disabled = this.props.disabled || this.props.isDisabled;
```
tdeekens: Dont we have to do the same for checked?

montezume: I think it happens in `getCheckboxProps` 🤔

montezume:
If you want you could also split out the non breaking changes and merge those into master...

src/components/inputs/checkbox-input/checkbox-input.js
montezume: maybe could also do

```

const checkboxProps = getCheckboxProps(this.props);
... 
<Label disabled={checkboxProps.disabled}>
```

jonnybel: We do, but because it's only used once under `render()`, it is done inline in L117 (apart from also being done in `getCheckboxProps`).

montezume: yeah, just meant so we could remove the disabled={isDisabled || disabled} thing

jonnybel: I like that alternative 👍

jonnybel:
I'll open a PR just to fix that problem with the hovered state when disabled, since its a bug and should be fixed asap.

PR #777, opened 2019-05-16
fix(collapsible-panel): disabled state fix

montezume:
#### Summary

When `CollapsiblePanel` is disabled, we should have the proper background colour for the icon button.

Should just be able to review the Percy diff.


montezume:
The version diff is easy to check through Percy @filippobocik

PR #783, opened 2019-05-17
feat(link-button): support isExternal prop

montezume:
#### Summary

Adds `isExternal` prop to `LinkButton`

#### Motivation

Some use cases in the MC (discount predicates for one) have this sort of external link, and this would make their code cleaner.

### DIFF
--- a/src/components/buttons/link-button/link-button.js
+++ b/src/components/buttons/link-button/link-button.js
@@ -44,8 +44,6 @@ const createStyledComponent = component => styled(component)`
 const StyledReactRouterLink = createStyledComponent(ReactRouterLink);
 const StyledExternalLink = createStyledComponent('a');
 
-const componentProps = ['isExternal', 'to', 'label', 'isDisabled', 'iconLeft'];
-
 const LinkBody = props => (
   <Spacings.Inline scale="xs" alignItems="center">
     {Boolean(props.iconLeft) &&
@@ -65,7 +63,10 @@ LinkBody.propTypes = {
 };
 
 const LinkButton = props => {
-  const remainingProps = getPassThroughProps(props, componentProps);
+  const remainingProps = getPassThroughProps(
+    props,
+    Object.keys(LinkButton.propTypes)
+  );
 
   if (props.isExternal) {
     return (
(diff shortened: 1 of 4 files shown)


src/components/buttons/link-button/link-button.js
```diff
@@ -1,65 +1,109 @@
...
+
+  cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
+
+  ${props => !props.disabled && hoverStyles}
+`;
+
+const StyledReactRouterLink = createStyledComponent(ReactRouterLink);
+const StyledExternalLink = createStyledComponent('a');
+
+const componentProps = ['isExternal', 'to', 'label', 'isDisabled', 'iconLeft'];
```
tdeekens: This is odd to me. Is this an established pattern? I thought we could do something like `Object.keys(LinkButton.propTypes)` to determine the pass through props?

montezume: good call!

src/components/links/link/link.js
```diff
@@ -19,10 +19,11 @@ const createStyledComponent = component => Styled(component)`
 const StyledReactRouterLink = createStyledComponent(ReactRouterLink);
 const StyledExternalLink = createStyledComponent('a');
 
-const componentProps = ['isExternal', 'to'];
-
 const Link = props => {
-  const remainingProps = getPassThroughProps(props, componentProps);
+  const remainingProps = getPassThroughProps(
+    props,
+    Object.keys(Link.propTypes)
```
montezume: thanks @tdeekens

src/components/buttons/link-button/link-button.js
```diff
@@ -1,65 +1,110 @@
...
+          isDisabled={props.isDisabled}
+          label={props.label}
+        />
+      </StyledExternalLink>
+    );
+  }
+
+  return (
+    <StyledReactRouterLink
+      to={props.to}
```
jonnybel: This can be omitted from here because it's also passed in `remainingProps`, right?

montezume: no, it's removed from remainingProps because it's a propType right?

jonnybel: Ah yes, you're right!

montezume:
Visual diff was just the extra proptype

PR #792, opened 2019-05-20
chore: update deps

montezume:
* Closes #791 
* Closes #790 
* Closes #789 
* Closes #788 
* Closes #787 
* Closes #786


package.json
```diff
@@ -122,10 +122,9 @@
     "colors": "1.3.3",
     "common-tags": "1.8.0",
     "conventional-changelog-cli": "2.0.21",
-    "core-js": "3.0.1",
```
montezume: wasn't used in UI-Kit, removed instead of updated.

package.json
```diff
@@ -185,7 +184,6 @@
     "storybook-readme": "5.0.3",
     "webpack": "4.31.0",
     "webpack-cli": "3.3.2",
-    "webpack-dev-server": "3.3.1",
```
montezume: ditto

PR #800, opened 2019-05-22
fix(i18n): build all locales

montezume:
#### Summary

Our build intl script was missing french and chinese.

### DIFF
--- a/i18n/data/en.json
+++ b/i18n/data/en.json
@@ -26,6 +26,7 @@
   "UIKit.LocalizedTextInput.expand": "Show all languages ({remainingLanguages})",
   "UIKit.LocalizedTextInput.missingRequiredField": "This field is required. Provide at least one value.",
   "UIKit.MoneyField.highPrecision": "High Precision Price",
+  "UIKit.MoneyInput.highPrecision": "High Precision Price",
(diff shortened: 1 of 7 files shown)


i18n/data/en.json
```diff
@@ -26,11 +26,12 @@
   "UIKit.LocalizedTextInput.expand": "Show all languages ({remainingLanguages})",
   "UIKit.LocalizedTextInput.missingRequiredField": "This field is required. Provide at least one value.",
   "UIKit.MoneyField.highPrecision": "High Precision Price",
+  "UIKit.MoneyInput.highPrecision": "High Precision Price",
```
montezume: I guess we hadn't run the script in awhile 🤷‍♂

scripts/extract-intl.js
```diff
@@ -46,7 +46,7 @@ const { presets, plugins } = babelConfig;
 // Resolve the absolute path of the caller location. This is necessary
 // to point to files within that folder.
 const rootPath = process.cwd();
-const locales = ['en', 'de', 'es'];
+const locales = ['en', 'de', 'es', 'fr-FR', 'zh-CN'];
```
montezume: this be the fix ☠️

PR #804, opened 2019-05-22
fix(multiline-input): Expand behaviour not working

jonnybel:
#### Summary

The `MultilineTextInput` expand behaviour is currently broken. This PR restores the functionality.

### DIFF
--- a/src/components/inputs/localized-multiline-text-input/translation-input.styles.js
+++ b/src/components/inputs/localized-multiline-text-input/translation-input.styles.js
@@ -26,7 +26,6 @@ const getTextareaStyles = (props, theme) => {
     props.isCollapsed &&
       css`
         overflow: hidden;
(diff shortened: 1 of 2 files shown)


src/components/inputs/localized-multiline-text-input/translation-input.styles.js
```diff
@@ -26,7 +26,6 @@ const getTextareaStyles = (props, theme) => {
     props.isCollapsed &&
       css`
         overflow: hidden;
-        word-break: break-all;
```
jonnybel: This was already above, so it wasn't doing anything.

montezume:
Can you check if this also effects our other components? LocalizedMultilineText input for one

jonnybel:
Yes, I already did and it was by checking it that I found how to fix the bug 😄

PR #805, opened 2019-05-23
fix(collapsible-panel): remove z-index

montezume:
#### Summary

Currently when you have a <SelectInput> above a CollapsiblePanel, the menu list is hidden by the CollapsiblePanel due to z-index setting. I don't believe we need to set a z-index here.

Screen Shot 2019-05-23 at 12 01 14 PM

Screen Shot 2019-05-23 at 11 44 42 AM

### DIFF
--- a/src/components/panels/collapsible-panel/collapsible-panel.styles.js
+++ b/src/components/panels/collapsible-panel/collapsible-panel.styles.js
@@ -53,7 +53,6 @@ const getHeaderContainerStyles = ({ isDisabled, isOpen, isSticky, theme }) => {
       css`
         position: sticky;
         top: 0;
-        z-index: 1;
         border-top-right-radius: ${vars.borderRadius6};
         border-top-left-radius: ${vars.borderRadius6};
       `,


montezume:
IMO it's a fix. We are changing internal styles of the CollapsiblePanel header.

montezume:
@jonnybel any thoughts ?

jonnybel:
Well, it does constitute a breaking change, although it might not affect anything, it's hard to tell without seeing how it's being used all over our apps 🤔 

But in doubt, it should be considered breaking. Perhaps we can add it to `10.0.0`

montezume:
Well, it sets the z-index of a header container. I can't see how it could effect anything unless there was a CollapsiblePanel with some element inside that overlapped it's header container. I guess we can put it in 10.0.0, since nobody but me noticed the bug in the MC...

tdeekens:
We can release it as a normal fix. Sorry for making the buzz. It was more about raising awareness for how verionsing can be different on a ui library than this being an apt case for following it.

PR #813, opened 2019-05-27
chore: update deps

montezume:
* Fixes #812 
* Fixes #811 
* Fixes #810 
* Fixes #809 
* Fixes #808
* Fixes #807


package.json
```diff
@@ -161,9 +161,9 @@
     "moment-timezone": "0.5.25",
     "mri": "1.1.4",
     "omit-empty": "1.0.0",
-    "pptr-testing-library": "0.3.2",
+    "pptr-testing-library": "0.4.0",
```
montezume: [ref](https://github.com/testing-library/dom-testing-library/releases/tag/v4.0.0)

now when you call getBy* and it returns a list, it throws.

tdeekens: That bubbles up from rtl itself and prevents us updating in the MC.

montezume: Yeah at some point we will need to refactor the tests that rely on this 😢

PR #815, opened 2019-05-27
feat: add stylelint for css-in-js

montezume:
#### Summary

When we migrated to @emotion, we removed stylelint. It makes sense now to add it back. This PR adds a script `yarn lint:css`, and the necessary configs.

It lints **inline styles**, styles with the **css** prop, and **styled** components.

Huzzah! 🧙

### Current status of stylelint report

```bash
Test Suites: 9 failed, 206 passed, 215 total
Tests:       9 failed, 206 passed, 215 total
Snapshots:   0 total
Time:        12.172s
```

Right now there are some failing tests, so my idea is to first add the script, and then fix the stylelint errors in another PR.

It even found some invalid css!

```
src/components/buttons/flat-button/flat-button.js
 64:11  ✖  Unexpected }   CssSyntaxError
```
There is an extra closing brace there. Cool!

Closes #815


.stylelintrc
```diff
@@ -0,0 +1,39 @@
...
+      true,
+      { "ignoreProperties": ["composes"] }
+    ],
+    "declaration-colon-newline-after": null,
+    "rule-empty-line-before": null,
+    "value-list-comma-newline-after": null,
+    "indentation": null,
+    "selector-descendant-combinator-no-non-space": null,
+    "no-descending-specificity": null,
+    "no-duplicate-selectors": null,
```
montezume: I took all of these up until here from mc-fe

.stylelintrc
```diff
@@ -0,0 +1,39 @@
...
+    "no-empty-source": null
```
montezume: I added this line because we are linting some files that ... don't have CSS. Like utils and such.

jest.stylelint.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  runner: 'jest-runner-stylelint',
+  displayName: 'stylelint',
+  moduleFileExtensions: ['js'],
+  modulePathIgnorePatterns: [
+    'dist',
+    '.spec.js',
```
montezume: I thought it made sense to exclude these

.stylelintrc
montezume: The issue with excluding this is you would get an error from let's say, `parseTime.js` for having no css...

montezume:
I think it would make sense to do the same setup on app-kit. @emmenko do you have any thoughts?

.stylelintrc
```diff
@@ -0,0 +1,39 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-standard"
+  ],
+    "plugins": [
```
emmenko:
```suggestion
  "plugins": [
```

montezume:
One caveat: this is only working with css made with string literals.

```js
// This works

<div css={css`
    color: green;
`}>
   Green
</div>

// this also works
const GreenText = styled.div`
   color: green;
`;

// this does not work

<div css={{
    color: 'green',
}}>
   Green
</div>

// this also does not work

<div style={{
    color: 'green',
}}>
   Green
</div>

**Comment by emmenko** [comment]
If only CSS could be typed... 😗😜

But anyway, shouldn’t VRTs provide feedback if the CSS is wrong? 🤔

**Comment by montezume** [comment]
Yeah, VRTs should give us an idea. Linting isn't just about finding errors though, it's also about finding duplicated css, etc. Stuff like this won't show up in VRTs but it's nice to get rid of it.

``` 47:3  ✖  Unexpected duplicate "display"   declaration-block-no-duplicate-properties```

File: test/jest-styles-runner/run.js [hunk 5]
```diff
@@ -0,0 +1,36 @@
+// copied from https://github.com/keplersj/jest-runner-stylelint/blob/master/src/run.js
+// because it uses an older version of stylelint as a dep (not peer dep)
+// we need new version of stylelint for css in js support
```
emmenko: Why not using yarn `resolutions`?

montezume: because I didn't know it existed 👍 🤣

package.json
```diff
@@ -196,6 +203,9 @@
     "react-intl": "2.x",
     "react-router-dom": "5.x"
   },
+  "resolutions": {
+    "jest-runner-stylelint/stylelint": "10.0.1"
```
montezume: gives this warning, but I think it's the intended result

```
warning Resolution field "stylelint@10.0.1" is incompatible with requested version "stylelint@^8.3.1"
```

montezume:
Alright, using the new version of stylelint solved the problems with trying to lint object styles

PR #819, opened 2019-05-31
chore: update react-select to version 3

montezume:
#### Summary

Updates dependency `react-select` to version 3. 

See this [issue](https://github.com/JedWatson/react-select/issues/3585) for more details on the changes

The main advantage for us is that with version 3, react-select now relies on @emotion 10. This means we won't be shipping two versions of emotion.

I'll update some other deps here too


src/components/inputs/select-input/select-input.js
```diff
@@ -242,22 +243,26 @@ export class SelectInput extends React.Component {
                   }
                 : undefined
             }
-            onChange={selectedOptions =>
-              // selectedOptions is either an array, or a single option
+            onChange={selectedOptions => {
+              // selectedOptions is either an array, or a single option, or null
               // depending on whether we're in multi-mode or not (isMulti)
```
montezume: before if you cleared your multi select, the value was empty array. now it's null

src/components/inputs/select-input/select-input.js
```diff
@@ -152,17 +152,18 @@ export class SelectInput extends React.Component {
...
-            // sorting is not lost
-            .map(value =>
-              optionsWithoutGroups.find(option => option.value === value)
-            )
-            .filter(Boolean)
-        : optionsWithoutGroups.find(
-            option => has(option, 'value') && option.value === props.value
-          ) || null,
+      selectedOptions:
+        props.isMulti && props.value
```
montezume: needed to add && props.value because when you clear multi select now, it sends null

.size-snapshot.json
```diff
@@ -1,15 +1,15 @@
...
-    "minified": 507454,
-    "gzipped": 97912,
+    "bundled": 866010,
+    "minified": 595824,
+    "gzipped": 121125,
     "treeshaked": {
       "rollup": {
-        "code": 354503,
-        "import_statements": 910
+        "code": 429358,
```
montezume: this goes up, but hopefully it's a win due to not shipping two versions of emotion 🤷‍♂ 🤔

montezume:
I'll do the react-testing-library update in another pr

src/components/inputs/select-input/select-input.js
montezume: I guess another option would be to continue sending empty array, that way we won't break anything in the MC 🤔

montezume: yeah I guess otherwise this would be a breaking change. 😱

.size-snapshot.json
jonnybel: Hmmm, but shouldn't that make the bundle size go down?

montezume: I don't think size snapshot includes dependencies.

jonnybel: Ah, alright, then it makes sense.

src/components/inputs/select-input/select-input.js
tdeekens: I would prefer to "wrap" the breaking change. I would assume that forms, mapping into and out of them, would break otherwise.

montezume: yeah, I'm currently doing this. Just investigating (console logging 😆) the old output of our selects when values are cleared, and going through (and adding tests 😄 ) to make sure new one matches. I did it for select-input already, need to do it for the other 3

src/components/inputs/select-input/select-input.js
montezume: going to remove this since we will wrap the breaking change

.size-snapshot.json
montezume: I'm not 100% sure how size snapshot works to be honest. But I wouldn't hold back a dependency update just to save a few kb 🤷‍♂

src/components/inputs/select-input/select-input.js
montezume: alright done

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -2,10 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
 import { withTheme } from 'emotion-theming';
-import {
-  components as defaultComponents,
-  AsyncCreatable as AsyncCreatableSelect,
-} from 'react-select';
+import { components as defaultComponents } from 'react-select';
+import AsyncCreatableSelect from 'react-select/async-creatable';
```
emmenko: Are they doing this on purpose to have better treeshaking?

montezume: Yeah 

this is what they said

"This should have no bundle-size impact on react-select consumers currently leveraging tree-shaking. However for consumers who aren’t leveraging tree-shaking, this should help alleviate some of the bundle-weight."

PR #826, opened 2019-06-03
chore: use prod in vrt

montezume:
#### Summary

Thanks to first class CSS in JS support in @percy/agent, we can remove this webpack workaround.


montezume:
The problematic tests ran fine in percy when I ran the command locally so 🙏

PR #832, opened 2019-06-04
fix(toggle-input): toggle input disabled styles

jonnybel:
Fixes the following reported bugs on the ToggleInput component for when it is `disabled`:

> - The handler (thumb-color) should be grey instead of white 
> - The track shadow (the inner part) should be NONE. Right now there is a shadow.

Also fixes the styles for when the input is `disabled` and `checked`.

### DIFF
--- a/src/components/inputs/toggle-input/toggle-input.js
+++ b/src/components/inputs/toggle-input/toggle-input.js
@@ -66,23 +66,39 @@ const Span = styled.span`
 `;
 
 const Input = styled.input`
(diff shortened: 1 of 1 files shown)


montezume:
@mariabarrena can you check the VRT diff?

montezume:
@mariabarrena can you look at the Percy diff and netlify deploy tomorrow? Thanks

mariabarrena:
Percy checked. All approved.

PR #840, opened 2019-06-06
feat: add version export

montezume:
#### Summary

Like similar PRs in App Kit, we also add a version export 😄

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -18,6 +18,9 @@ const configureRollupPlugins = (options = {}) =>
   [
     replace({
       'process.env.NODE_ENV': JSON.stringify('production'),
+      'process.env.npm_package_version': JSON.stringify(
+        process.env.npm_package_version
+      ),
     }),
     // To use the nodejs `resolve` algorithm
     resolve(),
--- a/src/index.bundlespec.js
+++ b/src/index.bundlespec.js
@@ -1,4 +1,4 @@
-import { i18n, customProperties } from 'ui-kit';
+import { i18n, customProperties, version } from 'ui-kit';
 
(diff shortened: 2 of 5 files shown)


rollup.config.js
```diff
@@ -18,6 +18,9 @@ const configureRollupPlugins = (options = {}) =>
   [
     replace({
       'process.env.NODE_ENV': JSON.stringify('production'),
+      'process.env.npm_package_version': JSON.stringify(
+        process.env.npm_package_version
+      ),
```
tdeekens: Nice. If this works I'll simplify app-kit again too again.

montezume: But we don't use lerna here for bumping versions... I think that's why we can have this simpler

montezume: Screen Shot 2019-06-06 at 10 51 42 AM

screenshot of cjs file

src/index.bundlespec.js
```diff
@@ -9,6 +9,12 @@ describe('exports', () => {
     expect(customProperties).toBeTruthy();
   });
 
+  describe('version', () => {
+    it('should match package.json version', () => {
+      expect(version).toMatch(process.env.npm_package_version);
```
montezume: bundle tests 😄

montezume: or should I directly import from package.json 🤔

rollup.config.js
tdeekens: Nice.

src/index.bundlespec.js
tdeekens: I think either is fine.

Issue #839, opened 2019-06-05
Storybook: use 5.1 features


montezume:
[See this article](https://medium.com/storybookjs/storybook-5-1-50afd02b5bbb)

Looks like we can use the `Contexts`  addon for setting locale.

PR #841, opened 2019-06-06
feat(storybook): use contexts for choosing locale

montezume:
#### Summary

We can use a new addon feature of storybook, contexts, for setting the global locale.

#### Before 

We had this knob in each story for setting locale

Screen Shot 2019-06-06 at 4 40 25 PM

It worked like this

before

#### After

Now we use `contexts`, and remove it as a knob.

coolio

#839


.storybook/configs/intl-context.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { IntlProvider, addLocaleData } from 'react-intl';
+import en from 'react-intl/locale-data/en';
+import de from 'react-intl/locale-data/de';
+import es from 'react-intl/locale-data/es';
+import frFR from 'react-intl/locale-data/fr';
+import zhCN from 'react-intl/locale-data/zh';
+import * as messages from '../../i18n';
+
```
montezume: this file is pretty much copied from our old decorator

.storybook/configs/intl-context.js
```diff
@@ -0,0 +1,77 @@
...
+    </IntlProvider>
+  );
+};
+
+IntlWrapper.propTypes = {
+  locale: PropTypes.string.isRequired,
+};
+
+const intlParams = locales.map(locale => ({
+  name: namifyLocale(locale),
```
montezume: name is a string, so I don't think we can localize it 🤔

.storybook/configs/intl-context.js
```diff
@@ -0,0 +1,77 @@
...
+    case 'en':
+      return 'English';
+    case 'es':
+      return 'Español';
+    case 'de':
+      return 'Deutsch';
+    case 'frFR':
+      return 'Français';
+    case 'zhCN':
+      return '简化字';
```
montezume: from https://en.wikipedia.org/wiki/Simplified_Chinese_characters

PR #855, opened 2019-06-13
fix(select-inputs): multi select disabled bug

### LINKED ISSUE #854: SelectInput - disabled and multi
#### Summary

When a `SelectInput` isMulti and isDisabled, then the selections are removeable

bug

#### Fix

The selections shouldn't be removable.

We should also add a test

montezume:
#### Summary

Fix for disabled multi selects

Fixes #854

### DIFF
--- a/src/components/inputs/select-input/select-input.spec.js
+++ b/src/components/inputs/select-input/select-input.spec.js
@@ -182,7 +182,7 @@ describe('in multi mode', () => {
   });
   describe('interacting', () => {
     describe('when disabled', () => {
-      it('should call not onChange when value is cleared', () => {
+      it('should not call onChange when value is cleared', () => {
         const onChange = jest.fn();
         const { getByLabelText, queryByText } = renderInput({
           onChange,
@@ -193,6 +193,21 @@ describe('in multi mode', () => {
         const input = getByLabelText('Fruit');
         fireEvent.focus(input);
         fireEvent.keyDown(input, { key: 'Backspace' });
+        expect(onChange).not.toHaveBeenCalled();
+        expect(queryByText('Mango')).toBeInTheDocument();
+      });
+      it('should call not call onChange when value is cleared by clicking delete button ', () => {
+        const onChange = jest.fn();
+        const { getByLabelText, getByText, queryByText } = renderInput({
+          onChange,
+          isMulti: true,
+          value: ['mango'],
+          isDisabled: true,
+        });
+        const input = getByLabelText('Fruit');
+        fireEvent.focus(input);
+        const deleteButton = getByText('Mango').nextSibling;
+        deleteButton.click();
         expect(onChange).not.toHaveBeenCalled();
         expect(queryByText('Mango')).toBeInTheDocument();
       });
--- a/src/components/inputs/select-input/select-input.spec.js
+++ b/src/components/inputs/select-input/select-input.spec.js
@@ -181,6 +181,22 @@ describe('in multi mode', () => {
     });
   });
   describe('interacting', () => {
+    describe('when disabled', () => {
+      it('should call not onChange when value is cleared', () => {
+        const onChange = jest.fn();
+        const { getByLabelText, queryByText } = renderInput({
+          onChange,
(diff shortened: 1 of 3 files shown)


src/components/inputs/select-input/select-input.spec.js
```diff
@@ -181,6 +181,22 @@ describe('in multi mode', () => {
...
+          value: ['mango'],
+          isDisabled: true,
+        });
+        const input = getByLabelText('Fruit');
+        fireEvent.focus(input);
+        fireEvent.keyDown(input, { key: 'Backspace' });
+        expect(onChange).not.toHaveBeenCalled();
+        expect(queryByText('Mango')).toBeInTheDocument();
+      });
+    });
```
montezume: I didn't add a test for clicking to delete because of [this](https://github.com/testing-library/react-testing-library/issues/356)

src/components/internals/create-select-styles.js
```diff
@@ -337,7 +337,7 @@ const multiValueRemoveStyles = (props, theme) => base => {
     } 0`,
     borderStyle: 'solid',
     borderWidth: '1px',
-
+    pointerEvents: state.isDisabled ? 'none' : base.pointerEvents,
```
montezume: another option would be in TagRemove removing the onClick from the spread props 🤔

tdeekens: No opinion on it. I think pointerEvents should be `none` anyway?

montezume: True, it's not one or the other, it's one or both 😄

tdeekens: If it makes it easier to test I'd `null` the `onChange` too.

src/components/inputs/select-input/select-input.spec.js
```diff
@@ -181,6 +181,22 @@ describe('in multi mode', () => {
     });
   });
   describe('interacting', () => {
+    describe('when disabled', () => {
+      it('should call not onChange when value is cleared', () => {
```
emmenko:
```suggestion
      it('should not call onChange when value is cleared', () => {
```

montezume: 😄

src/components/internals/tag-remove/tag-remove.js
```diff
@@ -2,14 +2,23 @@ import React from 'react';
...
-  <div {...props.innerProps}>
-    <CloseBoldIcon
-      theme={props.selectProps.isDisabled ? 'grey' : 'black'}
-      size="medium"
-    />
-  </div>
-);
+const TagRemove = props => {
+  let innerProps = props.innerProps;
+  if (props.selectProps.isDisabled) {
```
montezume: if it's disabled, then we don't spread the `onClick, onTouchStart and onTouchEnd` events onto the surrounding div.

src/components/inputs/select-input/select-input.spec.js
```diff
@@ -181,6 +181,37 @@ describe('in multi mode', () => {
...
+          value: ['mango'],
+          isDisabled: true,
+        });
+        const input = getByLabelText('Fruit');
+        fireEvent.focus(input);
+        fireEvent.keyDown(input, { key: 'Backspace' });
+        expect(onChange).not.toHaveBeenCalled();
+        expect(queryByText('Mango')).toBeInTheDocument();
+      });
+      it('should call not call onChange when value is cleared by clicking delete button ', () => {
```
montezume: test is back 😄

src/components/internals/tag-remove/tag-remove.js
emmenko: Can you leave a code comment with that explanation?

montezume: sure

montezume: I think [this](https://github.com/commercetools/ui-kit/pull/855/commits/a186362383f1eb7d5edcaa2b79727b203d28d62d) is more clear

Issue #853, opened 2019-06-12
Expose TS declarations inferred from prop-types


emmenko:
We can try to expose TS declarations, which can be inferred from prop types (https://dev.to/busypeoples/notes-on-typescript-inferring-react-proptypes-1g88)

Doing so will improve the usage of ui-kit components by leveraging TS declarations, while we think about migrating the codebase to TypeScript.

Roughly, what needs to be done is the following:
-  setup typescript tooling (which is a bit easier here since it's not a monorepo)
- add script to generated type declarations
- define the "infer" function (https://github.com/commercetools/merchant-center-application-kit/blob/b46a985c655383c02d773c63776e60e9249b987a/packages/sentry/src/type-utils.ts#L1-L12)
- export inferred types from each component, based on the prop-types
- make sure the package.json reference the types

That's what I would expect to happen, unless we encounter unexpected problems.

Maybe we can plan this as a tech day or something.

emmenko:
Not relevant anymore as we plan to eventually migrate to TS.

PR #860, opened 2019-06-15
feat: add initial type declarations for exported components

emmenko:
This PR aims to provide TS declarations for the exported ui-kit components and variables.

It uses a tool called [dtslint](https://github.com/microsoft/dtslint). Other libraries (e.g. dom-testing-library) use it as well.

**I tested the types in appkit** and so far things look good.

To avoid too many changes here, we can gradually add types to the remaining exports as follow ups.

I consider this a first step of providing type declarations for ui-kit, which is currently necessary in appkit (to avoid [this](https://github.com/commercetools/merchant-center-application-kit/blob/master/%40types/commercetools-frontend__ui-kit/index.d.ts)).


package.json
```diff
@@ -60,7 +62,8 @@
     "percy": "percy exec -- jest --config jest.visual.config.js --runInBand",
     "visual-testing-app:build": "NODE_ENV=production webpack --config visual-testing-app/webpack.config.js",
     "visual-testing-app:build:watch": "NODE_ENV=production webpack --config visual-testing-app/webpack.config.js --watch",
-    "visual-testing-app:serve": "serve visual-testing-app/dist --no-clipboard --single --listen 3001"
+    "visual-testing-app:serve": "serve visual-testing-app/dist --no-clipboard --single --listen 3001",
+    "typecheck": "dtslint typings"
```
montezume: I wouldn't call the script "typecheck" since we don't have any types to check...

emmenko: I don't understand your point, we do have types. 🤔

.circleci/config.yml
```diff
@@ -48,6 +48,7 @@ commands:
     description: 'Running linters'
     steps:
       - run: yarn run jest --projects jest.eslint.config.js jest.stylelint.config.js --maxWorkers=3 --reporters jest-silent-reporter
+      - run: yarn typecheck
```
montezume: So, if the typings update, does it also commit the changes to the file? 😕

package.json
montezume: dom-testing-library calls it dtslint, which is a better name.

https://github.com/testing-library/dom-testing-library/blob/4b2e8a535d8e88bc03126307cbee48a74dbf547e/package.json

We aren't using typescript, so calling a script "typecheck" feels weird. Does this script test our PropTypes? No...

package.json
```diff
@@ -84,6 +87,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.4.5",
+    "@babel/parser": "7.4.5",
```
montezume: why do we need this?

package.json
emmenko: From the https://github.com/microsoft/dtslint readme:

> `dtslint` tests a TypeScript declaration file for style and correctness. It will install `typescript` and `tslint` for you, so this is the only tool you need to test a type definition.

So yes, it obviously uses typescript to be able to perform type checks.

> dom-testing-library calls it dtslint

I know, but in all our repos when we use types we call this script `typecheck`, so I just went with that.

I think you're misunderstanding a bit what this tool does...

.circleci/config.yml
emmenko: I don't understand your question. Are you sure you understand what this tool does?

package.json
emmenko: One of the dependency's declaration files (don't recall which one) requires that.
You can leave it out and you'll see the error if you want to know more about it.

.circleci/config.yml
montezume: What does this tool do then? Does this PR not intend to deliver https://github.com/commercetools/ui-kit/issues/853?

package.json
montezume: If you create a pull request with one line of text for it's summary, it's not unusual for people to misunderstand it...

.circleci/config.yml
emmenko: No this is a different approach from the prop-types. Probably it won't really work because we need to have typescript files in order for the compiler to extract the declarations, even if we infer them from the prop-types.

This approach though requires some manual work in terms of maintenance but allows to expose declarations without having to setup typescript and migrate stuff.
The `dtslint` tool allows to "check" the type declarations that we define if it's valid syntax or not. It obviously does not guarantee that the types are "in sync" with the components but it's a good start.
This is also what you would use to publish type declarations of a JS library to [definitely typed](https://github.com/DefinitelyTyped/DefinitelyTyped).

package.json
emmenko: Sure but to be fair it's a "draft" PR, I haven't asked for a review yet and it's marked as experimental 🙂

.circleci/config.yml
montezume: Ah okay, so to clarify.

Step 1, you write your Typescript types for your component
Step 2, you run `dtslint` to check those values.

Okay yeah I get it now.

But now we need to maintain our typings, our readmes, and our proptypes, all independently. I don't really see how it helps us that much.

emmenko: Well, eventually we would probably want to only have the types, so we might need to migrate the codebase to TS.

The readme stuff I think is a general problem regardless if we use TS or not. Maybe we could also write our own docs parser if we don't find the right tool.

montezume: > eventually we would probably want to only have the types, so we might need to migrate the codebase to TS

Like I have mentioned before, I would like to triage our UI-Kit infrastructure refactorings, and prioritize them. Justifying the TS migration this way feels a bit like jumping the queue...

montezume: Since we can't generate the typings, manually making them seems like too much work to do without including our PO and defining the task.

emmenko: Sorry but no one said to migrate anything nor to prioritize this. I'm not sure where did you get that impression.

That's why I looked into this approach which is very lightweight and does not involve any change in the existing "infrastructure". It's also marked as "experiment" to emphasis that.

montezume: Alright, so to confirm, you don't plan on manually adding types for all of the components?

montezume:
Can you update the PR Description?

emmenko:
> Can you update the PR Description?

Already did...

montezume:
Can we have a talk about this when I get back?

montezume:
Woops sorry

montezume:
So for the dumb ones like me, what this does is it keeps the types we write manually in check? It verified that they are correct ? So we don’t need to worry about them going out of date?

emmenko:
I’m not aware of a way to keep the types in sync with the JS code. So to answer your question, there is no assurance that the types are correct all the times. This is a general caveat though with all the manually maintained types of JS libraries out there.

If this is the main concern, the solution is to simply have the codebase typed.

However, for the time being I think this is a good solution that pleases both sides and will buy us time to eventually migrate to a fully typed codebase. We just need to be aware and remember to update the types definition when we change the components API. Even if we forget about it, it’s not the end of the world.

scripts/generate-custom-properties.js
```diff
@@ -152,6 +179,11 @@ fs.writeFileSync(
   prettier.format(printJavaScript(tokens), prettierConfig)
 );
 
+fs.writeFileSync(
+  path.join(__dirname, '../materials/custom-properties.ts'),
+  prettier.format(printTypeScript(tokens), prettierConfig)
```
montezume: 👍

src/components/typography/text/text.js
```diff
@@ -266,7 +266,6 @@ Detail.displayName = 'TextDetail';
 Detail.propTypes = {
   isBold: PropTypes.bool,
   isItalic: PropTypes.bool,
-  isInline: PropTypes.bool,
```
montezume: is this a breaking change?

emmenko: Ah sorry, my bad

emmenko:
> I think it would be useful if this PR added typings for a "indicative" component, like an input.

I added a section in the `CONTRIBUTING.md` file to document how to add type declarations for a component: https://github.com/commercetools/ui-kit/pull/860/commits/1a5769f2cfbb1a817e800602a2d9f0ef25b99480

Let me know if that's enough.

jonnybel:
Looks like I really have to start doing my TS homework 😁

emmenko:
We decided to not proceed further with this (for now), as it's not a priority for the uikit team at the moment.
Instead we will provide type declarations for ui-kit as part of the appkit repo (which already uses TS).

PR #869, opened 2019-06-18
style(flat-button): implement inverted tone

ahmehri:
#### Summary

Add `inverted` tone to `FlatButton` component.

#### Description

This is needed when placing `FlatButton` on a dark background which makes the content not visible enough.

### DIFF
--- a/src/components/buttons/flat-button/flat-button.js
+++ b/src/components/buttons/flat-button/flat-button.js
@@ -16,14 +16,19 @@ const getIconElement = props => {
   });
 };
 
-const getTextColor = (tone, isHover = false) => {
+const getTextColor = (tone, isHover = false, theme) => {
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
+  };
+
   switch (tone) {
     case 'primary':
-      return isHover ? vars.colorPrimary25 : vars.colorPrimary;
+      return isHover
+        ? overwrittenVars.colorPrimary25
+        : overwrittenVars.colorPrimary;
     case 'secondary':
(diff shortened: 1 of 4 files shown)


src/components/buttons/flat-button/flat-button.js
```diff
@@ -16,12 +16,19 @@ const getIconElement = props => {
...
+
   switch (tone) {
     case 'primary':
-      return isHover ? vars.colorPrimary25 : vars.colorPrimary;
+      return isHover
+        ? overwrittenVars.colorPrimary25
+        : overwrittenVars.colorPrimary;
     case 'secondary':
-      return vars.colorSolid;
+      return overwrittenVars.colorSolid;
```
montezume: if it's a dark background, it makes more sense to wrap that background in a ThemeProvider, overriding the colorSolid.

montezume:
This is what it looks like in the VRT app
Screen Shot 2019-06-18 at 11 52 58 AM

montezume:
This way, if you included other UI-Kit elements that support theming, like text for example, you could do something like this.

```js
<ThemeProvider theme={{colorSurface: 'black', colorSolid: 'white'}}>
    <FlatButton theme='secondary' label="Click me" />
    <Text.Body>Hello world</Text.Body>
</ThemeProvider>

```

montezume:
I guess we could also add "inverted" prop, with duplicated functionality, but I would like to remove the inverted prop from the other components in the next breaking change release, so I don't really see the point 🤷‍♂

PR #871, opened 2019-06-18
fix(icon-button): hover state fix, and remove HOC

montezume:
#### Summary

Fixes a bug with hovering IconButtons with "green" theme. Also, refactors the component to remove the usage of the HOC.

#### Before

before-green

#### After
after-green

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,15 +1,15 @@
 {
   "dist/ui-kit.esm.js": {
-    "bundled": 879366,
-    "minified": 604150,
-    "gzipped": 123195,
+    "bundled": 879277,
+    "minified": 603873,
+    "gzipped": 123184,
     "treeshaked": {
       "rollup": {
-        "code": 434499,
+        "code": 434486,
         "import_statements": 919
       },
       "webpack": {
-        "code": 458773
+        "code": 458770
       }
     }
   }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,15 +1,15 @@
 {
   "dist/ui-kit.esm.js": {
-    "bundled": 866961,
-    "minified": 596248,
-    "gzipped": 121235,
+    "bundled": 879366,
+    "minified": 604150,
+    "gzipped": 123195,
     "treeshaked": {
       "rollup": {
-        "code": 429763,
+        "code": 434499,
         "import_statements": 919
       },
       "webpack": {
-        "code": 454131
+        "code": 458773
       }
     }
   }
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -73,21 +73,16 @@ export const IconButton = props => {
         css={css`
           width: 100%;
           height: 100%;
+          display: flex;
+          align-items: center;
+          justify-content: center;
         `}
       >
-        <div
-          css={css`
-            display: flex;
-            align-items: center;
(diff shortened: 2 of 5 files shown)


src/components/buttons/icon-button/icon-button.styles.js
```diff
@@ -169,6 +169,7 @@ const getThemeStyles = theme => {
   switch (theme) {
     case 'green':
       return css`
+        &:hover,
         &:active {
```
montezume: this was missing hover, so you never saw green on hover

src/components/buttons/icon-button/icon-button.js
```diff
@@ -73,21 +68,16 @@ export const IconButton = props => {
         css={css`
           width: 100%;
           height: 100%;
+          display: flex;
```
montezume: my hope is this fixes @adnasa's issue

src/components/buttons/icon-button/icon-button.visualroute.js
```diff
@@ -123,22 +123,24 @@ export const component = () => (
...
     </Spec>
 
-    <Spec label="theme - when blue">
+    <Spec label="theme - when blue - when toggled">
       <IconButton
         icon={<InformationIcon />}
         label="A label text"
         onClick={() => {}}
         isToggleButton={true}
+        isToggled={true}
```
montezume: these tests make more sense now because the output will actually be blue...

src/components/buttons/icon-button/icon-button.js
```diff
@@ -42,10 +40,6 @@ export const IconButton = props => {
   const isActive = props.isToggleButton && props.isToggled;
   return (
     <div
```
jonnybel: Is it also possible to get rid of this `div` altogether and apply the styles to the AccessibleButton component?

src/components/buttons/icon-button/icon-button.styles.js
```diff
@@ -169,7 +169,7 @@ const getThemeStyles = theme => {
   switch (theme) {
     case 'green':
       return css`
-        &:active {
```
montezume: this was the bug

src/components/buttons/icon-button/icon-button.js
montezume: giving it a shot now

.size-snapshot.json
```diff
@@ -1,15 +1,15 @@
 {
   "dist/ui-kit.esm.js": {
-    "bundled": 866961,
-    "minified": 596248,
-    "gzipped": 121235,
+    "bundled": 879277,
```
montezume: the problem is we don't run "yarn build" enough so this is always outdated.

src/components/buttons/icon-button/icon-button.js
montezume: okay removed it, and VRTs passed 🤷‍♂

jonnybel: Nice!
I always found it weird that we had an AccessibleButton, which already deals with the mouse events, but then wrap it on a dull `div` which then reimplemented those mouse events with the HoC... I guess there must have been some other reason before, but something that got refactored? 🤔

 It makes sense that it can be removed given how useless it seemed at the moment 🤷‍♂

montezume: do you want to approve then 🧐

PR #876, opened 2019-06-19
feat(storybook): add context add for theme

montezume:
#### Summary

As theming is still a beta feature, we don't have much documentation about it. Not all components support theming. Most inputs do at the moment. 

This PR adds the ability in storybook to change to a (very basic) dark theme.

```js
const darkTheme = {
  colorSolid: vars.colorSurface,
  colorSurface: vars.colorSolid,
};
```

dark theme

### DIFF
--- a/.storybook/configs/theme-context.js
+++ b/.storybook/configs/theme-context.js
@@ -10,9 +10,9 @@ const darkTheme = {
 
 const defaultTheme = vars;
 
-const ThemeWrapper = props => {
-  return <ThemeProvider theme={props.theme}>{props.children}</ThemeProvider>;
-};
+const ThemeWrapper = props => (
+  <ThemeProvider theme={props.theme}>{props.children}</ThemeProvider>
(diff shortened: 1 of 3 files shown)


.storybook/configs/theme-context.js
```diff
@@ -0,0 +1,45 @@
...
+
+const darkTheme = {
+  colorSolid: vars.colorSurface,
+  colorSurface: vars.colorSolid,
+};
+
+const defaultTheme = vars;
+
+const ThemeWrapper = props => {
+  return <ThemeProvider theme={props.theme}>{props.children}</ThemeProvider>;
```
tdeekens: Nit: you just implicit return.

montezume:
FYI, what I would like to do is hook into the event emitted by the context addon and change the overall Storybook theme.

https://github.com/storybookjs/storybook/issues/6418

But that can be a follow up.

PR #884, opened 2019-06-21
feat: require react 16.8

montezume:
#### Summary

Since App Kit will be requiring React 16.8 in it's next major release, we should do the same. [ref](https://github.com/commercetools/merchant-center-application-kit/issues/748)

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,8 +201,8 @@
   "peerDependencies": {
     "moment": "2.x",
     "moment-timezone": "0.5.x",
-    "react": "16.x",
-    "react-dom": "16.x",
+    "react": ">= 16.8.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -201,8 +201,8 @@
   "peerDependencies": {
     "moment": "2.x",
     "moment-timezone": "0.5.x",
-    "react": "16.x",
-    "react-dom": "16.x",
+    "react": ">= 16.8.0",
```
montezume: I wasn't 100% sure on this, but this is the same syntax that `styled-components` are using on their `5.0.0` branch 🤷‍♂

tdeekens: Yeah we use the same syntax on react-router and other deps I think.

PR #893, opened 2019-06-24
feat: add postcss codemod

montezume:
#### Summary

To make it easier to update to the upcoming UI-Kit version 10, we can provide a postcss codemod to replace outdated custom properties with their replacements.

Before

Screen Shot 2019-06-24 at 11 49 18 AM

Run the command

Screen Shot 2019-06-24 at 11 47 07 AM

After

Screen Shot 2019-06-24 at 11 49 38 AM

What do you think?

I was thinking if this is useful to create it as a separate repo / package, and then we would just keep the variable-mapping jsons in UI Kit.


emmenko:
Shouldn't we target the v10 branch?

montezume:
@emmenko Why? This way consumers can migrate prior to the release of the branch. There's nothing stopping them from doing it now. Both new and old custom properties exist.

emmenko:
Ah you want to keep it separated? I thought we just include it in the migration script with the other codemods for v10 so that we only need to run 1 migration script once.

montezume:
Since I'm not sure when exactly we will release v10, I think it makes sense to get this in first. As part of the v10 release we can update the mapping file as we make more changes 🤷‍♂.

codemods/custom-property-replacer/README.md
```diff
@@ -0,0 +1,31 @@
+#### Custom Property Replacer
+
+The purpose of this PostCSS plugin is to replace deprecated custom properties with their replacement custom properties. To use this plugin, follow the following instructions.
```
jonnybel: "replace x with their replacement y" sounds a bit redundant... How about:
```suggestion
The purpose of this PostCSS plugin is to replace deprecated custom properties with the new custom properties. To use this plugin, follow the following instructions.
```

codemods/custom-property-replacer/README.md
```diff
@@ -0,0 +1,31 @@
+#### Custom Property Replacer
+
+The purpose of this PostCSS plugin is to replace deprecated custom properties with the new custom properties. To use this plugin, follow the following instructions.
+
+First, start by installing `postcss-cli`. You will use this to run the css transformation.
+
+`yarn add -D postcss-cli`
```
emmenko: I would maybe recommend to use `npx` to run the command, as it's usually a one time thing, so there is no need to have an extra dependency installed

codemods/custom-property-replacer/variable-mapping.json
```diff
@@ -0,0 +1,46 @@
+{
```
emmenko: How about we scope those changes into "version folders", so we make them "immutable"?

```
codemods/custom-property-replacer/v10/variable-mapping.json
```

Then in the future we can have more of them...

codemods/custom-property-replacer/tests/custom-property-replace.spec.js
```diff
@@ -0,0 +1,40 @@
+import postcss from 'postcss';
+import path from 'path';
```
emmenko: Nitpick: define imports for nodejs modules first

codemods/custom-property-replacer/index.js
```diff
@@ -0,0 +1,34 @@
...
+module.exports = postcss.plugin('postcss-var-replacer', opts => {
+  let variables = {};
+  if (opts && opts.file) {
+    variables = JSON.parse(fs.readFileSync(opts.file));
+  }
+
+  return css => {
+    css.walkDecls(decl => {
+      if (!opts || !opts.file) {
+        throw new Error(
```
emmenko: This error should be thrown above, since you already try to read from the file.

montezume:
@emmenko your changes have been implemented 😄

PR #894, opened 2019-06-25
refactor(breaking): remove font-family

montezume:
#### Summary

As discussed, we want to remove the font-family declarations from UI-Kit and to make it the responsibility of the consumer.

I'll aim the PR at `10.0.0` after Percy runs 😄

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -5,7 +5,6 @@
     margin: 0;
     padding: 0;
     font-size: 13px;
-    font-family: 'Open Sans', sans-serif;
   }
 
   body {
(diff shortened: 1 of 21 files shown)


montezume:
Ah, this is annoying

https://stackoverflow.com/questions/6080413/input-doesnt-inherit-the-font-from-body

montezume:
Ah I guess we can specify font-family: inherit on these kind of components...

jonnybel:
Hmm, if we don't set a `font-family`, isn't it inherited by default? (from the nearest parent with a font-family setting)

montezume:
@jonnybel yes but not for the element types I posted

jonnybel:
Oh, nvm, I see that is precisely the problem with inputs.

montezume:
Would be nice if Percy disabled itself when PRs aren't pointed at master...

PR #899, opened 2019-06-27
feat(spacings.inline): support flex wrap

### LINKED ISSUE #898: `Spacings.Inline` should support `flex-wrap`
#### Summary

The `Spacings.Inline` component should support the flex-wrap property.

montezume:
#### Summary

Sometimes when we want to make things "flex-wrap". IMHO this is a nice addition to this component, which means we can make more UI elements with needing custom css

[more info](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap)
Fixes #898

### DIFF
--- a/src/components/spacings/inline/inline.js
+++ b/src/components/spacings/inline/inline.js
@@ -22,6 +22,13 @@ Inline.propTypes = {
     'flexStart',
     'flexEnd',
   ]),
+  flexWrap: PropTypes.oneOf([
+    'nowrap',
+    'wrap',
+    'wrap-reverse',
+    'initial',
+    'inherit',
+  ]),
   justifyContent: PropTypes.oneOf([
     'flex-start',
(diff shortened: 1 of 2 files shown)


montezume:
Now that I think about it, maybe this doesn't make sense.

Take a look at storybook output

Screen Shot 2019-06-27 at 3 31 47 PM

In cases where you wrap, you wouldn't want margin like this, right?

emmenko:
Maybe then you can just use `Grid` if you want elements to wrap, then we don't need `flex-wrap` anymore

jonnybel:
Hmm, yes, the wrapping is not very "compatible" with the margins.

In cases like these I'd rather use a custom styled div.

emmenko:
> In cases like these I'd rather use a custom styled div

Just use css grid 😉

PR #910, opened 2019-07-03
feat(date-inputs): add tooltip to "today" button

montezume:
#### Summary

Adds tooltip for the `today` button for DateInputs.

#### Limitations

Because the header is quite small, we cannot use placement "top" or placement "left", as there isn't enough space, so the positioning library (popper) will automatically place it on the bottom.

today


src/components/internals/calendar-header/calendar-header.js
```diff
@@ -1,11 +1,19 @@
...
 import vars from '../../../../materials/custom-properties';
 import Text from '../../typography/text';
 import Spacings from '../../spacings';
 import { AngleLeftIcon, AngleRightIcon, CircleIcon } from '../../icons';
+import Tooltip from '../../tooltip';
 import SecondaryIconButton from '../../buttons/secondary-icon-button';
+import messages from './messages';
+
+const WrapperComponent = styled.div`
+  display: flex;
```
montezume: this is required because the default wrapper is `display: inline-block`

i18n/data/de.json
```diff
@@ -6,6 +6,11 @@
   "UIKit.AsyncSelectInput.noOptionsMessageWithInputValue": "",
   "UIKit.AsyncSelectInput.noOptionsMessageWithoutInputValue": "",
   "UIKit.AsyncSelectInput.placeholder": "Auswählen...",
+  "UIKit.CalendarHeader.nextMonthTooltip": "nächstes Monat",
```
montezume: 😅

montezume:
FYI I opened a JIRA ticket for the missing translations.

montezume:
Is it okay from your side @Shecki ?

Shecki:
es, its fine

PR #911, opened 2019-07-03
fix(link): apply underlined styles

montezume:
#### Summary

Previously we did not set `text-decoration: underline` in the `Link` component. This was an oversight, as consumers who have competing global styles could inadvertently disable the underline. 

I also added a `hasUnderline` prop in the rare case that consumers want to remove the underline.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,15 +1,15 @@
 {
   "dist/ui-kit.esm.js": {
-    "bundled": 874809,
-    "minified": 600844,
-    "gzipped": 122315,
+    "bundled": 882007,
+    "minified": 604581,
+    "gzipped": 123515,
     "treeshaked": {
       "rollup": {
-        "code": 428635,
+        "code": 431686,
         "import_statements": 893
       },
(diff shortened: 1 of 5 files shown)


src/components/links/link/link.js
```diff
@@ -69,6 +72,7 @@ Link.propTypes = {
 };
 
 Link.defaultProps = {
+  hasUnderline: true,
```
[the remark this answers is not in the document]
adnasa: Looks underlined to me

.size-snapshot.json
```diff
@@ -1,15 +1,15 @@
...
+    "gzipped": 123515,
     "treeshaked": {
       "rollup": {
-        "code": 428635,
+        "code": 431686,
         "import_statements": 893
       },
       "webpack": {
-        "code": 455406
+        "code": 458492
```
tdeekens: ...

montezume: Because we often make PRs and merge them without having ran yarn build. It gets out of sync. I think I’ll just remove it

Issue #829, opened 2019-06-04
Keeping track of bundle size


montezume:
Size snapshot is only ran when you run

`yarn build`

This can lead to the size-snapshot.json getting out of sync.

Let's take an example.

* Person A adds new component to UI-Kit
* Person A works with storybook, and never runs `yarn build`
* Person A makes pull request
* Person A's new component is merged to master
* Person B makes a different contribute, but said person runs `yarn build`
* Now size-snapshot is ran and the json is updated, but the "real" changes come from both Person A's new component and Person B's work.

tdeekens:
That's also why I don't like to have this maintained manutally. On flopflip I use bundlephobia.

https://bundlephobia.com/result?p=@commercetools-frontend/ui-kit@9.4.0

emmenko:
I think we can use `matchSnapshot` option to enforce the file to stay in sync.

In any case, from my side I don't mind much which solution we use. My goal with that was to have an indicator of the sizes etc. If bundlephobia is good enough, we can also use that.

montezume:
I tried using `matchSnapshot` and it just failed everytime. I guess I need to delete the snapshot file? 🤔

emmenko:
We can use this: https://github.com/siddharthkp/bundlesize

It comes with a github app/bot.

tdeekens:
Yeah. We talked on Slack about adding bundlesize or wrting a rollup-plugin for packtracker. I think we'll go with bundlesize for now.

PR #913, opened 2019-07-04
chore: remove size snapshot

montezume:
#### Summary

The size snapshot file gets out of sync and it doesn't provide value. 

We can readd it if we come up with a solution where in every PR, the size snapshot accurately reflects the changes in that PR. But for now, this is not the case.


montezume:
I renamed the issue https://github.com/commercetools/ui-kit/issues/829

We can track this there

montezume:
Going to merge this as we are going with `bundlesize`

PR #914, opened 2019-07-04
feat: setup bundlesize

### LINKED ISSUE #829: Keeping track of bundle size

montezume:
#### Summary

Adds `bundlesize` to track our bundlesize in PRs

Closes #829

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,10 @@ commands:
     description: 'Running bundle tests'
     steps:
       - run: yarn run jest --projects jest.bundle.config.js --maxWorkers=3 --reporters jest-silent-reporter
+  bundle_size_test:
+    description: 'Running bundle size'
+    steps:
+      - run: yarn run test:bundlesize
   vrt_test:
     description: 'Running Visual Regression Tests'
     steps:
@@ -131,6 +135,12 @@ jobs:
       - install_dependencies
       - restore_workspace
       - bundle_test
+  bundle_size_test:
+    executor: node_10
+    steps:
+      - install_dependencies
+      - restore_workspace
+      - bundle_size_test
   vrt_test:
     executor: node_10
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,6 +172,12 @@ workflows:
           filters:
             tags:
               only: /v[\.0-9]+.*/
+      - bundle_size_test:
+          requires:
+            - build
+          filters:
+            tags:
+              only: /v[\.0-9]+.*/
       - vrt_test:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ commands:
     description: 'Building artifacts'
     steps:
       - run: yarn build
+      - run: yarn bundlesize
   save_workspace:
     steps:
       - persist_to_workspace:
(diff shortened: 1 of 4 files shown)


tdeekens:
Huuzaaah.

montezume:
I added the token to our CI variables. I'm not sure if anything is missing 🤔

bundlesize.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "files": [
+    {
+      "path": "./dist/ui-kit.esm.js",
+      "maxSize": "175 kB"
```
emmenko: Hmm are you sure this is correct?

https://bundlephobia.com/result?p=@commercetools-frontend/ui-kit@9.9.0

package.json
```diff
@@ -122,6 +122,7 @@
     "babel-plugin-transform-rename-import": "2.3.0",
     "babel-preset-jest": "24.6.0",
     "browserslist": "4.6.3",
+    "bundlesize": "0.18.0",
```
emmenko: We also need to trigger the bundlesize on CI...

bundlesize.config.json
montezume: Can you give more information? What is correct? The max size?

package.json
montezume: how do we do this? I can't find instructions

montezume: I added the token to our env vars

emmenko: Just add a step on CI to execute `bundlesize`. That's it.

After this: https://github.com/commercetools/ui-kit/blob/20cd7a4f5a2b2b75c6e9052712631253a98a5c6b/.circleci/config.yml#L36

Just add 

```
- run: yarn run bundlesize
```

bundlesize.config.json
emmenko: Yes the max size...currently it's like >700KB

.circleci/config.yml
```diff
@@ -131,6 +135,12 @@ jobs:
       - install_dependencies
       - restore_workspace
       - bundle_test
+  bundle_size_test:
```
emmenko: No new job please, just keep it simple: https://github.com/commercetools/ui-kit/pull/914#discussion_r300316270

bundlesize.config.json
montezume: No, it's correct. When I run bundlesize it gives ~165.

montezume: Screen Shot 2019-07-04 at 11 40 34 AM
which is very close to your links result...

package.json
montezume: done

bundlesize.config.json
emmenko: Ah ok it gzippes it

image

montezume:
Should I add this to app kit as well? 🤔

bundlesize.config.json
montezume: In fact I think we should lower this. It should be maybe 2-5%?

PR #916, opened 2019-07-04
v10 - support react-intl v3

montezume:
#### Summary

Preparation for next release. Now UI-Kit can be used with the latest (beta) version of react-intl. 

This PR is aimed at v10


package.json
```diff
@@ -53,11 +53,11 @@
     "format:js": "prettier --write '**/*.js'",
     "format:md": "prettier --write --parser markdown '**/*.md'",
     "format:yaml": "prettier --write --parser yaml '**/*.yaml'",
-    "test": "jest --config jest.test.config.js",
-    "test:watch": "jest --config jest.test.config.js --watch",
-    "test:bundle": "jest --config jest.bundle.config.js",
+    "test": "cross-env NODE_ICU_DATA=node_modules/full-icu jest --config jest.test.config.js",
```
montezume: From here

https://github.com/formatjs/react-intl/issues/1179#issuecomment-428183019

montezume: Otherwise tests that rely on locale stuff will fail

package.json
```diff
@@ -53,11 +53,11 @@
...
+    "test:watch": "cross-env NODE_ICU_DATA=node_modules/full-icu jest --config jest.test.config.js --watch",
+    "test:bundle": "cross-env cross-env NODE_ICU_DATA=node_modules/full-icu jest --config jest.bundle.config.js",
     "bundlesize": "bundlesize",
-    "test:bundle:watch": "jest --config jest.bundle.config.js --watch",
+    "test:bundle:watch": "cross-env NODE_ICU_DATA=node_modules/full-icu jest --config jest.bundle.config.js --watch",
```
montezume: maybe don't need this for these

package.json
tdeekens: Sadly npm hooks don't support wildcards (e.g. `pretest*`) but maybe a `setup` run script is nice to ease this and then just have `yarn setup && jest ...`.

montezume: Hm... can you clarify what `setup` would look like? 😄

emmenko: In appkit I did it like this:

https://github.com/commercetools/merchant-center-application-kit/blob/b6200086597c97640852052618cedc7cdb17f3b5/packages/jest-preset-mc-app/setup-test-framework.js#L26-L31

package.json
```diff
@@ -206,7 +207,7 @@
     "moment-timezone": "0.5.x",
     "react": "16.x",
     "react-dom": "16.x",
-    "react-intl": "2.x",
+    "react-intl": "2.x || 3.x",
```
emmenko: Are you sure we can make it backwards compatible? Can't we just include it in v10?

package.json
tdeekens: `"setup": "cross-env NODE_ICU_DATA=node_modules/full-icu"` :D

montezume: I tried this first 😢 

I think it's a different issue.

package.json
montezume: I think the only real breaking change is the dropping of `intlShape` and the polyfills.

montezume: Maybe to err on the safe side we should aim this at v10 though...

package.json
montezume: I tried this and it didn't work 🤔 ? In the test I put yarn setup && jest ...

emmenko: I don't think installing the full icu is necessary. Patching the Intl in the jest setup fixed the problem for me.

What is the problem you're having?

tdeekens: Ah spawned in different shell maybe.

montezume: Turns out just needed more polyfills 😓

montezume:
Alright this is ready for review 😇

montezume:
Does nobody want to review this 🤔

PR #926, opened 2019-07-08
10.0.0 - Update codemod

montezume:
#### Summary

Updates the codemods to be specific for our v10.0.0 release.


montezume:
Hm, this doesn't update the new values for Icons though. Any ideas @jonnybel ? Do you know an easy way to not only update props, but their values as well? 🤔

jonnybel:
> Hm, this doesn't update the new values for Icons though. Any ideas @jonnybel ? Do you know an easy way to not only update props, but their values as well? 🤔

Ah, not out the top of my head, but it must be readable and settable from some variation of the "path.value.name **[value]** " or "path.value.name.name **[value]** "

montezume:
Alright I updated the codemod to update the values for our Icon codemod

PR #927, opened 2019-07-09
fix(date-time-input): Today button not highlighting today's date

jonnybel:
We found a bug where the Today button in `DateTimeInput` would not highlight today's date, but the last day of the month instead.

This fixes that, and also for `DateTimeField`, as it is built with the input component.


src/components/inputs/date-time-input/date-time-input.js
```diff
@@ -99,7 +98,7 @@ class DateTimeInput extends React.Component {
         calendarDate: today,
         highlightedIndex:
           prevState.suggestedItems.length +
-          getDaysInMonth(today, this.props.timeZone) -
```
montezume: 😆

PR #934, opened 2019-07-12
chore: add missing german translations

montezume:
Translations from Franzi

@tdeekens can you do me a favour and double check that I didn't copy paste into the wrong spot? Last PR on a friday afternoon and I might have made a mistake. 🙏 🇩🇪


montezume:
looks like I made some mistake with escaping the "" in the translations. Can someone help out on Monday 🙏

PR #951, opened 2019-07-22
chore: update deps

montezume:
#### Summary

Updates deps


scripts/setup-test-framework.js
```diff
@@ -5,7 +5,7 @@
 // of the preset is on NPM.
 
 // react-testing-library setup
-import 'jest-dom/extend-expect';
+import '@testing-library/jest-dom/extend-expect';
```
emmenko: ah right, I'll change that in appkit as well

PR #955, opened 2019-07-25
[UKIT-85] fix: truncate select-input placeholder text is long

jonnybel:
Fixes Jira Issue: [UKIT-85](https://jira.commercetools.com/browse/UKIT-85)

Before:
image

After:
image


filippobocik:
Thanks for the quick fix

PR #957, opened 2019-07-25
refactor: use hooks

montezume:
#### Summary

Refactors the `select` components to use hooks, as well as `PasswordField` and some smaller components.

### DIFF
--- a/src/components/fields/password-field/password-field.js
+++ b/src/components/fields/password-field/password-field.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { injectIntl } from 'react-intl';
+import { useIntl } from 'react-intl';
 import requiredIf from 'react-required-if';
 import Constraints from '../../constraints';
 import Spacings from '../../spacings';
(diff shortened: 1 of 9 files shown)


src/components/fields/password-field/password-field.story.js
```diff
@@ -54,7 +54,7 @@ storiesOf('Components|Fields', module)
               isRequired={boolean('isRequired', false)}
               touched={boolean('touched', false)}
               name={text('name', '')}
-              value={text('value', value)}
```
montezume: this was annoying, it's better to use the value so that the user can type into the field itself and not into the knob.

tdeekens:
Nice. So UIKit will have the same React version as a peer dep as app-kit and this will be released as breaking or did we include that in the last release?

montezume:
@tdeekens we included in the last breaking UI-Kit version (10) the requirement for version 16.7 of React and version 3 of react-intl

src/components/fields/password-field/password-field.js
```diff
@@ -18,127 +18,105 @@ const sequentialId = createSequentialId('password-field-');
...
+          />
+          {!props.isDisabled && !props.isReadOnly && (
+            <FlatButton
+              icon={isPasswordVisible ? <EyeCrossedIcon /> : <EyeIcon />}
+              label={
+                isPasswordVisible
+                  ? intl.formatMessage(messages.hide)
+                  : intl.formatMessage(messages.show)
+              }
+              onClick={() => setIsPasswordVisible(!isPasswordVisible)}
```
tdeekens: Nit: could be a tiny memoized fn togglePasswordVisibility.

src/components/fields/password-field/password-field.js
```diff
@@ -18,127 +18,105 @@ const sequentialId = createSequentialId('password-field-');
...
-            renderError={this.props.renderError}
-          />
-        </Spacings.Stack>
-      </Constraints.Horizontal>
-    );
-  }
-}
+PasswordField.defaultProps = {
+  horizontalConstraint: 'scale',
+};
```
tdeekens: Assuming these just moved.

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -14,233 +14,222 @@ import DropdownIndicator from '../../internals/dropdown-indicator';
...
   LoadingIndicator,
 };
 
-export class AsyncCreatableSelectInput extends React.Component {
-  // Formik will set the field to an array on submission, so we always have to
-  // deal with an array. The touched state ends up being an empty array in case
-  // values were removed only. So we have to treat any array we receive as
-  // a signal of the field having been touched.
-  static isTouched = touched => Boolean(touched);
+const AsyncCreatableSelectInput = props => {
```
tdeekens: Nice to see them all being React.FCs.

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -14,233 +14,222 @@ import DropdownIndicator from '../../internals/dropdown-indicator';
...
+              newValue = [];
+            }
+            props.onChange(
+              {
+                target: { name: props.name, value: newValue },
+                persist: () => {},
+              },
+              info
+            );
+          }}
```
tdeekens: It is hard to trace changes but I assume it is all about the top.

src/components/inputs/select-input/select-input.js
```diff
@@ -15,277 +15,255 @@ import DropdownIndicator from '../../internals/dropdown-indicator';
...
-  static getDerivedStateFromProps = props => {
-    // Options can be grouped as
-    //   const colourOptions = [{ value: 'green', label: 'Green' }];
-    //   const flavourOptions = [{ value: 'vanilla', label: 'Vanilla' }];
-    //   const groupedOptions = [
-    //     { label: 'Colours', options: colourOptions },
-    //     { label: 'Flavours', options: flavourOptions },
-    //   ];
-    // So we "ungroup" the options by merging them all into one list first.
-    const optionsWithoutGroups = flatMap(props.options, option =>
```
tdeekens: In L40 this has been replaced without using state. Why was this in a gDSFP before?

montezume: I'm not really sure why it was there in the first place. We were using a gDSFP without a condition, so it always updated "state" to the new value on every render (including first render), and then we never set the state or did anything with it. Same thing in the other components.

montezume:
Yeah most of the diffs came from moving the proptypes + default props + etc outside of the class component

src/components/fields/password-field/password-field.js
montezume: exactly

src/components/inputs/select-input/select-input.js
tdeekens: Odd. We might just have adopted it cause it was the new thing.

montezume: It's possible that our first implementation of the component had setState happening but then we refactored it out but left the gDSFP 🤷‍♂ ?

src/components/fields/password-field/password-field.js
montezume: is this what you mean?

https://github.com/commercetools/ui-kit/pull/957/commits/4200a7b358180cd3aed72d996a1ed08e277e22c7

I'm not a huge hook expert yet 🤓

src/components/fields/password-field/password-field.js
```diff
@@ -24,6 +24,10 @@ const PasswordField = props => {
   const id = getFieldId(props, {}, sequentialId);
   const hasError = props.touched && hasErrors(props.errors);
 
+  const togglePasswordVisibility = React.useCallback(() => {
```
tdeekens: Exactly this. Cheers.

PR #958, opened 2019-07-25
refactor(localized-text-input): use hooks

montezume:
#### Summary

Refactors the `LocalizedTextInput` container to use react hooks

PR #960, opened 2019-07-25
refactor(time-input): use hooks

montezume:
#### Summary

Refactors the `TimeInput` component to use hooks

### DIFF
--- a/src/components/inputs/time-input/time-input.js
+++ b/src/components/inputs/time-input/time-input.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { injectIntl } from 'react-intl';
+import { useIntl } from 'react-intl';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import getFieldId from '../../../utils/get-field-id';
 import createSequentialId from '../../../utils/create-sequential-id';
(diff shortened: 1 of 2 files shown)


src/components/inputs/time-input/time-input.js
```diff
@@ -23,144 +23,133 @@ const format24hr = ({ hours, minutes, seconds, milliseconds }) => {
...
+      if (formattedTime !== props.value) emitChange(formattedTime);
+
+      // forward the onBlur call
+      if (props.onBlur) props.onBlur(event);
+    },
+    [intl.locale, props.value]
+  );
+
+  // so that it doesn't run on initial mount
+  const mounted = React.useRef();
```
montezume: basically the idea is to "imitate" componentDidUpdate.

src/components/inputs/time-input/time-input.js
```diff
@@ -23,144 +23,133 @@ const format24hr = ({ hours, minutes, seconds, milliseconds }) => {
...
+  const intl = useIntl();
+
+  const emitChange = React.useCallback(
+    value => {
+      const event = {
+        target: { id, name: props.name, value },
+      };
+      props.onChange(event);
+    },
+    [id, props.name]
```
tdeekens: Is not onChange also a dep?

emmenko: Do you think it's relevant to put `onChange` as a dependency? When would the function ever change? 🤔

tdeekens: Yeah, I noted that on another PR.

emmenko: Link?

tdeekens: You reacted to it darling :) https://github.com/commercetools/ui-kit/pull/958#discussion_r307601310

emmenko: Ah, I meant using the callback and having lang as a dep though. I don't think we need to pass `onChange` as a dep...

tdeekens: Why not? Why can we underspecify `onChange`? To my understanding it closes over it so we won't bust given it changes?

emmenko: I'm just trying to understand why you would need to specify it. Yes, it's a prop and "potentially" it might change but I never saw a case when changing a callback function on purpose would cause the UI to rerender.
It's just confusing for me as a mental model.

We can add it though if you prefer so.

tdeekens: I just thought one also has to specify things on `props` as otherwise things won't invalidate the memoization leading to bugs.

PR #961, opened 2019-07-26
chore(ci): add kodiak

tdeekens:
#### Summary

This pull request suggests adding kodiak. A nice and working merge bot. This only adds the config. Later we would install the app.

https://github.com/chdsbd/kodiak#prior-art

### DIFF
--- a/.kodiak.toml
+++ b/.kodiak.toml
@@ -1,0 +1,11 @@
+version = 1
+
+[merge]
+automerge_label = "🚀 Status: ship it"
+blacklist_title_regex = "^WIP.*"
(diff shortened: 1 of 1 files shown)


tdeekens:
I will merge this and install the app. I hope @montezume is fine with that. The config can be adjusted anyway and he seemed to like the idea of a bit and more automation around these things.

montezume:
Looks cool!

montezume:
So just to confirm, the reason why it isn't working is because [this](https://github.com/chdsbd/kodiak/issues/119)?

tdeekens:
Exactly. I think we have to ask how much we want that rule. obviously nobody external can push I guess. I got a bit confused by the rule but as I understand it, it overlays the contributors.

PR #963, opened 2019-07-26
refactor(date-input): use hooks

montezume:
#### Summary

Refactors the `DateInput` component to use hooks.

### DIFF
--- a/src/components/inputs/date-input/date-input.js
+++ b/src/components/inputs/date-input/date-input.js
@@ -36,9 +36,10 @@ const DateInput = props => {
   );
   const inputRef = React.useRef();
 
+  const { onChange } = props;
   const emit = React.useCallback(
     value =>
-      props.onChange({
+      onChange({
         target: {
           id: props.id,
           name: props.name,
@@ -47,7 +48,7 @@ const DateInput = props => {
           value: value || '',
         },
       }),
-    [props.id, props.name]
+    [onChange, props.id, props.name]
   );
 
   const handleChange = React.useCallback(
@@ -55,18 +56,19 @@ const DateInput = props => {
       inputRef.current.setSelectionRange(0, 100);
       emit(date);
     },
-    [inputRef]
+    [inputRef, emit]
   );
 
+  const { onBlur } = props;
   const handleBlur = React.useCallback(() => {
-    if (props.onBlur)
-      props.onBlur({
+    if (onBlur)
+      onBlur({
         target: {
           id: props.id,
           name: props.name,
         },
       });
-  }, [props.id, props.name]);
+  }, [onBlur, props.id, props.name]);
 
   const showToday = () => {
     const today = getToday();
--- a/src/components/inputs/date-input/date-input.js
+++ b/src/components/inputs/date-input/date-input.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Downshift from 'downshift';
-import { injectIntl } from 'react-intl';
+import { useIntl } from 'react-intl';
 import CalendarBody from '../../internals/calendar-body';
 import CalendarMenu from '../../internals/calendar-menu';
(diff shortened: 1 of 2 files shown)


src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,237 @@ import {
...
-  };
+    [props.id, props.name]
+  );
+
+  const handleChange = React.useCallback(
+    date => {
+      inputRef.current.setSelectionRange(0, 100);
+      emit(date);
+    },
+    [inputRef]
```
montezume: I'm not sure if this makes sense 🤔

src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,237 @@ import {
...
-  inputRef = React.createRef();
-
-  state = {
-    calendarDate: this.props.value || getToday(),
-    suggestedItems: [],
-    highlightedIndex:
-      this.props.value === '' ? null : getDateInMonth(this.props.value) - 1,
-  };
-
-  showPrevMonth = () => {
```
montezume: this was never used

src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,237 @@ import {
...
+              if (date === '') {
+                setSuggestedItems([]);
+                setHighlightedIndex(null);
               } else {
-                // input changed because user selected a date
-                this.setState({
-                  suggestedItems: [],
-                  highlightedIndex: null,
-                });
+                setSuggestedItems([date]);
```
montezume: I'm wondering if it makes sense to "batch" these updates? I guess one choice is to go with a "useReducer" but I feel like the readability is a bit lost then.

src/components/inputs/date-input/date-input.js
tdeekens: To my understanding yes. useCallback does not have smart exceptions like any value of props or any ref. It is just a closure and memoization as far as I understood it.

src/components/inputs/date-input/date-input.js
tdeekens: Won‘t React batch them? Not 100% sure but maybe.

src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,233 @@ import {
...
-        // when cleared the value is null, but we always want it to be an
-        // empty string when there is no value.
-        value: value || '',
-      },
-    });
+  const jumpMonth = amount => {
+    const nextDate = changeMonth(calendarDate, amount);
+    setCalendarDate(nextDate);
+    setHighlightedIndex(0);
+  };
```
tdeekens: Is there a specific reason to not useCallback these two?

src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,235 @@ import {
...
-  };
-
-  showToday = () => {
-    const today = getToday();
-    this.setState(
-      prevState => ({
-        calendarDate: today,
-        highlightedIndex:
-          prevState.suggestedItems.length + getDateInMonth(today) - 1,
+  const { onChange } = props;
```
emmenko: Same here

src/components/inputs/date-input/date-input.js
```diff
@@ -25,267 +25,235 @@ import {
...
+      onBlur({
         target: {
-          id: this.props.id,
-          name: this.props.name,
+          id: props.id,
+          name: props.name,
         },
       });
-  };
+  }, [onBlur, props.id, props.name]);
```
emmenko: And here

src/components/inputs/date-input/date-input.js
tdeekens: Yeah, this is how I would go about it too. Thanks.

PR #965, opened 2019-07-26
refactor(primary-action-dropdown): use hooks

montezume:
#### Summary

Refactors the component `PrimaryActionDropdown` to use hooks

### DIFF
--- a/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
+++ b/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
@@ -225,7 +225,6 @@ const PrimaryActionDropdown = props => {
     },
     [ref]
   );
-
   React.useEffect(() => {
     window.addEventListener('click', handleGlobalClick);
     return () => {
@@ -237,6 +236,21 @@ const PrimaryActionDropdown = props => {
   const primaryOption =
     childrenAsArray.find(option => !option.props.isDisabled) ||
     childrenAsArray[0];
+
+  const { onClick } = primaryOption.props;
+  const handleClickOnHead = React.useCallback(
+    event => {
+      if (isOpen) {
+        setIsOpen(false);
+      } else {
+        onClick(event);
+      }
+    },
+    [isOpen, onClick]
+  );
+  const handleClickOnChevron = React.useCallback(() => {
+    setIsOpen(!isOpen);
+  }, [isOpen]);
 
   invariant(
     childrenAsArray.length > 1,
@@ -254,19 +268,11 @@ const PrimaryActionDropdown = props => {
       <DropdownHead
         iconLeft={primaryOption.props.iconLeft}
         isDisabled={primaryOption.props.isDisabled}
-        onClick={isOpen ? () => setIsOpen(false) : primaryOption.props.onClick}
+        onClick={handleClickOnHead}
         chevron={
           <DropdownChevron
             ref={ref}
-            onClick={
-              isOpen
-                ? () => {
-                    setIsOpen(false);
-                  }
-                : () => {
-                    setIsOpen(true);
-                  }
-            }
+            onClick={handleClickOnChevron}
             isDisabled={primaryOption.props.isDisabled}
             isOpen={isOpen}
           />
--- a/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
+++ b/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
@@ -216,6 +216,7 @@ const PrimaryActionDropdown = props => {
     event => {
       const dropdownButton = ref.current;
       if (
+        dropdownButton &&
         event.target !== dropdownButton &&
         !dropdownButton.contains(event.target)
       ) {
--- a/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
+++ b/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
@@ -230,7 +230,7 @@ const PrimaryActionDropdown = props => {
     return () => {
       window.removeEventListener('click', handleGlobalClick);
     };
-  });
+  }, [handleGlobalClick]);
 
   const childrenAsArray = React.Children.toArray(props.children);
   const primaryOption =
--- a/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
+++ b/src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
@@ -39,67 +39,63 @@ const getButtonStyles = isDisabled => {
   ];
 };
 
-class DropdownHead extends React.PureComponent {
-  static displayName = 'DropdownHead';
-  static propTypes = {
-    iconLeft: PropTypes.element.isRequired,
-    onClick: PropTypes.func,
-    children: PropTypes.node.isRequired,
-    isDisabled: PropTypes.bool.isRequired,
-    chevron: PropTypes.element.isRequired,
-  };
-
-  render() {
-    return (
-      <div
+const DropdownHead = props => (
+  <div
+    css={css`
+      display: flex;
+      align-items: center;
+    `}
+  >
+    <AccessibleButton
+      label={props.children}
+      onClick={props.onClick}
+      isDisabled={props.isDisabled}
+      css={[
+        ...getButtonStyles(props.isDisabled),
+        css`
+          padding: 0 ${vars.spacingS};
+          border-radius: ${vars.borderRadius6} 0 0 ${vars.borderRadius6};
+        `,
+      ]}
+    >
+      <span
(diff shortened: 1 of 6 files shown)


src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
+      }
+    },
+    [ref]
+  );
 
-  handleClose = () => {
-    this.setState({ isOpen: false });
-  };
+  React.useEffect(() => {
+    window.addEventListener('click', handleGlobalClick);
```
montezume: this replaces the `componentDidMount` and `componentWillUnmount`

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.spec.js
```diff
@@ -142,8 +143,10 @@ describe('when dropdown is open and body is clicked', () => {
     expect(getByLabelText('Secondary Action')).toBeInTheDocument();
 
     // click on doucment body
-    document.body.click();
 
+    act(() => {
```
montezume: needed this or else got warnings in the test

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
+    return () => {
+      window.removeEventListener('click', handleGlobalClick);
+    };
+  }, [handleGlobalClick]);
```
tdeekens: I dont even think this has to be a dep. Dep talk all hook lomg.

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
-
-  static propTypes = {
-    children: PropTypes.node.isRequired,
-  };
-
-  state = {
-    isOpen: false,
-  };
+const PrimaryActionDropdown = props => {
+  const ref = React.useRef();
```
tdeekens: Can we name the hook better?

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.spec.js
tdeekens: Cool that it worked.

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -230,7 +230,7 @@ const PrimaryActionDropdown = props => {
     return () => {
       window.removeEventListener('click', handleGlobalClick);
     };
-  }, [handleGlobalClick]);
+  });
```
tdeekens: We want it to run once not on each render right. So empty array?

montezume: good call

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
+    event => {
+      const dropdownButton = ref.current;
+      if (
+        event.target !== dropdownButton &&
+        !dropdownButton.contains(event.target)
+      ) {
+        setIsOpen(false);
+      }
+    },
+    [ref]
```
emmenko: I think I saw an issue or something related to defining DOM elements (refs) as dependencies. I’ll see if I can find more about it but maybe we should double check if it makes sense to do so.

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
-
-  handleOpen = () => {
-    this.setState({ isOpen: true });
-  };
+  const handleGlobalClick = React.useCallback(
+    event => {
+      const dropdownButton = ref.current;
+      if (
+        event.target !== dropdownButton &&
+        !dropdownButton.contains(event.target)
```
emmenko: Should we also check if `dropdownButton` is defined? 🤔

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
-  static propTypes = {
-    children: PropTypes.node.isRequired,
-  };
-
-  state = {
-    isOpen: false,
-  };
+const PrimaryActionDropdown = props => {
+  const ref = React.useRef();
+  const [isOpen, setIsOpen] = React.useState(false);
```
emmenko: Use the new collapsible/toggle hook?

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
+      css={css`
+        position: relative;
+        display: inline-flex;
+        align-items: column;
+      `}
+    >
+      <DropdownHead
+        iconLeft={primaryOption.props.iconLeft}
+        isDisabled={primaryOption.props.isDisabled}
+        onClick={isOpen ? () => setIsOpen(false) : primaryOption.props.onClick}
```
emmenko: Nitpick: memoized callback?

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,82 @@ Option.defaultProps = {
...
+        chevron={
+          <DropdownChevron
+            ref={ref}
+            onClick={
```
emmenko: Same here?

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.spec.js
emmenko: Since RTL already includes this, can you maybe try to click on the rendered container?

```js
fireEvent.click(rendered.container)
```

There should be no need for using `act` directly

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
montezume: sure, but need to merge that first 😉

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.spec.js
emmenko: Fixed here: https://github.com/commercetools/ui-kit/pull/965/commits/b83d74f503cd5d44322219ca1b0b6b2a7c2f624f

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
emmenko: Actually we can't, because in the `handleGlobalClick` we force to be `false`, so it's not really a toggle anymore

src/components/dropdowns/primary-action-dropdown/primary-action-dropdown.js
```diff
@@ -211,89 +208,89 @@ Option.defaultProps = {
...
-  };
+  const childrenAsArray = React.Children.toArray(props.children);
+  const primaryOption =
+    childrenAsArray.find(option => !option.props.isDisabled) ||
+    childrenAsArray[0];
 
-  componentDidMount() {
-    window.addEventListener('click', this.handleGlobalClick);
-  }
+  const { onClick } = primaryOption.props;
```
emmenko: Same drill here again

src/index.js
```diff
@@ -141,3 +140,5 @@ export { default as version } from './version';
 export {
   default as ContentNotification,
 } from './components/notifications/content-notification';
+
+export * from './hooks';
```
emmenko: Extra unrelated little change.

tdeekens: Does UIKit really want to export it's hooks already? This would maybe even a feature releasea as a result.

emmenko: It was already exposed actually, I just refactored it to `export *` instead of `export { default as useToggleState }`

tdeekens: Hm. I, personal opinion ahead, liked it before more then. The library opts into what it wants to expose instead of defaulting to everything.

emmenko: Sure my bad, I'll revert this change then

emmenko: https://github.com/commercetools/ui-kit/pull/965/commits/8361bc17e7c9cd7d84db70adf633bceb0d1c5821

PR #966, opened 2019-07-26
refactor(multiline-text-input): use hooks & expose `useCollapsible` hook

montezume:
#### Summary

Refactors `MutilineTextInput` to use hooks, and exposes a new `useCollapsible` hook.


src/components/use-collapsible/use-collapsible.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
```
tdeekens: Why do we put this into components? Would a hooks folder be more apt?
We could also call it useToggle I guess.

emmenko: Yeah, I agree with Tobi

montezume: ya good call 🎈

src/hooks/use-collapsible/use-collapsible.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const useCollapsible = (defaultValue = true) => {
```
emmenko: I’m wondering if we should name it like this or just something like `useToggleState`, because the hook itself does not do any “collapsible” work... 🤔

montezume: yeah that makes sense.

montezume: [here](https://github.com/commercetools/ui-kit/pull/966/commits/5fb91aef6008a84dfea79a779f3539737d6dbeb7)

src/components/collapsible/collapsible.js
```diff
@@ -1,61 +1,76 @@
...
+      isOpen: !props.isClosed,
+      toggle: props.onToggle,
+    })}
+  </>
+);
+ControlledCollapsible.displayName = 'ControlledCollapsible';
+ControlledCollapsible.propTypes = collapsiblePropTypes;
+
+const UncontrolledCollapsible = props => {
+  const [isOpen, toggle] = useToggleState(!props.isDefaultClosed);
```
emmenko: I split the rendering into two components (controlled / uncontrolled) to fix the "conditional" rendering of the hook in the main component

PR #967, opened 2019-07-27
chore: add eslint-plugin-react-hooks

montezume:
#### Summary

The only issue here is that the plugin will squak at you if you **don't** destructure props.

Any ideas @tdeekens @emmenko ?

Screen Shot 2019-07-27 at 10 21 38 AM
Screen Shot 2019-07-27 at 10 21 45 AM

### DIFF
--- a/src/components/inputs/time-input/time-input.js
+++ b/src/components/inputs/time-input/time-input.js
@@ -26,28 +26,29 @@ const hasMilliseconds = parsedTime => parsedTime.milliseconds !== 0;
 const TimeInput = props => {
   const id = getFieldId(props, {}, sequentialId);
   const intl = useIntl();
+  const { name, value, onBlur, onChange } = props;
 
   const emitChange = React.useCallback(
-    value => {
+    newValue => {
       const event = {
-        target: { id, name: props.name, value },
+        target: { id, name, value: newValue },
       };
-      props.onChange(event);
+      onChange(event);
     },
-    [id, props.name]
+    [id, name, onChange]
   );
 
   const handleBlur = React.useCallback(
     event => {
       // check formatting and reformat when necessary
-      const formattedTime = TimeInput.toLocaleTime(props.value, intl.locale);
+      const formattedTime = TimeInput.toLocaleTime(value, intl.locale);
 
-      if (formattedTime !== props.value) emitChange(formattedTime);
+      if (formattedTime !== value) emitChange(formattedTime);
 
       // forward the onBlur call
-      if (props.onBlur) props.onBlur(event);
+      if (onBlur) onBlur(event);
     },
-    [intl.locale, props.value]
+    [intl.locale, value, onBlur, emitChange]
   );
 
   // so that it doesn't run on initial mount
@@ -56,9 +57,9 @@ const TimeInput = props => {
     if (!mounted.current) {
       mounted.current = true;
     } else {
-      emitChange(TimeInput.toLocaleTime(props.value, intl.locale));
+      emitChange(TimeInput.toLocaleTime(value, intl.locale));
     }
-  }, [intl.locale]);
+  }, [intl.locale, emitChange, value]);
 
   return (
     <Constraints.Horizontal constraint={props.horizontalConstraint}>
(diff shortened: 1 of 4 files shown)


tdeekens:
Thanks I was about to add this later :)

src/components/inputs/time-input/time-input.js
```diff
@@ -26,39 +26,42 @@ const hasMilliseconds = parsedTime => parsedTime.milliseconds !== 0;
 const TimeInput = props => {
   const id = getFieldId(props, {}, sequentialId);
   const intl = useIntl();
+  const { name, value, onBlur, onChange } = props;
 
   const emitChange = React.useCallback(
-    value => {
+    newValue => {
```
tdeekens: Absolute nit: `nextValue`.

montezume:
@tdeekens just sort of annoying that we will need to move towards destructuring and against our not destructuring convention?

montezume:
but the advantage is it will make reviewing easier 😄

tdeekens:
> @tdeekens just sort of annoying that we will need to move towards destructuring and against our not destructuring convention?

Yeah seems with hooks it makes sense though. As otherwise, according to what it says, the `props` outer ref will change and... wait does that even make sense. I am confused now a bit. Still, I guess we have to live with that?

tdeekens:
Ah, maybe if you have `props` with `{ foo: 'bar', baz: { foo: 'bar' } }` then if would still work on `props.foo` but not `props.baz` as the whole object changed. Still, it would work on `props.baz.foo` as it's just comparing primitives.

montezume:
So the idea would be that instead of using defaultProps, you would just destructure and set defaults then?

tdeekens:
> So the idea would be that instead of using defaultProps, you would just destructure and set defaults then?

To how I recall reading it yes. Kind of like: "Do the shallow merge yourself.". https://github.com/reactjs/rfcs/pull/107.

```js
const Thing = props => {
  const { foo = 'defaulted' } = props;
}
```

I've been digging into the source of the hooks plugin. It feels a bit like just an extra/super careful warning. Not sure though. https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L853

src/components/inputs/time-input/time-input.js
```diff
@@ -26,39 +26,42 @@ const hasMilliseconds = parsedTime => parsedTime.milliseconds !== 0;
...
 
-      if (formattedTime !== props.value) emitChange(formattedTime);
+      if (formattedTime !== value) emitChange(formattedTime);
 
       // forward the onBlur call
-      if (props.onBlur) props.onBlur(event);
+      if (onBlur) onBlur(event);
     },
-    [intl.locale, props.value]
+    [intl.locale, value, onBlur, emitChange]
```
montezume: should we also be destructuring intl? 🤔

emmenko:
Thanks for setting up the plugin. I haven’t look into this particular problem yet but at a first glance it feels weird that it does not work by not destructuring props.

Can we maybe look into this a bit more before making decisions such as “let’s destructure props from now on”, as it goes a bit against what we’ve been advocating so far?

montezume:
I think [this](https://github.com/facebook/react/issues/14920#issuecomment-467494468) is the reason why

```
This is because technically props.foo() passes props itself as this to foo call. So foo might implicitly depend on props. We'll need a better message for this case though. The best practice is always destructuring.

```

tdeekens:
@emmenko I did not want to advocate to blindly use it and wonder where it came across like that. I just felt that blindly not using it is also questionable as the plugin in itself has obviously nice benefits. 

I think we are hitting a limitation of how the plugin can determine currently what is used inside of the hook from props. Given you could access anything of it and it can not, but may later, detect it it bails out to recommend you to change the coding style. 

We do not have to destructure all props in the argument position of any component. We can also just destructure right in front of a hook what we use inside to comply with the rules capability. If it them ever changes w can adapt again. Personally I still like not destructuring. 

Note that this is a bit like the problems one has with immer and ESLint. There the rule recommendig not to mutate things also defeats the purpose as inside a produce you „are free“. Still, ESLint currently doesn’t know that context.

emmenko:
Thanks for the explanation, it makes more sense to me now.

Let’s try that then!

emmenko:
I'll merge this and update the other PRs

PR #981, opened 2019-08-01
refactor(collapsible-motion): use hooks

montezume:
#### Summary

Refactors `CollapsibleMotion` to use hooks.

I think the code got quite a bit simpler.

A follow up could be to create a `useCollapsibleMotion` hook

### DIFF
--- a/src/components/collapsible-motion/collapsible-motion.js
+++ b/src/components/collapsible-motion/collapsible-motion.js
@@ -5,6 +5,13 @@ import { keyframes, ClassNames } from '@emotion/core';
 import isNil from 'lodash/isNil';
 import useToggleState from '../../hooks/use-toggle-state';
 
+const collapsibleMotionPropTypes = {
+  children: PropTypes.func.isRequired,
+  isClosed: PropTypes.bool,
+  onToggle: PropTypes.func,
+  isDefaultClosed: PropTypes.bool,
+};
+
 const usePrevious = value => {
   const ref = React.useRef();
   React.useEffect(() => {
@@ -118,12 +125,7 @@ const ControlledCollapsibleMotion = props => {
 };
 
 ControlledCollapsibleMotion.displayName = 'ControlledCollapsibleMotion';
-ControlledCollapsibleMotion.propTypes = {
-  children: PropTypes.func.isRequired,
-  isClosed: PropTypes.bool,
-  onToggle: PropTypes.func,
-  isDefaultClosed: PropTypes.bool,
-};
+ControlledCollapsibleMotion.propTypes = collapsibleMotionPropTypes;
 
 const UncontrolledCollapsibleMotion = props => {
   const [isOpen, toggle] = useToggleState(!props.isDefaultClosed);
@@ -166,19 +168,9 @@ const UncontrolledCollapsibleMotion = props => {
 };
 
 UncontrolledCollapsibleMotion.displayName = 'UncontrolledCollapsibleMotion';
-UncontrolledCollapsibleMotion.propTypes = {
-  children: PropTypes.func.isRequired,
-  isClosed: PropTypes.bool,
-  onToggle: PropTypes.func,
-  isDefaultClosed: PropTypes.bool,
-};
+UncontrolledCollapsibleMotion.propTypes = collapsibleMotionPropTypes;
 
 CollapsibleMotion.displayName = 'CollapsibleMotion';
-CollapsibleMotion.propTypes = {
-  children: PropTypes.func.isRequired,
-  isClosed: PropTypes.bool,
-  onToggle: PropTypes.func,
-  isDefaultClosed: PropTypes.bool,
-};
+CollapsibleMotion.propTypes = collapsibleMotionPropTypes;
 
 export default CollapsibleMotion;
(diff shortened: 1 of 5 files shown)


montezume:
Actually I can probably make this simpler by using the hook

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -17,144 +25,116 @@ const createClosingAnimation = height =>
...
+ToggleAnimation.displayName = 'ToggleAnimation';
+ToggleAnimation.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  toggle: PropTypes.func.isRequired,
+  children: PropTypes.func.isRequired,
+};
+
+const CollapsibleMotion = props => {
+  return (
+    <Collapsible
```
montezume: I want to use `useToggleState` here, but it's a bit complicated because `CollapsibleMotion` can be both uncontrolled and controlled...

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -4,6 +4,14 @@ import invariant from 'tiny-invariant';
 import { keyframes, ClassNames } from '@emotion/core';
 import Collapsible from '../collapsible';
 
+const usePrevious = value => {
+  const ref = React.useRef();
```
montezume: useRef is the bomb dot com

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -17,144 +27,225 @@ const createClosingAnimation = height =>
...
+
+  const [
+    animation,
+    containerStyles,
+    animationToggle,
+    registerContentNode,
+  ] = useToggleAnimation(isOpen, toggle);
+
+  return (
+    <ClassNames>
```
montezume: I wish there was a `useClassNames` hook  😢

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -2,7 +2,16 @@ import PropTypes from 'prop-types';
...
 import { keyframes, ClassNames } from '@emotion/core';
-import Collapsible from '../collapsible';
+import isNil from 'lodash/isNil';
+import useToggleState from '../../hooks/use-toggle-state';
+
+const usePrevious = value => {
+  const ref = React.useRef();
+  React.useEffect(() => {
+    ref.current = value;
+  });
```
tdeekens: You want this on every render or only when `value` changes?

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -17,144 +26,159 @@ const createClosingAnimation = height =>
...
-  }
-
-  handleToggle = () => {
+const useToggleAnimation = (isOpen, toggle) => {
+  const nodeRef = React.useRef();
+  const prevIsOpen = usePrevious(isOpen);
+
+  React.useEffect(
+    () => {
+      invariant(
```
tdeekens: We might want to wrap this into a dev mode condition. PropTypes are often compiled away. This not. So it throwing might break people's apps.

montezume: this is the same behaviour as before -> the invariant was called in componentDidMount

src/components/collapsible-motion/collapsible-motion.js
montezume: good call, should have [value] as deps array

src/components/collapsible-motion/collapsible-motion.js
tdeekens: Wait a bit :)

montezume: I think it's not that simple 😢 
https://github.com/emotion-js/emotion/issues/1321

src/hooks/use-previous/use-previous.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  const { getByTestId } = render(<TestComponent />);
+  expect(getByTestId('openState')).toHaveTextContent('open');
+  expect(getByTestId('prevOpenState')).toHaveTextContent('undefined');
+});
+
+it('should maintain the previous state after changing state', () => {
+  const { getByTestId } = render(<TestComponent />);
+  expect(getByTestId('openState')).toHaveTextContent('open');
+  getByTestId('toggle').click();
+  expect(getByTestId('prevOpenState')).toHaveTextContent('closed');
```
montezume: hm this does not seem right lol

src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -17,144 +33,144 @@ const createClosingAnimation = height =>
...
+
+const UncontrolledCollapsibleMotion = props => {
+  const [isOpen, toggle] = useToggleState(!props.isDefaultClosed);
+
+  const [
+    animation,
+    containerStyles,
+    animationToggle,
+    registerContentNode,
+  ] = useToggleAnimation(isOpen, toggle);
```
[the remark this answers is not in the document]
montezume: Yeah, it makes it more readable, easy to understand. We could move the toggle logic into `useToggleAnimation` though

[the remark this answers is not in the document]
montezume: The reason I don't put the toggle logic into useToggleAnimation is because this used for both the controlled and uncontrolled components. If we put the toggle logic in there, then the hook will be "conditional"

montezume:
I would like to get this merged. Any final thoughts @emmenko ?

emmenko:
Yeah sorry, all good from my side. Thanks

PR #983, opened 2019-08-02
refactor(tooltip): use hooks

montezume:
#### Summary

Refactors the `Tooltip` component to use hooks.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { isValidElementType } from 'react-is';
 import styled from '@emotion/styled';
+import isNil from 'lodash/isNil';
 import { Manager, Reference, Popper } from 'react-popper';
 import getFieldId from '../../utils/get-field-id';
 import createSequentialId from '../../utils/create-sequential-id';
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -23,221 +24,230 @@ TooltipWrapper.propTypes = {
...
-    }
-
-    if (this.props.closeAfter) {
-      this.leaveTimer = setTimeout(() => {
-        this.handleClose();
-      }, this.props.closeAfter);
-    } else {
-      this.handleClose();
-    }
+  React.useEffect(() => {
```
tdeekens: Should we build a `useTimeout` for this? Seeems like a frequent case?

src/components/tooltip/tooltip.js
```diff
@@ -23,221 +24,230 @@ TooltipWrapper.propTypes = {
...
+      if (event.type === 'mouseleave' && onMouseLeave) {
+        onMouseLeave(event);
+      }
+
+      if (event.type === 'blur' && onBlur) {
+        onBlur(event);
+      }
+
+      if (closeAfter) {
+        leaveTimer.current = setTimeout(() => {
```
tdeekens: `leaveTimerRef` would have helped me. Wondering if 

```js
const [ triggerTimeout ] = useTimeout(close, 1000) 
```

or something alike could help.

src/components/tooltip/tooltip.spec.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
+import { act } from 'react-dom/test-utils';
```
montezume: I've spent so much time battling this over the last week...

emmenko: Hmm interesting...

montezume: this thread is helpful

https://github.com/facebook/react/issues/15379

TL:DR they have added an async act in the react 16.9 which should drop shortly

emmenko: Yeah but RTL already uses `act`. But I guess updating the timers lies "outside" of RTL scope...

montezume: the worst is combining async actions + timers. It's really very hard to test with RTL

montezume:
Can you take a look @jonnybel ?

PR #984, opened 2019-08-02
refactor: improve `useToggleState` and use it across our components

montezume:
#### Summary

Updates `useToggleState` hook's `toggle` function to accept an option arg that can be used to force the state.

Refactors various components (`MoneyInput`, `PrimaryActionDropdown`, `PasswordField` and `CalendarBody`) to use this hook instead of custom toggle logic.

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -7,6 +7,7 @@ import Select, { components } from 'react-select';
 import { useIntl } from 'react-intl';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
+import useToggleState from '../../../hooks/use-toggle-state';
 import vars from '../../../../materials/custom-properties';
 import DropdownIndicator from '../../internals/dropdown-indicator';
 import isNumberish from '../../../utils/is-numberish';
@@ -311,16 +312,40 @@ const getCurrencyDropdownName = name =>
 
 const MoneyInput = props => {
   const intl = useIntl();
-  const [currencyHasFocus, setCurrencyHasFocus] = React.useState(false);
-  const [amountHasFocus, setAmountHasFocus] = React.useState(false);
+  const [
+    currencyHasFocus,
+    // eslint-disable-next-line no-unused-vars
+    toggleCurrencyHasFocus,
+    setCurrencyHasFocusOn,
+    setCurrencyHasFocusOff,
+  ] = useToggleState(false);
+  const [
+    amountHasFocus,
+    // eslint-disable-next-line no-unused-vars
+    toggleAmountHasFocus,
+    setAmountHasFocusOn,
+    setAmountHasFocusOff,
+  ] = useToggleState(false);
 
   const containerRef = React.useRef();
   const amountInputRef = React.useRef();
 
+  const { onFocus } = props;
+  const handleAmountFocus = React.useCallback(() => {
+    if (onFocus)
+      onFocus({
+        target: {
+          id: MoneyInput.getAmountInputId(props.id),
+          name: getAmountInputName(props.name),
+        },
+      });
+    setAmountHasFocusOn();
+  }, [setAmountHasFocusOn, onFocus, props.id, props.name]);
+
   const { onChange } = props;
   const handleAmountBlur = React.useCallback(() => {
     const amount = props.value.amount.trim();
-    setAmountHasFocus(false);
+    setAmountHasFocusOff();
     // Skip formatting for empty value or when the input is used with an
     // unknown currency.
     if (amount.length > 0 && currencies[props.value.currencyCode]) {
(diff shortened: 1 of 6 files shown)


src/components/inputs/money-input/money-input.js
```diff
@@ -311,16 +312,40 @@ const getCurrencyDropdownName = name =>
...
+  const [
+    currencyHasFocus,
+    // eslint-disable-next-line no-unused-vars
+    toggleCurrencyHasFocus,
+    setCurrencyHasFocusOn,
+    setCurrencyHasFocusOff,
+  ] = useToggleState(false);
+  const [
+    amountHasFocus,
+    // eslint-disable-next-line no-unused-vars
```
montezume: this is a sort of annoying part of using hooks 🤔

src/components/inputs/money-input/money-input.js
```diff
@@ -311,16 +312,40 @@ const getCurrencyDropdownName = name =>
...
+    toggleAmountHasFocus,
+    setAmountHasFocusOn,
+    setAmountHasFocusOff,
+  ] = useToggleState(false);
 
   const containerRef = React.useRef();
   const amountInputRef = React.useRef();
 
+  const { onFocus } = props;
+  const handleAmountFocus = React.useCallback(() => {
```
montezume: these are just moved to using callbacks

src/components/inputs/money-input/money-input.js
jonnybel: I think you can do this:
```
const [amountHasFocus, ,setAmountHasFocusOn, setAmountHasFocusOff] = useToggleState(false);
```
to ignore the returned `toggleCurrencyHasFocus`, since you don't need it.
[source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Ignoring_some_returned_values)

src/hooks/use-toggle-state/use-toggle-state.js
```diff
@@ -1,12 +1,21 @@
...
+    setIsToggled(false);
+  }, [setIsToggled]);
+
+  const toggle = useCallback(() => {
     setIsToggled(!isToggled);
-  }, [isToggled]);
+  }, [isToggled, setIsToggled]);
 
-  return [isToggled, toggle];
+  return [isToggled, toggle, setOn, setOff];
```
emmenko: Hmm I'm not sure I like this new API.

Can't we just accept an argument to the `toggle` function, which takes precedence over the `!isToggled`?

```js
const toggle = useCallback((forceIsToggled) => {
  setIsToggled(typeof forceIsToggled === 'boolean' ? forceIsToggled : !isToggled);
}, [isToggled]);
```

montezume: hm, cool idea 👍

tdeekens: Hm. I don't like that force boolean arg. I like those tiny functions.

montezume: this is the result of the force bool arg [here](https://github.com/commercetools/ui-kit/pull/984/commits/c56b5355d6672f95e2421a938ee983ba617d6d5b)

emmenko: Yeah I also prefer to have one setter and therefore one tuple as the return value. Having additional values feels fragile to me and it's really not clear that those functions are available.

I also assume that forcing the state is not something we would do regularly, so I'm fine with this being the exception.

src/hooks/use-toggle-state/use-toggle-state.js
```diff
@@ -1,12 +1,26 @@
...
 
 const useToggleState = (defaultValue = true) => {
-  const [isToggled, setIsToggled] = React.useState(defaultValue);
-  const toggle = React.useCallback(() => {
-    setIsToggled(!isToggled);
-  }, [isToggled]);
+  const [isToggled, setIsToggled] = useState(defaultValue);
 
-  return [isToggled, toggle];
+  const setOn = useCallback(() => {
```
emmenko: Shouldn't we remove those extra functions?

montezume: err yeah

src/hooks/use-toggle-state/use-toggle-state.js
montezume: I got the original idea to add the isOn and isOff functions from this [hook](https://github.com/sgenoud/quarolib/blob/master/hooks/useToggle.js)

... because I was looking at his `usePopper` hook 😆

jonnybel: I also like having the smol fns, but I agree it might be better to keep it simple with just a tuple.
 -  -  -

> this way I need to remember the order of the setOn and setOff functions

Alternatively, we could also make it return an object instead of an array, but its a bit less convenient for naming the fns.

montezume:
@emmenko do you have any more concerns?

montezume:
I'm going to rebase master and that should fix the percy hiccup

PR #993, opened 2019-08-05
chore(select-input): add test for when autoFocussed is true

### LINKED ISSUE #897: Select inputs cannot be interacted with in tests when using autoFocus
**Describe the bug**
When testing (with RTL) select fields that have the `autoFocus` enabled, the fields do not seem to react to the triggered events, making it impossible to perform a test.

**To Reproduce**
See https://github.com/JedWatson/react-select/issues/3655

montezume:
#### Summary

Adds test case with `isAutofocussed` to `SelectInput` 

Closes #897 

Big thanks to @iamchanii and the help in https://github.com/JedWatson/react-select/issues/3655#issuecomment-518158533

### DIFF
--- a/src/components/inputs/select-input/select-input.spec.js
+++ b/src/components/inputs/select-input/select-input.spec.js
@@ -116,6 +116,26 @@ describe('in single mode', () => {
     });
   });
   describe('interacting', () => {
+    describe('when isAutofocussed is `true`', () => {
+      it('should open the list and all options should be visible', () => {
+        const { getByLabelText, getByText } = renderInput({
+          isAutofocussed: true,
+        });
+
+        const input = getByLabelText('Fruit');
+
+        fireEvent.blur(input);
+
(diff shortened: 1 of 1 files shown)


src/components/inputs/select-input/select-input.spec.js
```diff
@@ -116,6 +116,26 @@ describe('in single mode', () => {
...
   describe('interacting', () => {
+    describe('when isAutofocussed is `true`', () => {
+      it('should open the list and all options should be visible', () => {
+        const { getByLabelText, getByText } = renderInput({
+          isAutofocussed: true,
+        });
+
+        const input = getByLabelText('Fruit');
+
+        fireEvent.blur(input);
```
emmenko: So this was the trick?

montezume: yeah... 🤦‍♂.

PR #994, opened 2019-08-05
refactor(tooltip): replace react-popper with use-popper

montezume:
#### Summary

Replaces the usage of `react-popper` with [`use-popper`](https://github.com/sandiiarov/use-popper)


src/components/tooltip/tooltip.js
```diff
@@ -148,40 +148,31 @@ const Tooltip = props => {
...
-          </WrapperComponent>
-        )}
-      </Reference>
+    <React.Fragment>
+      <WrapperComponent {...eventListeners} ref={reference.ref}>
+        {React.cloneElement(props.children, { ...childrenProps })}
+      </WrapperComponent>
       {tooltipIsOpen && (
         <TooltipWrapperComponent>
-          <Popper placement={props.placement} positionFixed={true}>
```
montezume: I think it was a mistake to have positionFixed set by default anyways, it's not really the behaviour we are looking for

montezume:
I love how we don't need to make changes to our specs after switching out the implementation from `react-popper` to a hook. Pretty great stuff 🎈

PR #995, opened 2019-08-06
chore: remove unused dep and move to dev

montezume:
#### Summary 

Now that we use hooks we don't need this to be an explicit dep anymore (we no longer use `withTheme` hoc.


montezume:
No change in lock file means that I was right 😉

PR #997, opened 2019-08-07
fix(tooltip): use defined variable

montezume:
#### Summary

Regression in the last release.


montezume:
I'll trigger a release after to fix our gatsby build in app kit

PR #999, opened 2019-08-08
fix(inputs): do not generate new field id on each render

montezume:
#### Summary

Fixes bug with inputs, where on every render (when an id is not provided), a new ID is generated.

Introduces new hook `useFieldId`

### DIFF
--- a/src/components/inputs/time-input/time-input.js
+++ b/src/components/inputs/time-input/time-input.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
+import useFieldId from '../../../hooks/use-field-id';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
-import getFieldId from '../../../utils/get-field-id';
 import createSequentialId from '../../../utils/create-sequential-id';
 import Constraints from '../../constraints';
 import { parseTime } from '../../../utils/parse-time';
@@ -24,8 +24,9 @@ const format24hr = ({ hours, minutes, seconds, milliseconds }) => {
 const hasMilliseconds = parsedTime => parsedTime.milliseconds !== 0;
 
 const TimeInput = props => {
-  const id = getFieldId(props, {}, sequentialId);
+  const [id] = useFieldId(props.id, sequentialId);
   const intl = useIntl();
+
   const { name, value, onBlur, onChange } = props;
 
   const emitChange = React.useCallback(
@@ -55,14 +56,17 @@ const TimeInput = props => {
 
   // so that it doesn't run on initial mount
   const mounted = React.useRef();
-  React.useEffect(() => {
-    if (!mounted.current) {
-      mounted.current = true;
-    } else {
-      emitChange(TimeInput.toLocaleTime(value, intl.locale));
-    }
-  }, [intl.locale, emitChange, value]);
-
+  React.useEffect(
+    () => {
+      if (!mounted.current) {
+        mounted.current = true;
+      } else {
+        emitChange(TimeInput.toLocaleTime(value, intl.locale));
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [intl.locale]
+  );
   return (
     <Constraints.Horizontal constraint={props.horizontalConstraint}>
       <TimeInputBody
--- a/src/components/inputs/time-input/time-input.js
+++ b/src/components/inputs/time-input/time-input.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
 import useFieldId from '../../../hooks/use-field-id';
+import usePrevious from '../../../hooks/use-previous';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 import createSequentialId from '../../../utils/create-sequential-id';
 import Constraints from '../../constraints';
@@ -26,6 +27,7 @@ const hasMilliseconds = parsedTime => parsedTime.milliseconds !== 0;
 const TimeInput = props => {
   const [id] = useFieldId(props.id, sequentialId);
   const intl = useIntl();
+  const prevLocale = usePrevious(intl.locale);
 
   const { name, value, onBlur, onChange } = props;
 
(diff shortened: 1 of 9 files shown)


src/hooks/use-field-id/use-field-id.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import getFieldId from '../../utils/get-field-id';
+
+const useFieldId = (id, sequentialId) => {
+  const [internalId, setId] = React.useState(id);
+
+  React.useEffect(() => {
+    setId(getFieldId({ id }, { id: internalId }, sequentialId));
```
tdeekens: Sorry, I fail to follow. Why all the arguments? What does this do?

montezume:
```js
const getFieldId = (props, state, createId) =>
  props.id || state.id || createId();

export default getFieldId;
```

so if props id set set, it uses that, else it uses the internal state one, if that's not set, it uses sequentialId to create a new id

src/hooks/use-field-id/use-field-id.js
```diff
@@ -0,0 +1,14 @@
...
+  }, [id, internalId, setId, sequentialId]);
+
+  return [internalId];
```
emmenko: I would just return a single value, no need for an array

src/components/inputs/time-input/time-input.js
```diff
@@ -55,14 +56,17 @@ const TimeInput = props => {
...
-
+  React.useEffect(
+    () => {
+      if (!mounted.current) {
+        mounted.current = true;
+      } else {
+        emitChange(TimeInput.toLocaleTime(value, intl.locale));
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
emmenko: What's happening here?

src/hooks/use-field-id/use-field-id.spec.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { render } from '../../test-utils';
```
emmenko: You know that there is a testing library version for testing hooks?

https://github.com/testing-library/react-hooks-testing-library

src/hooks/use-field-id/use-field-id.js
tdeekens: I thought the same at first but then felt array is fine as it would be a breaking change to the API otherwise if later we return multiple values.

src/hooks/use-field-id/use-field-id.spec.js
montezume: Hm, but doesn't this break the test-library philosophy of testing like a user? A user doesn't render hooks 🤔

montezume: I think I would rather add this in a separate PR

tdeekens: I looked at `react-hooks-testing-library` too when implementing the `useHubSpotIdentity`-hook. I decided against it cause it's testing hooks more like a unit. However, giving the "advantage" of not having to "wrap" a hook into a `TestComponent`. I didn't mind the additional setup though for the benefit of not testing the fact that something is a hook.

src/hooks/use-field-id/use-field-id.js
emmenko: Isn't this hook only used internally?

src/hooks/use-field-id/use-field-id.spec.js
emmenko: Well, the "user test" should happen in the component right? The hook is just a function, which can be simply unit tested.

montezume: Not really, what's the point of testing the hook in every single component that uses it? It makes sense for me to test is as a user once.

montezume: BTW, I'm just mentioning this because it's come up in the past that we misunderstand each other / get into arguments.

Saying "You know that there is a testing library version for testing hooks?" comes across as rude in English. I don't think this is your intention, but that's my first impression upon reading your comment as a native speaker...

src/hooks/use-field-id/use-field-id.js
montezume: I didn't export the hook so we can make this change

src/components/inputs/time-input/time-input.js
montezume: You're right, this isn't good. I was trying to call emitChange when the locale changed.

I changed the approach in the last commit to this

```js

  const prevLocale = usePrevious(intl.locale);

  // if locale has changed
  if (typeof prevLocale !== 'undefined' && prevLocale !== intl.locale) {
    emitChange(TimeInput.toLocaleTime(value, intl.locale));
  }

```

src/hooks/use-field-id/use-field-id.js
montezume: [here](https://github.com/commercetools/ui-kit/pull/999/commits/046438d27b716566d8be8c619903f09317b4f8ff)

src/hooks/use-field-id/use-field-id.spec.js
emmenko: Yes of course this wasn't my intention, I wanted to make you aware that there was this library for testing hooks specifically. Just that.

As for the approach, it's fine either way. Thanks

src/components/inputs/time-input/time-input.js
```diff
@@ -53,15 +56,10 @@ const TimeInput = props => {
...
-  const mounted = React.useRef();
-  React.useEffect(() => {
-    if (!mounted.current) {
-      mounted.current = true;
-    } else {
-      emitChange(TimeInput.toLocaleTime(value, intl.locale));
-    }
-  }, [intl.locale, emitChange, value]);
+  // if locale has changed
+  if (typeof prevLocale !== 'undefined' && prevLocale !== intl.locale) {
```
emmenko: Hmm this looks wrong. As far as I understand it, the dependencies of `useEffect` or `useCallback` are meant exactly for checking if a value changed. So when the body function of those hooks gets executed, it means that one of the deps values changed.

Maybe I'm missing a bit the intention here. Can you clarify when exactly the `emitChange` should be called?

src/hooks/use-field-id/use-field-id.js
```diff
@@ -0,0 +1,14 @@
...
+import getFieldId from '../../utils/get-field-id';
+
+const useFieldId = (id, sequentialId) => {
+  const [internalId, setId] = React.useState(id);
+
+  React.useEffect(() => {
+    setId(getFieldId({ id }, { id: internalId }, sequentialId));
+  }, [id, internalId, setId, sequentialId]);
+
+  return internalId;
```
emmenko: Thanks 💯

src/components/inputs/time-input/time-input.js
montezume: Sure, `emitChange` should only be called with `intl.locale` changes.

Here's an gif :)

like this

montezume: It's to reformat for the currency.

montezume: another option is removing the undefined check and also do it on mount...

montezume: AFAIK, either we use a `useEffect` where we do NOT define, `value` in the deps array, or we do it like this. There's no other way.

I would like to merge this and release ui-kit before I go on vacation tonight...

emmenko: Yeah I don't see any other option at the moment.

PR #1000, opened 2019-08-08
tests: introduce `testing-library/react-hooks`

montezume:
#### Summary

Replaces the test for `useToggleState` to use `testing-library/react-hooks`


package.json
```diff
@@ -181,6 +182,7 @@
     "react-dom": "16.8.6",
     "react-intl": "3.1.1",
     "react-router-dom": "5.0.1",
+    "react-test-renderer": "16.8.6",
```
montezume: peer dep of testing-library/react-hooks

tdeekens:
...so do you like this unit testing appraoch of hooks?

montezume:
@tdeekens I'm not sure. I think both approaches have value. Maybe I should have titled this "technical exploration". This PR doesn't need to be merged 😇

tdeekens:
> @tdeekens I'm not sure. I think both approaches have value. Maybe I should have titled this "technical exploration". This PR doesn't need to be merged 😇

Ah alright missed that. I guess the general question is: test hooks as functions or with a component. Both sides have pros and cons as you said. I liked testing it with components as it gives a bit of environmental expectations as to how the should work etc.

montezume:
Maybe we could do both 🤔 WDYT @emmenko ?

tdeekens:
What specific benefit would that bring? Maybe we should just list the props and cons we each see instead of just talking about 'em 😄

emmenko:
I think it depends.

Generally speaking, if I implement a hook/function in a library, it does not make much sense to test it "as a user" because there is no way to know how this function is going to be used and in which context. For example, if I implement a `sum` function, I would just unit test it, there is no point in having complex tests setup.

It also depends on the type of hook that we write. A simple hook can just be unit tested, a more complex hook (e.g. data fetching and so on) can also be tested with user interactions.

On the other hand, we usually have integration tests for components using those functions, so you already have them implicitly tested there.

So my take would be to see where it makes sense to have a more complex setup, otherwise just test it as a unit, since "it's just a function".

montezume:
I think I prefer the older approach, I'll close for now.

PR #1008, opened 2019-08-12
chore: update deps

jonnybel:
#### Summary

* Closes #1003  
* Closes #1004
* Closes #1005
* Closes #1006
* Closes #1007

Does not update `@testing-library/react` to `v8.0.9` (#1002) because it will throw Warnings in every single test (which themselves throw Errors when running in the CI) until we upgrade to react@16.9, which I think we should do separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,7 @@
     "mri": "1.1.4",
     "omit-empty": "1.0.0",
     "postcss": "7.0.17",
-    "postcss-value-parser": "4.0.1",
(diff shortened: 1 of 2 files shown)


jonnybel:
Percy wasn't triggered for some reason 🤔

emmenko:
Yeah, I had the same in app-kit...maybe some problems on their side

jonnybel:
Ran Percy locally and it also triggered the test for this branch 🤷‍♂ At least it's working now.

PR #1015, opened 2019-08-20
chore: update (most) deps

montezume:
#### Summary

Updates deps, but leaves the `test-library` and `react` updates for another PR.

* Closes #1014 
* Closes #1013 
* Closes #1012 
* Closes #1011 
* Closes #1010 
* Closes #1009


.eslintrc.yaml
```diff
@@ -101,3 +101,4 @@ rules:
   'lines-between-class-members': 0
   'react-hooks/rules-of-hooks': 'error'
   'react-hooks/exhaustive-deps': 'warn'
+  'max-classes-per-file': 0
```
montezume: I guess this was added in the latest version...

tdeekens: Yeah. I had this in the mc-be too. I just added it to the files having multiple classes so we create more of them. If you don't care about the rule disabling it entirely is sound obviously.

montezume: Yeah doing it on a case by case basis might be better

.eslintrc.yaml
```diff
@@ -26,6 +26,9 @@ plugins:
   - 'react'
   - 'react-hooks'
 overrides:
+  - files: ['*.story.js']
+    rules:
+      'max-classes-per-file': 0
```
montezume: I think it makes more sense to add an override for stories, as they had a bunch of these, and we don't really care about them as much as we do our components

PR #1018, opened 2019-08-22
chore: enable clear mocks

montezume:
https://twitter.com/RobinDrexler/status/1164515151418142721


tdeekens:
Can you maybe open a PR on the MC see if we fail there?

Issue #1017, opened 2019-08-22
Parsing values in MoneyInput 


islam3zzat:
**Describe the bug**
Different locals have different ways for formatting numbers to name 3 examples:
- Some locals use the `,` for thousands separation and `.` for fractions separation (e.g  `EN`),
- others the other way around (e.g `DE`), and others
- use ` ` for thousands separation and `.` for fractions separation (e.g `FR`).

Because of that, we had to put locale into consideration when parsing numbers.
To find out what are the throwaway and what is the fraction symbol based on the locale we used to format a random number **`9999.999`** with fractions (to get the fraction symbol) and more than 3 digits in the non-fraction part (to get the thousands symbol).
So that after formatting it can be for examle
- `"9,999.999"`
- `"9.999,999"`
- `"9 999.999"`
- `"9 999,999"`
Then we remove the number and split, the result would be for examle
- `[',', '.']`
- `['.', ',']`
- `[' ', '.']`
- `[' ', ',']`

From the output, we'll find out the `thousands symbol` (the first item) and the `fraction symbol` (the last item)
Then we remove the thousands separator and replace the fraction symbol with a `.`  from the number to be parsed (e.g **`12 345,67`)**
So that it'll be **`"12345.67"`**
Then parse the result to float  **`12345.67`.**
https://github.com/commercetools/ui-kit/blob/4d64e24f00335e3177aa6dabcab4bd8eb53f7571/src/components/inputs/money-input/money-input.js#L183

This worked fine until we found out that some locals have Exceptions to digit grouping and don't  show thousands separator for four-digits number 

https://en.wikipedia.org/wiki/Decimal_separator#Exceptions_to_digit_grouping

Which causes our solution to fail because the random number that we initially choose was 4-digits (9999.999 )and formatting it with this locale will result `"9999,999"` 
and after the removing the numbers and splitting, it'll be [","]
And when assigning the first item to the thousands separator and the second to the fractions separator we'll end up with.
thousands separator = "," 
fractions separator = undefined

**To Reproduce**
1. Switch to the `es-PE` locale
2. go to [Price details](https://mc.commercetools.com/test-with-big-data-44/products/7af8191b-cc75-4376-be73-4959f2bfe61f/variants/4/prices/0d8a5766-c0f9-41c9-86a4-10b970331075)
3. Choose the `EUR` currency
4. Enter `7070707` for the amount
3. keep focusing in and out

[...]

PR #1019, opened 2019-08-22
Fix MoneyInput's parsing

islam3zzat:
#### Summary

In this PR we fix the issue explained thoroughly in  #1017 
which causes wrong parsing in `es` locale
The issue only happens to my knowledge in the `es` locale, and that's because in `es` there is no thousands separator for 4 digits numbers and the number that we used for figuring out the separation symbol was 4 digits number.
The issue could've been solved easily by making the number 5 or 6 digits, but the new implementation which uses the fractions separator only made more sense to me.

Currently, our test env can only provide `en` locale, that's why we can't add unit tests for this issue

### DIFF
--- a/src/components/inputs/money-input/money-input.js
+++ b/src/components/inputs/money-input/money-input.js
@@ -185,26 +185,23 @@ const createCurrencySelectStyles = ({
 // position of `.` and `,`. Whatever occurs later is used as the decimal
 // separator.
 export const parseRawAmountToNumber = (rawAmount, locale) => {
-  let throwaway;
-  let separator;
+  let fractionsSeparator;
 
   if (locale) {
-    [throwaway, separator] = (9999.999) // we need any number that has more than 3 digits to show the thousand sepirator, and has fractions
-      .toLocaleString(locale) // after this step (localizing it) it'll be either 9,999.999 or 9.999,999 based in the locale
-      .replace(/9/g, '') // then we remove the number `9` to endup with `,.` or `.,`
-      .split('')
-      .map(symbol => (symbol === '.' ? '\\.' : symbol)); // here we escape the '.' to use it as regex
+    fractionsSeparator = (2.5) // we need any number with fractions, so that we know what is the fraction
+      .toLocaleString(locale) // "symbol" for the provided locale
+      .replace(/\d/g, ''); // then we remove the numbers and keep the "symbol"
   } else {
     const lastDot = String(rawAmount).lastIndexOf('.');
     const lastComma = String(rawAmount).lastIndexOf(',');
-
-    separator = lastComma > lastDot ? ',' : '.';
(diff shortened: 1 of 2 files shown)


montezume:
If we can't use jsdom to test this, could we use a puppeteer test?

montezume:
I'm wondering if we should just get rid of all this fancy formatting of money / time. It causes more trouble than it's worth and breaks in a lot of weird ways in different languages.

montezume:
I think we should look into including full-icu and then we can have a test for this, for our date pickers, and for our time input.

montezume:
Now that https://github.com/commercetools/ui-kit/pull/1020 is merged, can you write a test 😄

islam3zzat:
Sorry for the late response got sick last friday.
> Now that #1020 is merged, can you write a test 😄

Thanks a lot.

Start writing the tests on master to make it fail first, but the tests passed.
After long time of confusion, found out that the formatting is not similar to the formatting in the browser. Which blocking me from reproducingg the issue in a unit test.

try adding to any test file, and observe the output
```js
console.log(
  'Formatting 4 digit number in `es` in our test env: ',
  (2222.222).toLocaleString('es')
);
```
For me the output was  ` Formatting 4 digit number in "es" in our test env:  2.222,222 `

now check the output in the browser console 
For me it was `Formatting 4 digit number in "es" in browser:  2222,222`

> If we can't use jsdom to test this, could we use a puppeteer test?

That's what I'm going to try next.
Thanks for the suggestion. Thought that `visualspec` is just for percy.

montezume:
@islam3zzat can you create a reproduction branch? I think it's important that we are able to test our formatting, and I'd rather fix what we have rather than just jump to puppeteer...

montezume:
Hi, @islam3zzat can you check this [issue](https://github.com/unicode-org/full-icu-npm/issues/21)?

If you look at the link they provide [here](https://www.unicode.org/cldr/charts/latest/by_type/numbers.symbols.html#a1ef41eaeb6982d), you will see that our implementation is doing the "correct" formatting.

In fact, if you test it in safari, it also has the same formatting as in our tests env

Screen Shot 2019-08-26 at 11 11 53 AM

To me this seems more like a bug with Chrome than a problem with our test setup.

montezume:
On a side note, I'm really starting to think that formatting on blur was a fundamental mistake, and we should have gone with something like cleave that formats as you type 😢

islam3zzat:
> If you look at the link they provide [here](https://www.unicode.org/cldr/charts/latest/by_type/numbers.symbols.html#a1ef41eaeb6982d), you will see that our implementation is doing the "correct" formatting.

I can't open the link, but right now we can't reproduce the issue that happens in chrome.
Our test setup, Firefox, and Safari are consistent.

PR #1020, opened 2019-08-23
chore: run tests with full-icu

montezume:
#### Summary

We want to be able to test formatting in different languages. Node doesn't support this out of the box. So we use `full-icu`. But `full-icu` is broken with yarn. So we bring the dependency internal to a "vendors" folder until they fix the problem.

cc @islam3zzat


package.json
```diff
@@ -52,8 +52,9 @@
     "format:js": "prettier --write '**/*.js'",
     "format:md": "prettier --write --parser markdown '**/*.md'",
     "format:yaml": "prettier --write --parser yaml '**/*.yaml'",
-    "test": "jest --config jest.test.config.js",
-    "test:watch": "jest --config jest.test.config.js --watch",
+    "postinstall": "cd vendors && npm install",
```
tdeekens: Nit: `yarn --cwd` or `npm --prefix` should work too instead of `cd.

.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ commands:
   unit_test:
     description: 'Running unit tests'
     steps:
-      - run: yarn cross-env NODE_ICU_DATA=./vendors/node_modules/full-icu run jest --projects jest.test.config.js --maxWorkers=3 --reporters jest-silent-reporter
+      - run: yarn run cross-env NODE_ICU_DATA=vendors/node_modules/full-icu jest --projects jest.test.config.js --maxWorkers=3 --reporters jest-silent-reporter
```
tdeekens: Nit: the run scripts use `npm` CI uses `yarn`? Maybe settle on one CLI?

montezume: which run script do you mean?

montezume: There is currently an issue with installing `full-icu` with `yarn`. See [issue](https://github.com/unicode-org/full-icu-npm/issues/9).

tdeekens: Ah sorry :)

PR #1042, opened 2019-09-02
feat(collapsible-motion): add minHeight prop

montezume:
#### Summary

Adds a `minHeight` property to `CollapsibleMotion` which allows us to animate between two different heights.


montezume:
I'm going to add a regular test. I can try a VRT, let's see if Percy works at disabling the animations

montezume:
I think testing it with the RTL test should be sufficient...

PR #1043, opened 2019-09-03
fix(collapsible-motion): do not set to null on rerender

montezume:
#### Summary

Fix for the `CollapsibleMotion` component. Currently if the component is rerendered without the isOpen prop changing, the animation is lost. Leading to bugs

### DIFF
--- a/src/components/collapsible-motion/collapsible-motion.js
+++ b/src/components/collapsible-motion/collapsible-motion.js
@@ -36,6 +36,7 @@ const createClosingAnimation = (height, minHeight) =>
 
 const useToggleAnimation = (isOpen, toggle, minHeight) => {
   const nodeRef = React.useRef();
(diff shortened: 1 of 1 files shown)


src/components/collapsible-motion/collapsible-motion.js
```diff
@@ -65,16 +66,14 @@ const useToggleAnimation = (isOpen, toggle, minHeight) => {
     ? { height: 'auto' }
     : { height: getMinHeight(minHeight), overflow: 'hidden' };
 
-  let animation = null;
-
   // if state has changed
   if (typeof prevIsOpen !== 'undefined' && prevIsOpen !== isOpen) {
```
montezume: before if the component re-rendered without a toggle, animation was reset to null. Now the animation stays.

PR #1056, opened 2019-09-11
fix(flat-button): fully support theming

montezume:
#### Summary

We were supporting theming, but not for the "neutral" state.


montezume:
@tdeekens not entirely sure to be honest. @pa3 is having some issue with theming on the FlatButton, and when I looked at the code I saw this...

PR #1068, opened 2019-09-23
chore: simplify localized text input

montezume:
#### Summary

Uses `TextInput` in our `LocalizedTextInput`, instead of our current approach of copy pasting the component.

Adds the `className` field to `TextInput`

### DIFF
--- a/src/components/inputs/text-input/README.md
+++ b/src/components/inputs/text-input/README.md
@@ -19,6 +19,7 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
 | ---------------------- | -------- | :------: | ---------------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
 | `id`                   | `string` |    -     | -                            | -       | Used as HTML `id` property                                                                             |
 | `name`                 | `string` |    -     | -                            | -       | Used as HTML `name` property                                                                           |
+| `className`            | `string` |    -     | -                            | -       | Used as HTML `className` property                                                                      |
 | `value`                | `string` |    ✅    | -                            | -       | Value of the input                                                                                     |
 | `autoComplete`         | `string` |    -     | -                            | -       | Used as HTML `autocomplete` property                                                                   |
 | `onChange`             | `func`   |    -     | -                            | -       | Called with the new value. Required when input is not read only. Parent should pass it back as `value` |
--- a/src/components/inputs/text-input/README.md
+++ b/src/components/inputs/text-input/README.md
@@ -19,7 +19,6 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
 | ---------------------- | -------- | :------: | ---------------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
 | `id`                   | `string` |    -     | -                            | -       | Used as HTML `id` property                                                                             |
 | `name`                 | `string` |    -     | -                            | -       | Used as HTML `name` property                                                                           |
-| `className`            | `string` |    -     | -                            | -       | Used as HTML `className` property                                                                      |
 | `value`                | `string` |    ✅    | -                            | -       | Value of the input                                                                                     |
 | `autoComplete`         | `string` |    -     | -                            | -       | Used as HTML `autocomplete` property                                                                   |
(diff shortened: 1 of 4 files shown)


src/components/inputs/text-input/README.md
```diff
@@ -19,6 +19,7 @@ import { TextInput } from '@commercetools-frontend/ui-kit';
 | ---------------------- | -------- | :------: | ---------------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
 | `id`                   | `string` |    -     | -                            | -       | Used as HTML `id` property                                                                             |
 | `name`                 | `string` |    -     | -                            | -       | Used as HTML `name` property                                                                           |
+| `className`            | `string` |    -     | -                            | -       | Used as HTML `className` property                                                                      |
```
tdeekens: Should we really document it. Feels like we just need some private APIs to compose things better but actually do not want to expose it as a public API.

montezume: Yeah, I wasn't sure about this. I guess it's also something that should be dealt with in #706.

tdeekens: Yeah it would be solved with it. Still spreading everything related to styling can also allow people to customize things a bit too much. It's nice for regular Html% spec'ed things but what `className` or `css` or other props might not always be intended?

montezume: I kind of like this pattern

 https://www.freecodecamp.org/news/introducing-the-single-element-pattern-dfbd2c295c5d/

But it's not really feasible to change all our components to match it 🙈

tdeekens: Yes, personally I do always value:

1. Forwarding props from web standards (just hard to maintain a complete list)
2. Forward all event handlers
3. Allow element customization through `as`-prop

What I think depends on the type of component library is

1. Allow passing `className` or `styles` and merge them

My "type of component library" I mean:

1. Is it totally brandable so something like MaterialUI or is it something which somehow should reseble and encapsulate an existing corporate identify
2. How much to you as a library author want to maintain/control the themability 

Historically, the MC has had just very bad experience with allowing passing `className` and `style` as overwrites as people went completely over board to hack things into the components intead of contributing back the the component library and maintaining a level of consistency.

montezume: So I guess here we have four options.

* Use `className` for this input and document it.
* Use `INTERNAL_className` or something and map that internally to to className in that component, and don't document it.
* Use `className` and don't document it.
* Keep the copy pasted component approach and do nothing.

jonnybel: FWIW, I don't oppose having `className` on the components, because there will almost always be special cases in the App that will require some kind of customization, as well as other different cases such as in this PR.
It should be discouraged (in the documentation) to use them, but they can be so very useful and spare a lot of effort.

But sad to hear that it was tried and a bad experience...
In the end, the developer can still customize the component if he really wants using other "hacks", but I guess the question is that by having the prop and documenting it we are somewhat promoting that behaviour?

---

Anyway, here I'd go for using `className` internally without documenting it.

tdeekens: > Anyway, here I'd go for using className internally without documenting it.

Same. I think it gives people the ability to customize but discourages in a way while not giving the regular API contract allowing us to evolve. There are just quite annoying cases we don't think of yet. Trying to come up with some. E.g.: 

1. Using `className` reaching into the `button > label` and style it
   - Now the component changes internal HTML structure and breaks somebodies styling. Is that is a breaking change?
2. Using `className` to change alignment of button and label
   - Not the component uses `css`-prop to change it too having higher specificity. Is that a breaking change?

I do the other side of the argument but really would keep to my point of: having to customize a component library through such low-level constructs against `className` and `styles` to me is a smell that the compnents allow to do what they should or the organization hasn't really settled on the systematic approach behind it.

montezume: Okay, so back on this PR, I used `className` internally and removed the changes to the readme.

Going to try to figure out why Percy is crying.

montezume: So I re-triggred circle on the last commit and now percy passes 🤷‍♂

PR #1069, opened 2019-09-24
chore(rich-text-input): improve performance

montezume:
#### Summary

I started playing around with creating a `LocalizedRichTextInput`, but found that after only two languages, the performance got horrendous.

After analyzing the situation and talking to the helpful people on the slateJS slack channel, we figured out that the component was always rerendering, even when there was no change to the `value` prop. Adding a `componentShouldUpdate` to the component seems to have improved the performance a lot, and now it does not rerender unless a prop actually changes.


src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -78,41 +81,65 @@ const plugins = [
...
-      renderEditor={renderEditor}
-    />
-  );
-};
+  shouldComponentUpdate(nextProps) {
+    let remainingPropsHaveChanged = false;
+    const remainingProps = omit(this.props, propsToOmit);
+
+    // check remaining props normally using !==
+    Object.entries(remainingProps).forEach(([key, val]) => {
```
montezume: this is just how `shouldComponentUpdate` usually works. However, we cannot rely on it for slate's value.

src/components/inputs/rich-text-input/rich-text-input.story.js
```diff
@@ -12,50 +11,42 @@ import jsonValue from './value';
...
-                onChange={evt => onChange(evt.target.value)}
-                value={value}
-              />
-            );
-          }}
-        />
-      </Spacings.Stack>
-    </Section>
-  ));
+    const [textValue, setTextValue] = React.useState('');
```
montezume: decided to use `useState` instead of react-value.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -78,41 +81,65 @@ const plugins = [
...
+      if (nextProps[key] !== val) remainingPropsHaveChanged = true;
+    });
```
tdeekens: This is "just" a shallow equality check, correct?

montezume: yeah, that would be the default functionality of a "PureComponent", right?

tdeekens: Yes, but we still seem to need to do some custom magic on `value.equals`.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -78,41 +81,65 @@ const plugins = [
...
+
+    if (remainingPropsHaveChanged) return true;
+
+    // since value uses immutableJS, we should use .equals
+    if (nextProps.value.equals(this.props.value)) {
+      return false;
+    }
+
+    return false;
```
tdeekens: Shouldn't this be true?

src/components/inputs/rich-text-input/rich-text-input.js
montezume: `value` is the prop that was changing all the time, yes.

Yeah, we need to use immutableJS's `equals` method.

src/components/inputs/rich-text-input/rich-text-input.js
montezume: hm, why? if no props have changed, then we should not update, correct?

montezume: ah, probably should do !equals

src/components/inputs/rich-text-input/rich-text-input.js
tdeekens: Thanks. Got it. Odd though as the whole point of immutable data structures is that there're referential equality holds.

tdeekens: Wondering if one would test this.

montezume:
Okay, changed just to use `PureComponent`. Now the PR only changes `rich-text-input` to be a class based component that inherits from `PureComponent` instead of `Component`

PR #1073, opened 2019-09-25
refactor(rich-text-input): move components internal

montezume:
#### Summary

In order to create a `LocalizedRichTextInput`, we need to move around some of our logic and functionality into an internal folder.

This PR just refactors the `RichTextInput` to have its files structured in a different way. It should not change the logic or presentation of the component.

#### Review notes

If our tests pass, then it should be fine. All of the functionality (redo, undo, setting block and marks, etc, is tested in our visualspec)


src/components/inputs/rich-text-input/editor.js
```diff
@@ -116,72 +29,7 @@ const Editor = props => {
...
-  const onChangeMoreStyles = val => {
-    if (!props.editor.value.selection.isFocused) {
-      props.editor.focus();
-    }
-    props.editor.toggleMark(val.value);
-  };
-
-  const hasUndos = props.editor.hasUndos();
-  const hasRedos = props.editor.hasRedos();
+  }, [props.editor.value.document, renderToggleButton]);
```
montezume: this is a slight change -> now we only rerun the effect when the document changes, read about this in the slate JS slack channel.

src/components/inputs/rich-text-input/editor.js
```diff
@@ -192,200 +40,27 @@ const Editor = props => {
         return (
           <Constraints.Horizontal constraint={props.horizontalConstraint}>
             <Spacings.Stack scale="xs">
-              <Container
+              <RichTextBody
```
montezume: I chose to separate out the "body", because the LocalizedInput will have a different look and functionality for it's expanding / collapsing buttons, etc.

src/components/inputs/rich-text-input/editor.js
```diff
@@ -192,200 +40,27 @@ const Editor = props => {
...
+                ref={{
+                  containerRef: ref,
```
montezume: passing multiple refs with useRef is pretty weird 🤔

tdeekens: Maybe the RichTextInput could have its own context which creates the refs (among other things) so we can just ‚const { refs } = useRichTextInputContext()‘ of which one eg forwards ‚refs.container‘.

montezume: One of the refs comes from `CollapsibleMotion` though. AFAIK I can't create a collapsibleMotion hook because of how emotion is built

PR #1074, opened 2019-09-26
fix(rich-text-input): do not fire on blur on dropdown open

montezume:
#### Summary

Fixes a bug with onBlur and onFocus firing when you open / close the dropdowns in the `RichTextInput`

### DIFF
--- a/src/components/inputs/rich-text-input/rich-text-input.story.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.story.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs, boolean, text } from '@storybook/addon-knobs/react';
 import { Value } from 'slate';
+import { action } from '@storybook/addon-actions';
 import withReadme from 'storybook-readme/with-readme';
 import Spacings from '../../spacings';
 import Section from '../../../../.storybook/decorators/section';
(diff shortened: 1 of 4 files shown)


src/components/internals/rich-text-body/dropdown.js
```diff
@@ -82,15 +81,12 @@ const Dropdown = props => {
...
               <AccessibleButton
-                ref={buttonRef}
                 label={props.label}
                 css={getButtonStyles()}
                 {...toggleButtonProps}
                 onMouseDown={event => {
                   event.preventDefault();
                   toggleMenu();
-                  // because we can't use onClick, let's fire the focus manually
-                  if (buttonRef.current) buttonRef.current.focus();
```
montezume: this was a bad idea

src/components/internals/rich-text-body/dropdown.styles.js
```diff
@@ -32,25 +32,22 @@ const getButtonStyles = (props = { isStyleButton: true }) => [
...
-      * {
-        fill: ${vars.colorSurface};
-      }
-    }
   `,
   props.isIndeterminate &&
     css`
       background-color: ${vars.colorAccent95};
     `,
+  props.isOpen &&
```
montezume: we can't use &:focus, so we just pass the downshift `isOpen` prop

src/components/inputs/rich-text-input/rich-text-input.story.js
```diff
@@ -36,6 +37,8 @@ storiesOf('Components|Inputs', module)
               false
             )}
             value={value}
+            onBlur={action('onBlur')}
```
jonnybel: Do we need these here?

montezume: Yeah, it's nice to have them in the story for debugging. We have the same in a bunch of other components.

PR #1082, opened 2019-10-02
refactor(rich-text-input): focus rework

montezume:
#### Summary

Some internal changes to how we deal with focus within the `RichTextInput`. Also makes the buttons in the toolbar accessible.


src/components/inputs/rich-text-input/editor.js
```diff
@@ -18,7 +18,7 @@ const Editor = props => {
   const intl = useIntl();
   const ref = React.useRef();
 
-  const prevIsFocused = usePrevious(props.isFocused);
+  const prevIsFocused = usePrevious(props.editor.value.selection.isFocused);
```
montezume: I removed the `isFocused` state variable, now we only use the internal slate isFocused. This leads to less unnecessary re-renders.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -10,31 +10,39 @@ import html from '../../internals/rich-text-utils/html';
...
 
   // this issue explains why we need to use next() + setTimeout
   // for calling our passed onBlur handler
   // https://github.com/ianstormtaylor/slate/issues/2434
   onBlur = (event, editor, next) => {
+    // we don't call next() if it's a button to stop our input from losing
+    // slate focus
+    if (
+      event.relatedTarget &&
+      event.relatedTarget.getAttribute('data-button-type') ===
```
montezume: we prevent slate from firing `onBlur` when it is triggered by a button within the component's toolbar.

src/components/internals/rich-text-body/multi-dropdown.js
```diff
@@ -63,15 +60,7 @@ const MultiDownshift = props => {
...
   return (
     <Downshift onSelect={props.onSelect} itemToString={itemToString}>
-      {({
-        getToggleButtonProps,
-        getMenuProps,
-        isOpen,
-        getItemProps,
-        toggleMenu,
-      }) => {
-        const toggleButtonProps = omit(getToggleButtonProps(), propsToRemove);
```
montezume: no need to omit because now we can just use `onClick`, as we are preventing the onBlur at the higher level.

src/components/internals/rich-text-body/rich-text-body-button.js
```diff
@@ -1,59 +1,71 @@
...
-      onMouseDown(event);
-    },
-    [onMouseDown]
-  );
+  const restOfProps = omit(props, propsToOmit);
+
   return (
-    <div>
-      <AccessibleButton
-        isDisabled={props.isDisabled}
```
jonnybel: Since you're no longer using the `AccessibleButton`, should the `MouseDown` prop be removed from it?

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -10,26 +10,24 @@ const AccessibleButton = React.forwardRef((props, ref) => (
...
-        padding: 0;
-        margin: 0;
-        white-space: nowrap;
-        cursor: pointer;
-        color: inherit;
-        font: inherit;
-        font-size: ${vars.fontSizeDefault};
-        font-family: inherit;
-      `,
-      props.isDisabled &&
```
montezume: no need to do this as we can just use css pseudo selector.

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -38,7 +36,6 @@ const AccessibleButton = React.forwardRef((props, ref) => (
     aria-disabled={props.isDisabled}
     {...(props.isToggleButton ? { 'aria-pressed': props.isToggled } : {})}
     {...props.buttonAttributes}
-    onMouseDown={props.onMouseDown}
```
montezume: not used anymore

src/components/internals/rich-text-body/rich-text-body-button.js
montezume: yep done

PR #1085, opened 2019-10-04
fix(rich-text-input): use `output` instead of `div` for contenteditable

montezume:
#### Summary

This way we can use `<label>` elements to refer to our `RichTextInput`.

Big thanks to @islam3zzat

### DIFF
--- a/src/components/inputs/rich-text-input/editor.js
+++ b/src/components/inputs/rich-text-input/editor.js
@@ -103,15 +103,9 @@ const Editor = props => {
 
 // eslint-disable-next-line react/display-name
 const renderEditor = (props, editor, next) => {
-  const children = next();
-
-  const transformedChildren = {
-    ...children,
-    props: {
-      ...children.props,
-      tagName: 'output',
-    },
-  };
+  const children = React.cloneElement(next(), {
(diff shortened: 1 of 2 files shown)


src/components/inputs/rich-text-input/editor.js
```diff
@@ -106,6 +105,14 @@ const Editor = props => {
 const renderEditor = (props, editor, next) => {
   const children = next();
 
+  const transformedChildren = {
+    ...children,
+    props: {
+      ...children.props,
+      tagName: 'output',
```
montezume: by default with slate it's `div`. this changes it to output

src/components/inputs/rich-text-input/editor.js
```diff
@@ -103,15 +103,9 @@ const Editor = props => {
...
-  const transformedChildren = {
-    ...children,
-    props: {
-      ...children.props,
-      tagName: 'output',
-    },
-  };
+  const children = React.cloneElement(next(), {
+    tagName: 'output',
+  });
```
tdeekens: This is far nicer than the fn on the right regardless of having to switch over to an output tag 🌟🤙🏼

PR #1089, opened 2019-10-07
fix(localized-multiline-text-input): margin fix

montezume:
#### Summary

Fixes a problem with margin with localized-multiline-text-input


montezume:
@filippobocik can you check the percy diff? if it's good I'll do the same in the localized rich text input.

PR #1090, opened 2019-10-08
chore: update yarn to 1.19.0 due to security vulnerability

tdeekens:
#### Summary

As you may know the older versions of yarn have a security vulnerability. Given that we now also use this version on dev machines and don't allow them to use their own there should be even more reason to keep is up-to-date.

Best,
T. Greekens.


montezume:
There's something off with renovate + kodiak + the Percy github integration. I've noticed a few times that Percy passes, but the github integration never updates.

https://percy.io/commercetools-GmbH/ui-kit/builds/2765754

montezume:
TL:DR robots are bad

PR #1093, opened 2019-10-08
fix(rich-text-input): fix onBlur and onFocus

montezume:
#### Summary

Regression caused by #1088 

I imagine that because we are using setTimeout, our synthetic event is gone and by not calling `persist()`, we no longer have access to `name` and `id` inside of the onBlur.

When I debug it in storybook I also get `null` as event.target for a non persisted event.

### DIFF
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -40,14 +40,26 @@ class RichTextInput extends React.PureComponent {
     next();
 
     if (this.props.onBlur) {
-      setTimeout(() => this.props.onBlur(event), 0);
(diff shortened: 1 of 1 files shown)


montezume:
I'm actually wondering now if we even need the setTimeout before calling onBlur. Playing around with it now.

jonnybel:
Hmm, to recap, why was it there in the first place?

montezume:
@jonnybel because we were having problems with focus hijacking, but it may have been fixed by refactoring the component 🤔

montezume:
wtf is this

montezume:
If I remove the setTimeout, then you need to click twice to focus it under some circumstances.

PR #1094, opened 2019-10-11
chore(rich-text-input): simplify testing

montezume:
#### Summary

Simplifies testing by not dealing directly with JSON values, and instead setting the values with HTML.

Also adds a test with some "unknown" tags, (img and anchor tag), to show that our Rich Text doesn't crash when it encounters HTML it doesn't "know"


montezume:
There should be some VRT diff due to using different "large value"

src/components/inputs/rich-text-input/rich-text-input.spec.js
```diff
@@ -0,0 +1,209 @@
...
+
+  it('should call onFocus when focused', () => {
+    const onFocus = jest.fn();
+    const { container } = render(
+      <RichTextInput {...baseProps} onFocus={onFocus} id="foo-bar" />
+    );
+    const input = container.querySelector('#foo-bar');
+    input.focus();
+    expect(onFocus).toHaveBeenCalled();
+    // expect(container.querySelector("[id='foo-bar']")).toBeInTheDocument();
```
tdeekens: ?

src/components/inputs/rich-text-input/rich-text-input.spec.js
```diff
@@ -0,0 +1,209 @@
...
+        const button = getByText('Expand');
+        expect(button).toBeInTheDocument(); // is this line necessary since getByText will fail if it's not found :thinking:
+        button.click();
+        expect(getByText('Collapse')).toBeInTheDocument();
+      });
+    });
+  });
+  describe('when defaultExpandMultilineText is disabled', () => {
+    describe('when height of text is less than 32', () => {
+      const originalOffsetHeight = Object.getOwnPropertyDescriptor(
```
tdeekens: Do you really have to restore this? Isn't each spec in it's on process anyway?

src/components/inputs/rich-text-input/rich-text-input.spec.js
montezume: oh, made a mistake in last commit.

PR #1098, opened 2019-10-15
refactor(rich-text-input): make rich text only use HTML

montezume:
#### Summary

I started playing around with using the `RichTextInput` in the MC and I found that forcing our users to use Slate values made it a lot more complicated to integrate.

This PR refactors the `RichTextInput` to still be a `controlled` component, but to not expose slate values to the consumer.

Thanks to @pa3 for questioning why we were using slate values in the first place...

### DIFF
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -113,7 +113,6 @@ class RichTextInput extends React.Component {
   };
 
   render() {
-    console.count('rendering here');
     return (
       <Editor
         {...filterDataAttributes(this.props)}
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -9,7 +9,7 @@ import plugins from '../../internals/rich-text-plugins';
 import html from '../../internals/rich-text-utils/html';
 import isEmpty from '../../internals/rich-text-utils/is-empty';
 
-const removeProps = ['value'];
+const omittedPropsForRender = ['value'];
 
 class RichTextInput extends React.Component {
   state = {
@@ -19,14 +19,20 @@ class RichTextInput extends React.Component {
   };
 
   shouldComponentUpdate(nextProps, nextState) {
-    // ignore updates for changes to `props.value`
-    const props = omit(this.props, removeProps);
+    // ignore updates for changes to `props.value` because this is
+    // actually an `HTML` value.
+
+    // instead we want to update when the slate `value` changes, which
+    // is stored in state.
+    const props = omit(this.props, omittedPropsForRender);
 
     const havePropsChanged = Object.entries(props).some(
       ([key, val]) => nextProps[key] !== val
     );
 
     if (havePropsChanged) return true;
+
+    if (this.props.value !== this.state.serializedValue) return true;
 
     if (this.state.value !== nextState.value) return true;
 
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import omit from 'lodash/omit';
 import requiredIf from 'react-required-if';
 import { Editor } from 'slate-react';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
@@ -8,12 +9,29 @@ import plugins from '../../internals/rich-text-plugins';
 import html from '../../internals/rich-text-utils/html';
 import isEmpty from '../../internals/rich-text-utils/is-empty';
 
-class RichTextInput extends React.PureComponent {
+const removeProps = ['value'];
+
+class RichTextInput extends React.Component {
   state = {
     // we keep track of the serialized (HTML) value
     serializedValue: this.props.value || '',
     value: html.deserialize(this.props.value || ''),
   };
+
+  shouldComponentUpdate(nextProps, nextState) {
+    // ignore updates for changes to `props.value`
+    const props = omit(this.props, removeProps);
+
+    const havePropsChanged = Object.entries(props).some(
+      ([key, val]) => nextProps[key] !== val
+    );
+
+    if (havePropsChanged) return true;
+
+    if (this.state.value !== nextState.value) return true;
+
+    return false;
+  }
 
   componentDidUpdate() {
     // if the value provided is not in sync with the RichTextInput,
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -15,7 +15,7 @@ class RichTextInput extends React.Component {
   state = {
     // we keep track of the serialized (HTML) value
     serializedValue: this.props.value || '',
-    value: html.deserialize(this.props.value || ''),
+    internalSlateValue: html.deserialize(this.props.value || ''),
   };
 
   shouldComponentUpdate(nextProps, nextState) {
@@ -34,7 +34,8 @@ class RichTextInput extends React.Component {
 
     if (this.props.value !== this.state.serializedValue) return true;
 
-    if (this.state.value !== nextState.value) return true;
+    if (this.state.internalSlateValue !== nextState.internalSlateValue)
+      return true;
 
     return false;
   }
@@ -44,7 +45,7 @@ class RichTextInput extends React.Component {
     // then reset our value by deserializing this new value
     if (this.props.value !== this.state.serializedValue) {
       this.setState({
-        value: html.deserialize(this.props.value),
+        internalSlateValue: html.deserialize(this.props.value),
         serializedValue: this.props.value,
       });
     }
@@ -64,7 +65,7 @@ class RichTextInput extends React.Component {
     this.props.onChange(fakeEvent);
 
     this.setState({
-      value: event.value,
+      internalSlateValue: event.value,
       serializedValue,
     });
   };
@@ -112,6 +113,7 @@ class RichTextInput extends React.Component {
   };
 
   render() {
+    console.count('rendering here');
     return (
       <Editor
         {...filterDataAttributes(this.props)}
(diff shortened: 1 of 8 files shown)


src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -10,16 +9,40 @@ import html from '../../internals/rich-text-utils/html';
...
+    serializedValue: this.props.value || '',
+    value: html.deserialize(this.props.value || ''),
+  };
+
+  componentDidUpdate() {
+    // if the value provided is not in sync with the RichTextInput,
+    // then reset our value by deserializing this new value
+    if (this.props.value !== this.state.serializedValue) {
+      this.setState({
+        value: html.deserialize(this.props.value),
```
montezume: while the component is controlled, it "acts" uncontrolled until the value it receives from the consumer differs from it's internal value. This way we don't have to deserialize on every render, and we keep slate's undo functionality, etc

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -10,16 +9,40 @@ import html from '../../internals/rich-text-utils/html';
...
+
   onValueChange = event => {
+    const serializedValue = html.serialize(event.value);
+
     const fakeEvent = {
       target: {
         id: this.props.id,
         name: this.props.name,
-        value: event.value,
+        value: serializedValue,
```
montezume: we expose only HTML to the consumer

src/components/inputs/rich-text-input/rich-text-input.spec.js
```diff
@@ -114,39 +114,13 @@ describe('RichTextInput', () => {
 });
 
 describe('RichTextInput static methods', () => {
-  describe('serialize / deserialize', () => {
```
montezume: don't need to test these as static methods as now it's not exposed.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -10,16 +9,40 @@ import html from '../../internals/rich-text-utils/html';
...
+  state = {
+    // we keep track of the serialized (HTML) value
+    serializedValue: this.props.value || '',
+    value: html.deserialize(this.props.value || ''),
+  };
+
+  componentDidUpdate() {
+    // if the value provided is not in sync with the RichTextInput,
+    // then reset our value by deserializing this new value
+    if (this.props.value !== this.state.serializedValue) {
```
tdeekens: How much can this just be a memoized value instead of keeping the syncing in a lifecycle.

montezume:
I'll add a test case to the visual spec, one sec.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -1,25 +1,66 @@
...
+const removeProps = ['value'];
+
+class RichTextInput extends React.Component {
+  state = {
+    // we keep track of the serialized (HTML) value
+    serializedValue: this.props.value || '',
+    value: html.deserialize(this.props.value || ''),
+  };
+
+  shouldComponentUpdate(nextProps, nextState) {
```
montezume: instead of using PureComponent, we do something similar but ignoring the prop 'value'

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -8,13 +9,30 @@ import plugins from '../../internals/rich-text-plugins';
 import html from '../../internals/rich-text-utils/html';
 import isEmpty from '../../internals/rich-text-utils/is-empty';
 
-class RichTextInput extends React.PureComponent {
+const removeProps = ['value'];
```
tdeekens: Nit: `omittedPropsForRender`

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -8,13 +9,30 @@ import plugins from '../../internals/rich-text-plugins';
...
+    // ignore updates for changes to `props.value`
+    const props = omit(this.props, removeProps);
+
+    const havePropsChanged = Object.entries(props).some(
+      ([key, val]) => nextProps[key] !== val
+    );
+
+    if (havePropsChanged) return true;
+
+    if (this.state.value !== nextState.value) return true;
```
tdeekens: I am confused but I presume it's just me. We omit `value` and if anything else changes want to render. THen we check if value changed and if so render. Isn't that the same as just not omitting `value` in the first place?

montezume: the `prop` value is actually a serializedValue. We just call it `value` so we don't expose to the consumer that under the hood we don't deal with HTML values.

If update if the state `value` changes, which is a `slate` value 😄

src/components/inputs/rich-text-input/rich-text-input.visualspec.js
```diff
@@ -336,5 +336,33 @@ describe('RichTextInput', () => {
 
     numOfTags = await getNumberOfTags('li');
     expect(numOfTags).toEqual(2);
+
+    // remove all the text
+    await selectAllText(input);
+    await input.press('Backspace');
+
```
montezume: this test basically shows that we can externally control the value, and then still use the input after

src/components/inputs/rich-text-input/rich-text-input.js
montezume: It would probably be clearer if we called `value` something else, but then it makes it weird for the consumer who doesn't really care about slate vs draftjs vs whatever...

tdeekens: It's not about the value. It's about the code no matter what value is so me it reads as

1. Exclude `value`: if anything else changed then render
2. If `value` changed then render

Which is the same as not excluding it in the first plae.

montezume: 1. Exclude `props.value`. 
2. If `state.value` changes then render. Not `props.value`

Ignore props.value. Clear?

src/components/inputs/rich-text-input/rich-text-input.story.js
```diff
@@ -56,14 +59,25 @@ storiesOf('Components|Inputs', module)
     return (
       <Section>
         <Spacings.Stack scale="l">
+          <Value
+            defaultValue={''}
+            render={(value, onChange) => (
+              <TextInput
```
jonnybel: Is this extra Input a leftover?

montezume:
@tdeekens would renaming `state.value` to something like `state.internalSlateValue` help?

src/components/inputs/rich-text-input/rich-text-input.visualspec.js
```diff
@@ -336,5 +336,33 @@ describe('RichTextInput', () => {
...
+    await resetButton.click();
+
+    // 1 strong tag should be in the document
+
+    numOfTags = await getNumberOfTags('strong');
+    expect(numOfTags).toEqual(1);
+
+    await wait(() => getByText(doc, 'Hello World'));
+
+    await input.click();
```
montezume: this refocusing the input, as we removed focus by clicking on our reset button.

src/components/inputs/rich-text-input/rich-text-input.js
jonnybel: I like this approach, but I've tested and found that we're re-rendering the RichTextInput a lot more times than in the previous "version".
For instance, I noticed that when out of focus, clicking inside the input causes 9 (!) more renders, without actually typing anything.

Haven't tracked down the cause yet 🤔

montezume: have you tried pulling the latest code?

jonnybel: Yes, I'm testing with the latest.

montezume: Yeah the problem is that `ReactValue`'s onChange is changing on every re-render.

src/components/inputs/rich-text-input/rich-text-input.story.js
montezume: no, it makes it easier to test the focus / blur of our input.

tdeekens:
Would be so dope if one could test the render count somehow. This would ensure that stuff like this doesn't trigger performance regressions on that side.

montezume:
@tdeekens yeah that's an interesting idea but it sort of goes against RTL's philosophy doesn't it? 🤔

tdeekens:
> @tdeekens yeah that's an interesting idea but it sort of goes against RTL's philosophy doesn't it? 🤔

I would argue it is even in its philosophy. The performance is a user concern and affects the user. One might test it with a test component and ref just counting up. Not sure though.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -1,25 +1,73 @@
...
+
+    const havePropsChanged = Object.entries(props).some(
+      ([key, val]) => nextProps[key] !== val
+    );
+
+    if (havePropsChanged) return true;
+
+    if (this.props.value !== this.state.serializedValue) return true;
+
+    if (this.state.internalSlateValue !== nextState.internalSlateValue)
```
pa3: Does this check make sense? Internal slate value gonna change whenever user types something, which means previous check would already return `true`.

In general looks like this approach is a bit over-complicated. Wouldn't something like this do the same with the same performance, but with the less complexity:

```js
const propsToState = props => ({
    serializedValue: props.value || '',
    internalSlateValue: html.deserialize(props.value || ''),
}); 

class RichTextInput extends React.PureComponent {
  state = propsToState(this.props);

  static getDerivedStateFromProps(props, state) {
    const nextState = propsToState(props);
    if (nextState.serializedValue == state.serializedValue) return null;
    return nextState;
  }  

...

}
```

without the need to implement `componentDidUpate` and `shouldComponentUpdate`? Or am I missing some use-case?

montezume: Yeah that seems to accomplish the same result in less confusing code 👍

montezume: [here](https://github.com/commercetools/ui-kit/pull/1098/commits/ba4134e57e1670f71f166f3410151f8c6f9f6cd7)

montezume: Thanks, I think this is a lot more clear

tdeekens: Nice suggestion. At least it wasn't just my head getting super confused by that logic there just not smart enough to give a suggestion 💃

Issue #875, opened 2019-06-19
Consider removing prop spreading from VRT output


montezume:
#### Summary

Right now, we show all of the props of the direct child of a `Spec` in our visual regression tests. 

```js
const Props = props => {
  const node = React.Children.only(props.children);
  const propEntries = Object.entries(node.props);
  return (
    <PropList>
      {propEntries
        .filter(([, value]) => typeof value !== 'function')
        .map(([key, value]) => (
          <Pill key={key} label={key} value={value} />
        ))}
    </PropList>
  );
};
```

This leads to visual output like this.

<img width="951" alt="Screen Shot 2019-06-19 at 10 55 10 AM" src="https://user-images.githubusercontent.com/2425013/59751416-c3cd6d80-9280-11e9-9361-08362b763dec.png">

#### Problems with this

While this was useful at the start for reviewing the initial baseline VRTs, it causes problems when we refactor our components. If you check this [PR](https://github.com/commercetools/ui-kit/pull/872/files), you will see that we didn't make any changes to the props accepted by the component. All we did was remove a surrounding HOC around the component. However, the following visual diff showed up

<img width="1909" alt="Screen Shot 2019-06-19 at 10 57 05 AM" src="https://user-images.githubusercontent.com/2425013/59751633-2f173f80-9281-11e9-8bfb-80a1ec345b32.png">

#### Suggestion approach

Remove display of props, and use the `Spec label` to properly display the component state.

jonnybel:
I think it would be nice if we could just display the props we wanted in the VRT, instead of all of them or none of them.

For instance, on this simple Toggled example, just pass the `isToggled` prop (and perhaps also the `isToggleButton`) to make it clear which are the props that cause the visual difference.

montezume:
Yeah that would make sense as well. 👍

emmenko:
@jonnybel what should we do about this?

tdeekens:
Feels to me, out of the blue, like we could add a `hideProps` so we can by default show all props but hide specific ones. We could also have a `showProps` depending on what's more useful maybe.

jonnybel:
Currently, we can either omit all props with `omitPropsList` and we can filter which props do display with `propsToList`, but most of the VRTs are still not using it.

I'm not sure if it's worth to go through all VRTs and make them to use the filter. I guess we should try to use the filter when we refactor components and their VRT that show a diff in the props list, and also to newly added VRTs.

[...]

PR #1101, opened 2019-10-15
feat(collapsible-panel): add prop to set alignment of header controls

jonnybel:
### Summary

We have a bunch of these `CollapsiblePanel` components in our MC which have a input/control on the left side of header. Currently, our component allows setting a `headerControls` but it’s always aligned to the right and there is no way to override that positioning. 
image
This example in the image is achieved by setting the input in the `header` prop, and we need to add a wrapper to make sure clicking on the input/control doesn’t trigger the collapsing of the panel. This seems unnecessary because the `headerControls` prop already does this for you, but we can’t use it because of the inflexible alignment.

### DIFF
--- a/src/components/panels/collapsible-panel/collapsible-panel.js
+++ b/src/components/panels/collapsible-panel/collapsible-panel.js
@@ -25,13 +25,14 @@ export default class CollapsiblePanel extends React.PureComponent {
     description: PropTypes.string,
     className: PropTypes.string,
     isSticky: PropTypes.bool,
-    headerControls: PropTypes.node,
     isDisabled: PropTypes.bool,
     children: PropTypes.node,
     tone: PropTypes.oneOf(['urgent', 'primary']),
     theme: PropTypes.oneOf(['dark', 'light']),
     condensed: PropTypes.bool,
     hideExpansionControls: PropTypes.bool,
+    headerControls: PropTypes.node,
+    headerControlsAlignment: PropTypes.string,
 
     // props when uncontrolled
     isDefaultClosed(props, propName, componentName, ...rest) {
@@ -73,6 +74,7 @@ export default class CollapsiblePanel extends React.PureComponent {
     theme: 'dark',
     condensed: false,
     isDisabled: false,
+    headerControlsAlignment: 'right',
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


src/components/panels/collapsible-panel/collapsible-panel.js
```diff
@@ -109,6 +111,7 @@ export default class CollapsiblePanel extends React.PureComponent {
                   css={getHeaderStyles({
                     isCondensed: this.props.condensed,
                     isDisabled: this.props.isDisabled,
+                    headerControlsAlignment: this.props.headerControlsAlignment,
```
tdeekens: `PropTypes.oneOf(['left', 'right'])`.

src/components/panels/collapsible-panel/collapsible-panel.styles.js
```diff
@@ -64,13 +64,19 @@ const getHeaderContainerStyles = ({ isDisabled, isOpen, isSticky, theme }) => {
...
+  isCondensed,
+  headerControlsAlignment,
+}) => {
   const baseStyles = css`
     display: flex;
     flex: 1;
     align-items: center;
     list-style-type: none;
-    justify-content: space-between;
+    justify-content: ${headerControlsAlignment === 'left'
```
tdeekens: I like the fact that alignment encapsulates the fact that flexbox is used instead of exposing all `justify-content` options through the prop.

src/components/panels/collapsible-panel/collapsible-panel.story.js
```diff
@@ -37,6 +37,7 @@ storiesOf('Components|Panels', module)
           theme={select('theme', ['dark', 'light'])}
           condensed={condensed}
           secondaryHeader={text('secondaryHeader', 'Subtitle')}
+          headerControlsAlignment={select('size', ['left', 'right'], 'right')}
```
montezume: shouldn't we make "left" the default?

montezume:
Looks reasonable to me

montezume:
Kind of annoying we can't really rely on our VRTs because of the new prop

https://github.com/commercetools/ui-kit/issues/875

PR #1102, opened 2019-10-16
refactor(rich-text-input): performance optimisation 

montezume:
#### Summary

Currently, the `RichTextInput` is double rendering on every key stroke.

#### Before

before

#### After
better

### DIFF
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -14,6 +14,14 @@ class RichTextInput extends React.PureComponent {
   internalSlateValue = html.deserialize(this.props.value || '');
 
   componentDidUpdate() {
+    // everytime we call `onChange`, we update `this.serializedValue`
+    // to the new HTML value
+    // this condition only occurs if the parent component takes `control`
+    // and resets the component to a different HTML value that what
+    // we expect
+    // in this case, we need to parse this new value into a value slate
+    // can understand, save this value to our class variable, and forceUpdate
+    // this keeps the component in sync.
     if (this.props.value !== this.serializedValue) {
       this.internalSlateValue = html.deserialize(this.props.value);
       this.serializedValue = this.props.value;
--- a/src/components/inputs/rich-text-input/rich-text-input.js
+++ b/src/components/inputs/rich-text-input/rich-text-input.js
@@ -9,17 +9,22 @@ import plugins from '../../internals/rich-text-plugins';
 import html from '../../internals/rich-text-utils/html';
 import isEmpty from '../../internals/rich-text-utils/is-empty';
 
-const propsToState = props => ({
-  serializedValue: props.value || '',
-  internalSlateValue: html.deserialize(props.value || ''),
-});
+class RichTextInput extends React.Component {
+  serializedValue;
+  internalSlateValue;
 
-class RichTextInput extends React.PureComponent {
-  state = propsToState(this.props);
+  constructor(props) {
+    super(props);
+    this.serializedValue = props.value || '';
+    this.internalSlateValue = html.deserialize(this.props.value || '');
+  }
 
-  static getDerivedStateFromProps(props, state) {
-    if (props.value === state.serializedValue) return null;
-    return propsToState(props);
+  componentDidUpdate() {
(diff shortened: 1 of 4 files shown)


src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -9,17 +9,22 @@ import plugins from '../../internals/rich-text-plugins';
...
+  }
 
-  static getDerivedStateFromProps(props, state) {
-    if (props.value === state.serializedValue) return null;
-    return propsToState(props);
+  componentDidUpdate() {
+    if (this.props.value !== this.serializedValue) {
+      this.internalSlateValue = html.deserialize(this.props.value);
+      this.serializedValue = this.props.value;
+      this.forceUpdate();
```
montezume: 🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪🤪

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -9,17 +9,22 @@ import plugins from '../../internals/rich-text-plugins';
 import html from '../../internals/rich-text-utils/html';
 import isEmpty from '../../internals/rich-text-utils/is-empty';
 
-const propsToState = props => ({
-  serializedValue: props.value || '',
-  internalSlateValue: html.deserialize(props.value || ''),
-});
-
 class RichTextInput extends React.PureComponent {
-  state = propsToState(this.props);
```
montezume: no more state, only refs.

src/components/inputs/rich-text-input/rich-text-input.js
```diff
@@ -9,17 +9,22 @@ import plugins from '../../internals/rich-text-plugins';
...
+  serializedValue;
+  internalSlateValue;
+
+  constructor(props) {
+    super(props);
+    this.serializedValue = props.value || '';
+    this.internalSlateValue = html.deserialize(this.props.value || '');
+  }
```
tdeekens: Why do we need a constructor and can't just `state = {}` as a property initializer?

src/components/inputs/rich-text-input/rich-text-input.js
tdeekens: Does this component still have a `shouldComponentUpdate` now?

src/components/inputs/rich-text-input/rich-text-input.js
montezume: because these are not state, they are refs?

montezume: 🤔

src/components/inputs/rich-text-input/rich-text-input.js
montezume: no, it's still a `PureComponent`

src/components/inputs/rich-text-input/rich-text-input.js
tdeekens: Sorry, I meant `internalSlateValue = props.value || ''` and ... not on state but outside of the constructor.

src/components/inputs/rich-text-input/rich-text-input.visualspec.js
```diff
@@ -350,5 +350,16 @@ describe('RichTextInput', () => {
     expect(numOfTags).toEqual(1);
 
     await wait(() => getByText(doc, 'Hello World'));
+
+    // remove bold from text
```
montezume: 🙏 that now these pass on CI

src/components/inputs/rich-text-input/rich-text-input.js
montezume: oh yeah ok

montezume:
How I Used One Simple Trick To Improve Performance By 50% 🚀 
3 min read

src/components/inputs/rich-text-input/rich-text-input.js
tdeekens: Can you add a comment why we do forceUpdate? I think nobody will know in two months time.

montezume: Sure 👍

PR #1114, opened 2019-10-21
fix(accessible-button): use not-allowed cursor on disabled

### LINKED ISSUE #1113: Buttons should have `not-allowed` cursor
#### Summary

When a Button is disabled, it should have a `not-allowed` cursor. 

They don't currently because we use `pointer-events: none`: which removes the cursor.

montezume:
#### Summary

Refactors our Buttons to display `cursor: not-allowed` when disabled.

Fixes #1113

### DIFF
--- a/src/components/buttons/accessible-button/accessible-button.spec.js
+++ b/src/components/buttons/accessible-button/accessible-button.spec.js
@@ -31,14 +31,17 @@ describe('rendering', () => {
     expect(container).toContainElement(ref.current);
   });
   it('should be marked as "disabled"', () => {
+    const onClick = jest.fn();
     const { getByLabelText } = render(
-      <AccessibleButton {...props} isDisabled={true} />
+      <AccessibleButton {...props} onClick={onClick} isDisabled={true} />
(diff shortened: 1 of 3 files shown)


src/components/buttons/accessible-button/accessible-button.spec.js
```diff
@@ -31,14 +31,17 @@ describe('rendering', () => {
...
-      'true'
+      <AccessibleButton {...props} onClick={onClick} isDisabled={true} />
     );
+    const button = getByLabelText('test-button');
+
+    expect(button).toHaveAttribute('disabled');
+    expect(button).toHaveAttribute('aria-disabled', 'true');
+
+    button.click();
+    expect(onClick).not.toHaveBeenCalled();
```
montezume: before we couldn't test this as `pointer-events: none` doesn't stop you from calling onClick manually 😆

montezume: although TBH testing `disabled` is kind of dumb

PR #1117, opened 2019-10-21
Ml fix broken yarn lock tests

montezume:
#### Summary

Updates `yarn.lock`, and fixes broken tooltip test.


.renovaterc.json
```diff
@@ -17,6 +17,6 @@
     }
   ],
   "lockFileMaintenance": {
-    "enabled": true
```
montezume: this is a huge pain and not worth it

tdeekens: I might agree but can you elaborate? This would have caught us just later, right?

montezume: I'm just not a fan of having my work day hijacked by 🤖 s. I don't like renovate because it makes updating dependencies seem "easier" than it really should be.

montezume: This is the second week in a row that I had to manually take over the renovate yarn.lock PR...

tdeekens: > I'm just not a fan of having my work day hijacked by 🤖 s.

I see but it hijacks you work day once a week.

> This is the second week in a row that I had to manually take over the renovate yarn.lock PR...

True that sounds no fun. Still if you loose a yarn.lock after not updating it for 5 months sounds like less fun even.

test/setup-tests.js
```diff
@@ -6,6 +6,17 @@ global.window.app = {
   mcApiUrl: 'http://localhost:8080',
 };
 
+if (global.document) {
```
jonnybel: This is needed for the tooltip now?

montezume: yeah, the old __mocks__ file was deleted and replaced with this which works better with the latest version of popperjs

jonnybel: OK, thanks!

.renovaterc.json
jonnybel: I have no strong opinion one way or another, so I guess let's try turning it off and see how it goes.

PR #1118, opened 2019-10-21
fix(buttons): fix onClick bug

montezume:
#### Summary

If the button is NOT disabled, then we should check that `onClick` exists before calling it 🤦‍♂ 

This usually affects link buttons.

### DIFF
--- a/src/components/buttons/secondary-button/secondary-button.spec.js
+++ b/src/components/buttons/secondary-button/secondary-button.spec.js
@@ -73,7 +73,7 @@ describe('rendering', () => {
   describe('when using "linkTo"', () => {
     it('should navigate to link when clicked', async () => {
       const { getByLabelText, history } = render(
-        <SecondaryButton {...props} linkTo="/foo/bar" />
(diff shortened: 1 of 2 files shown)


src/components/buttons/secondary-button/secondary-button.spec.js
```diff
@@ -73,7 +73,7 @@ describe('rendering', () => {
   describe('when using "linkTo"', () => {
     it('should navigate to link when clicked', async () => {
       const { getByLabelText, history } = render(
-        <SecondaryButton {...props} linkTo="/foo/bar" />
+        <SecondaryButton {...props} onClick={null} linkTo="/foo/bar" />
```
montezume: this test failed in same way as the current issues in MC-FE before making the code change 👍

montezume:
I think the failing tests were due to onClick being undefined and not being disabled.

jonnybel:
Hmm yeah, I didn't understand it correctly before 👍

PR #1121, opened 2019-10-21
feat(select-inputs): support inputValue

montezume:
Allows you to do things like this

hhere
With code that looks like

```
  const [inputValue, setInputValue] = React.useState('');

  return (
    <SelectInput
      {...props}
      inputValue={inputValue}
      onInputChange={(value, { action }) => {
        if (action === 'input-change' || action === 'set-value') {
          setInputValue(value);
        }
      }}
    />
  );
```

### DIFF
--- a/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
+++ b/src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
@@ -57,6 +57,7 @@ const AsyncCreatableSelectInput = props => {
           // This makes it easier to less confusing to use with <label />s.
           id={props.containerId}
           inputId={props.id}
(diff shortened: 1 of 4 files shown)


montezume:
Oh I should add for other selects

tdeekens:
...should it be documented?

montezume:
It's sort of already in the docs

```
This input is built on top of react-select v2. It supports mostly same properties as react-select. Behaviour for some props was changed, and support for others was dropped.

In case you need one of the currently excluded props, feel free to open a PR adding them.

```

tdeekens:
Oki doks. Should have checked. Thanks.

PR #1122, opened 2019-10-22
fix(typography): fix Text.Body to use intlMessage when 'as' prop is present

jonnybel:
Fixes a bug where `intlMessage` prop is being ignored when the `as` prop is present for the `Text.Body` component.

### DIFF
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -180,7 +180,11 @@ const Body = props => {
         title={props.title}
         {...filterDataAttributes(props)}
       >
-        {props.children}
+        {props.intlMessage ? (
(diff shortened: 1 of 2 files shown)


src/components/typography/text/text.js
```diff
@@ -180,7 +180,11 @@ const Body = props => {
         title={props.title}
         {...filterDataAttributes(props)}
       >
-        {props.children}
+        {props.intlMessage ? (
+          <FormattedMessage {...props.intlMessage} />
+        ) : (
+          props.children
+        )}
```
jonnybel: This was missing.

montezume:
Can you add a test?

PR #1125, opened 2019-10-22
feat(localized-rich-text-input): adds `LocalizedRichTextInput.createLocalizedString`

montezume:
#### Summary

Adds static function `createLocalizedString` to `LocalizedRichTextInput`.


src/components/internals/rich-text-utils/localized/localized.spec.js
```diff
@@ -0,0 +1,62 @@
...
+        en: '<p>hello world</p>',
+        de: '<p>Hallo</p>',
+      });
+    });
+  });
+});
+
+describe('isEmpty', () => {
+  describe('when calling with an undefined value', () => {
+    it('should return true', () => {
```
tdeekens: Nit: maybe `should indicate the value being empty` etc

tdeekens: Danke!

PR #1127, opened 2019-10-22
chore: build i18n

montezume:
#### Summary

Builds i18n add makes a German translation switch as requested by Maria.


i18n/data/core.json
```diff
@@ -39,6 +44,39 @@
   "UIKit.MultilineTextInput.expand": "Expand",
   "UIKit.PasswordField.hide": "hide",
   "UIKit.PasswordField.show": "show",
+  "UIKit.RichTextBody.boldButtonLabel": "Bold",
+  "UIKit.RichTextBody.boldTooltipTitle": "Bold",
+  "UIKit.RichTextBody.expandButtonLabel": "Expand",
+  "UIKit.RichTextBody.expandTooltipTitle": "Expand",
+  "UIKit.RichTextBody.italicButtonLabel": "Italic",
+  "UIKit.RichTextBody.italicTooltipTitle": "Italic",
```
emmenko: Q: why do we have so many duplicate messages? Is that on purpose? 🤔

montezume: Well, I made different messages for the tooltip message and for the button aria-label, in case we wanted different messages. But then design ended up using the same message for both. It might make sense to combine them actually, and only split if we do end up going with diff messages

montezume: @emmenko dumped them in the last commit

emmenko: Thanks. As follow up we should also do the same for all other duplicate messages (I see there are still some).

montezume: Yeah, things like `expand`, `collapse` and `show` and `hide` are really copy pasted 🤔 . We should have like a generic message for those.

`LocalizedInput.show` instead

montezume: https://github.com/commercetools/ui-kit/issues/1129

PR #1128, opened 2019-10-22
fix(rich-text-input): disabled fix

montezume:
#### Summary

Slate does not seem to respect `disabled`. So instead we pass `readOnly` 

[ref](https://docs.slatejs.org/slate-core/editor#readonly)


montezume:
VRT passed on percy, not sure why github integration isn't updating

https://percy.io/commercetools-GmbH/ui-kit/builds/2899784

PR #1134, opened 2019-10-24
fix(localized-rich-text-input): fix omit empty translations, add readme

montezume:
#### Summary

Defines an HTML centric `omitEmptyTranslations` for `LocalizedRichTextInput`.

Finishes up the README for `LocalizedRichTextInput`.

### DIFF
--- a/src/components/inputs/localized-multiline-text-input/README.md
+++ b/src/components/inputs/localized-multiline-text-input/README.md
@@ -60,7 +60,7 @@ LocalizedMultilineTextInput.createLocalizedString(languages);
 // -> { en: '', de: '' }
 ```
 
-In case existingTranslations is passed, it will merge an empty localized field with the exisiting translations. Usually this is used to ensure that a localized string contains at least the project's languages.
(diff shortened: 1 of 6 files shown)


src/components/inputs/localized-multiline-text-input/README.md
```diff
@@ -60,7 +60,7 @@ LocalizedMultilineTextInput.createLocalizedString(languages);
 // -> { en: '', de: '' }
 ```
 
-In case existingTranslations is passed, it will merge an empty localized field with the exisiting translations. Usually this is used to ensure that a localized string contains at least the project's languages.
+In case existingTranslations is passed, it will merge an empty localized field with the existing translations. Usually this is used to ensure that a localized string contains at least the project's languages.
```
montezume: just a typo fix

src/components/inputs/localized-rich-text-input/localized-rich-text-input.js
```diff
@@ -11,14 +11,14 @@ import {
   getHasErrorOnRemainingLanguages,
   getHasWarningOnRemainingLanguages,
   isTouched,
-  omitEmptyTranslations,
```
montezume: this would not omit <p></p>

Issue #1136, opened 2019-10-25
Add a component prop to RadioInput.Option that wraps its contents


dogayuksel:
Adding a Tooltip around a single RadioInput.Option is not possible since RadioInput.Group expects RadioInput.Option to be a direct child.

Possibly adding a `component` prop to RadioInput.Option and wrapping its contents with that given component on render would provide more flexibility for such use cases.

`<RadioInput.Option component={<Tooltip />} />`

tdeekens:
Thanks for creating the issue. I think it might be worth having a `components` prop so we call supply multiple, if needed, in the future. so `components={{Tooltip}}

tdeekens:
The more generic naming also sounds legit. :)

PR #1139, opened 2019-10-25
[WIP] feat(radio-input): add components prop to radio input option

dogayuksel:
#### Summary

Allows passing in a react element to wrap RadioInput.Option with.


dogayuksel:
https://github.com/commercetools/ui-kit/issues/1136

src/components/inputs/radio-input/radio-group.js
```diff
@@ -29,7 +29,7 @@ const Group = props => {
   const optionElements = React.Children.map(props.children, (child, index) => {
     // NOTE: Allowing to intersperse other elements than `Option`.
     if (child && child.type.displayName === Option.displayName) {
-      return React.cloneElement(child, {
+      const option = React.cloneElement(child, {
```
tdeekens: Nit: maybe `children` as a name would be apt here.

src/components/inputs/radio-input/radio-group.js
```diff
@@ -41,6 +41,13 @@ const Group = props => {
         onFocus: props.onFocus,
         onBlur: props.onBlur,
       });
+      return child.props.components.WrapperComponent ? (
```
tdeekens: Nice! Wondering, as I always struggle to read `<a` with a lowercase character if it is actually a React component. Would 

```js
const Wrapper = child.props.components.WrapperComponent || React.Fragment`
```

work to then just do:

```js
<Wrapper>{children}</Wrapper>
```

src/components/inputs/radio-input/radio-group.js
montezume: const `clonedChild`?

src/components/inputs/radio-input/radio-option.js
```diff
@@ -66,6 +67,16 @@ Option.propTypes = {
     PropTypes.element,
     PropTypes.func,
   ]).isRequired,
+  components: PropTypes.shape({
+    WrapperComponent: (props, propName) => {
```
montezume: is this really necessary? Can't we just make it optional?

dogayuksel:
@montezume @tdeekens 
Any ideas how to pass in props to WrapperComponent in this setup? For example, wrapper being a Tooltip and user wanting to control the `off` prop.
- Add an other prop to RadioInput.Option, like WrapperComponentProps and spread them on to WrapperComponent?
- Otherwise, WrapperComponent can be a JSX.Element but I don't know, if it is possible to append children to that.

src/components/inputs/radio-input/radio-option.js
dogayuksel: You mean the `children` prop right?

tdeekens:
Hm, good question. So far we only swaped out default components while applying "internal" props to them. Correct me if that's wrong @montezume.

1. A prop-getter like `getWrapperComponentProps` feels sort of backwards
2. A `props.props.WrapperComponent` feels also a bit odd
3. A `props.components.WrapperComponent.props` and `props.components.WrapperComponent.component` also looks like an odd API

Sorry I am just thinking out loud. 🤔 would it make sense just not to render the wrapper component in the parent? It doesn't really solve the "issue" though.

src/components/inputs/radio-input/radio-group.js
montezume: Hm, another approach we could go for just allows any `child` but leave it up to the consumer to not allow prop spreading. That way we don't need a new components prop. 

So for your case, you would do something like this. Basically capture the props in your wrapper then pass them to the component that should receive them (RadioOption).
```
const Wrapper = props => {
 <ToolTip off={isTooltipOff}>
    <RadioInput.Option {...props} />
  <ToolTip>
}
```

dogayuksel: I like this self-service approach. Would we be recursively searching for an RadioInput.Option as a child for the invariant, or removing the check all together, what do you think?

tdeekens: I would be fine with removing it. A recursive search could also be misleading. As in: somewhere in the component tree there is a `Radio.Option` but it wasn't the one the developer but "nearby" his concern. So the invariant will maybe rarely give an appropriate hint.

montezume: Hm, I'm not sure how we could do this though without making it a breaking change.

src/components/inputs/radio-input/radio-option.js
```diff
@@ -68,11 +68,23 @@ Option.propTypes = {
     PropTypes.func,
   ]).isRequired,
   components: PropTypes.shape({
-    WrapperComponent: (props, propName) => {
-      if (props[propName] && !isValidElementType(props[propName])) {
-        return new Error(
-          `Invalid prop 'components.WrapperComponent' supplied to 'RadioInput.Option': the prop is not a valid React component`
-        );
+    wrapper: (props, propName) => {
```
tdeekens: Wondering. Is this almost the same as `PropTypes.oneOf([PropTypes.func, PropTypes.array, PropTypes.elementType])`?

dogayuksel: Tried to restrict it to a function (1st if) that receives one argument (2nd if) and returns a valid react.element (3rd if)

dogayuksel: Maybe a bit over engineered and fails stupidly, thought it could help/guide the user maybe

tdeekens: True. It's a nice idea! Maybe it's also something UIKit could share and reuse internally.

montezume:
@tdeekens do you want to review?

montezume:
Not sure why Percy didn't report ... it passed https://percy.io/commercetools-GmbH/ui-kit/builds/2967506

src/components/inputs/radio-input/radio-group.js
```diff
@@ -41,6 +41,8 @@ const Group = props => {
         onFocus: props.onFocus,
         onBlur: props.onBlur,
       });
+      const { wrapper } = child.props.components;
+      return wrapper ? wrapper(clonedChild) : clonedChild;
```
tdeekens: This is cool.

PR #1145, opened 2019-10-29
fix(rich-text-input): dropdown focus bug

montezume:
#### Summary

The bug has to do with calling editor.focus() on any event propagated up to the Toolbar. The inline comment should explain the bug and the fix.

#### Before
before

#### After

after

### DIFF
--- a/src/components/internals/rich-text-body/rich-text-body.js
+++ b/src/components/internals/rich-text-body/rich-text-body.js
@@ -258,7 +258,7 @@ const RichTextEditorBody = React.forwardRef((props, ref) => {
     // if clicking on a dropdown, we need to both focus the RichTextInput
     // and open the dropdown.
     // if we don't use setTimeout, our dropdown opening will be hijacked by the
-    // editors rerendering.
+    // editors rerendering. (a so called `race condition`)
     // the reason we keep this here, is because any onClick from our dropdown or our mark buttons
     // will propogate to here.
     if (!props.editor.value.selection.isFocused) {
--- a/src/components/internals/rich-text-body/rich-text-body.js
+++ b/src/components/internals/rich-text-body/rich-text-body.js
@@ -252,6 +252,20 @@ const RichTextEditorBody = React.forwardRef((props, ref) => {
     );
   }
 
+  const { focus } = props.editor;
+
+  const onToolbarClick = React.useCallback(() => {
+    // if clicking on a dropdown, we need to both focus the RichTextInput
+    // and open the dropdown.
+    // if we don't use setTimeout, our dropdown opening will be hijacked by the
+    // editors rerendering.
+    // the reason we keep this here, is because any onClick from our dropdown or our mark buttons
+    // will propogate to here.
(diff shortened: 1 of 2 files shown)


src/components/internals/rich-text-body/rich-text-body.js
```diff
@@ -252,6 +252,20 @@ const RichTextEditorBody = React.forwardRef((props, ref) => {
...
+
+  const onToolbarClick = React.useCallback(() => {
+    // if clicking on a dropdown, we need to both focus the RichTextInput
+    // and open the dropdown.
+    // if we don't use setTimeout, our dropdown opening will be hijacked by the
+    // editors rerendering.
+    // the reason we keep this here, is because any onClick from our dropdown or our mark buttons
+    // will propogate to here.
+    if (!props.editor.value.selection.isFocused) {
+      setTimeout(focus, 0);
```
montezume: setTimeout allows our dropdown button onClick to "win" the race condition

jonnybel: Just curious, would this also work with `setImmediate(focus)` instead of `setTimeout(focus, 0)`? If so, it should be slightly more performant.

tdeekens: CleanShot 2019-10-29 at 14 03 11@2x

tdeekens: So we want the focus to happen on the next tick essentially?

montezume: @tdeekens yeah. There are lots of questions like this in the Slate slack channel. Usually people recommend to use setTimeout.

montezume: https://caniuse.com/#feat=setimmediate

montezume: I would not use it.

montezume: Screen Shot 2019-10-29 at 2 06 20 PM
No chrome support

montezume: Screen Shot 2019-10-29 at 2 06 51 PM
No Chrome support...

montezume: I also wouldn't worry about performance. This only happens when 

A) The editor is out of focus
B) The user clicks on a mark button or a dropdown button.

montezume:
Percy passing 😌

src/components/internals/rich-text-body/rich-text-body.js
tdeekens: Also like `const = scheduleFocusAfter...` would also be cool.

montezume: Err, I added like 5 lines of comments? I can add more...

jonnybel: Oh wow, I had a completely wrong idea about `setImmediate` on the browser 🤦‍♂ This ain't node 😅

PR #1147, opened 2019-10-31
feat(field-label): add tone

tdeekens:
#### Summary

This pull request adds the `tone` prop to the `<FieldLabel /`.

#### Description

We refactored parts of the MC to use the FieldLabel. Good stuff. However, the FieldLabel is also used in things like table headers. There the background is dark. As a result we would need the tone to be inverted.

Noticing that the label already accepts a tone I felt we can just forward that.


tdeekens:
Sorry for not creating an issue first but the time on the issue write up felt the same as implementing it :)

src/components/field-label/field-label.visualroute.js
```diff
@@ -58,5 +58,8 @@ export const component = () => (
         horizontalConstraint="m"
       />
     </Spec>
+    <Spec label="with inverted tone">
+      <FieldLabel title="Hello" hasRequiredIndicator={true} tone="invereted" />
```
jonnybel:
```suggestion
      <FieldLabel title="Hello" hasRequiredIndicator={true} tone="inverted" />
```

tdeekens: Thanks. I feel I mistype every third word. I need some pro keyboard. Something with Cherry clicks and stuffs.

jonnybel:
We can also pass a `Text` component to the `title` of FieldLabel and set the tone there, so I'm wondering if we actually need this prop here 🤔

tdeekens:
Yeah just felt like maybe too much wrapping. Somehow I felt that label and fieldlabel should somehow share the same props to influence their look.

src/components/field-label/README.md
```diff
@@ -55,6 +55,7 @@ The `hintIcon` also accepts a custom `color` while defaulting to `warning` in th
...
 | `description`          | `string` or `node` |    -     | -                            | -       | Provides a description for the title.                                                                                                                                                                                                                                 |
 | `onInfoButtonClick`    | `function`         |    -     | -                            | -       | Function called when info button is pressed. Info button will only be visible when this prop is passed.                                                                                                                                                               |
+| `tone`                 | `string`           |    -     | `['primary', 'inverted']`    | \_      | Indicates the tone to be applied to the label                                                                                                                                                                                                                         |
```
jonnybel:
```suggestion
| `tone`                 | `string`           |    -     | `['primary', 'inverted']`    |  _      | Indicates the tone to be applied to the label                                                                                                                                                                                                                         |
```

tdeekens:
Do visual specs have the ability to deal with white text? Do invert the colour of their canvas?

montezume:
@tdeekens yes, check other examples. `text-input.visualroute` has one.

montezume:
We can already do this on master. Is that what you are trying to do?
Screen Shot 2019-10-31 at 1 03 29 PM

tdeekens:
CleanShot 2019-10-31 at 13 11 05@2x

I need to achive that this is white. How do you suggest moving forward to do so?

montezume:
```js

const invertedTheme = {
  colorSolid: vars.colorSurface,
  colorSurface: vars.colorSolid,
};
<ThemeProvider theme={invertedTheme}><FieldLabel title="Return quantity" /></ThemeProvider>

```

montezume:
You might want to wrap the `CheckboxInput` as well with the ThemeProvider. It looks weird now.

tdeekens:
Will close as this is not the preferred solution to achieve this with UIKit.

montezume:
@tdeekens I understand your use case now. 

Maybe using `theming` as a tool is not the best in each individual case. In this case, you would need to have a `ThemeProvider` for each column definition. Which is not ideal.

If you want to continue with this that's fine, but either rename the prop to `textTone` or have it apply to `description` as well.

src/components/field-label/field-label.visualroute.js
```diff
@@ -58,5 +58,8 @@ export const component = () => (
         horizontalConstraint="m"
       />
     </Spec>
+    <Spec label="with inverted tone">
```
montezume: you can access the `themes` prop that is passed in here.
```js
export const component = ({ themes }) => (
```

and then you can do 

```js

        <Spec label="with inverted tone">
//
        </Spec>
      </ThemeProvider>

```

montezume: hm, maybe this won't really work because now the theme will apply to the "Text" inside of your component. In that case you would need to use ThemeProvider twice 🤔

tdeekens: Let's see. Two commits incoming.

src/components/field-label/field-label.visualroute.js
```diff
@@ -58,5 +59,15 @@ export const component = () => (
         horizontalConstraint="m"
       />
     </Spec>
+    <ThemeProvider theme={themes.darkTheme}>
```
montezume:
```suggestion
      <ThemeProvider theme={themes.darkTheme}>
        <Spec label="with inverted tone">
          <ThemeProvider theme={customProperties}>
            <FieldLabel
              title="Hello"
              description="description"
              hasRequiredIndicator={true}
              tone="inverted"
            />
          </ThemeProvider>
        </Spec>
      </ThemeProvider>
```

montezume: This way we are actually testing tone -> inverted. Else we are just testing that theming works 😉

src/components/field-label/field-label.js
```diff
@@ -51,11 +52,13 @@ export const FieldLabel = props => {
...
             )}
-            {props.hint && <Text.Detail>{props.hint}</Text.Detail>}
+            {props.hint && (
+              <Text.Detail tone={props.tone}>{props.hint}</Text.Detail>
+            )}
           </Spacings.Inline>
         )}
         {props.description && (
-          <Text.Detail>
+          <Text.Detail tone={props.tone}>
```
montezume:
```suggestion
          <Text.Detail>
           <Text.Wrap tone={props.tone}>
          
```

src/components/field-label/field-label.visualroute.js
```diff
@@ -58,5 +59,17 @@ export const component = () => (
...
+      <Spec label="with inverted tone">
```
montezume:
```suggestion
      <Spec label="with inverted tone" omitPropsList>
```

src/components/field-label/field-label.js
montezume: Screen Shot 2019-10-31 at 2 25 14 PM
We can tell from Percy that the tone is not being properly applied

montezume:
Sorry for the massive amount of nitpicking on this tiny PR 🙈

PR #1149, opened 2019-10-31
chore(vrts): define new spec `listPropsOfChild` prop

montezume:
#### Summary

We sometimes have Specs that look like this.

```js
      <Spec label="when inverted" omitPropsList={true}>
          <Label tone="inverted">Hello</Label>
        </ThemeProvider>
      </Spec>
```

In these cases, we would usually use `omitPropsList`, because otherwise, our props list will come from `ThemeProvider`, instead of from the component that we are testing.

By adding this prop, we can tell our `Spec` to use the child, and therefore display the correct props.

### DIFF
--- a/test/percy/README.md
+++ b/test/percy/README.md
@@ -191,5 +191,6 @@ A `Spec` should render your component in a specific state.
 
 You may use the prop `propsToList` to specify which props you want to display within the snapshot. Ideally, these should be the ones which you are testing. If `propsToList` is not defined, every prop will be displayed.
 If don't want to display any props at all, use `omitPropsList`.
+If you want to display of the child of the children you pass it, use `listPropsOfChild`. This can be useful if you need to wrap your component in a `ThemeProvider`.
 
 You can use multiple specs within a `Suite`.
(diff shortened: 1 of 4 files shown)


test/percy/README.md
```diff
@@ -191,5 +191,6 @@ A `Spec` should render your component in a specific state.
 
 You may use the prop `propsToList` to specify which props you want to display within the snapshot. Ideally, these should be the ones which you are testing. If `propsToList` is not defined, every prop will be displayed.
 If don't want to display any props at all, use `omitPropsList`.
+If you want to display of the child of the children you pass it, use `listPropsOfChild`. This can be useful if you need to wrap your component in a `ThemeProvider`.
```
jonnybel: This phrase sounds a bit confusing to me.. It's hard to refer to a "grand-child" of a component 😆 

Maybe something like
```suggestion
If you want to display the props of a nested component which is wrapped by the direct child of `Spec`, use `listPropsOfChild`. This can be useful if you need to wrap your component in a `ThemeProvider`.
```

test/percy/spec.js
```diff
@@ -100,14 +103,20 @@ const Props = props => {
 Props.displayName = 'Props';
 Props.propTypes = {
   children: PropTypes.node.isRequired,
+  listPropsOfChild: PropTypes.bool.isRequired,
```
jonnybel: Nit: How about `listPropsOfNestedChild`? Because this is a prop of `Spec`, and we're referring not to its child, but to the child of its child 👶 So it's like further _nested_  😃🤷‍♂

montezume: ya that makes sense, switched it.

PR #1150, opened 2019-10-31
fix(rich-text-input): fix for safari

montezume:
#### Summary

Currently the `RichTextInput` and the `LocalizedRichTextInput` are completely broken in Safari.

After a lot of debugging, I figured out that combining `slate` + `contenteditable` + `<output>` lead to this problem.

bug

In case anyone has forgotten, the reason why we wanted to use `output` in the first place, is because it natively supports working with `label`s and `htmlFor`.

Instead, I added a hidden input and I manually focus the `RichTextInput` based on that.

### DIFF
--- a/src/components/inputs/localized-rich-text-input/editor.js
+++ b/src/components/inputs/localized-rich-text-input/editor.js
@@ -9,6 +9,7 @@ import filterDataAttributes from '../../../utils/filter-data-attributes';
 import usePrevious from '../../../hooks/use-previous';
 import CollapsibleMotion from '../../collapsible-motion';
 import Spacings from '../../spacings';
+import accessibleHiddenInputStyles from '../../internals/accessible-hidden-input.styles';
 import { AngleUpIcon, AngleDownIcon } from '../../icons';
 import Text from '../../typography/text';
 import FlatButton from '../../buttons/flat-button';
@@ -177,12 +178,14 @@ const Editor = props => {
 
 // eslint-disable-next-line react/display-name
 const renderEditor = (props, editor, next) => {
+  const internalId = `${props.id}__internal__id`;
+
   const children = React.cloneElement(next(), {
-    tagName: 'output',
+    id: internalId,
   });
 
   const passedProps = {
-    id: props.id,
+    id: internalId,
     isDisabled: props.disabled,
     isReadOnly: props.readOnly,
     ...pick(props.options, [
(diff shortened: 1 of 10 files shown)


montezume:
I would appreciate a quick review since this is a critical bug. cc @jonnybel

src/components/inputs/localized-rich-text-input/editor.js
```diff
@@ -177,8 +178,10 @@ const Editor = props => {
 
 // eslint-disable-next-line react/display-name
 const renderEditor = (props, editor, next) => {
+  const internalId = `${props.id}__internal__id`;
+
   const children = React.cloneElement(next(), {
-    tagName: 'output',
+    id: internalId,
```
montezume: we don't want two elements with the same id, so we change the `<div contenteditable>` id to an internal id

src/components/inputs/localized-rich-text-input/editor.js
```diff
@@ -201,9 +204,16 @@ const renderEditor = (props, editor, next) => {
...
 
+  const isFocused = props.editor.value.selection.isFocused;
+
   return (
     <Editor editor={editor} {...passedProps}>
       {children}
+      <HiddenInput
+        isFocused={isFocused}
+        handleFocus={editor.focus}
+        id={props.id}
```
montezume: the hidden input receives the "real" id, so that it can receive label focus

src/components/inputs/rich-text-input/editor.js
```diff
@@ -129,6 +134,11 @@ const renderEditor = (props, editor, next) => {
   return (
     <Editor editor={editor} {...passedProps}>
       {children}
+      <HiddenInput
```
tdeekens: Is this like a focus trap?

montezume: Can you explain what a focus trap is?

tdeekens: It's a trap for the page's focus or the user's ability to tab around. So only any child within the focus-trap can be focussed. I wasn't sure if this was what we're doing here just couldn't derive fromt the name `HiddenInput`.

montezume: To understand this, we need to step back.

When you write this in HTML, what does it mean?

```<label htmlFor="name">Name</label>```

It associates this label with the text "Name", with the `<input>` who's id is "name". Right?

<input>. Not div. Not span. You can't associate a label with a <div> or a <span>

**That's the problem.**

Before we were not rendering a <div>, we were rendering an `<output>`.

You **can** associate labels with <outputs> 

However, contenteditable outputs break in Safari.

montezume: https://jsbin.com/nuqetifoli/edit?html,output

tdeekens: Ah got it. Thanks. So it's just the "label-anchor" or something like that.

PR #1151, opened 2019-11-01
refactor(rich-text-input): remove focusability from buttons

montezume:
#### Summary

Having "focusable" inputs "inside" of our `RichTextInput` is a nightmare and really hard to deal with. This solves a few bugs, including this one.

current


montezume:
If it was somehow possible to have two elements focused at the same time in HTML, then this would be a lot easier 😱

Issue #695, opened 2019-04-18
Styling bug when applying `isReadOnly` to inputs


qmateub:
**Describe the bug**
Some of the inputs in UIKit are not applying correctly the `isReadOnly` style when the prop is defined. This is a little bit unfortunate because according the the docs of each input, all of them accept the prop.

**To Reproduce**
Just create a test view and define all of the inputs with `isReadOnlyProp` to true. This is the list of the status for each input.

* AsyncCreatableSelectInput  ✅
* AsyncSelectInput  ✅
* CheckboxInput ✅
* CreatableSelectInput  ✅
* DateInput ✅
* DateRangeInput ✅
* DateTimeInput ✅
* LocalizedMoneyInput ✅
* LocalizedMultilineTextInput ✅
* LocalizedTextInput ✅
* MoneyInput ✅
* MultilineTextInput ✅
* NumberInput ✅
* PasswordInput ✅
* RadioInput ✅
* SelectInput  ✅
* TextInput ✅
* TimeInput ✅
* ToggleInput ❌

**Expected behavior**
All the inputs should follow the same styling behavior when defining `isReadOnly`

**Screenshots**
![image](https://user-images.githubusercontent.com/11442895/56356932-77ab7180-61da-11e9-8922-75abd9122441.png)

**Additional Notes**
Didn't test if this also affects the Field components that forwards the prop to the inputs 😅

montezume:
Hm, `react-select` doesn't support the 'readonly' prop... I'm not sure if makes sense for selects to be readonly. We will need to talk to design about this...

montezume:
`react-select` doesn't support read-only out of the box. So there's no easy way of making our `SelectInputs` readonly

https://github.com/JedWatson/react-select/issues/1802

montezume:
FYI @qmateub I'm going to keep updating your post as I progress on this 🙃

montezume:
FYI @jonnybel I updated the description to reflect your latest changes in #1192

PR #1157, opened 2019-11-04
feat(checkbox-input): support isReadOnly prop

montezume:
#### Summary

Addresses one of the cases [here](https://github.com/commercetools/ui-kit/issues/695)


src/components/inputs/checkbox-input/checkbox.js
```diff
@@ -11,48 +11,50 @@ const Input = styled.input`
...
-      />
-    );
-  }
-}
+  }, [props.isIndeterminate]);
+
+  const { onChange } = props;
+
+  const handleChange = React.useCallback(
+    event => !props.isReadOnly && onChange && onChange(event),
```
montezume: Why use JS? Why isn't setting `readOnly` on the input enough?

https://stackoverflow.com/questions/155291/can-html-checkboxes-be-set-to-readonly

src/components/inputs/checkbox-input/checkbox-input.styles.js
```diff
@@ -33,6 +33,21 @@ const getCheckboxWrapperStyles = (props, theme) => {
       `,
     ];
   }
+  if (props.isReadOnly) {
+    return [
+      baseStyles,
+      css`
+        svg [id$='borderAndContent'] > [id$='content'] {
+          fill: ${overwrittenVars[designTokens.fontColorForInputWhenReadonly]};
```
montezume: same look as for RadioInputs when readOnly

src/components/inputs/checkbox-input/checkbox.js
tdeekens: Thanks for sharing. I often wanted to ask UX "what's the different line in communication towards the user between readonly and disabled? When use which?"

montezume:
@tdeekens yeah, I like to think about it like 

readonly: submitted in form
disabled: not submitted in form

[ref](https://stackoverflow.com/questions/1355728/values-of-disabled-inputs-will-not-be-submitted#comment23440330_1355728)

PR #1160, opened 2019-11-05
fix(date-inputs): header focus problem fix

### LINKED ISSUE #1159: DateInputs: onBlur problems
#### Summary

Currently when you switch months inside of our DateInputs, onBlur / onFocus is called.

This is probably due to using `SecondaryIconButton` and `onClick` inside of our DateInput.

We should fix this so that onBlur and onFocus are `only` called when the input is focused / blurred, not when "internal" buttons are focused / blurred.

bug

bug in mc

Thanks for the 🎣 @qmateub (#1 MC bug catcher)

montezume:
#### Summary

Fixes #1159 

This is very similar to how we deal with buttons inside of our `RichTextInput`'s Toolbar. 😇

### DIFF
--- a/src/components/inputs/date-input/date-input.js
+++ b/src/components/inputs/date-input/date-input.js
@@ -60,15 +60,28 @@ const DateInput = props => {
   );
 
   const { onBlur } = props;
-  const handleBlur = React.useCallback(() => {
-    if (onBlur)
-      onBlur({
-        target: {
-          id: props.id,
-          name: props.name,
-        },
-      });
-  }, [onBlur, props.id, props.name]);
+  const handleBlur = React.useCallback(
+    event => {
+      if (
+        event.relatedTarget &&
+        event.relatedTarget.getAttribute &&
+        event.relatedTarget.getAttribute('data-button-type') ===
+          'calendar-header'
+      ) {
+        event.preventDefault();
+        return;
+      }
+
+      if (onBlur)
+        onBlur({
+          target: {
+            id: props.id,
+            name: props.name,
+          },
+        });
+    },
+    [onBlur, props.id, props.name]
+  );
 
   const showToday = () => {
     const today = getToday();
--- a/src/components/inputs/date-input/date-input.js
+++ b/src/components/inputs/date-input/date-input.js
@@ -60,28 +60,15 @@ const DateInput = props => {
   );
 
   const { onBlur } = props;
-  const handleBlur = React.useCallback(
-    event => {
-      if (
-        event.relatedTarget &&
-        event.relatedTarget.getAttribute &&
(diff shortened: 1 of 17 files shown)


src/components/inputs/date-input/date-input.js
```diff
@@ -60,15 +60,28 @@ const DateInput = props => {
...
-        target: {
-          id: props.id,
-          name: props.name,
-        },
-      });
-  }, [onBlur, props.id, props.name]);
+  const handleBlur = React.useCallback(
+    event => {
+      if (
+        event.relatedTarget &&
```
tdeekens: #comment?

src/components/inputs/date-input/date-input.spec.js
```diff
@@ -44,6 +45,29 @@ class Story extends React.Component {
...
+        value: '2018-09-18',
+        onFocus,
+        onBlur,
+      });
+      const dateInput = getByLabelText('Date');
+      userEvent.click(dateInput);
+      expect(onFocus).toHaveBeenCalledTimes(1);
+      expect(getByText('September')).toBeInTheDocument();
+      const previousMonthButton = getByLabelText('show prev month');
+      userEvent.click(previousMonthButton);
```
tdeekens: Would you suggest using this in the MC too or apply it everywhere in UIKit?

montezume: @tdeekens yes. Now that I realize that `fireEvent.click` doesn't actually ... click, I think it's the worst and should be deleted.

src/components/inputs/date-input/date-input.visualspec.js
```diff
@@ -11,5 +14,21 @@ describe('DateInput', () => {
...
     await percySnapshot(page, 'DateInput - open');
+
+    const doc = await getDocument(page);
+    const input = await getByTestId(doc, 'date-input');
+
+    const prevMonthButton = await getByLabelText(doc, 'show prev month');
+    await prevMonthButton.click();
+
+    await wait(() => getByText(doc, 'October'));
+    // our input should still be focused even though we clicked a header button
```
montezume: here is the updated test

src/components/internals/calendar-header/calendar-header.js
```diff
@@ -17,8 +17,16 @@ const WrapperComponent = styled.div`
 
 const CalendarHeader = props => {
   const intl = useIntl();
+  // https://codepen.io/mudassir0909/pen/eIHqB
+
+  // we prevent all our defined onClicks inside of the CalendarHeader
+  // from blurring our input.
+  const onMouseDown = React.useCallback(event => {
```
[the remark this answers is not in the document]
tdeekens: No, you can build one 😄

PR #1164, opened 2019-11-06
feat(icon-button): support `as` 

### LINKED ISSUE #1162: add linkTo property on IconButton
I was working on the addresses panel on MC and I needed an icon button with a link and the original implementation was an `IconButton`  inside a `Link` and that added more code to a simple piece of functionality.
Please refer to the code snippet added to get a better view of the issue.

Adding a `linkTo` property on the `IconButton` will make things concise and more readable.

Note: if needed I can do a PR for it.

image

montezume:
#### Summary

Adds support for `as` to the `IconButton`.

Also supports prop forwarding to `IconButton` and `SecondaryButton`. Without this, it would not be possible to use `anchor` tags.

Closes #1162

### DIFF
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -81,18 +81,6 @@ export const IconButton = props => {
 };
 
 IconButton.propTypes = {
-  to(props, propName, componentName, ...rest) {
-    if (props[propName] != null) {
-      if (!props.as) {
-        return new Error(oneLine`
-          Invalid prop "${propName}" supplied to "${componentName}".
-          "${propName}" does not have any effect when "as" is not defined`);
-      }
-      return PropTypes.string(props, propName, componentName, ...rest);
-    }
-
-    return PropTypes.string(props, propName, componentName, ...rest);
-  },
   as: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
   type: (props, propName, componentName, ...rest) => {
     // the type defaults to `button`, so we don't need to handle undefined
--- a/src/components/buttons/icon-button/icon-button.js
+++ b/src/components/buttons/icon-button/icon-button.js
@@ -1,10 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { oneLine } from 'common-tags';
 import isNil from 'lodash/isNil';
 import { css } from '@emotion/core';
 import vars from '../../../../materials/custom-properties';
-import filterAriaAttributes from '../../../utils/filter-aria-attributes';
-import filterDataAttributes from '../../../utils/filter-data-attributes';
+import filterInvalidAttributes from '../../../utils/filter-invalid-attributes';
 import AccessibleButton from '../accessible-button';
 import {
   getStateStyles,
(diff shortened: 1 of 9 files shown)


src/utils/filter-invalid-attributes.js
```diff
@@ -0,0 +1,11 @@
+import isPropValid from '@emotion/is-prop-valid';
```
montezume: this is used internally by emotion

src/components/buttons/icon-button/icon-button.spec.js
```diff
@@ -45,4 +46,34 @@ describe('rendering', () => {
...
   });
+  describe('when used with `as`', () => {
+    describe('when as is a valid HTML element', () => {
+      it('should render as that HTML element', () => {
+        const { container } = render(
+          <IconButton
+            {...props}
+            as="a"
+            href="https://www.kanyetothe.com"
+            target="_BLANK"
```
montezume: if we don't support passing all valid props, then using regular "external" links would be super annoying.

src/utils/filter-invalid-attributes.js
jonnybel: It's great!

src/components/buttons/icon-button/icon-button.js
```diff
@@ -33,15 +33,17 @@ const getIconThemeColor = props => {
...
+    ...filterInvalidAttributes(props),
   };
+
   const isActive = props.isToggleButton && props.isToggled;
+
   return (
     <AccessibleButton
-      buttonAttributes={buttonAttributes}
+      as={props.as}
+      buttonAttributes={attributes}
```
jonnybel: I have a concern about forwarding all props, given we already "reimplement" some of the HTML props:
- It is possible to pass both the props `disabled` and `isDisabled` to a component, and in this case, because the [`AccessibleButton`](https://github.com/commercetools/ui-kit/blob/67dcdef9397d6a37ba1e2d268489ab1df21397fa/src/components/buttons/accessible-button/accessible-button.js) spreads the `buttonAttributes` prop _after_ the `isDisabled` prop is assigned to `disabled`, which means the native html prop will override the `isDisabled` one. Same for all other props that we manually assign before the `buttonAttributes` spread. Is this an intended behaviour?

montezume: I don't get it. You mean if the user passed in `disabled={true}` and `isDisabled={false}`?

montezume: Kind of feels like a play stupid games win stupid prizes situation.

tdeekens: Oh nice catch!

montezume: What would you prefer to happen? I guess we could add a prop-type warning if a user passes both?

jonnybel: Yeah, it shouldn't happen, but it _can_ lead to weird bugs that are hard to track.
It's more a question of: should we worry about it, and can we do something to mitigate this?

montezume: I would just ignore it. The plan is to deprecate all the `isDisabled`, etc stuff right? 

If someone writes the code
```
<IconButton disabled={true} isDisabled={false} />
```

and it passes pull request, then we deserve the bugs.

jonnybel: Yeah, it totally makes sense to deprecate the `isDisabled` and others as the next step.

It could happen accidentally after some layers of abstraction, but I also wouldn't worry about it too much, as for now we're still supposed to use the explicit props that are documented.

I'm OK with keeping it as it is, as long as we're aware and we hopefully deprecate those redundant props soon 😇.

montezume: Having two "sources of truth" is a bad idea. 😄

tdeekens: One could have like a prop-zipper or something so that always one has precedence. However, I think it's a super nice theoretical catch but given the scale we are the practical implications can be negated.

PR #1165, opened 2019-11-07
refactor(tag): simplify code by using `as` internally

montezume:
#### Summary

The `Tag` code was needlessly complicated as we rendered different internal "containers" based on whether or not it was a `Link`. This simplifies it quite a bit while preserving the same API.

### DIFF
--- a/src/components/tag/tag.js
+++ b/src/components/tag/tag.js
@@ -10,6 +10,8 @@ import AccessibleButton from '../buttons/accessible-button';
 import Text from '../typography/text';
 import { CloseBoldIcon } from '../icons';
 
+const Body = styled.div``;
+
 const getTextDetailColor = (isDisabled, theme) => {
   const overwrittenVars = {
     ...vars,
@@ -82,9 +84,7 @@ const getClickableContentWrapperStyles = ({ type, theme }) => {
       ];
 };
 
-const Body = styled.div``;
-
-export const TagNormalBody = props => (
+const TagBody = props => (
   <Body
     to={props.to}
     as={props.as}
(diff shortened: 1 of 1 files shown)


montezume:
Hm, looks like we are missing VRTs for the `link` style Tags. I'll add them in another PR.

src/components/tag/tag.js
```diff
@@ -202,97 +140,96 @@ TagNormalBody.propTypes = {
...
-                &:hover,
-                &:focus {
-                  border-color: ${overwrittenVars[
-                    designTokens.borderColorForTagWarning
-                  ]};
+        {Boolean(props.onRemove) && (
+          <AccessibleButton
+            label="Remove"
+            isDisabled={props.isDisabled}
+            onClick={props.isDisabled ? undefined : props.onRemove}
```
jonnybel: We no longer need to do this, since now the `AccessibleButton` takes care of "ignoring" onClicks when disabled, right?

src/components/tag/tag.js
```diff
@@ -120,7 +120,7 @@ const TagBody = props => (
         `,
       props.styles.body,
     ]}
-    onClick={props.isDisabled ? undefined : props.onClick}
```
jonnybel: Wrong one 😆 

I think we still need this one because it's not an `AccessibleButton`

montezume:
@tdeekens can you take a look at this? See how the bot merged master, and since then, Percy is not properly communicating

montezume:
See, when I rebased and pushed now Percy integration works.

PR #1166, opened 2019-11-07
chore(tag): add missing vrts

montezume:
#### Summary

I noticed in #1165 that we didn't have any VRTs for `Tag` with `linkTo` set.

Here I add them


src/components/tag/tag.visualroute.js
```diff
@@ -76,5 +78,113 @@ export const component = ({ themes }) => (
         <Tag type="normal">Tag</Tag>
       </Spec>
     </ThemeProvider>
+    <Spec
+      label="Normal - with linkTo"
+      propsToList={[...basePropsToList, 'linkTo']}
```
jonnybel: nice

PR #1169, opened 2019-11-08
refactor(flat-button): add as, deprecate linkbutton

montezume:
#### Summary

* Deprecates `LinkButton`
* Adds `as` prop to `FlatButton`.
* Fixes a bug in `AccessibleButton` where `handleClick` was being overwritten by the passed in `onClick`

### DIFF
--- a/src/components/buttons/accessible-button/accessible-button.js
+++ b/src/components/buttons/accessible-button/accessible-button.js
@@ -14,7 +14,7 @@ const Button = styled.button`
   padding: 0;
   margin: 0;
   white-space: nowrap;
-  cursor: pointer;
+  cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
   color: inherit;
   font: inherit;
   font-size: ${vars.fontSizeDefault};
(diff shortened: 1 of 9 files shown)


src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -14,7 +14,7 @@ const Button = styled.button`
   padding: 0;
   margin: 0;
   white-space: nowrap;
-  cursor: pointer;
+  cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
```
montezume: we need this because `a` doesn't have a &:disabled state

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -30,6 +30,10 @@ const AccessibleButton = React.forwardRef((props, ref) => {
 
   const handleClick = React.useCallback(
     event => {
+      if (props.isDisabled) {
+        event.preventDefault();
```
montezume: this way links `a` tags can be disabled

src/components/buttons/accessible-button/accessible-button.js
```diff
@@ -45,6 +49,9 @@ const AccessibleButton = React.forwardRef((props, ref) => {
 
   return (
     <Button
+      {...(props.isToggleButton ? { 'aria-pressed': props.isToggled } : {})}
```
montezume: move these to the top so they don't overwrite what we pass

montezume:
Malcolm fights the VRTs

montezume:
okay I officially don't like CSS

montezume:
Not nice

montezume:
malcolm battles the visual regression robots

pictured: me, and percy. battling

montezume:
Okay the VRTs look fine to me now.

PR #1174, opened 2019-11-12
fix: primary button active and toggle state

jonnybel:
## Summary

Fixes the `active` and `toggled` state of `PrimaryButton`, because of some leftover deprecated design tokens we forgot to update on version 10.0.0 😅

**Before**: (the button remains the same when clicked or toggled)
**After**:
image


jonnybel:
Re-ran the CI workflow because Percy got clogged or something 🚽

jonnybel:
The Percy diffs are the expected changes. I'm approving and 🚢

PR #1177, opened 2019-11-13
feat: split codebase into packages and set up monorepository

montezume:


tdeekens:
I like this approach of exporting a `combined` package. Maybe something the MC will always use or teams decide to only pick what they want. 

I think generally exporting multiple bundles and modules is nice and the approach to gradually migrate there is very nice.

Have you checked for our MC apps what porportion of the bundle size is uikit so we can evaluate the gains for us?

packages/multiline-text-input/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ui-kit/multiline-text-input",
```
tdeekens: Nit but I bet this namespace will be taken. Also nested namespaces don't work. So `@ct-fe/uikit/text-input` won't do. I guess `@ct-fe/uikit-textinput` could be something.

montezume: yeah good point.

`@commercetools-frontend/ui-kit-text-input` should work?

montezume:
@tdeekens I'll do some research on the FE side, good call.

packages/multiline-text-input/package.json
tdeekens: Yes, I think so. Or a completley new scope `#commercetools-uikit/text-input`.

montezume:
@tdeekens with `application-orders`, ui-kit or ui-kit related dependencies (react-select, immutable js, slate, etc etc) make up (all gzipped numbers) seem to make up about 1/2 of the bundle size.

tdeekens:
Oh wow. Immutable.js? Gotta do `yarn why` on that one I guess.

montezume:
@tdeekens it's from the rich-text-input...

package.json
```diff
@@ -86,7 +89,8 @@
     "slate-react": "0.22.10",
     "tiny-invariant": "1.0.6",
     "use-popper": "1.1.6",
-    "warning": "4.0.3"
+    "warning": "4.0.3",
```
emmenko: In a monorepo, the root `package.json` should not have any runtime dependencies, as it simply does not make sense to. Instead runtime deps should be defined by each package.

packages/avatar/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "scripts": {
+    "prebuild": "rimraf dist/**",
+    "build": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js -i ./index.js"
+  },
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@emotion/core": "10.0.22",
```
emmenko: I'm wondering "how easy" it would be here to miss some runtime dep.

package.json
montezume: yeah I hadn't got to that party yet 😄

packages/avatar/package.json
montezume: So far it seems like you get warnings in the build but not failures.

package.json
```diff
@@ -8,8 +8,14 @@
     "type": "git",
     "url": "https://github.com/commercetools/ui-kit.git"
   },
+  "private": true,
+  "workspaces": [
+    "packages/*",
+    "packages/buttons/*",
+    "packages/typography/*",
+    "packages/inputs/*"
```
emmenko: Should we keep the packages flat?

packages/avatar/package.json
emmenko: I mean within the monorepo it shouldn't matter, as packages get hoisted. I'm talking about consumers of the package. If some runtime dep is not defined, you will probably get an error then.

I guess this would make more sense to do once the source code is moved into its own package.

rollup.config.js
```diff
@@ -26,7 +32,7 @@ const configureRollupPlugins = (options = {}) =>
     // See also https://medium.com/@kelin2025/so-you-wanna-use-es6-modules-714f48b3a953
     // Transpile sources using our custom babel preset.
     babel({
-      exclude: ['node_modules/**'],
+      exclude: path.join(__dirname, '/node_modules/**'),
```
emmenko: Much better now 😉

rollup.config.js
```diff
@@ -67,37 +73,41 @@ const defaultExternal = deps.concat(peerDeps);
...
-    plugins: configureRollupPlugins(),
-  },
-  {
-    input: 'src/index.js',
-    external: defaultExternal,
-    output: {
-      file: pkg.module,
-      format: 'esm',
+const createConfig = cliArgs => {
+  return [
```
emmenko: Nit: unnecessary `return` block

src/components/avatar/avatar.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
+/** @jsx jsx */
```
emmenko: What's this for? 🤔

src/components/buttons/flat-button/flat-button.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { omit } from 'lodash';
-import vars from '../../../../materials/custom-properties';
+import Text from '@commercetools-uikit/text';
```
emmenko: Wondering if it would make sense to have the package named `@commercetools-uikit/typography`, which exports individual components like `import { Header, Body, ... }`

src/components/icons/create-styled-icon.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import invariant from 'tiny-invariant';
 import styled from '@emotion/styled';
-import vars from '../../../materials/custom-properties';
+import vars from '@commercetools-uikit/custom-properties';
```
emmenko: Should we name this package something like `@commercetools-uikit/design-system` or `@commercetools-uikit/design-tokens`? 🤔

src/components/inputs/multiline-text-input/multiline-text-input.js
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import requiredIf from 'react-required-if';
 import { useIntl } from 'react-intl';
 import { css } from '@emotion/core';
+import { AngleUpIcon, AngleDownIcon } from '@commercetools-uikit/icons';
```
emmenko: As far as I remember, one of the issues with tree shaking was related to the svgs...should we maybe publish each icon individually? If we bundle them all together we might end up with the same problem 🤔

src/components/table/base-table/base-table.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { CellMeasurer, CellMeasurerCache, MultiGrid } from 'react-virtualized';
 import sortBy from 'lodash/sortBy';
-import getScrollbarSize from 'dom-helpers/scrollbarSize';
+import getScrollbarSize from 'dom-helpers/util/scrollbarSize';
```
emmenko: Was this a bug fix? 🤔

visual-testing-app/webpack.config.js
```diff
@@ -54,6 +54,40 @@ module.exports = {
 
   module: {
     rules: [
+      // For svg icons, we want to get them transformed into React components
+      // when we import them.
+      {
+        test: /\.react\.svg$/,
```
emmenko: Can you clarify why we need this now?

src/components/inputs/multiline-text-input/multiline-text-input.js
montezume: yeah, that's what I wanted to mess around with now. 👍 Going to investigate in the VRT app how treeshaking works with the icons.

src/components/table/base-table/base-table.js
montezume: I'm confused about this. There must be some weird resolution happening, because when I open `node_modules` on master, the file is at `scrollbarSide`, yet when I open node modules here, it's at `utils/scrollbarSize` and you get a lint warning 🤔

src/components/inputs/multiline-text-input/multiline-text-input.js
montezume: Ah yeah, so right now if you use `MultilineTextInput`, you end up with all of the icons. I'll look into alternatives now.

visual-testing-app/webpack.config.js
montezume: we don't, can remove it

src/components/avatar/avatar.js
montezume: because the only JSX rendered by this component using the css prop, the built version drops all usages of react.createELement and replaces it with JSX

so we get warnings about unused 'react' import

montezume: Here's an example of what `Text` looks like

```
import 'react';
import { FormattedMessage } from 'react-intl';
import requiredIf from 'react-required-if';
import { isNil } from 'lodash-es';
import { oneLine } from 'common-tags';
import { css, jsx } from '@emotion/core';
import vars from '@commercetools-uikit/custom-properties';
```

... more stuff

```jsx
var Text = function Text(_ref) {
  var intlMessage = _ref.intlMessage,
      children = _ref.children;
  return intlMessage ? jsx(FormattedMessage, intlMessage) : children;
};
```

montezume:
still a work in progress

montezume:
obviously need a new way to test the bundlesize 😉

package.json
```diff
@@ -207,17 +188,10 @@
     "stylelint-order": "3.1.1",
     "webpack": "4.41.2",
     "webpack-cli": "3.3.10",
+    "wsrun": "^5.1.0",
```
emmenko: What's this?

packages/calendar-utils/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@commercetools-uikit/calendar-utils",
```
emmenko: Would it make sense to group such packages together? For example `calendar-utils` and `calendar-time-utils`

packages/components/buttons/preset/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@commercetools-uikit/buttons",
```
emmenko: Maybe we should be explicit in marking those packages as "presets". For example `@commercetools-uikit/preset-buttons`.

packages/components/buttons/primary-button/index.js
```diff
@@ -0,0 +1 @@
+export { default } from '../../../../src/components/buttons/primary-button';
```
emmenko: Can we try having a flat structure for the packages? So that every package is under `packages/*`

emmenko: We can also have other top level workspace folders, like:

```
packages/
presets/
utils/
```

and so on

packages/components/spacings/stack/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@commercetools-uikit/spacings-stack",
```
emmenko: Hmm is there a particular reason for having the spacings components split?

visual-testing-app/webpack.config.js
emmenko: Is this still valid?

package.json
montezume: oh we can drop it, sorry.

packages/calendar-utils/package.json
montezume: my idea was to split these two because they rely on two different moment dependencies.

packages/components/buttons/primary-button/index.js
montezume: I left it like this so that it matches the current src structure, which made it a lot easier to ensure I didn't miss any packages...

packages/components/spacings/stack/package.json
montezume: treeshaking 😄

src/components/avatar/avatar.js
montezume: can remove this

visual-testing-app/webpack.config.js
montezume: ah yeah, should not noeed that sorry

packages/components/spacings/stack/package.json
montezume: otherwise you end up with too many static properties to treeshake right?

emmenko: Yeah ok. Should we do the same for the typography `Text` components? 🤔

packages/components/buttons/primary-button/index.js
emmenko: Ok, I guess we can reorganize things as a next step

packages/calendar-utils/package.json
emmenko: I see, ok

packages/components/spacings/stack/package.json
montezume: hm, yeah we could.

rollup.config.js
```diff
@@ -65,22 +72,38 @@ const deps = Object.keys(pkg.dependencies || {});
...
+    onwarn(warning, warn) {
+      // skip certain warnings
+      if (
+        warning.code === 'UNUSED_EXTERNAL_IMPORT' &&
+        (warning.source === 'react' || warning.source === '@emotion/core')
+      )
+        return;
+
+      // always throw when unresolved import
+      if (warning.code === 'UNRESOLVED_IMPORT') {
```
montezume: because Lerna hides warnings, it's a good idea to fail early if a package is missing a dep.

montezume:
cc @jonnybel @emmenko 
There's still a few problems that need to be solved, but they should be pretty straightforward

1) For some reason, the version 3 of `dom-helpers` is being resolved by jest and eslint in the package `base-table`. This version is coming from react-intl. We need to be using the latest version to match MC-FE. 

2) We need to make sure that the current "all-in-one" preset of ui-kit `@commercetools-frontend/ui-kit` has the same `files` as the current version. Because some apps are relying on the i18n or the shared `materials` folder.

Other than that it should work. I tried it to the MC and got it working by resolving 2) and 1) manually 😄

emmenko:
FYI: I'm taking over this PR

bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
   "files": [
     {
       "path": "visual-testing-app/dist/ui-kit.js",
-      "maxSize": "130 kB"
+      "maxSize": "150 kB"
```
tdeekens: Wasn't the point of this to... just kidding 😄.

emmenko:
This is ready to go now from my side. We can start moving code into their packages as follow ups.

emmenko:
Great, I’ll merge this later today then.

PR #1178, opened 2019-11-13
chore: change lodash imports

montezume:
Stop bundling lodash.

This way use 'lodash' for our CJS build, and 'lodash-es' for the es build, and our rename plugin actually works.

It should lower the bundle size as well.

example from dist/cjs 

```js

var lodash = require('lodash');
```

and dist/es

```js

import { omit, isNil, uniq, has, flatMap, pick, memoize as memoize$2 } from 'lodash-es';
```


montezume:
Bundle size down 4kb

PR #1180, opened 2019-11-14
refactor(icons): treeshaking

montezume:
#### Summary

I spent most of the morning investigating our icons and whether or not they could be treeshaken.

#### Current situation.

```jsx
import { GearIcon } from 'ui-kit';

const App = () => (
 <GearIcon />
)
```

results in a total ui-kit bundle size (using the VRT app) of 470kb with ~131 bundled icons

#### After this PR

```jsx
import { GearIcon } from 'ui-kit';

const App = () => (
 <GearIcon />
)
```

results in a total ui-kit bundle size (using the VRT app) of 361kb. We end up with around 16 icons (mostly from the rich-text-input and the date inputs).


scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+const util = require('util');
 const path = require('path');
 const glob = require('glob');
+const del = require('del');
+const svgr = require('@svgr/core').default;
+
+// Convert fs.readFile / writeFile into Promise version of same
+const readFile = util.promisify(fs.readFile);
+const writeFile = util.promisify(fs.writeFile);
+const mkdir = util.promisify(fs.mkdir);
```
tdeekens: Nit/FYI: you can use fsPromises here I think [ref](https://nodejs.org/api/fs.html#fs_fspromises_access_path_mode).

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+          'info',
+          'primary',
+          'primary40',
+          'warning',
+          'error',
+        ]),
+        size: PropTypes.oneOf(['small', 'medium', 'big', 'scale']),
+      };
+
+      if (process.env.NODE_ENV !== "production") {
```
tdeekens: 😕 Will they not be stripped anyway?

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+
+       \${getSizeStyle(props.size)}
+      \`} />;
+
+
+      Component.displayName = '${displayName}';
+
+      const componentPropTypes = {
+        color: PropTypes.oneOf([
+          'solid',
```
tdeekens: Nit: wondering if we could import all cases from somewhere else so they adapt?

scripts/generate-icon-exports.js
montezume: They will be but it will break treeshaking if you don't add this...
[ref](https://github.com/alex996/react-css-spinners/issues/1)

montezume:
> So essentially we only "default bunde" what UIKit uses itself while everything else is added on if consumed?

@tdeekens I'm not sure what you mean here? Do you mean because we still end up with "some" icons? That's more to do with other components breaking treeshaking 😄. The idea will be to combine this + the monorepo approach to get "true" treeshaking.

montezume:
I think I should just add the generated components to the git history because using gitignore and the needing to run it on build / test / lint and storybook build is annoying

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+  mkdir(rawExportsPath);
+  mkdir(iconComponentsPath);
+
+  /* eslint-disable no-restricted-syntax, no-await-in-loop */
+  for (const fileName of files) {
+    const componentName = upperFirst(
+      camelCase(path.basename(fileName, iconFileExt))
+    );
+
+    const svgCode = await readFile(fileName, 'utf8');
```
montezume: this reads the raw svg file

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+    const jsCode = await svgr(
```
montezume: this transforms the svg code into a react component

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+          '@svgr/plugin-svgo',
+          '@svgr/plugin-jsx',
+          '@svgr/plugin-prettier',
+        ],
+      },
+      { componentName }
+    );
+
+    await writeFile(
+      path.join(rawExportsPath, `/${componentName}.js`),
```
montezume: this writes that react component onto disk

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+      prettier.format(jsCode, prettierConfig)
+    );
+
+    const displayName = `${componentName}Icon`;
+
+    const importStatement = stripIndents`
+      import React from 'react';
+      import PropTypes from 'prop-types';
+      import { css } from '@emotion/core';
```
montezume: we use { css } instead of styled because styled seems to break treeshaking

scripts/generate-icon-exports.js
tdeekens: Oh wow. Please add a comment ontop the file as to that this unsafe wrap and using styled would break things.

montezume: yeah will do 👍

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
 const fs = require('fs');
+const util = require('util');
 const path = require('path');
 const glob = require('glob');
+const del = require('del');
+const svgr = require('@svgr/core').default;
+
+// Convert fs.readFile / writeFile into Promise version of same
```
emmenko: Q: is there a reason to not use the `sync` version of the functions?

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
 const fs = require('fs');
+const util = require('util');
 const path = require('path');
 const glob = require('glob');
+const del = require('del');
```
emmenko: I would avoid using a new dep and use `shelljs` (which we already use).

scripts/generate-icon-exports.js
montezume: no not really, I just thought it was better practice not to. I guess for a script it doesn't matter?

.gitignore
```diff
@@ -12,6 +12,9 @@ pids
 *.seed
 *.pid.lock
 
+src/components/icons/raw-components
+src/components/icons/components
```
emmenko: Sorry to be picky, can you move those at the end of the file? Just so that we keep the default grouping

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
 
-glob(importPath, (err, files) => {
+const prettierConfig = rcfile('prettier');
+
+glob(importPath, async (err, files) => {
+  await del([rawExportsPath, iconComponentsPath]);
+  mkdir(rawExportsPath);
+  mkdir(iconComponentsPath);
+
+  /* eslint-disable no-restricted-syntax, no-await-in-loop */
```
emmenko: Why not using `Promise.all` and so avoid silencing `no-await-in-loop`?

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,139 @@
...
+);
+
+const indexPath = path.join(__dirname, '../src/components/icons/index.js');
+
 const iconFileExt = '.react.svg';
 
-glob(importPath, (err, files) => {
+const prettierConfig = rcfile('prettier');
+
+glob(importPath, async (err, files) => {
```
emmenko: Please describe what we're trying to do/achieve here

scripts/generate-icon-exports.js
emmenko: Yeah I would also prefer if we can simplify the setup here.

scripts/generate-icon-exports.js
emmenko: Yeah a script usually run synchronously. At least if feels a bit weird to force synchronous function into async functions.

montezume:
I need to fix the broken icons that Percy revealed

montezume:
Thanks for the comments @emmenko I'll make the fixes once I fix the problem with broken icons 😅

package.json
```diff
@@ -135,6 +141,7 @@
     "common-tags": "1.8.0",
     "conventional-changelog-cli": "2.0.27",
     "cross-env": "6.0.3",
+    "del": "5.1.0",
```
emmenko: This can be removed now

scripts/generate-icon-exports.js
```diff
@@ -1,49 +1,115 @@
...
+
+    const displayName = `${componentName}Icon`;
+
+    const importStatement = stripIndents`
+      import React from 'react';
+      import { css } from '@emotion/core';
+      import { getColor, getSizeStyle, iconPropTypes } from '../create-styled-icon';
+      import ${componentName} from '../raw-components/${fileNameWithoutExtension}';
+
+      const Component = props => <${componentName} {...props} css={theme => css\`
```
emmenko: We can extract these styles as well into a `get...Style` function

montezume:
Percy failing because I accidentally deleted an icon

montezume:
As a follow up I would like to upgrade this to transform the "internal" icons as well (rich-text-input icons), then we can remove the rollup svgr plugin.

montezume:
Thanks all for the reviews. I'll take this one step further tomorrow and see how far I can go 😄

PR #1185, opened 2019-11-18
fix(toggle-input): allow to pass value prop

emmenko:
TL;DR: in Formik v2 they changed a bit how checkbox and radio inputs deal with values

> Formik 2 "fixes" React checkboxes and multi-selects with built-in array binding and boolean behavior

https://github.com/jaredpalmer/formik/releases/tag/v2.0.1

As we are upgrading our apps to formik v2, we started bumping into weird test failures and we narrowed the issue down to checkbox input types. It seems that if no `value` is provided, the default value is `on`, [as described in the MDN <input type="checkbox">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value).

You can also see in the following codesandbox how the behavior differs:

image

https://codesandbox.io/s/formik-checkbox-value-i40kv

## The fix

It appears that by passing both `value` and `checked` to the input, Formik handles the boolean value correctly.

### DIFF
--- a/src/components/inputs/toggle-input/toggle-input.js
+++ b/src/components/inputs/toggle-input/toggle-input.js
@@ -119,6 +119,7 @@ const ToggleInput = props => {
         onChange={props.onChange}
         disabled={props.isDisabled}
         checked={props.isChecked}
+        value={props.value}
         type="checkbox"
         size={props.size}
         {...filterDataAttributes(props)}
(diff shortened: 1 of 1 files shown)


montezume:
I guess Percy didn't run because the PR wasn't open when the VRTs ran

PR #1186, opened 2019-11-19
chore: rollup config fix

montezume:
#### Summary

When we bundle UI-Kit, we only want to bundle **our** code. We don't want to include vendor code.

For that reason, I have removed `rollup-plugin-node-resolve`.


bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
   "files": [
     {
       "path": "./dist/ui-kit.esm.js",
-      "maxSize": "167 kB"
+      "maxSize": "129 kB"
```
montezume: 😄

package.json
```diff
@@ -68,7 +68,10 @@
   },
   "dependencies": {
     "@emotion/core": "10.0.22",
+    "@emotion/css": "10.0.22",
```
emmenko: Isn't this part of `@emotion/core`? 🤔

bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
   "files": [
     {
       "path": "./dist/ui-kit.esm.js",
-      "maxSize": "167 kB"
+      "maxSize": "115 kB"
```
montezume: 😄

package.json
```diff
@@ -68,7 +68,10 @@
...
+    "@emotion/is-prop-valid": "0.8.5",
     "@emotion/styled": "10.0.23",
+    "@emotion/styled-base": "10.0.24",
```
emmenko: Can you explain why do we need all those deps?

package.json
montezume: otherwise we will get unresolved dependency errors as rollup compiles the code to this

```
import _styled from '@emotion/styled-base';
...
import _css from '@emotion/css';
```

otherwise we can add them to the "external" array

package.json
montezume: gonna move them to the ignored externals instead

src/components/table/base-table/base-table.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { CellMeasurer, CellMeasurerCache, MultiGrid } from 'react-virtualized';
-import sortBy from 'lodash/sortBy';
```
amine-benselim: I thought this way ensures smaller bundle size 🤔

montezume: Well we use `lodash-es` for our ESM bundle which ensures smaller bundle size. 

Consumers who use the cjs bundle will have a larger bundle-size, but we only use that in test environment...

montezume: Plus doing this doesn't work with the lodash-es switcher

montezume: that's not a bad idea actually.

montezume: Okay I tried both approaches, they result in same bundle sizes for consumer and for UI-Kit.

amine-benselim: thank you for going the extra mile 🙏

montezume: I want to make another change first though so we can have a better idea of what's happening.

montezume:
@amine-benselim okay I messed around a bunch with the lodash imports and I think I came up with the simplest approach

montezume:
Now that I removed the different approach for lodash on the ESM / CJS bundles, it builds twice as fast.

src/components/buttons/icon-button/icon-button.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { oneLine } from 'common-tags';
-import { isNil } from 'lodash';
```
montezume: this way we don't need to bother with using lodash-es

PR #1189, opened 2019-11-20
fix(buttons): make onClick not required if as is passed

montezume:


src/components/buttons/flat-button/flat-button.js
```diff
@@ -116,7 +117,9 @@ FlatButton.propTypes = {
   tone: PropTypes.oneOf(['primary', 'secondary', 'inverted']),
   type: PropTypes.oneOf(['submit', 'reset', 'button']),
   label: PropTypes.string.isRequired,
-  onClick: PropTypes.func.isRequired,
+  onClick: requiredIf(PropTypes.func, props => {
```
emmenko: Nit: one line? `props => !props.as`

PR #1192, opened 2019-11-21
feat(select): add isReadOnly prop to select inputs and fields

jonnybel:
### Summary

This PR adds the `isReadOnly` prop to our `SelectInput`.
It also adds the same prop to `AsyncSelectInput`, `CreatableSelectInput` and `AsyncCreatableSelectInput`.
And then also to their 4 corresponding `*SelectField` components.

There's a lot changes because it's the same change in 8 different components, but the focus of the review can be on only one of them and on the `create-select-styles`, which is shared across all of them.

#### Description

`react-select` doesn't support `isReadOnly` because it doesn't have any functional difference from `isDisabled`, it's only a matter of styling it differently. So basically I'm disabling it internally and giving it the same `read-only` visuals as we have for the other inputs.

But the main point of adding the prop is because almost all of our inputs support `readOnly` and users of the library may erroneously expect it for this component as well, as it happened [here](https://github.com/commercetools/merchant-center-frontend/pull/7977). So it's a matter of consistency.


src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -59,7 +61,7 @@ const AsyncCreatableSelectInput = props => {
           inputId={props.id}
           inputValue={props.inputValue}
           isClearable={props.isClearable}
-          isDisabled={props.isDisabled}
+          isDisabled={props.isDisabled || props.isReadOnly}
```
tdeekens: What's the benefit of the prop if it maps to this?

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
```diff
@@ -47,6 +47,8 @@ const AsyncCreatableSelectInput = props => {
               hasError: props.hasError,
               showOptionGroupDivider: props.showOptionGroupDivider,
               menuPortalZIndex: props.menuPortalZIndex,
+              isDisabled: props.isDisabled,
+              isReadOnly: props.isReadOnly,
```
jonnybel: We're now also passing the `isDisabled` prop because we will use it for the styles instead of the state, to differentiate between `isDisabled` and `isReadOnly`.

src/components/inputs/async-creatable-select-input/async-creatable-select-input.js
jonnybel: `react-select` doesn't support `isReadOnly` because it doesn't have any functional difference from `isDisabled`, it's only a matter of styling it differently. So basically I'm disabling it internally and giving it the same `read-only` visuals as we have for the other inputs.

But the main point of adding the prop is because almost all of our inputs support `readOnly` and users of the library can erroneously use it for this component as well, as it happened [here](https://github.com/commercetools/merchant-center-frontend/pull/7977). So it's a matter of consistency.

src/components/internals/create-select-styles.js
```diff
@@ -20,12 +20,14 @@ const controlStyles = (props, theme) => (base, state) => {
   return {
     ...base,
     fontSize: overwrittenVars[designTokens.fontSizeForInput],
-    backgroundColor: state.isDisabled
+    backgroundColor: props.isDisabled
```
jonnybel: We're now using `props.isDisabled` instead of the internal `state.isDisabled`, because `isReadOnly` is also making the SelectInput become internally disabled (see [this line](https://github.com/commercetools/ui-kit/pull/1192/files#diff-8dd7f63a278e2dc6ae659f5036bafbb3R64)), but we can check the props to differentiate between the two modes for the styles.

PR #1196, opened 2019-11-25
fix(tooltip): do not use native title when tooltip is off

montezume:
#### Summary

When `Tooltip` is set to off, we should not add `aria-describedby` or `title`.

### DIFF
--- a/src/components/tooltip/tooltip.spec.js
+++ b/src/components/tooltip/tooltip.spec.js
@@ -448,7 +448,7 @@ describe('when off is true', () => {
     // should not be visible
     expect(queryByText('What kind of bear is best?')).not.toBeInTheDocument();
     // should not have title
-    expect(button).not.toHaveProperty('title', 'What kind of bear is best?');
+    expect(button).not.toHaveAttribute('title');
 
     fireEvent.mouseLeave(button);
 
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.spec.js
```diff
@@ -430,8 +430,8 @@ describe('when off is true', () => {
 
     // should not be visible
     expect(queryByText('What kind of bear is best?')).not.toBeInTheDocument();
-    // should not remove title
-    expect(button).toHaveProperty('title', 'What kind of bear is best?');
+    // should not have title
+    expect(button).not.toHaveProperty('title', 'What kind of bear is best?');
```
emmenko:
```suggestion
    expect(button).not.toHaveAttribute('title');
```

I think it's a bit more safe like this, as we just want to check that the `title` attribute is not set and not some specific title.

src/components/tooltip/tooltip.spec.js
```diff
@@ -447,8 +447,8 @@ describe('when off is true', () => {
...
```
emmenko:
```suggestion
    expect(button).not.toHaveAttribute('title');
```

this one as well 😅

PR #1205, opened 2019-12-04
docs(changelog): for release v10.13.0 [skip ci]

emmenko:


emmenko:
> the commits in the changelog now have a different format?

Yes, because of the monorepo. It's the same as in app-kit now.

PR #1207, opened 2019-12-05
chore: move to zeit now

emmenko:


emmenko:
I already wrote in Slack `#ui-kit` channel 😉

packages/components/table/index.js
```diff
@@ -0,0 +1,4 @@
+// This file exists to allow tools to use our non-compiled code to e.g. run tests, build storybook.
+// If this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) the tool will fail if the bundle is not yet built.
```
jonnybel: Ahhh, so that's why this was happening before building 😄

PR #1211, opened 2019-12-09
fix(flat-button): icon size and position for multi-line anchors

jonnybel:
[UKIT-95](https://jira.commercetools.com/browse/UKIT-95)
#### Summary

This PR adjusts the `FlatButton` to allow breaking the text into multiple lines when the element is not a `button`, and changes the position and size of the icon, not allowing it to shrink and keeping it on the same line as the text.

Before:
image-2019-12-04-15-33-05-618

After:
image

### DIFF
--- a/src/components/buttons/flat-button/flat-button.js
+++ b/src/components/buttons/flat-button/flat-button.js
@@ -77,11 +77,7 @@ export const FlatButton = props => {
         };
 
         return css`
-          align-items: center;
-          font-size: 1rem;
-          border: none;
-          background: none;
-          padding: 0;
+          display: inline-block;
           min-height: initial;
           ${props.as && props.as !== 'button' ? 'white-space: normal' : ''};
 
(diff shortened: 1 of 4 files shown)


jonnybel:
Yes, I'll add a test for this.

src/components/buttons/flat-button/flat-button.js
```diff
@@ -6,25 +6,37 @@ import requiredIf from 'react-required-if';
...
 const propsToOmit = ['type'];
 
-const getIconElement = props => {
+const ButtonIcon = props => {
   if (!props.icon) return null;
 
   let iconColor = 'solid';
   if (props.isDisabled) iconColor = 'neutral60';
   else if (props.tone === 'primary') iconColor = 'primary';
-  else if (props.tone === 'secondary' && props.isMouseOver)
```
jonnybel: This was a leftover because there is no `isMouseOver` prop here.

src/components/buttons/flat-button/flat-button.js
```diff
@@ -84,29 +93,31 @@ export const FlatButton = props => {
...
-                : getTextColor(props.tone, true, overwrittenVars)};
-            }
-
-            svg * {
-              fill: ${props.isDisabled
-                ? overwrittenVars.colorNeutral
-                : getTextColor(props.tone, true, overwrittenVars)};
-            }
+          span + span {
+            margin-left: ${vars.spacingXs};
```
jonnybel: Almost the same as using an `Inline` spacings component, but we can't use flexbox here.

PR #1213, opened 2019-12-10
feat: add min and max value props to date-input and field

jonnybel:
#### Summary
This PR adds a `minValue` and `maxValue` props to the `DateInput` and `DateField` components.

The point is providing the possibility to create a Date Range selector by combining two `DateInput` components, as an alternative to `DateRangeInput`, which is why the same props weren't added to the latter. See [UKIT-67](https://jira.commercetools.com/browse/UKIT-67) for more details.
In the near future, they might be added to all Date-related inputs but this is yet to be discussed, as it is likely that the `DateRangeInput` will be re-done to use two `DateInputs` internally.


src/utils/calendar.js
```diff
@@ -79,3 +79,29 @@ export const parseInputToDate = (text, locale) => {
...
+  if (min && !max) {
+    return itemDate.isSameOrAfter(min);
+  }
+  if (!min && max) {
+    return itemDate.isSameOrBefore(max);
+  }
+  return true;
+};
+
+export const getExampleDateStrings = () => ({
```
jonnybel: This is only used for the Storybook examples and the values are opinionated, but it's nice to keep the examples updated and easy to understand.

src/components/inputs/date-input/date-input.js
```diff
@@ -215,7 +216,11 @@ const DateInput = props => {
                         key={item}
                         isToday={isSameDay(today, item)}
                         {...getItemProps({
-                          disabled: props.isDisabled,
+                          disabled: !isDateInRange(
```
islam3zzat: Will this mean that `props.isDisabled` will not affect the disable state?

tdeekens: Nit: `getIsDateInRange` is nicer when it comes then to distinguish between a boolean and the respective getter.

jonnybel: It was wrong before. When the component `isDisabled`, the calendar menu is never opened/rendered, so there was no point disabling anything inside of it.

src/utils/calendar.js
```diff
@@ -79,3 +79,29 @@ export const parseInputToDate = (text, locale) => {
...
+  currentDate: moment().format('YYYY-MM-DD'),
+  preselectedDate: moment()
+    .add(3, 'days')
+    .format('YYYY-MM-DD'),
+  minDate: moment().format('YYYY-MM-DD'),
+  maxDate: moment()
+    .add(1, 'year')
+    .format('YYYY-MM-DD'),
+});
```
tdeekens: Please add tests for this? Would make sense, right?

jonnybel: Yeah. There are no existing tests for these calendar utils, but I'll start by adding for this one in this PR.

jonnybel: Dun. The values returned from `getExampleDateStrings` are opinionated, so I didn't assert all their values, but only if they "made sense" for the examples.

src/utils/calendar.spec.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+  it('preselectedDate should be between minDate and maxDate', () => {
+    expect(
+      moment(exampleDates.preselectedDate).isBetween(
+        exampleDates.minDate,
+        exampleDates.maxDate
+      )
+    ).toBeTruthy();
+  });
+});
```
tdeekens: Wonderful!

PR #1214, opened 2019-12-10
fix(text-detail): proptypes warning when providing an empty string as…

### LINKED ISSUE #1212: PropTypes warning when providing an empty string to `TextDetail`
**Describe the bug**
When providing an empty string `''` to `TextDetail` component, it will result in a PropTypes warning.

```
Warning: Failed prop type: The prop `intlMessage` is marked as required in `TextDetail`, but its value is `undefined`.
```

This is because of the check done here.

https://github.com/commercetools/ui-kit/blob/860410698e38c0934fb28ba47138a25aaf15d073/src/components/typography/text/text.js#L266

**To Reproduce**

https://codesandbox.io/s/exciting-lehmann-sgpgt?expanddevtools=1&fontsize=14&hidenavigation=1&theme=dark

**Expected behavior**
There should not be a warning when providing an empty string.

ahmehri:
… children to TextDetail component

#### Summary

A PropTypes warning will be thrown when providing an empty string as the children of `Text.Detail` component.

#### Description

The bug is caused by the empty string being coerced to `false`, this PR fixes it by counting the children instead.

Fixes #1212

### DIFF
--- a/src/components/typography/text/text.js
+++ b/src/components/typography/text/text.js
@@ -263,7 +263,10 @@ Detail.propTypes = {
   ]),
   title: nonEmptyString,
   truncate: PropTypes.bool,
-  intlMessage: requiredIf(intlMessageShape, props => !props.children),
+  intlMessage: requiredIf(
+    intlMessageShape,
+    props => !React.Children.count(props.children)
+  ),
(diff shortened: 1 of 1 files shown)


ahmehri:
@jonnybel Yes, I will do the changes for other components as well 👍 .

Issue #1210, opened 2019-12-09
Cannot click on ClearIndicator of SelectInput in RTL tests


ahmehri:
**Describe the bug**
In RTL tests, clicking on the `ClearIndicator` of the `SelectInput` component has no effect.

**To Reproduce**

1. Write an RTL test that renders a `SelectInput` component with the default `ClearIndicator` component, i.e the X icon
2. get the `ClearIndicator` element
3. fire a click event on it
4. nothing happens

**Expected behavior**
The component should response to the fired event.

jonnybel:
Thanks. I think we could make it easier to get the `ClearIndicator` by giving a title attribute to the SVG, and also by making its container actually a button instead of a div.

Meanwhile, see if you can make it work by following the example on this UI-kit test: https://github.com/commercetools/ui-kit/blob/master/src/components/inputs/select-input/select-input.spec.js#L320

ahmehri:
I followed the example and was able to retrieve the `ClearIndicator` element. Clicking on it, however, didn't trigger the related callback.

jonnybel:
OK, I've found the problem: `react-select` is passing `onMouseDown` and `onTouchEnd` handlers to the element, but not "onClick".
As such, I've tested and it does respond to `fireEvent.mouseDown(ClearIndicatorElement)`. Let me know if it works for you 🙂

Meanwhile, I'll open a PR to add a `label` and `title` to this button to make it easier to select for the tests, and I'll also update the `SelectInput` tests to illustrate the example.

ahmehri:
Yes, `fireEvent.mouseDown` works perfectly. Thanks for finding this out 🙂 .

PR #1215, opened 2019-12-11
fix(select-input): make the ClearIndicator a button

jonnybel:
Related to #1210 

#### Summary

This PR changes the `ClearIndicator`, used by `SelectInput`, to be a button and have a title and label, becoming screen-readable and easier to select (e.g. for tests).

Some tests for this element were fixed, because they were giving false positives.

### DIFF
--- a/src/components/inputs/select-input/select-input.spec.js
+++ b/src/components/inputs/select-input/select-input.spec.js
@@ -215,20 +215,13 @@ describe('in multi mode', () => {
         expect(onChange).not.toHaveBeenCalled();
         expect(queryByText('Mango')).toBeInTheDocument();
       });
-      it('should call not call onChange when value is cleared by clicking delete button ', () => {
-        const onChange = jest.fn();
-        const { getByLabelText, getByText, queryByText } = renderInput({
-          onChange,
+      it('should not render the clear button', () => {
+        const { queryByTitle } = renderInput({
           isMulti: true,
           value: ['mango'],
           isDisabled: true,
         });
-        const input = getByLabelText('Fruit');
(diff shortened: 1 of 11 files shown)


src/components/internals/clear-indicator/clear-indicator.js
```diff
@@ -1,28 +1,39 @@
...
+        cursor: pointer;
+        background: none;
+        box-sizing: border-box;
+        text-decoration: none;
+      `}
       style={getStyles('clearIndicator', props)}
+      title={intl.formatMessage(messages.clearButtonLabel)}
+      aria-label={intl.formatMessage(messages.clearButtonLabel)}
     >
-      <CloseIcon color={props.isDisabled && 'neutral60'} size="medium" />
```
jonnybel: Removed this because it is never disabled. When the Input is disabled/readonly, this component isn't rendered.

src/components/inputs/select-input/select-input.spec.js
```diff
@@ -215,20 +215,13 @@ describe('in multi mode', () => {
...
+        const { queryByTitle } = renderInput({
           isMulti: true,
           value: ['mango'],
           isDisabled: true,
         });
-        const input = getByLabelText('Fruit');
-        fireEvent.focus(input);
-        const deleteButton = getByText('Mango').nextSibling;
-        deleteButton.click();
-        expect(onChange).not.toHaveBeenCalled();
```
jonnybel: These were giving a false positive: onChange wasn't called because the ClearIndicator isn't rendered when the component is disabled, so clicking this "nextSibling" wouldn't do anything anyway.

PR #1233, opened 2020-01-06
[UKIT-96] feat(table): add new component DataTable

jonnybel:
#### Summary

The new `DataTable` component.
[UKIT-96](https://jira.commercetools.com/browse/UKIT-96)

#### Description

The new table uses CSS-grid, doesn't use virtualization, and tries to keep a similar API as the existing table component.

Check out the latest version and be sure to play around with all the available knobs and column options:
https://ui-kit-git-ja-new-table.commercetools.now.sh/?path=/story/components-table-new--datatable

#### Feature Checklist
- [x] Basic Table functionality with the current API, with `items` and `columns` definition arrays, and `renderItem` prop replacing `itemRenderer`.
- [x] Table uses all available space
- [x] Adjustable max and height and width
- [x] Cell alignment: configurable at table level and column level -- setting at column level takes precedence.
- [x] Sticky Header
- [x] "Condensed" mode: compact mode with less spacings for the cells, maximizing the usage of available space
- [x] Column width is fully adjustable: accepts auto or fixed width as well as a minmax pair (using the [css-grid template column syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns#Syntax)).
- [x] Uses semantic `<table>` markup under the hood.
- [x] Truncated text cell option
- [x] Row hovering 
- [x] Row click
- [x] Cell click (and option to ignore rowClick)
- [x] Row Selection (with `useRowSelection` hook. This can also be done by the consumer, like it is on the old table)
- [x] Row Sorting (also with a hook)

Follow-up: 
- Add VRTs
- Test in MC to discover potential bugs and enhancements "in the field"

<hr/> 

A heads up about this commit: https://github.com/commercetools/ui-kit/pull/1233/commits/00996783d05258012e0b03fe6549577909fe3b8c

I've changed the API a little bit:
* `items` prop renamed to `rows`, to create less confusion regarding the difference with actual _items_ (cells) of a table. Also, I think it pairs better with the `columns` prop.
* Changed the `renderItem` prop of the table back to `itemRenderer`, and this becomes the default renderer for each item. The default is set to simply render whatever is the value of an item property (usually it's a string). See: https://github.com/commercetools/ui-kit/pull/1233/files#diff-aa4a4319c62f56d7a3a775b74edbcb06R94
* Added a `renderItem` prop to the column definitions. This is the renderer to be used for custom cases per-column, taking precedence over the default `itemRenderer`. I find it's easier to have this on the column definition (among with others like `onClick`), instead of having to implement a `switch (case)` logic on the `itemRenderer` prop. However, that is still possible which may actually help migrate things from the old table faster.

I dediced for these changes because it makes rendering a table much easier:
* The only two required props to render the simplest responsive table are `columns` and `rows` 🎉


packages/components/simple-table/simple-table.styles.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  ${props => (props.maxHeight ? `max-height: ${props.maxHeight}px;` : '')}
+  ${props =>
+    props.maxWidth ? `max-width: ${props.maxWidth}px;` : ''}
+
+  overflow: auto;
+`;
+
+const Header = styled.div`
```
tdeekens: Would it make sense to just use th, td and in general semantic markup for the new table and not a divs? What lead to the decision of using divs in general?

jonnybel: That is a good question, as I am an adept of using semantic markup whenever possible 😄
I started by using it at first, but after studying some approaches, I realized it has some limitations, the biggest one being that it can't be combined with CSS-grid (which enables a lot of really cool layout magic) unless by using the css rule `display: contents` [which is not very well supported by all browsers](https://caniuse.com/#search=display%20contents).

Also, I didn't find the semantic markup of tables to be particularly useful, in terms of a11y navigation and screen readers. We can add some aria attributes to ease job of screen readers.

Lastly, the current table also doesn't use semantic `table` markup, so I didn't consider this a critical requirement.

That being said, it is still an option on the table (pun intended 👀) and I will continue to study this possibility, but currently, the ⚖️ of "table markup" vs "css-grid" is tending towards the latter 😄.

emmenko: Do we plan to still have **one single** table?

If not, we can have different kind of tables, with e.g. one of the basic ones being built as an HTML table.

Can you clarify this a bit?

tdeekens: I think we will have multiple tables. I assume from discussions and the name BasicTable here. Which ones specifically we need/want and how to split interactions/use cases across them we should dicuss if not already somewhere :)

jonnybel: The plan is to have a few exported tables with different added functionalities (such as sorting), and perhaps also exported "cell types" that one could use for the `renderItem` for common use cases.

But the "base" table would be the same for all of them, and the main challenge is to keep its layout flexible and responsive (and this is where css-grid shines).

I will put a bit more focus on trying to use `table` markup. In fact, with my current challenge of hovering/selecting whole rows, having an element for a whole row can be quite useful (but it breaks the css grid), so I will test more thoroughly how the `display: contents` rule behaves on all major browsers to see if it could be usable.

emmenko: Cool, thanks for the explanation 💯

packages/components/simple-table/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './src';
```
emmenko: export * from './src';

packages/components/simple-table/src/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './simple-table';
```
emmenko:
```suggestion
export { default } from './simple-table';
export { default as version } from './version';
```

packages/components/simple-table/src/simple-table.story.js
```diff
@@ -0,0 +1,178 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, number, boolean, select } from '@storybook/addon-knobs';
+import { Value } from 'react-value';
+import SimpleTable from './simple-table';
+
+const items = [
+  {
+    key: 'camilo-jimenez',
```
emmenko: Maybe we should come up with other names 😅

jonnybel: Yeah, I was going to change this at some point 😆 For now I just took the same set of data that we had

src/hooks/use-row-selection/use-row-selection.spec.js
```diff
@@ -0,0 +1,144 @@
...
+            </button>
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+};
+
+it('should have no selected items by default', () => {
+  const { getByTestId } = render(<TestComponent items={testData} />);
```
tdeekens: Nice flat tests. Maybe assign it to `const rendered`.

src/hooks/use-row-selection/use-row-selection.spec.js
```diff
@@ -0,0 +1,144 @@
...
+
+  testData.forEach(item => {
+    expect(getByTestId(item.key)).toHaveTextContent('false');
```
tdeekens: Maybe queryBy gives you better errors?

packages/components/simple-table/src/simple-table.styles.js
```diff
@@ -0,0 +1,39 @@
...
+  display: contents;
+  position: ${props => (props.isSticky ? 'sticky' : 'static')};
+  top: 0;
+`;
+
+const Body = styled.tbody`
+  display: contents;
+`;
+
+const Row = styled.tr`
```
tdeekens: 🏩

src/hooks/use-row-selection/use-row-selection.js
```diff
@@ -0,0 +1,81 @@
...
+    [keyName]: state[item.key],
+  }));
+
+  const getIsRowSelected = key => state[key];
+  const toggleRow = key => dispatch({ type: 'toggle', payload: key });
+  const selectRow = key => dispatch({ type: 'select', payload: key });
+  const deselectRow = key => dispatch({ type: 'deselect', payload: key });
+  const selectAllRows = () => dispatch({ type: 'selectAll' });
+  const deselectAllRows = () => dispatch({ type: 'deselectAll' });
+  const getCountSelectedRows = () =>
```
tdeekens: Sorry, just nitting around cause this is super useful :) 

```suggestion
  const getNumberOfSelectedRows = () =>
```

jonnybel: That was the name I gave it at first 😆

tdeekens: Haha. Okay. Totally your choice and needlessly subjective.

jonnybel:
A heads up about this commit: https://github.com/commercetools/ui-kit/pull/1233/commits/00996783d05258012e0b03fe6549577909fe3b8c

I've changed the API a little bit:
* `items` prop renamed to `rows`, to create less confusion regarding the difference with actual _items_ (cells) of a table. Also, I think it pairs better with the `columns` prop.
* Changed the `renderItem` prop of the table back to `itemRenderer`, and this becomes the default renderer for each item. The default is set to simply render whatever is the value of an item property (usually it's a string).
* Added a `renderItem` prop to the column definitions. This is the renderer to be used for custom cases per-column, taking precedence over the default `itemRenderer`. I find it's easier to have this on the column definition (among with others like `onClick`), instead of having to implement a `switch (case)` logic on the `itemRenderer` prop. However, that is still possible which may actually help migrate things from the old table faster.

I dediced for these changes because it makes rendering a table much easier:
* The only two required props to render the simplest responsive table are `columns` and `rows` 🎉

jonnybel:
Added the PCM team as reviewers because our first real test for the Table is using it on the Products List 🙂 

(I will finish writing the README soon)

Share your thoughts about the current API.

ahmehri:
Kudos for implementing this 💯 
`tableMaxHeight` and `tableMaxWidth` knobs don't have any effect or am I missing something?

Kapture 2020-01-31 at 16 58 22

And can you tell me how to test the stickiness of the header?

jonnybel:
Oh, I might have broken those on one of the last commits when I renamed the props 😅 

I'll add a knob to control the stickiness, forgot about it 👍 

Thanks @ahmehri !

jonnybel:
Fixed the table dimensions knobs.

The sticky behaviour seems to be broken :( I suspect it doesn't use in conjunction with `display: contents`... I'll have to figure out a solution for this.

src/hooks/use-row-selection/use-row-selection.spec.js
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
+import { render } from '../../test-utils';
```
ahmehri: Any reason to not use [@testing-library/react-hooks](https://github.com/testing-library/react-hooks-testing-library)?

src/hooks/use-row-selection/use-row-selection.spec.js
```diff
@@ -142,3 +142,19 @@ it('should be possible to deselect all rows', () => {
...
+    { key: 'woman-at-war' },
+  ];
+
+  const rendered = render(
+    <TestComponent keyName="selected" items={customRows} />
+  );
+
+  expect(rendered.queryByTestId('system-crasher')).toHaveTextContent('false');
+  expect(rendered.queryByTestId('birds-of-passage')).toHaveTextContent('true');
+  expect(rendered.queryByTestId('system-crasher')).toHaveTextContent('false');
```
ahmehri:
```suggestion
  expect(rendered.queryByTestId('woman-at-war')).toHaveTextContent('false');
```

packages/components/simple-table/src/simple-table.story.js
```diff
@@ -3,7 +3,7 @@ import { storiesOf } from '@storybook/react';
 import { withKnobs, number, boolean, select } from '@storybook/addon-knobs';
 import { Value } from 'react-value';
 import SimpleTable from './simple-table';
-import { useRowSelection } from '../../../hooks';
+import { useRowSelection } from '.';
```
ahmehri: 💯

packages/components/simple-table/src/simple-table.story.js
```diff
@@ -126,10 +128,115 @@ const columns = [
...
   .addDecorator(withKnobs)
   // .addDecorator(withReadme(Readme))
   .add('SimpleTable', () => {
+    const [tableData, setTableData] = React.useState({
+      rows: items,
+      columns: initialColumnsState,
+    });
+
+    const handleUpdateColumn = (column, colIndex) => {
+      const newColumns = tableData.columns;
```
ahmehri: I'm not sure but I think this assignment is mutable, i.e changes to `newColumns[colIndex]` will also affect `tableData.columns[colIndex]`.

ahmehri:
What happens to the idea of using react-table? I assume it will be done in the more advanced tables, right?

packages/components/simple-table/src/simple-table.story.js
jonnybel: You're right! This was a bit half-assed 😳

packages/components/simple-table/simple-table.story.js
```diff
@@ -0,0 +1,178 @@
...
+                    onChange={() => onChange(event.target.value)}
+                  />
+                )}
+              />
+            );
+          default:
+            return item[column.key];
+        }
+      }}
+      tableMaxHeight={number('tableMaxHeight', 0, {
```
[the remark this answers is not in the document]
jonnybel: It's ignored when it's `0`, and I didn't want to set a fixed default, because I wanted the default to be the auto size (responsive).

jonnybel:
@ahmehri I experimented integrating it with this implementation when I worked on Row Selection, but I felt it brought a bit too much bloat and required too much "glue" to use it's API while keeping ours, and also to integrate with our layout system.
I realized it wasn't that complicated to implement our own `useRowSelection` hook and took some inspiration from them, but stripped it down to just what we needed and ended up happy with the result.

I might still use it for more advanced cases (like row expansion), but for now I'm still not convinced I really need it, and so we can avoid pulling a new dependency.
Next up, even for sorting, it's not that complicated to do it on our own.

src/hooks/use-row-selection/use-row-selection.spec.js
jonnybel: Hmm, not really. However, I do like the idea of testing this particular hook with a layout component, providing an example of how to use it in practice.

packages/components/simple-table/src/cell.js
```diff
@@ -0,0 +1,118 @@
...
+        * {
+          fill: ${vars.colorNeutral};
+        }
+      }
+    }
+  `;
+
+  return (
+    <BaseHeaderCell
+      {...omit(props, ['onSortChange'])}
```
tdeekens: Nit: maybe `const baseHeaderCellProps` ontop could ease reading this.

packages/components/simple-table/src/simple-table.js
```diff
@@ -0,0 +1,107 @@
...
+      {props.rows.map((row, rowIndex) => (
+        <Row
+          key={row.id}
+          onClick={
+            props.onRowClick ? () => props.onRowClick(row, rowIndex) : undefined
+          }
+        >
+          {props.columns.map(column => (
+            <DataCell
+              key={`${rowIndex}-{row.id}/${column.key}`}
```
tdeekens: Why the rowIndex if we have a row.id? Isn't the row.id enough? BTW: who generates this id?

packages/components/simple-table/src/simple-table.spec.js
```diff
@@ -0,0 +1,86 @@
...
+const baseProps = { rows: testRows, columns: testColumns };
+
+describe('SimpleTable', () => {
+  it('should forward data-attributes', () => {
+    const { container } = render(<SimpleTable {...baseProps} data-foo="bar" />);
+    expect(container.querySelector("[data-foo='bar']")).toBeInTheDocument();
+  });
+
+  it('should render the column labels', () => {
+    const { queryByText } = render(<SimpleTable {...baseProps} />);
```
tdeekens: Below we save the return in `rendered` here not. Should we also use `const rendered = ` here?

src/hooks/use-row-selection/use-row-selection.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  const {
+    rows,
+    toggleRow,
+    selectRow,
+    deselectRow,
+    selectAllRows,
+    deselectAllRows,
+    getIsRowSelected,
+    getNumberOfSelectedRows,
+    // eslint-disable-next-line react/prop-types
```
tdeekens: On this line? Should we disable it for the file or specs in general?

src/hooks/use-sorting/use-sorting.js
```diff
@@ -0,0 +1,60 @@
...
+
+  if (direction === 'desc') {
+    return sortedItems.reverse();
+  }
+
+  return sortedItems;
+};
+
+const useSorting = (items, column, sortDirection, sortMethod) => {
+  const initialState = {
```
tdeekens: Nit: this could live outside the hook as `const getInitialState = ({...}) =>` then inside the hook we cna pass the function to React.useState. It will only be invoked once. The fn avoids a redefinition of the obejct and a `sortItems` invocation which is then actually not used by the hook anymore.

Additionally we could send the state ourside the hook itself. So:

```js
const useSortingState = () => {
   const [sortState, setSorting] = React.useState(getInitialSortingState);
   
   return [sortState, setSorting]; 
}
```

This allow us to `useDebugValue`

```js
const useSortingState = () => {
   const [sortState, setSorting] = React.useState(getInitialSortingState);
   
   React.useDebugValue({
      sortState
   })

   return [sortState, setSorting]; 
}
```

which can be helpful for consumers in the apps as it shows up nicely in the React DevTools.

src/hooks/use-sorting/use-sorting.js
```diff
@@ -0,0 +1,60 @@
...
+    sortDirection,
+  };
+
+  const [sortState, setSorting] = React.useState(initialState);
+
+  const onSortChange = columnKey => {
+    let nextSortDirection;
+
+    // if the intented column is not already sorted, the initial direction is 'asc'
+    if (sortState.sortedBy !== columnKey) {
```
tdeekens: What if I resort by the same columnKey? Will we recompute it all or should it be a noop?

packages/components/simple-table/src/cell.js
jonnybel: Yes, I'm not too happy with this and I'm actually going to refactor the `BaseHeaderCell`, and also make them fully accessible when they're clickable.

Might still use your tip though!

packages/components/simple-table/src/simple-table.js
jonnybel: No particular reason actually...

The ID is given by the consumer (and the only required field of the `rows` objects). Usually it's already part of the items that you want to render on the table and not for displaying.

Anyway, I'll remove the `rowIndex`

src/hooks/use-row-selection/use-row-selection.spec.js
jonnybel: It just complains about the props on the next line. But this is just a test component so idc about adding extra boilerplate.

I'll move to the top of the file though.

src/hooks/use-sorting/use-sorting.js
jonnybel: This is interesting! Thanks, I'll try it

src/hooks/use-sorting/use-sorting.js
jonnybel: When you click on the same one, it just reverses the `sortDirection`.
I just added a test for this specific case 😇

jonnybel: But you're right, in that case can skip the sort (which will still run at least one iteration through all the array) and just do `.reverse()`.

packages/components/simple-table/src/simple-table.js
```diff
@@ -13,9 +13,9 @@ const SimpleTable = props => (
   >
     <Header isSticky={props.isHeaderSticky}>
       <Row>
-        {props.columns.map((column, columnIndex) => (
+        {props.columns.map(column => (
           <HeaderCell
-            key={`${columnIndex}-${column.key}`}
```
tdeekens: Yay for less mounting and unmounting!

src/hooks/use-row-selection/use-row-selection.js
```diff
@@ -1,6 +1,16 @@
 import React from 'react';
 
-function reducer(state, action) {
+const getInitialState = (rows, keyName) =>
```
tdeekens: 💯

src/hooks/use-sorting/use-sorting.js
```diff
@@ -15,14 +15,29 @@ const sortItems = (items, field, direction, sortingFunction = sortBy) => {
...
 
+const getInitialState = (items, field, sortDirection, sortingFunction) => ({
+  items: sortItems(items, field, sortDirection, sortingFunction),
+  sortedBy: field,
+  sortDirection,
+});
+
+const useSortingState = (items, field, sortDirection, sortingFunction) => {
+  const [sortState, setSorting] = React.useState(
+    getInitialState(items, field, sortDirection, sortingFunction)
```
tdeekens: This should return a function maybe? So `getInitialSortState = ({...}) => () => ({...})`. Then we avoid the sorting inside the func on each render.

jonnybel: Ah, got it! It's the [lazy state initialization](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state) thing 😄 
Thanks!

tdeekens: Yup, exactly that. Sorry for not calling the beauty by name earlier.

jonnybel:
Hello my dear reviewers 😄 

If we are happy with the core implementation, I would like to merge this PR soon to release in a "beta" state and test in our beloved MC in a draft PR.
This way, both us and the design team can see how the Table behaves on a production-level use case, enabling us to discover potential bugs, ideas for enhancements, performance bottlenecks, as well understanding how migrating process goes.

Further developments for the Table would be done in additional PRs. This PR is already pretty long anyway, so I think it would be easier (and less scary) to review the upcoming additional changes in new PRs 🙂 

Thanks!

jonnybel:
Renamed this component to `DataTable`.

This PR will finally be merged after this. [See this comment.](https://github.com/commercetools/ui-kit/pull/1233#issuecomment-595719008)

PR #1238, opened 2020-01-14
fix(select-input): multi select tag borders

jonnybel:
#### Summary

Small adjustment to the `SelectInput` styles, for multi-values: When there is no `multiValueRemove` component being displayed (by passing an empty component to it), the `multiValueLabel` doesn't have a right border.
This PR fixes that, to prevent the need to also override the `multiValueLabel` just for this styling detail, or other "hacky" workarounds.

#### Before:
image

#### After:
image


adnasa:
One thing I've noticed is that `MultiValueRemove` must be a component. So either `TagRemove` or a `div`. Have you considered the possibility of accepting just `null` and from there passdown a dumb `div` to `Select`

adnasa:
nvm.. one could do this:

> So basically it will work if you pass  () => null to components.MultiValueRemove (or a React.Fragment)

PR #1242, opened 2020-01-15
chore(tag): fix storybook example

jonnybel:
#### Summary

If you go to: https://uikit.commercetools.com/?path=/story/components-tags--tag
And go to `Knobs`, the page crashes.

This PR fixes the problem.

### DIFF
--- a/src/components/tag/tag.story.js
+++ b/src/components/tag/tag.story.js
@@ -11,7 +11,7 @@ const Story = () => (
   <Section>
     <Tag
       type={select('type', ['normal', 'warning'], 'normal')}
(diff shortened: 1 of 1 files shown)


src/components/tag/tag.story.js
```diff
@@ -11,7 +11,7 @@ const Story = () => (
   <Section>
     <Tag
       type={select('type', ['normal', 'warning'], 'normal')}
-      linkTo={select('linkTo', ['/foo', null], '/foo')}
```
jonnybel: The option `null` breaks the selector and crashes the page ⚡️

PR #1246, opened 2020-01-23
feat(contraints-horizontal): modify to pass down data props

dogayuksel:
#### Summary

Modify `Constraints.Horizontal` to pass down `data-*` props.

### DIFF
--- a/src/components/constraints/horizontal/horizontal.spec.js
+++ b/src/components/constraints/horizontal/horizontal.spec.js
@@ -6,19 +6,19 @@ import Horizontal from './horizontal';
 // That tests also ensure this component accepts a "constraint" prop.
 
 it('should render children', () => {
-  const { getByTestId } = render(
+  const rendered = render(
     <Horizontal>
       <div data-testid="child" />
     </Horizontal>
   );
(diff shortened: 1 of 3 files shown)


src/components/constraints/horizontal/horizontal.spec.js
```diff
@@ -13,3 +13,12 @@ it('should render children', () => {
   );
   expect(getByTestId('child')).toBeInTheDocument();
 });
+
+it('should pass down `data` prop', () => {
+  const { getByTestId } = render(
```
tdeekens: Nit: `const rendered =`

src/components/constraints/horizontal/horizontal.spec.js
```diff
@@ -13,3 +13,12 @@ it('should render children', () => {
...
+    <Horizontal data-testid="child">
+      <div />
+    </Horizontal>
+  );
+  expect(getByTestId('child')).toBeInTheDocument();
```
tdeekens: Recommend use `renderd.queryBy`. We have an ESLint rule in the MC for it.

PR #1251, opened 2020-01-27
fix(secondary-button): to prop-type

tdeekens:
#### Summary

This pull request intends to clear up some confusion about the `to` prop-type.

#### Description

The `linkTo`-prop was recently being deprecated. In favour of using the `as`-prop with a paring of the `to`-prop. However, the `linkTo` prop either accepted a string of an object of `{ pathname, search, query }` for the underlying `react-router`. The `to`-prop doesn't support this. This overall makes migrating to the `to` prop in internal apps not possible as far as I can tell. As a result I propose to accept the same shape on the `to` prop and forward it.


tdeekens:
Note, that this "blocks" a non-urgent internal PR. I would also silence the console there dependening on when the next release is planned 😸.

PR #1253, opened 2020-01-29
[UKIT-97] fix(tag): on remove button hover when disabled

jonnybel:
#### Summary

Fix for [UKIT-97](https://jira.commercetools.com/browse/UKIT-97)

> When the user hovers the removable area on a disabled label, the button still displays the orange hover effect that indicates the label is removable. There should not display any hover effect when disabled.

### DIFF
--- a/src/components/tag/tag.js
+++ b/src/components/tag/tag.js
@@ -198,8 +198,8 @@ const Tag = props => {
                   background: inherit;
                   border-style: solid;
                   border-width: 1px 1px 1px 1px;
-                  &:hover,
-                  &:focus {
+                  :not(:disabled)&:hover,
+                  :not(:disabled)&:focus {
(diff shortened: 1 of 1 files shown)


src/components/tag/tag.js
```diff
@@ -198,8 +198,8 @@ const Tag = props => {
                   background: inherit;
                   border-style: solid;
                   border-width: 1px 1px 1px 1px;
-                  &:hover,
-                  &:focus {
+                  :not(:disabled)&:hover,
+                  :not(:disabled)&:focus {
```
tdeekens: How dare you use css for this!

jonnybel: If it can be done with CSS, I will do it with CSS 😎

Issue #1269, opened 2020-02-28
Add ability for hidden input labels to be able to target inputs without visual labels


tdeekens:
Our current setup of inputs and labels and their combination in labels works great. However, at times we do not have the visual space to place a label next to the input. 

This case here

<img width="288" alt="CleanShot 2020-02-28 at 15 53 39@2x" src="https://user-images.githubusercontent.com/1877073/75558775-9e654600-5a42-11ea-8dd9-fcd841732508.png">

Is an input without an label. The same applied to

<img width="385" alt="CleanShot 2020-02-28 at 15 54 57@2x" src="https://user-images.githubusercontent.com/1877073/75558817-b1781600-5a42-11ea-8a3b-66d6baa39190.png">

This is not ideal from an accessibility perspective and also impacts how we can test these inputs. Without a label inputs are hard to target in our rtl and cypress testing setup.

As a result I propose that:

1. The <Name>Field receive an `isLabelHidden` (or named differently) to hide labels visually but have them present in the DOM
2. Add a `isVisible` to our `Label` component

This could allow us to render a invisible label next to an input while targeting it in testing.

Note that this is just an idea and I am myself not really sure if the approach is right.

jonnybel:
Thanks for opening this issue.

I understand the problem and I agree with the concept of adding a hidden label to make tests easier (as well as improving a11y while we're at it).

Adding an `isLabelHidden` prop to the `Fields` could actually also be useful to a few use cases (which we've already seen) where a developer wants to take advantage of the validation logic that is built-in for `Fields` but not `Inputs`, while not wanting to render the `Label`.

I think my approach would be exporting an `HiddenAccessibleLabel`, or even a `HiddenAccessibleComponent` which you could decide what element to render with an `as` prop, and it would just add the necessary CSS to make it [hidden but screen-readable](https://webaim.org/techniques/css/invisiblecontent/#techniques). Then this component would be used by the Field components as well when you set the `isLabelHidden` prop.

tdeekens:
Thanks for understanding the issue. This is an interesting approach. Having a AccessibleLabel and a HiddenAccessibleLabel could be a good separation. Question only would be how much they differ after all apart form some CSS.  Maybe the HiddenAccessibleLabel uses the AccessibleLabel below? Is that what you mean? Sorry, if this is already too implementation specific.

jonnybel:
[...]

PR #1271, opened 2020-03-02
feat: add accessible-hidden component

jonnybel:
#### Summary

This is a new component proposal for addressing #1269 

The idea is providing an easy way to [render invisible, but machine-readable content](https://webaim.org/techniques/css/invisiblecontent/).

#### Usage

```js
import AccessibleHidden from '@commercetools-uikit/accessible-hidden';

<AccessibleHidden>
  <label htmlFor="sample-input">Sample Input</label>
</AccessibleHidden>;
```

##### Follow-up:
 Add a prop to Field components to allow hiding the `label` using this component.

### DIFF
--- a/packages/components/accessible-hidden/README.md
+++ b/packages/components/accessible-hidden/README.md
@@ -1,0 +1,19 @@
+# Accessible Hidden
+
+#### Description
+
+Visually hides a component, while keeping it accessible for screen readers and other assistive technology, as well as for testing purposes with tools such as react-testing-library and cypress.
+
+#### Usage
+
+```js
+import AccessibleHidden from '@commercetools-uikit/accessible-hidden';
+
+<AccessibleHidden>
+  <label htmlFor="sample-input">Sample Input</label>
+</AccessibleHidden>;
+```
+
+#### Properties
+
+This component only accepts the prop `children`, which is the component you want to hide.
--- a/packages/components/accessible-hidden/README.md
+++ b/packages/components/accessible-hidden/README.md
@@ -2,14 +2,14 @@
 
 #### Description
 
-This component is used to hide content offscreen, removing it from sighted users, while keeping the still accessible to screenreaders and other assistive technology.
-It can also be useful for testing with tools like react-testing-library and cypress, which query elements through content which you might not want to be showing explicitly, such as querying for a input by its label while the label is visually hidden.
+This component is used to hide content offscreen, removing it from sighted users, while keeping it still accessible to screenreaders and other assistive technology.
+It can also be useful for testing with tools like react-testing-library and cypress, which requires quering elements through content which might not be intended to be visible on screen, such as querying for a input by its label while the label is visually hidden.
 
 It's the logical opposite of [the `aria-hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute).
 
 #### Usage
 
-##### Example: A11y
+##### Example: Accessibility (a11y)
 
 In this example, we're showing additional text specifically to be read only by screen readers, to make it more contextualized and easier to understand for non-sighted users (as well "translating" the numeronym which might confuse automatic screen-readers).
 
@@ -23,7 +23,7 @@ import AccessibleHidden from '@commercetools-uikit/accessible-hidden';
 
 ##### Example: Tests
 
-This is an example for when you would want to render an Input without visually displaying a label visually, but still having a label to query by, when testing with RTL, as well as being a11y friendly.
+This is an example for when you would want to render an input without visually displaying a label. You still would want a label to be present so it can be used when testing with RTL as well as supporting a11y.
 
 ```js
 const rendered = render(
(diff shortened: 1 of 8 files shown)


packages/components/accessible-hidden/src/accessible-hidden.js
```diff
@@ -0,0 +1,29 @@
...
+import { css } from '@emotion/core';
+
+const AccessibleHidden = props => {
+  return (
+    <div
+      css={css`
+        clip: rect(0 0 0 0);
+        width: 1px;
+        height: 1px;
+        margin: -1px;
```
tdeekens: Nifty.

packages/components/accessible-hidden/src/accessible-hidden.spec.js
```diff
@@ -0,0 +1,22 @@
...
+it('should find text inside AccessibleHidden', () => {
+  const rendered = render(<AccessibleHidden>Spies!</AccessibleHidden>);
+
+  expect(rendered.getByText('Spies!')).toBeInTheDocument();
+});
+
+it('should find the input of a label inside AccessibleHidden', () => {
+  const rendered = render(
+    <>
+      <AccessibleHidden>
```
tdeekens: Super test!

jonnybel:
I called that way because it's for rendering a component both Accessible & Hidden.

I'm not such a fan of "HiddenAccessibility" or "HideAccessibility" because it gives me the impression it's "hiding the accessibility" 🤔

tdeekens:
Yeah that makes sense. Still somehow it is hiding it. InvisbleAccessibility maybe? Anyway, too much naming hassle maybe? 🤓😎

jonnybel:
Let's leave it to marinate overnight and make a decision tomorrow 👨‍🍳

packages/components/accessible-hidden/src/accessible-hidden.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+
+const AccessibleHidden = props => {
```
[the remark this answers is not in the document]
jonnybel: I improved the explanation in the readme, and added more examples and a reference link 🙂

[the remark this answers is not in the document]
jonnybel: @emmenko Do you think the readme is clearer now or should I add more examples/references?

packages/components/accessible-hidden/README.md
```diff
@@ -0,0 +1,50 @@
+# Accessible Hidden
+
+#### Description
+
+This component is used to hide content offscreen, removing it from sighted users, while keeping the still accessible to screenreaders and other assistive technology.
```
tdeekens:
```suggestion
This component is used to hide content offscreen, removing it from sighted users, while keeping it still accessible to screenreaders and other assistive technology.
```

packages/components/accessible-hidden/README.md
```diff
@@ -0,0 +1,50 @@
...
+It can also be useful for testing with tools like react-testing-library and cypress, which query elements through content which you might not want to be showing explicitly, such as querying for a input by its label while the label is visually hidden.
```
tdeekens:
```suggestion
It can also be useful for testing with tools like react-testing-library and cypress, which requires quering elements through content which might not be intended to be visible on screen, such as querying for a input by its label while the label is visually hidden.
```

packages/components/accessible-hidden/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+It's the logical opposite of [the `aria-hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute).
+
+#### Usage
+
+##### Example: A11y
```
tdeekens:
```suggestion
##### Example: Accessibility (a11y)
```

packages/components/accessible-hidden/README.md
```diff
@@ -0,0 +1,50 @@
...
+import AccessibleHidden from '@commercetools-uikit/accessible-hidden';
+
+<h3>An Article on A11y</h3>
+<p>A summary of the article</p>
+<button>Read More<AccessibleHidden> from An Article on Accessibility</AccessibleHidden></button>
+```
+
+##### Example: Tests
+
+This is an example for when you would want to render an Input without visually displaying a label visually, but still having a label to query by, when testing with RTL, as well as being a11y friendly.
```
tdeekens:
```suggestion
This is an example for when you would want to render an input without visually displaying a label. You still would want a label to be present so it can be used when testing with RTL as well as supporting a11y.
```

packages/components/accessible-hidden/README.md
```diff
@@ -0,0 +1,50 @@
...
+    </AccessibleHidden>
+    <input id="maiden-name-input" type="text"></input>
+  </>
+);
+
+expect(rendered.getByLabelText('Enter your Maiden Name')).toBeInTheDocument();
+// ✓ True
+```
+
+It is a common use case for inputs inside Tables, where a sighted user understands the context of the input through the column header and a label element taking space inside the cell is undesirable.
```
tdeekens:
```suggestion
It is a common requirement to show inputs inside tables. In such a case a sighted user should be able understand context of the input through the column header. As label element taking space inside the cell is undesirable.
```

jonnybel: Just adjusted a couple of words in your suggestion to add it to the batch:

```suggestion
It is a common requirement to show inputs inside tables. In such a case a sighted user should be able to understand the context of the input through the column header. A label element taking space inside the cell is undesirable.
```

PR #1272, opened 2020-03-06
fix(collapsible-panel): improve accessibility

### LINKED ISSUE #1263: Improve accessibility of CollapsiblePanel
The `CollapsiblePanel` component [has been reported](https://github.com/commercetools/merchant-center-frontend/pull/8383/files#r379316051) to not being accessible (no aria labels) and hard to test because of its DOM structure: it should be easier to select the component that represents the content of a given header, for testing purposes.

My suggestion for fixing the `a11y`:
* We should use `aria-hidden` and `aria-expanded` for the 2 states.
* Ensure that the Header behaves like a `button`, being focusable and triggerable from the keyboard.
* We could add an `id` to the content of the panel and associate it with `aria-controls` on the header button.

For the DOM structure, it has been suggested to make the content a child of the header, but I would try to avoid this, because collapsibles are usually just plain buttons which show/hide another element regardless of its location in the DOM (almost always its next direct sibling).

A couple of sources for helping with this:
* https://inclusive-components.design/collapsible-sections/
* https://accessibility.huit.harvard.edu/%E2%9C%8E-practice-expandable-sections

jonnybel:
Fixes #1263 

I recommend checking the diff commit-by-commit, to understand better each of the changes, especially because I converted the whole component into a function.

* Adds an export to `AccessibleButton` to get the normalized `button` styles, called `getNormalizedButtonStyles`, so we can re-use them in places where we don't want all of the component's functionality, but just to reset the styles of a `button` element, removing the vendor's default styles.
This will become more useful in the future as there are more plans to improve the overall a11y of our components, where we should use `button` components to enable tab navigation and focusability but don't want that ugly default button look.

* Uses the getter above to transform the Header of the CollapsiblePanel into a focusable button, and adds aria attributes(`aria-expanded`, `aria-hidden` and `aria-controls`) to the Header and Content of the Panel, among with an `id` to the Content, improving the a11y while also making them easier to query for testing purposes.

Additional minor changes:
* Refactored the component to become a function.
* Simplified the layout by reducing the number of wrappers and combining styles.
* Allow passing `data-` and `id` attributes to the Panel Header.

<details><summary>Fixed a tiny visual bug where the bottom border being added/removed when opening/closed caused a little jump in the Panel size. (Click to see slow-motion gif of the bug)</summary>

Kapture 2020-03-06 at 18 23 17

[Ha! This element is _Collapsible_ 😏 ]
</details>


packages/components/collapsible-panel/src/collapsible-panel.js
```diff
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import isNil from 'lodash/isNil';
+import uniqueId from 'lodash/uniqueId';
```
tdeekens: Is this a new dep? Maybe using uuid can help. We use it everywhere else. Not sure, just a thought.

jonnybel: `uuid` isn't a dependency in UI-kit either.

However, since it's used in the MC it might be preferable to use the same tool 👍

tdeekens: On the other hand UIKit uses lodash and the MC is not the only user. Sorry, the discussion I started is a bit side-tracking. Lodash uniqueid is fine. Proceed 🤓.

jonnybel: Yeah, also, I was thinking again and I prefer the simple sequential `id` from `lodash/uniqueId` rather than the much longer `uuid`, for HTML elements.

jonnybel:
Will add tests to the aria attributes and also fix the issues detected by Percy 🔍

packages/components/collapsible-panel/src/collapsible-panel.spec.js
```diff
@@ -146,6 +146,109 @@ it('should not call "onToggle" when header is clicked while disabled', () => {
...
+    const panelContentId = CollapsiblePanel.getPanelContentId('example');
+
+    // assert that the header button has the aria attribute
+    const headerButton = rendered.container.querySelector('button');
+    expect(headerButton).toHaveAttribute('aria-controls', panelContentId);
+
+    // find the correspondent panel content
+    expect(
+      rendered.container.querySelector(`[id=${panelContentId}]`)
+    ).toBeInTheDocument();
```
tdeekens: Nit: we could encapsulate this a bit bette rinto the render:

```js
const render = () => {
   const rendered = render()
   const panelContentId = CollapsiblePanel.getPanelContentId('example');
   const queryByPanelId = () => rendered.container.querySelector(`...`);

   return {
      ...rendered,
      queryByPanelId
   }
}
```

packages/components/collapsible-panel/src/collapsible-panel.spec.js
```diff
@@ -159,94 +159,72 @@ describe('getPanelContentId', () => {
...
         Children
       </CollapsiblePanel>
     );
+    const getPanelHeader = () => rendered.container.querySelector('button');
 
-    const panelContentId = CollapsiblePanel.getPanelContentId('example');
+    const panelContentId = CollapsiblePanel.getPanelContentId(
+      props.id || 'example'
+    );
+    const getPanelContent = () =>
```
tdeekens: Nit: get* throws in rtl. That's why I felt `query` (which doesn't) wording wise was more apt.

jonnybel: I only didn't call these `query` because they don't accept any args to _query by_, they "automatically" **get** the element we want without exposing the query 😄 Hope that's okay

tdeekens: Totally. Thanks. You're a beach bum!

PR #1278, opened 2020-03-12
chore(card): fix documentation and small refactor

jonnybel:
#### Summary

@YahiaElTai raised the fact that the `className` prop was documented with the wrong casing. This is fixed in this PR.

Also took the opportunity to remove one unnecessary wrapper and styles from the component.

### DIFF
--- a/packages/components/card/src/card.js
+++ b/packages/components/card/src/card.js
@@ -8,10 +8,10 @@ const Card = props => (
   <div
     {...filterDataAttributes(props)}
     css={css`
-      display: flex;
-      font-size: 1rem;
-      flex-direction: column;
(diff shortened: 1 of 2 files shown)


packages/components/card/src/card.js
```diff
@@ -8,10 +8,10 @@ const Card = props => (
   <div
     {...filterDataAttributes(props)}
     css={css`
-      display: flex;
-      font-size: 1rem;
-      flex-direction: column;
+      box-sizing: border-box;
```
jonnybel: This rule makes it unnecessary to have a wrapper just to set a padding.

Also, without the extra wrapper there's no need for flex here.

PR #1283, opened 2020-03-16
fix(collapsible-panel): Accessible panel Header element

jonnybel:
#### Summary

This PR addresses this problem: https://github.com/commercetools/merchant-center-frontend/pull/8565#issuecomment-598637568.

It occurs when the `CollapsiblePanel` is passed any child elements for the Header that are buttons, because the Header of the CollapsiblePanel is also a `button` element.

#### Description

* Fix the `AccessibleButton` component to become truly accessible when the `as` prop is used:
  * Add `role` and `tabIndex` attributes.
  * Add `onKeyPress` event for `Enter` and `Space` to trigger the `onClick`, for full keyboard navigation support.
* Use this adjusted `AccessibleButton` as a `div` for the CollapsiblePanel header, instead of a `button` element but keeping the a11y.

### DIFF
--- a/packages/components/buttons/accessible-button/src/accessible-button.js
+++ b/packages/components/buttons/accessible-button/src/accessible-button.js
@@ -9,11 +9,15 @@ import getNormalizedButtonStyles from './accessible-button.styles';
 
 const propsToOmit = ['onClick'];
 
+const getIsEnterOrSpace = e =>
+  e.key === ' ' || e.key === 'Spacebar' || e.which === 13 || e.which === 32;
+
 const Button = styled.button`
   ${getNormalizedButtonStyles}
   display: inline-flex;
-  outline: 0;
   font-size: ${vars.fontSizeDefault};
+
+  ${props => (!props.as || props.as === 'button' ? 'outline: none' : '')}
 
   cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
   &:disabled {
@@ -30,7 +34,9 @@ const AccessibleButton = React.forwardRef((props, ref) => {
         event.preventDefault();
         return false;
       }
-      if (!props.isDisabled && onClick) return onClick(event);
+      if (!props.isDisabled && onClick) {
+        return onClick(event);
+      }
       // eslint-disable-next-line no-useless-return, consistent-return
       return;
     },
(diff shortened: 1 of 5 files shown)


packages/components/buttons/accessible-button/src/accessible-button.js
```diff
@@ -9,12 +9,16 @@ import getNormalizedButtonStyles from './accessible-button.styles';
 
 const propsToOmit = ['onClick'];
 
+const getIsEnterOrSpace = e =>
+  e.key === ' ' || e.key === 'Spacebar' || e.which === 13 || e.which === 32;
```
tdeekens: What is `' '`? :). Is there maybe like a itty-bity util library we can use for this? Or do you think it's fine inlining it?

packages/components/buttons/accessible-button/src/accessible-button.js
```diff
@@ -39,9 +45,18 @@ const AccessibleButton = React.forwardRef((props, ref) => {
...
 
-  const buttonProps = {
-    type: props.type,
-  };
+  let buttonProps = {};
+  if (isButton) {
+    buttonProps = {
+      type: props.type,
+    };
+  } else if (!props.isDisabled) {
```
tdeekens: Nit: do-expression?

jonnybel: It gets a little weird-looking with the object braces for the logic I am trying to achieve, am i doing this right?

```
const buttonProps = 
do {
   if (isButton) {
     {
      type: props.type,
    }
   }
  else if (!props.isDisabled) {
    {
      role: 'button',
      tabIndex: '0',
      onKeyPress: event => getIsEnterOrSpace(event) && handleClick(event),
    }
  }
   else {
  { }
 }
```

packages/components/buttons/accessible-button/src/accessible-button.js
jonnybel: `' '` is for the Spacebar.

I just double-checked a bit deeper and simplified this line, according to this spec that React uses for the `key` field of their synthetic events: https://www.w3.org/TR/uievents-key/#keys-whitespace

packages/components/buttons/accessible-button/src/accessible-button.js
tdeekens: True that also don't make it much simpler either.

tdeekens: ...just tried to avoid the let in a way.

PR #1305, opened 2020-03-29
fix(data-table): allow disabling text wrap for column header label

jonnybel:
#### Summary

Adds an option to set whether the column header labels should be allowed to break in several lines or not. If true, the header may grow in height, otherwise the column will guarantee a larger minimum width, enough to contain the label in one line.


packages/components/data-table/README.md
```diff
@@ -72,6 +72,7 @@ return <DataTable rows={rows} columns={columns} />;
...
 | `isSortable`           | `boolean` | -        | `false` | If this is `true`, this column's header will be a become a button which calls `onSortChange` when clicked. You should enable this flag for every column you want to be able to sort. When at least one column is sortable, the table props `sortBy`, `sortDirection` and `onSortChange` should be set.                                                                                                                                                                |
+| `wrapHeaderLabel`      | `boolean` | -        | `true`  | Sets whether the column header labels should be allowed to break in several lines or not. If true, the header may grow in height, otherwise the column will guarantee a larger minimum width, enough to contain the label in one line.                                                                                                                                                                                                                                |
```
tdeekens: Struggled a bit with the name (sorry) on my end.  

Wondering if we could also flip the model: fitText or fitColumnHeader to be default true. 

If it wraps, whatever the prop is called, will the table still take 100% of the available space and not wrap text if not needed?

jonnybel: I prefer flipping it too, but have to discuss with the Design which shall be the default. Currently, it's wrapping (tho on the old table it wouldn't, but it was buggy and would even be clipped in some cases).

Meanwhile, I'm adding the flag to control it and then maybe change the default in a subsequent PR.

With wrapping on, it will only wrap if there isn't enough space. These are the steps:
1. Table grows horizontally as needed until 100% of parent.
2. Table reaches 100% and starts wrapping text which is allowed.
3. Table can't break any more text, stops shrinking and adds horizontal scrollbar.

jonnybel: The Design team prefers having the default to wrap text (at least for now).

I can still "flip" the naming to `fitColumnLabel` with the default of `false`.

tdeekens: Okay for the defaulting. For the namng unsure. Anything which describes what it does best I think. So `expandToColumnLabel`, `fitColumnLabel` could be options. But also wrap is good if that's the language used in ui-kit.

jonnybel: I prefer the term `wrap` as it is quite explicit and it translates directly to the css rule.

Do you prefer something like `allowWrapHeaderText` ? Or `allowHeaderLineBreak`?

There are plans for adding row _expansion_ as a different behaviour, so I would avoid using that term here. "Fit" also doesn't sound clear enough for me 🤔

tdeekens: Makes total sense. Then I am also fine with what we have.

jonnybel: Thanks for the input 🙂

PR #1310, opened 2020-04-01
fix(data-table): change background on hover only for clickable rows

jonnybel:
#### Summary

Currently, when hovering a row of `DataTable`, it will change to a grey color.

This PR changes the row so it only behaves like this if the row itself is clickable -- which means the table has an `onRowClick` prop.


tdeekens:
Let's not use internal ticket numbers on PRs I'd say. Maybe just increment branch numbers could be enough?

jonnybel:
@tdeekens I do it when the ticket exists because it makes the PR become automatically linked in the Jira ticket 😛 

But it's not rly necessary

PR #1312, opened 2020-04-02
fix(data-table): add footer prop

jonnybel:
#### Summary

Adds a `footer` prop to `DataTable` to allow placing content on the `tfoot`  element.

### DIFF
--- a/packages/components/data-table/src/data-table.spec.js
+++ b/packages/components/data-table/src/data-table.spec.js
@@ -63,6 +63,16 @@ describe('DataTable', () => {
     });
   });
 
+  it('should allow rendering a footer', () => {
+    const rendered = render(
+      <DataTable {...baseProps} footer="This is in the footer" />
+    );
+
+    const footerElement = rendered.container.querySelector('tfoot');
+
+    expect(footerElement.textContent).toBe('This is in the footer');
+  });
+
   describe('when setting an action for onRowClick', () => {
     // we will be expecting this mock function to be called with (row: object, rowIndex: number)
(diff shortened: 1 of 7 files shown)


packages/components/data-table/src/data-table.spec.js
```diff
@@ -63,6 +63,16 @@ describe('DataTable', () => {
     });
   });
 
+  it('should allow rendering a footer', () => {
+    const rendered = render(
+      <DataTable {...baseProps} footer="This is in the footer" />
+    );
+
+    const footerElement = rendered.container.querySelector('tfoot');
```
tdeekens, line 71: Would it make sense to give this the `aria-role` `contentinfo` so we can `getByRole('contentinfo')`? Not sure myself I am just wondering how we can allow selecting it niver, semantically if possible.

jonnybel: I researched a bit about this, and I couldn't find usages of `contentinfo` for table footers, only for page footers, as it usually relates to the whole document.
I also couldn't find any role that suits the `tfooter` according to standards.

However, I could add an `aria-label="table-footer"` attribute to the footer and it would make it selectable with the `getByLabelText`.
From what I've seen, this seems a valid usage, do you agree?

I can also add other labels/roles for other elements of the table to make them more easily selectable.

tdeekens: Maybe let's keep it as is for now and keep it in mind. Maybe the footer can also be labelled-by a hidden field?

PR #1314, opened 2020-04-07
fix(data-table): remove onClick from column options

jonnybel:
#### Summary

Removing the `onClick` prop from columns of the `DataTable`.

This was a Design decision, to prevent wrong usages. If a cell needs an action, it should render a button or link.


packages/components/data-table/README.md
```diff
@@ -70,10 +70,9 @@ return <DataTable rows={rows} columns={columns} />;
...
 | `width`                | `string`  | -        | `auto`  | Sets a `width` for this column. Accepts the same values as the ones specified for individual [grid-template-columns](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns). Some common useful examples are `minmax` pairs (e.g. `minmax(200px, 400px)`), combinations of fraction values (`1fr`/`2fr`/etc), or fixed values such as `200px`. By default, the column grows according to the content and respecting the total table available width. |
-| `onClick`              | `func`    | -        | -       | Function to make the cells of this column clickable by the user. Keep in mind that if used on a row that implements `onRowClick`, you might want to set the column property `shouldIgnoreRowClick` to ignore it for this column, otherwise, upon click this function will be called first and then `onRowClick` immediately afterwards.                                                                                                                               |
```
tdeekens, line 73: No more onClick?

packages/components/data-table/README.md
```diff
@@ -70,10 +70,9 @@ return <DataTable rows={rows} columns={columns} />;
...
-| `shouldIgnoreRowClick` | `boolean` | -        | `false` | Set this to `true` to prevent click event propagation for this cell. You might want this if you need the column to have its own `onClick` while the row also has a defined `onRowClick`.                                                                                                                                                                                                                                                                              |
+| `shouldIgnoreRowClick` | `boolean` | -        | `false` | Set this to `true` to prevent click event propagation for this cell. You might want this if you need the column to have its own call-to-action or input while the row also has a defined `onRowClick`.                                                                                                                                                                                                                                                                |
```
tdeekens, line 76: How is this different to a isDisabled?

packages/components/data-table/README.md
jonnybel: Not for individual columns. This was a Design decision, to prevent wrong usages.

packages/components/data-table/README.md
jonnybel: I don't want to convey the wrong idea of disabling the cell altogether (it may contain inputs, buttons, etc, which won't be disabled).
This prop really only exists to ignore the `onRowClick` for specific columns.

tdeekens: Ah. Got it.

jonnybel: Thanks :)

PR #1338, opened 2020-05-11
chore: setup changesets, move to gh actions

emmenko:
https://github.com/atlassian/changesets

TL;DR: changesets are a replacement of lerna changelog, as they provide a more convenient way of writing changelogs incrementally. In fact, the tool is primarily built to work in monorepositories.

Additionally, using the bot and the github action, the workflow of releasing becomes almost fully automated.


.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+        env:
+          # https://github.com/bahmutov/cypress-gh-action-split-install/blob/ca3916d4e7240ebdc337825d2d78eb354855464b/.github/workflows/tests.yml#L14-L18
+          # https://github.com/marketplace/actions/cypress-io#custom-install
+          CYPRESS_INSTALL_BINARY: 0
```
tdeekens, line 58: This is not needed here is it?

emmenko:
> Are the s3r3ts set up already

Yep

image

.github/workflows/release.yml
emmenko: Nope, no Cypress here 🙂

emmenko:
> I am still feeling itchy about having lerna and changesets and yarn workspaces but when changesets maybe improves their canary release story a bit that ahould go away.

Yeah that part should improve. They seem to be pretty active in maintaining the project, so I hope that if we give the proper input (e.g. RFC) we can find a more permanent solution.

emmenko:
PS: there is a [WIP PR](https://github.com/atlassian/changesets/pull/359) that implements a [`pre snapshot`](https://github.com/atlassian/changesets/issues/327#issuecomment-614932833) command, which might do what we need for the `canary` releases.

emmenko:
FYI: I "removed" the integration with CircleCI, we only run on GitHub Actions now.

PR #1340, opened 2020-05-12
chore: init changesets

emmenko:
Leftover of #1338


.changeset/config.json
```diff
@@ -0,0 +1,100 @@
...
+      "@commercetools-uikit/text-field",
+      "@commercetools-uikit/text-input",
+      "@commercetools-uikit/time-field",
+      "@commercetools-uikit/time-input",
+      "@commercetools-uikit/toggle-input",
+      "@commercetools-uikit/tooltip",
+      "@commercetools-uikit/ui-kit",
+      "@commercetools-uikit/utils"
+    ]
+  ],
```
emmenko, line 97: This should keep more or less the same behavior of lerna "locked" mode, where all packages stay more or less at the same version.

If we don't want this behavior we can remove the list.

PR #1341, opened 2020-05-12
fix(data-table): contain table with horizontal scroll when maxWidth is set

jonnybel:
#### Summary

`DataTable`: Fixes a bug where even with `maxWidth` set, the table could still grow out of its container, whereas it should be contained inside with a horizontal scrollbar.


tdeekens:
Don't forget to generate a changseset :)

jonnybel:
I approved the diffs in Percy, as they all showed the expected changes.

.changeset/violet-cougars-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/data-table': patch
+---
+
+fixed `maxWidth` prop to actually contain the table within the width value adding a horizontal scrollbar if necessary
```
emmenko, line 5: 💯

PR #1347, opened 2020-05-19
chore(text): fix wrong util fn causing warnings in dev

jonnybel:
#### Summary

This PR simply fixes a false `Warning` being thrown in development because of a wrong fn name.


emmenko:
No changelog?

jonnybel:
It was a leftover from an earlier refactor that just threw some warnings when running storybook in development, so I thought it wasn't necessary as there are no real changes.

Should I add a changeset for that?

emmenko:
Ah ok, you only need the changelog if you want to release the package.

PR #1349, opened 2020-05-25
feat(src/components/icons): Add AngleUpDown icon to the icons set

mohib0306:
#### Summary

`src/components/icons` => `AngleUpDown` icon was add to the available icons so that it could be used later. i.e. in `DataTable`


jonnybel:
The code seems OK!
Let's wait for the VRTs to confirm if the icon is shown as expected 🙂

PR #1351, opened 2020-05-25
feat(data-table/cell): Improve data table's sortable columns

mohib0306:
#### Summary

- `DataTable` => Sortable columns should be clearly indicated (showing `AngleUpDown` icon in the header cell) so that user can see it without hover on each column and finding it out that way. If a column is focused/active, either `AngleUp` or `AngleDown` icon is shown, depending on its current sort direction.

### DIFF
--- a/packages/components/data-table/src/cell.styles.js
+++ b/packages/components/data-table/src/cell.styles.js
@@ -103,21 +103,21 @@ const getSortableHeaderStyles = (props) => css`
   * THEN AngleUpDown icon is hidden
   * AND AngleUp or AngleDown icon is shown
   */
-  svg:nth-last-of-type(2) {
+  svg[id='nonActiveSortingIcon'] {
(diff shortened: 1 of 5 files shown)


packages/components/data-table/src/cell.styles.js
```diff
@@ -94,15 +94,31 @@ const getSortableHeaderStyles = (props) => css`
...
+  }
   svg:last-of-type {
-    visibility: ${props.isActive ? 'visible' : 'hidden'};
+    display: ${props.isActive ? 'inline-block' : 'none'};
     margin-left: ${vars.spacingS};
   }
+
   :hover,
   :focus {
+    svg:first-of-type {
```
jonnybel, line 117: This one should also be `svg:nth-last-of-type(2)`

jonnybel:
Alright, looks great now 💯

The VRTs diffs were as expected 😎

PR #1353, opened 2020-05-27
feat(data-table): Enable rows to expand and collapse in case they have truncated columns

mohib0306:
#### Summary

- `DataTable` => Allow table users to expand or collapse data per rows when the table content is truncated

### DIFF
--- a/packages/components/data-table/src/cell.js
+++ b/packages/components/data-table/src/cell.js
@@ -5,6 +5,8 @@ import {
   AngleUpIcon,
   AngleDownIcon,
   AngleUpDownIcon,
+  RightTriangleFilledIcon,
+  RightTriangleLinearIcon,
 } from '@commercetools-uikit/icons';
 import {
   BaseCell,
@@ -13,7 +15,7 @@ import {
   CellInner,
   HeaderCellInner,
   SortableHeaderInner,
-  CellCollapseButtonWrapper,
+  RowExpandCollapseButton,
 } from './cell.styles';
 
 const HeaderCell = (props) => {
@@ -81,7 +83,9 @@ const DataCell = (props) => {
     [shouldIgnoreRowClick]
   );
 
-  const Icon = props.isRowCollapsed ? AngleDownIcon : AngleUpIcon;
+  const Icon = props.isRowCollapsed
+    ? RightTriangleFilledIcon
+    : RightTriangleLinearIcon;
   return (
     <BaseCell isTruncated={props.isTruncated}>
       <CellInner
@@ -89,16 +93,15 @@ const DataCell = (props) => {
         onClick={props.shouldIgnoreRowClick ? onClickHandler : undefined}
       />
       {props.shouldRenderCollapseButton && (
-        <CellCollapseButtonWrapper>
-          <Icon
-            id="rowExpandCollapseButton"
-            size="medium"
-            onClick={(event) => {
-              props.handleRowCollapseClick();
-              event.stopPropagation();
-            }}
-          />
-        </CellCollapseButtonWrapper>
+        <RowExpandCollapseButton
+          icon={<Icon size="small" />}
+          id="rowExpandCollapseButton"
+          label="Expand/Collapse"
+          onClick={(event) => {
+            props.handleRowCollapseClick();
+            event.stopPropagation();
+          }}
+        />
       )}
     </BaseCell>
   );
--- a/packages/components/data-table/src/cell.js
+++ b/packages/components/data-table/src/cell.js
@@ -95,7 +95,6 @@ const DataCell = (props) => {
       {props.shouldRenderCollapseButton && (
         <RowExpandCollapseButton
           icon={<Icon size="small" />}
-          id="rowExpandCollapseButton"
           isRowCollapsed={props.isRowCollapsed}
           label="Expand/Collapse Row"
           onClick={(event) => {
@@ -114,13 +113,13 @@ DataCell.propTypes = {
   isCondensed: PropTypes.bool,
   isTruncated: PropTypes.bool,
   shouldIgnoreRowClick: PropTypes.bool,
-  handleRowCollapseClick: requiredIf(
-    PropTypes.func,
-    (props) => props.isTruncated
-  ),
   shouldRenderCollapseButton: requiredIf(
     PropTypes.bool,
     (props) => props.isTruncated
+  ),
+  handleRowCollapseClick: requiredIf(
+    PropTypes.func,
+    (props) => props.shouldRenderCollapseButton
   ),
   isRowCollapsed: requiredIf(PropTypes.bool, (props) => props.isTruncated),
 };
--- a/packages/components/data-table/src/cell.js
+++ b/packages/components/data-table/src/cell.js
@@ -113,15 +113,15 @@ DataCell.propTypes = {
   isCondensed: PropTypes.bool,
   isTruncated: PropTypes.bool,
   shouldIgnoreRowClick: PropTypes.bool,
-  shouldRenderCollapseButton: requiredIf(
-    PropTypes.bool,
-    (props) => props.isTruncated
-  ),
+  shouldRenderCollapseButton: PropTypes.bool.isRequired,
   handleRowCollapseClick: requiredIf(
     PropTypes.func,
     (props) => props.shouldRenderCollapseButton
   ),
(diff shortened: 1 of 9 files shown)


packages/components/data-table/src/data-table.story.js
```diff
@@ -23,7 +23,7 @@ import NumberInput from '../../inputs/number-input';
 const items = [
   {
     id: '5e188c29791747d9c54250e2',
-    name: 'Morgan Bean',
+    name: 'Barry Allen - aka Flash - aka Fastest Man Alive',
```
jonnybel, line 26: 😆 

(I have no objections)

packages/components/data-table/src/cell.styles.js
```diff
@@ -172,11 +173,18 @@ const SortableHeaderInner = styled.button`
   ${getSortableHeaderStyles}
 `;
 
+const CellCollapseButtonWrapper = styled.div`
```
jonnybel, line 176: I think we should try to make this control an accessible `button` element, and also with a hover/focus effect (like our [SecondaryIconButton](https://uikit.commercetools.com/?path=/story/components-buttons--secondaryiconbutton)), so that the user has a better perception of whether he's clicking the row or the little button, and avoiding accidental clicks on one or the other.

packages/components/data-table/src/data-table.story.js
mohib0306: 😆 Just wanted to show the UX/UI people how would a long name be affected by the truncation feature

packages/components/data-table/src/data-table.spec.js
```diff
@@ -44,6 +46,19 @@ describe('DataTable', () => {
     expect(rendered.queryByText('2-woman')).not.toBeInTheDocument();
   });
 
+  it('should render only one expand-collapse button per row', () => {
+    const rendered = render(<DataTable {...baseProps} />);
+    /**
+     * Even though two columns are marked as truncatable in each row, but only button should be shown
+     * in each row that will control the expand and collapse of the whole row.
```
jonnybel, line 53: Nitpick:
```suggestion
     * Even though two columns are marked as truncatable in each row, only one button should be shown
     * per row, which controls the expand and collapse of its whole row.
```

packages/components/data-table/src/data-table.styles.js
```diff
@@ -42,6 +42,11 @@ const Body = styled.tbody`
 const Row = styled.tr`
   display: contents;
   ${getClickableRowStyle}
+  :hover {
+    button[id='rowExpandCollapseButton'] {
```
jonnybel, line 46: ids to the rescue once more! 😄

packages/components/data-table/src/data-table.spec.js
```diff
@@ -12,10 +12,12 @@ const testColumns = [
   {
     key: 'title',
     label: 'Title',
+    isTruncated: true,
   },
   {
     key: 'year',
     label: 'Year',
+    isTruncated: true,
```
jonnybel, line 20: Do you think we should keep this change the story, or it was just for testing? In practice, it almost makes no difference anyway :)

packages/components/data-table/src/cell.js
```diff
@@ -97,6 +114,15 @@ DataCell.propTypes = {
   isCondensed: PropTypes.bool,
   isTruncated: PropTypes.bool,
   shouldIgnoreRowClick: PropTypes.bool,
+  handleRowCollapseClick: requiredIf(
+    PropTypes.func,
+    (props) => props.isTruncated
```
jonnybel, line 119: Shouldn't these be _requiredIf_ `shouldRenderCollapseButton` is true, and not the `isTruncated` prop?

jonnybel, line 119: Shouldn't these be _requiredIf_ `shouldRenderCollapseButton` is true, and not the `isTruncated` prop?

packages/components/data-table/src/data-table.styles.js
jonnybel: FYI, I think in this case you could also import the `RowExpandCollapseButton`component defined in `cell.styles.js` and use it as the selector, [as seen here](https://emotion.sh/docs/styled#targeting-another-emotion-component). It would allow us to avoid setting an id, except you are also using the id for the unit test, so we can keep the id.

This comment is just for awareness and in case you wanna try it as an experiment for your CSS-in-JS skills 😄

packages/components/data-table/src/data-table.spec.js
mohib0306: These changes were on purpose to test the expand/collapse button rendering. We could create another set of test data, but to me that looked a bit redundant. Therefore I updated the `default` test data to check the rendering of the expand/collapse buttons.

packages/components/data-table/src/data-table.styles.js
mohib0306: Using the approach from `emotions.sh` (the one that you mentioned) make much more sense. Also since the `RowExpandCollapseButton` has a label and we can use the `queryAllByLabelText` in the tests, to find all the expand/collapse CTAs and count them. 

These changes will make the usage of `id` on `RowExpandCollapseButton` absolute and therefore I will remove it in the next commit.

packages/components/data-table/src/data-table.spec.js
mohib0306: Good catch

packages/components/data-table/src/cell.js
mohib0306: yes

packages/components/data-table/src/data-table.styles.js
jonnybel: Cool!

packages/components/data-table/src/data-table.spec.js
jonnybel: Oh, nevermind, I thought I was reading the `.story.js` file 🤦‍♂️ 
All good!

mohib0306:
Rebase the branch to master and resolved conflicts

packages/components/data-table/src/cell.js
```diff
@@ -114,14 +113,14 @@ DataCell.propTypes = {
...
-  ),
   shouldRenderCollapseButton: requiredIf(
     PropTypes.bool,
     (props) => props.isTruncated
   ),
+  handleRowCollapseClick: requiredIf(
+    PropTypes.func,
+    (props) => props.shouldRenderCollapseButton
+  ),
   isRowCollapsed: requiredIf(PropTypes.bool, (props) => props.isTruncated),
```
jonnybel, line 124: Here it should also be dependent on the `shouldRenderCollapseButton` prop and not `isTruncated`, right?
```suggestion
  isRowCollapsed: requiredIf(PropTypes.bool, (props) => props. shouldRenderCollapseButton),
```

packages/components/data-table/src/cell.js
```diff
@@ -114,14 +113,14 @@ DataCell.propTypes = {
   isCondensed: PropTypes.bool,
   isTruncated: PropTypes.bool,
   shouldIgnoreRowClick: PropTypes.bool,
-  handleRowCollapseClick: requiredIf(
-    PropTypes.func,
-    (props) => props.isTruncated
-  ),
   shouldRenderCollapseButton: requiredIf(
```
jonnybel, line 116: I wouldn't make this one required. It's only `true` when a DataCell is in the last column (and only in combination with `isTruncated`), but otherwise I think it can be omitted safely.

packages/components/data-table/src/cell.js
```diff
@@ -97,6 +113,15 @@ DataCell.propTypes = {
   isCondensed: PropTypes.bool,
   isTruncated: PropTypes.bool,
   shouldIgnoreRowClick: PropTypes.bool,
+  shouldRenderCollapseButton: requiredIf(
+    PropTypes.bool,
+    (props) => props.isTruncated
+  ),
```
mohib0306, line 119:
```suggestion
  shouldRenderCollapseButton: PropTypes.bool.isRequired
```

mohib0306: we can make it   `shouldRenderCollapseButton: PropTypes.bool.required` because even it is not the last column it will have a value (`false`)

jonnybel: Fine by me 👍

PR #1357, opened 2020-05-28
feat(label): add intlMessage prop

jonnybel:
#### Summary

This PR adds a `intlMessage` prop to `Label` component, which can replace the `children` prop for translated text, in the same fashion as the `Text` components.


packages/components/label/src/label.js
```diff
@@ -17,10 +23,18 @@ Label.displayName = 'Label';
...
   isBold: PropTypes.bool,
   isRequiredIndicatorVisible: PropTypes.bool,
   htmlFor: PropTypes.string,
+  intlMessage: requiredIf(
+    PropTypes.shape({
+      id: PropTypes.string.isRequired,
+      description: PropTypes.string,
+      defaultMessage: PropTypes.string.isRequired,
+    }),
+    (props) => !React.Children.count(props.children)
```
tdeekens, line 35: Nifty. I wonder how TypeScript would do this honestly right now.

emmenko: We can't use `react-required-if` with TypeScript. We would need to use `invariant` or something on render.

As for the types, it would be something like this: https://github.com/commercetools/merchant-center-application-kit/blob/c4f556dbf269a6cc35d74ba2c3cd91d24994f92f/packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx#L11-L16

PR #1363, opened 2020-05-28
feat(data-table): enable column resizing by user mouse drag

jonnybel:
#### Summary

`DataTable`: enables a new feature that allows users to adjust column sizes by dragging the Header edge with their mouse! 🎉 

Kapture 2020-05-28 at 18 42 55

#### Description

To achieve this behaviour, the column widths, which are set by the `grid-template-columns` style rule, is set inline when the user starts dragging a column edge.
This allows us to measure the distance (in the horizontal axis) travelled by the mouse while dragging, calculate the new width every time the screen gets updated as long as the user is still holding the mouse button, and then write it to the new `grid-template-columns` value.

We don't want to perform a re-render for each fraction of a second that the screen gets updated, as it would hurt performance a lot and we don't need it, since we only need to change CSS, which is why these calculations and settings are being done outside of the normal React component life-cycle, and directly with Javascript references and inline styles.

**Note:**
When the user starts manually resizing a column, the table columns stop becoming responsive (resizing automatically), and acquire their current column width as a fixed value. This is measured directly on the layout as soon as the user starts dragging.
This is intended, as we couldn't have both responsiveness and manual resizing at the same time, for they are in conflict with each other -- a user would be dragging a column to resize it while other columns would automatically and unexpectedly change their widths at the same time, resulting in bad UX.
 

#### Follow-up:
*  Add a method to return the current sizes of the columns, so that, for instance, they can be saved as a user setting.


packages/components/data-table/src/cell.styles.js
```diff
@@ -130,30 +133,49 @@ const BaseHeaderCell = styled.th`
...
 
   /* remove user-agent styles */
   padding: 0;
   font-weight: normal;
 
-  /* adds borders between header cells */
-  :not(:last-of-type) {
-    border-right: 1px solid ${vars.colorNeutral90};
-  }
+  /* right border that doesn't count towards the column width */
```
jonnybel, line 145: Because otherwise, the `clientWidth` we get from measuring the columns will be one pixel smaller than the actual `width` it was supposed to have, because the border takes that space.

Visually, there's no difference between the border and this "fake border" box-shadow.

packages/components/data-table/src/cell.styles.js
```diff
@@ -170,6 +192,7 @@ const SortableHeaderInner = styled.button`
   ${getCellInnerStyles}
   ${getButtonStyle}
   ${getSortableHeaderStyles}
+  ${(props) => (props.shouldWrap ? '' : 'white-space: nowrap')}
```
jonnybel, line 195: This is unrelated to the PR but was missing for non-sortable header cells.

packages/components/data-table/src/column-size-utils.js
```diff
@@ -0,0 +1,35 @@
...
+  return minSize > newSize ? minSize : newSize;
+};
+
+// returns an array of column widths
+const getColumnWidthsArray = (columns) =>
+  columns.reduce((widths, current) => [...widths, current.width], []);
+
+const setColumnWidth = (columns, position, value) => {
+  // eslint-disable-next-line no-param-reassign
+  columns[position] = value;
```
jonnybel, line 21: I think this is one of those cases where mutating the value directly is advantageous because it's for adjusting the layout quickly on-the-fly, before any re-renders or reconciliations.

packages/components/data-table/src/data-table.js
```diff
@@ -1,75 +1,122 @@
...
+
+  return (
+    <TableGrid
+      ref={tableRef}
+      {...filterDataAttributes(props)}
+      columns={props.columns}
+      maxWidth={props.maxWidth}
+      maxHeight={props.maxHeight}
+      style={{
+        gridTemplateColumns: gridTemplateColumnsStyle,
```
jonnybel, line 48: Inline style because we're going to be changing the values on-the-fly, without re-renders, so its more performant 🙂

packages/components/data-table/src/cell.styles.js
```diff
@@ -99,18 +99,21 @@ const getSortableHeaderStyles = (props) => css`
...
-  * 
+  *
   * GIVEN column is sortable and foucsed
   * THEN AngleUpDown icon is hidden
   * AND AngleUp or AngleDown icon is shown
   */
+  svg[id='nonActiveSortingIcon'],
+  svg[id='activeSortingIcon'] {
+    margin-left: ${vars.spacingS};
+    flex-shrink: 0;
```
jonnybel, line 110: This was the only added this line in these svg styles, to make sure the icon doesn't shrink when the column is narrow.

jonnybel:
Added a few paragraphs to the description to give a better picture of the approach.

packages/components/data-table/src/cell.js
```diff
@@ -17,44 +22,138 @@ import {
...
+const HeaderCellWrapper = (props) => {
+  const [colResizingState, setColResizingState] = React.useState({
+    isResizing: false,
+    initialColWidth: undefined,
+    initialMousePosition: undefined,
+  });
+  const tableRef = props.tableRef;
+  const headerRef = React.useRef(null);
+
+  const onStartResizing = (e) => {
```
mohib0306, line 36: It is not a required change, but would be more clear if we could name the variable `event`

packages/components/data-table/src/column-size-utils.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return columns;
+};
+
+const getGridTemplateColumnsStyle = (columns) =>
+  `${columns.map((width) => `${width}px`).join(' ')}`;
```
tdeekens, line 27: Isn't this the same? Late day brain but... 🤔 

```suggestion
  `${columns.join(('px ')`;
```

packages/components/data-table/src/data-table.js
```diff
@@ -1,57 +1,102 @@
...
-    </Header>
-    <Body>
-      {props.rows.map((row, rowIndex) => (
-        <DataRow {...props} row={row} key={rowIndex} rowIndex={rowIndex} />
-      ))}
-    </Body>
-    {props.footer && (
-      <Footer>
+const DataTable = (props) => {
+  const [columnSizing, setcolumnSizing] = React.useState({
```
tdeekens, line 13:
```suggestion
  const [columnSizing, setColumnSizing] = React.useState({
```

packages/components/data-table/src/data-table.js
```diff
@@ -1,57 +1,102 @@
...
+      const renderedLayoutSizes = [];
+      tableRef.current.querySelectorAll('th').forEach((header, index) => {
+        renderedLayoutSizes[index] = header.clientWidth;
+      });
+      setcolumnSizing({
+        isManualControlled: true,
+        sizes: renderedLayoutSizes,
+      });
+    }
+  };
```
tdeekens, line 35: My main wonder here is if we should hoist state management into a `useReducer` and put it onto the context if needed. It's a separate, additional suggestion to the other suggestion of using a reducer.

jonnybel:
I've moved most of the resizing logic to the reducer and the reducer itself is shared via React Context.

I've also improved the resizing indicator, fixing its bugs and improving its draggability.

packages/components/data-table/src/cell.js
```diff
@@ -18,46 +18,98 @@ import {
...
+  };
+
+  const onDrag = (event) =>
+    columnResizingReducer.onDragResizing(event, headerRef.current.cellIndex);
+
+  const onDragEnd = () => {
+    columnResizingReducer.finishResizing();
+
+    window.removeEventListener('mousemove', onDrag);
+    window.removeEventListener('mouseup', onDragEnd);
```
tdeekens, line 40: Maybe for the future we can see how to wrap this into React hooks. Not sure myself now but this adding and removing could maybe be.

PR #1368, opened 2020-06-02
feat(components/icons): Update grid icon

mohib0306:
#### Summary

- `icons`: => update grid icon


jonnybel:
Thanks.

Don't forget to add a changeset :)

mohib0306:
I think it does not required changeset release. Right?

jonnybel:
We should add one, just in case someone gets confused about the icon change, it should be on the changelog. We can consider this `patch`, I think.

PR #1369, opened 2020-06-04
fix(tooltip): Hide tooltip on mouseleave event for disabled inputs/buttons

mohib0306:
#### Summary

-`tooltip` => Hide tooltip on mouseleave event for disabled inputs/buttons

In-order to proper catch the `mouseleave` events, the `pointer-events` should be `none`.


packages/components/tooltip/src/tooltip.js
```diff
@@ -2,22 +2,17 @@
...
-import styled from '@emotion/styled';
 import isNil from 'lodash/isNil';
 import usePopper from 'use-popper';
 import { useFieldId, useToggleState } from '@commercetools-uikit/hooks';
 import {
   createSequentialId,
   SafeHTMLElement,
 } from '@commercetools-uikit/utils';
-import { Body, getBodyStyles } from './tooltip.styles';
+import { Body, getBodyStyles, Wrapper } from './tooltip.styles';
```
tdeekens, line 12: Nit

```suggestion
import { Body, Wrapper, getBodyStyles } from './tooltip.styles';
```

jonnybel:
I could still reproduce the bug on the test deployment 😕

PR #1376, opened 2020-06-11
fix(select-fields): Fix RTL tests

mohib0306:
#### Summary

- **_All select fields_** => The select fields uses the `title` prop to render the labels of a select field. Therefore, the tests should use that for testing the rendered label instead of adding another separate label on top.

### DIFF
--- a/.changeset/fair-icons-talk.md
+++ b/.changeset/fair-icons-talk.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-uikit/async-creatable-select-field': patch
+'@commercetools-uikit/async-select-field': patch
+'@commercetools-uikit/creatable-select-field': patch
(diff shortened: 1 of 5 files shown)


mohib0306:
Regarding the issue that the `id` is not propagated down to the `input` itself, according to the documentation of `react-select` (https://react-select.com/props => checkout id prop), the ID prop will point to the select container holding the select input component. However, even though it does not pass down to the input itself, `getByLabelText` from the `react-testing-library` still query the correct select input component.

There also `inputId` prop form the `react-select` that then passes down the `id` prop down to the `input` element but it says that is for the search input. Also IMO it might not be necessary to use that, since the `*ByLabelText` works.

tdeekens:
Thanks @mohib0306 makes total sense. Thanks for looking into this! ❤️

PR #1379, opened 2020-06-11
fix(data-table): add data-testid attributes to headers and cells

jonnybel:
#### Summary

This PR adds `data-testid` attributes to the headers and cells of the table, to improve their testability.

I used the same sequential id that was being used in the "old" table, but with the `data-testid` attribute instead of just `data-test`. This means the migration steps for the tests that query for these test ids will be minimal.


jonnybel:
Same tbh. I only added this to make it easier to adapt the pre-existing tests, otherwise we should avoid using them on new tests.

PR #1380, opened 2020-06-12
fix(data-table): Improve DataTable row's expand/collapse accessibility

mohib0306:
#### Summary

- `DataTable` => Improve accessibility of row's expand/collapse functionality. The implementation is based on the suggestions from https://a11yproject.com/posts/how-to-hide-content.


packages/components/data-table/src/data-table.styles.js
```diff
@@ -43,9 +43,9 @@ const Body = styled.tbody`
 const Row = styled.tr`
   display: contents;
   ${getClickableRowStyle}
-  :hover {
+  :hover, :focus-within {
```
jonnybel, line 46: I think this is the only line we need to change.

The "visibility trick" isn't needed because it's usually for content that remains hidden but we want to show to screen readers. In this case, the component is machine-readable when the row is focused, and that is adequate.
In fact, we shouldn't even bother about keeping this button machine-readable because someone who uses screen-readers won't even see the content as truncated: the truncated part still remains always fully (machine-)readable, and only visually hidden 😄 

Therefore, we only need to worry about making it focusable by the keyboard, and the `:focus-within` selector does that 🙂

mohib0306: Let me double check.

PR #1386, opened 2020-06-16
feat(data-table): Sticky footer 🩹

jonnybel:
#### Summary

This PR makes the `DataTable` footer become sticky on the bottom when the inner part of the table is scrollable. Similar to the behaviour of the header.

Kapture 2020-06-16 at 18 18 46


mohib0306:
Looks nice.
Just one suggestion. Wouldn't it be nice to have a flag for it with the default as false. Similar as `disableHeaderStickiness` prop

jonnybel:
@mohib0306 Agreed!

PR #1387, opened 2020-06-16
fix(data-table): column resizer taking extra space

jonnybel:
#### Summary

This PR fixes a small issue with the DataTable column resizer taking too much space on the edge of the table:
Screenshot 2020-06-16 at 19 02 21
This bug makes the table have a horizontal scrollbar because of that little extra space 😠

### DIFF
--- a/packages/components/data-table/src/column-resizer.js
+++ b/packages/components/data-table/src/column-resizer.js
@@ -3,12 +3,9 @@ import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 
-const resizerWidth = 3;
-const resizerHoverArea = 6;
-
 const ResizerIndicator = styled.div`
   height: 100%;
-  width: ${resizerWidth}px;
+  width: 3px;
   background: ${vars.colorInfo};
   visibility: hidden;
   cursor: col-resize;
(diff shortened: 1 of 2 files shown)


packages/components/data-table/src/column-resizer.js
```diff
@@ -20,11 +17,13 @@ const ResizerIndicator = styled.div`
 `;
 
 const DraggableArea = styled.div`
+  display: flex;
+  justify-content: flex-end;
   height: 100%;
   position: absolute;
   top: 0;
-  width: ${resizerHoverArea}px;
-  right: -${resizerHoverArea - resizerWidth}px;
```
jonnybel, line 27: Instead of keeping the draggable area centered on the edge, overlapping the right side for a few pixels, I moved it to stay on the inner edge of the header cell.

packages/components/data-table/src/column-resizer.js
```diff
@@ -35,14 +34,13 @@ const DraggableArea = styled.div`
...
 
   /* when dragging, we increase the area of the invisible draggable zone
-  to prevent accidentally hovering other elements which may cause flickers */
+  to prevent accidentally hovering other elements which causes flickering */
   &:active {
-    display: flex;
-    justify-content: center;
-    width: 50px;
-    right: -23px;
+    width: 20px;
```
jonnybel, line 39: The area didn't have to become so large and with rounded values this is easier to understand.

PR #1388, opened 2020-06-17
feat(data-table): add new prop verticalCellAlignment

jonnybel:
#### Summary

`DataTable`: This PR adds a prop to control the vertical alignment of the content of all data-cells (thus, excluding Headers).

Because there was already a prop called `cellAlignment`, it was renamed to `horizontalCellAlignment`, and the new prop is called `verticalCellAlignment`.

I'm not considering this a breaking change given the component's  "Beta" status (which is soon to end, as we're done with most of the new features).


packages/components/data-table/src/cell.js
```diff
@@ -196,15 +197,18 @@ DataCell.defaultProps = {
...
+    >
       {props.children}
     </CellInner>
   </BaseFooterCell>
 );
 FooterCell.displayName = 'FooterCell';
 FooterCell.propTypes = {
   children: PropTypes.node.isRequired,
-  alignment: PropTypes.oneOf(['left', 'center', 'right']),
+  horizontalCellAlignment: PropTypes.oneOf(['left', 'center', 'right']),
```
tdeekens, line 211: Are or shoudl we defaulting this?

jonnybel: Yeah, forgot to add the default on the Footer.

packages/components/data-table/src/cell.js
```diff
@@ -212,5 +213,8 @@ FooterCell.propTypes = {
   isCondensed: PropTypes.bool,
   numberOfColumns: PropTypes.number.isRequired,
 };
+FooterCell.propTypes = {
```
tdeekens, line 216: Get a ☕ 😄 

```suggestion
FooterCell.defaultProps = {
```

jonnybel: 🙈

jonnybel: `fix: remove stupidity`

PR #1389, opened 2020-06-17
fix: to add `eslint-plugin-jest-dom`

tdeekens:
#### Summary

This pull request adds the `eslint-plugin-jest-dom`.

#### Description

The `eslint-plugin-jest-dom` helps to verify that correct attributes are set on the correct elements. For instance a `<a />` not being `disabled` through the prop but an "nulled" `href`. 

Adding the plugin required to move enzyme matchers into the two specs of the legacy table which use shallow rendering at the moment.

### DIFF
--- a/jest.eslint.config.js
+++ b/jest.eslint.config.js
@@ -2,6 +2,6 @@ module.exports = {
   runner: 'jest-runner-eslint',
   displayName: 'eslint',
   testMatch: ['<rootDir>/**/*.js'],
-  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist/es;'],
+  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist;'],
   watchPlugins: ['jest-watch-typeahead/filename'],
 };
--- a/jest.eslint.config.js
+++ b/jest.eslint.config.js
@@ -2,6 +2,6 @@ module.exports = {
   runner: 'jest-runner-eslint',
   displayName: 'eslint',
   testMatch: ['<rootDir>/**/*.js'],
-  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist/'],
+  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist/es;'],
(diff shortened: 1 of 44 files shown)


tdeekens:
Feel free to add a changeset. Not sure we have to as it's a non user facing change and consumers might not mind how we test.

src/components/inputs/localized-rich-text-input/localized-rich-text-input.spec.js
```diff
@@ -75,9 +75,11 @@ describe('LocalizedRichTextInput', () => {
           />
         );
         getByLabelText(/show all languages/i).click();
+        // eslint-disable-next-line jest-dom/prefer-enabled-disabled
```
tdeekens, line 78: CleanShot 2020-06-17 at 13 40 45

packages/components/table/src/base-table.spec.js
```diff
@@ -1,3 +1,4 @@
+import 'jest-enzyme';
```
tdeekens, line 1: CleanShot 2020-06-17 at 13 35 35

scripts/setup-test-framework.js
```diff
@@ -19,15 +18,6 @@ import * as commerceToolsEnzymeMatchers from '@commercetools/jest-enzyme-matcher
 
 Enzyme.configure({ adapter: new Adapter(), disableLifecycleMethods: true });
 
-expect.extend({
```
tdeekens, line 22: Not used at all.

jest.eslint.config.js
```diff
@@ -2,6 +2,6 @@ module.exports = {
   runner: 'jest-runner-eslint',
   displayName: 'eslint',
   testMatch: ['<rootDir>/**/*.js'],
-  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist/'],
+  testPathIgnorePatterns: ['/node_modules/', '.public/', 'public/', 'dist;'],
```
mohib0306, line 5: is this on purpose?

tdeekens: Nope. It's my rule of 👍 and life philosophy: no PR without a typo by Tobi. 3c869cb

PR #1391, opened 2020-06-17
rich-text-inputs: Fix not-allowed cursor rendering on disabled rich text inputs

mohib0306:
#### Summary

- `localized-rich-text-input` & `rich-text-input` => The rich text input's editor components were wrapped in a separate `div` to enable different cursor rendering when inputs are disabled. The reason it was not rendering before because `pointer-events` are disabled on underlying component used from the `internals` (`RichTextBody` => `Container`)


src/components/inputs/rich-text-input/editor.styles.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable import/prefer-default-export */
+import styled from '@emotion/styled';
+
+const EditorWrapper = styled.div`
+  cursor: ${(props) =>
+    props.isDisabled || props.isReadOnly ? 'not-allowed' : 'inherit'};
+`;
+const CollapseExpandButtonWrapper = styled.label`
```
jonnybel, line 8: I couldn't follow why is this a `label` 🤔 Shouldn't it just be a `div`?
(In fact, I wonder if this wrapper is even needed if we change the parent Stack to `align-items: flex-end` and then add `align-self: stretch` to the EditorWrapper)

Keep in mind that the whole `label` element is clickable (because of the button inside it), which makes the whole block below the editor clickable when it should just be the button itself (because it's having the _stretched_ behaviour applied from the parent Stack).

image

mohib0306: Good point, actually it was from the old code. I will investigate it

mohib0306: My bad. It was not from old code. I made mistake while copying over the styles. It should have been `div`. I also tried the suggestion that you made and it works and is also clean since the `Stack` component already expose `alignItems` prop and that can be used. I will push the changes in next commit

tdeekens:
Nice. No conflicts from "my" work. Was worried I'd create some for you here.

PR #1392, opened 2020-06-17
fix: not pass `to` prop to `SecondaryButton` when disabled

tdeekens:
#### Summary

This pull request overwrites the `to` prop which is being passed to the `as`-component when using a `SecondaryButton`.

#### Description

Given a `SecondaryButton` is used with a `RouterLink` it would render a `<a />`. The `<a />` receives `aria-disabled` etc but should also not receive a `href` to be the full path.

### DIFF
--- a/packages/components/buttons/secondary-button/src/secondary-button.js
+++ b/packages/components/buttons/secondary-button/src/secondary-button.js
@@ -34,7 +34,8 @@ export const SecondaryButton = (props) => {
   const buttonAttributes = {
     'data-track-component': 'SecondaryButton',
     ...filterInvalidAttributes(props),
-    ...(shouldUseLinkTag ? { to: props.linkTo } : {}),
+    ...(shouldUseLinkTag && { to: props.linkTo }),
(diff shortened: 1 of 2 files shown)


tdeekens:
@jonnybel this would be an attempt. Oddly passing `#` as `to` the `a` receives an `href='/'`.

jonnybel:
Meh, is there no way to _nullify_ the `href` attribute? Then we should prevent the navigation through the `onClick`, with `e.preventDefault();`

Ideally, we should actually just render plain text instead of an anchor when a link is supposed to be disabled... But that's probably not doable right now without breaking a bunch of things.

tdeekens:
I think we're testing and building something we shouldn't support after all. The fact that we make it work internally and have to should not affect the ideals of our component library here and what parts of the design systems are meant to be used for. Closing as a result. Sorry for the disturbance :)

jonnybel:
No disturbance at all! Thanks for making us reflect on this :)

PR #1393, opened 2020-06-17
feat: add intlMessage to notification

Rombelirk:
#### Summary

This PR adds a intlMessage prop to ContentNotification component, which can replace the children prop for translated text, in the same fashion as the Text components.


Rombelirk:
> Thanks for contributing 😄
> 
> Could you add a changeset to the PR?
> Run `yarn changeset`, consider this a `minor` (tho I don't mind if we consider this a `patch` either), and add a simple message, similar to the commit message, that will appear in the changelog 🙂

Already done :)

jonnybel:
Oh, and seems that we need to add the new used dependencies to the `notifications` `package.json`: (`react-intl` and `required-if`)

PR #1396, opened 2020-06-18
fix: storybook data-table example

jonnybel:
#### Summary

This PR fixes a bug in the storybook DataTable example:
When you set a value for the `align` option of a column, you can't change it back to the default. It keeps the previously selected option.

No changeset as this is just a bug in a storybook example.

### DIFF
--- a/packages/components/data-table/src/data-table.story.js
+++ b/packages/components/data-table/src/data-table.story.js
@@ -204,7 +204,7 @@ const ColumnConfigForm = (props) => {
       const updatedColumn = {
         ...props.column,
         width: values.width,
-        align: values.align,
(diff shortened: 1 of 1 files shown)


packages/components/data-table/src/data-table.story.js
```diff
@@ -238,7 +238,7 @@ const ColumnConfigForm = (props) => {
               onChange={formik.handleChange}
               value={formik.values.align}
             >
-              <option value={undefined}>default</option>
```
jonnybel, line 241: I was dumb and didn't realize that setting `value` to undefined is the same as it not being there, so the passed value would be the string `default`, which isn't a valid value for alignment and wouldn't do anything.

PR #1403, opened 2020-06-30
fix(data-table): clip cells when table is resized

jonnybel:
#### Summary

There are a few edge cases when a DataTable is resized that cause content to overflow out of their cell. This is more noticeable on the last column of a table and especially if the background isn't white.
Kapture 2020-06-22 at 17 37 07

This PR addresses this problem by making sure cell content is clipped when a column may be narrower than its content, and the table edge (seen by its background color) is moved along with the edge of the last column:
Kapture 2020-06-30 at 11 38 56

### DIFF
--- a/packages/components/data-table/src/cell.styles.js
+++ b/packages/components/data-table/src/cell.styles.js
@@ -148,6 +148,30 @@ const getSortableHeaderStyles = (props) => css`
     }
   }
 `;
+
+const CellInner = styled.div`
+  box-sizing: border-box;
+  flex: 1;
+
+  ${(props) => getPaddingStyle(props, false)}
+  ${getCellInnerStyles}
+  ${(props) => (props.shouldClipContent ? 'overflow: hidden;' : '')}
+`;
+
+const HeaderCellInner = styled.div`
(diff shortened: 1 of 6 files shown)


packages/components/data-table/src/cell.styles.js
```diff
@@ -169,10 +193,11 @@ const BaseHeaderCell = styled.th`
   :hover,
   :active {
     z-index: 2;
+  }
 
-    & > * {
-      overflow: hidden;
-    }
+  ${HeaderCellInner},
```
jonnybel, line 198: These selectors are more accurate.

packages/components/data-table/src/cell.styles.js
```diff
@@ -148,6 +148,30 @@ const getSortableHeaderStyles = (props) => css`
     }
   }
 `;
+
```
jonnybel, line 151: Just moved this whole block up so that these components can be used as selectors below.

packages/components/data-table/src/cell.styles.js
tdeekens: 💯

PR #1404, opened 2020-07-01
Chore: Move `fields` from `src/components` to `packages/components`

mohib0306:
#### Summary
- Move `fields` from `src/components/fields` to `packages/components/fields`


packages/components/fields/async-creatable-select-field/package.json
```diff
@@ -22,6 +22,7 @@
     "@commercetools-uikit/design-system": "10.18.4",
     "@commercetools-uikit/field-errors": "10.22.1",
     "@commercetools-uikit/field-label": "10.22.1",
+    "@commercetools-uikit/icons": "10.22.0",
```
jonnybel, line 25: Since the `icons` are only used by the Storybook example (correct me if I'm wrong), I think we can remove them of the dependencies of the package and import locally (with relative path) in the story.

mohib0306: will do that

mohib0306:
Just one question: Wouldn't it better to have a PR for each component, rather than having one PR for all the fields? Reviewing smaller PRs will be easy

jonnybel:
> Just one question: Wouldn't it better to have a PR for each component, rather than having one PR for all the fields? Reviewing smaller PRs will be easy

Sure, we can. But because there's dozens of similar fields and inputs, and since refactoring them is more or less the same, we grouped components together in some PRs instead of opening one for each, cuz it would become quite boring 😆 

I don't mind if you work on some individually at least for now 🙂

mohib0306:
@jonnybel All the fields are moved to `packages/components/fields`. May you can have one last look at the PR. Thanks

PR #1406, opened 2020-07-01
Adding isRowCollapsed to call of column renderItem and table itemRenderer

NickDevG:
#### Summary
Adding `isRowCollapsed` to call of column `renderItem` and table `itemRenderer`

#### Description
Allowing the rendering functions to differentiate when the row is collapsed or not, allowing to set a different value in each case resulting in a customize truncation.


packages/components/data-table/src/data-table.story.js
```diff
@@ -245,17 +245,6 @@ const ColumnConfigForm = (props) => {
             </select>
           </label>
         </div>
-        <div>
```
jonnybel, line 248: Could you undo this change? Seems to be unnecessary.

PR #1408, opened 2020-07-02
feat(droppable-panel): adding droppable panel

NickDevG:
#### Summary
Adding the droppable panel in order to be used in the column manager

#### Description
Adding the component on separate PR in order to be easier to review. Next PR will include the `<ManagedDataTable>` including the Display Settings and the Column Manager.


tdeekens:
Thanks for opening! Just to bring up the thought: how much is this ui-kit and not application-components in app-kit?

NickDevG:
> Thanks for opening! Just to bring up the thought: how much is this ui-kit and not application-components in app-kit?

Hi @tdeekens! We are adding a component that wraps the new `DataTable` adding a table settings dropdown that allows to use a display settings (wrapping text on all cells, condensed style over the table...)  and the column manager. As you know the column manager uses the droppable panel. That's why we need to have it in the UIKit even though we are not going to expose it.

Does this makes sense?

packages/components/droppable-panel/src/droppable-panel.js
```diff
@@ -0,0 +1,75 @@
...
+                          ])
+                      : undefined
+                  }
+                  isDisabled={props.isDisabled}
+                />
+              ))}
+              {props.isSearchable && (
+                <Spacings.Inset scale="xs">
+                  <Text.Detail tone="secondary">
+                    {formatMessage(messages.useInputSearch)}
```
jonnybel, line 47: Does it make sense to include this here as a static message? The message seems to be tied to some business logic. 

Perhaps it could be replaced by a render prop?

packages/components/droppable-panel/src/droppable-panel.styles.js
```diff
@@ -0,0 +1,13 @@
+import styled from '@emotion/styled';
+import { customProperties as vars } from '@commercetools-uikit/design-system';
+
+const TagContainerEditable = styled.div`
+  background-color: ${vars.backgroundColorForInput};
+  border: ${vars.borderRadius1} solid #999;
```
jonnybel, line 6: We should use the colors from our design tokens. There should be an available colour that corresponds to this one, since I don't think it's justifiable to add a new one.

packages/components/droppable-panel/src/messages.js
```diff
@@ -0,0 +1,10 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  useInputSearch: {
```
jonnybel, line 4: It feels like this translation doesn't belong to a component library as it's too tied to a more specific behaviour / business logic (hence my other comment above).

packages/components/droppable-panel/src/draggable-tag/draggable-tag.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactDOM from 'react-dom';
+import { Draggable } from 'react-beautiful-dnd';
+import { DragIcon } from '@commercetools-uikit/icons';
+import Spacings from '@commercetools-uikit/spacings';
+import Tag from '@commercetools-uikit/tag';
+import DraggingContainer from './draggable-tag.styles';
+// import styles from './draggable-tag.mod.css';
```
jonnybel, line 9: 🕵️

packages/components/droppable-panel/src/droppable-panel.styles.js
NickDevG: I'll check the one that is closest

packages/components/droppable-panel/src/messages.js
NickDevG: This input it's a usability thing: When there are more than 50 columns in the table the search input appears

packages/components/droppable-panel/src/droppable-panel.styles.js
NickDevG: `colorNeutral60` same color in hsl format

packages/components/droppable-panel/src/messages.js
jonnybel: Ah I see, this makes sense then. But perhaps we should be more consistent with the namings: here we are calling them "attributes" but in the codebase we call them "columns" (because we know they will be used for the table).

As a standalone component, I feel we should use a more generic name and just call them "items" (or something equivalent), both here and in the codebase.

NickDevG:
Closing in order to open a new one with updated content

PR #1419, opened 2020-07-06
chore(packages/text): move typography to packages

mohib0306:
#### Summary

- Move `typography` to from `src/components` to `packages/components`


packages/components/text/typography.story.js
```diff
@@ -3,8 +3,8 @@ import styled from '@emotion/styled';
 import { storiesOf } from '@storybook/react';
```
mohib0306, line 3: This file can be renamed to `text.story.js` and placed in the `text/src`

jonnybel: Yeah, let's do that :)

PR #1422, opened 2020-07-10
docs(text): use correct import

emmenko:


emmenko:
@mohib0306 please make sure also that the other readmes have the correct import.

On a side note: we should also start documenting a bit better to use the direct import, or to use one of the presets.

emmenko:
Regarding the puppeteer error, I get the same in appkit. I'm already looking into that.

emmenko:
Fix for puppeteer in #1424

PR #1423, opened 2020-07-10
chore: update pptr-testing-library

jonnybel:
#### Summary

`pptr-testing-library` version was locked because of https://github.com/testing-library/pptr-testing-library/issues/31
That has been fixed with https://github.com/testing-library/pptr-testing-library/pull/34.

This PR unlocks the version and updates.

<del>EDIT: Actually, we need to wait for the new version with the fix to be released 😅 </del>

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -23,7 +23,6 @@
     "slate",
     "slate-html-serializer",
     "slate-react",
-    "@svgr/plugin-svgo",
-    "pptr-testing-library"
+    "@svgr/plugin-svgo"
   ]
 }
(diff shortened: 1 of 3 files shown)


jonnybel:
This is weird, after updating dependencies, the interactive VRT passed, but then I gave it a second run just to make sure it's not flaky and it failed again 😞

emmenko:
Is it only failing for the `RichTextInput` test?

jonnybel:
Yeah, only the interactive test, it's a timeout. I still haven't figured out why.

Sometimes it also times out on the main branch, but it's a lot less frequent.
With this update, it times-out much more frequently (only passed once out of ~10runs) 🤔

jonnybel:
Alright, so this finally seems to work now.

PR #1425, opened 2020-07-10
fix(data-table): Row collapsing behaviour

jonnybel:
#### Summary

This PR fixes the `DataTable` to properly update row collapsed state when `isTruncated` options are changed for the whole row.

#### Description

A bit of context:
* `isTruncated` is a flag that only controls whether a row can be truncated.
* `isRowCollapsed` actually controls if a column with isTruncated=true is in fact truncated(=collapsed)
When a table is initially rendered, the row’s `isRowCollapsed` state is initialized according to the `isTruncated` props of the columns. 
Changing the `isTruncated` prop doesn’t re-initialize that state, which is why the table remains uncollapsed if it was initialized like that (and vice versa). Only the expand/collapse button controls that state.

This was a problem for developing the ManagedDataTable, which allows imperatively toggling the `isTruncated` of the columns but it wasn't actually toggling the collapsed state of the rows.

### DIFF
--- a/packages/components/data-table/src/data-row.js
+++ b/packages/components/data-table/src/data-row.js
@@ -25,11 +25,8 @@ const DataRow = (props) => {
     }
   }, [rowHasTruncatedColumn]);
 
-  const shouldRenderCollapseButton = (
-    isTruncated,
-    totalColumnsLength,
-    currentColumnIndex
-  ) => isTruncated && totalColumnsLength - 1 === currentColumnIndex;
+  const shouldRenderCollapseButton = (totalColumnsLength, currentColumnIndex) =>
+    rowHasTruncatedColumn && totalColumnsLength - 1 === currentColumnIndex;
 
   return (
     <Row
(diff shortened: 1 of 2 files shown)


packages/components/data-table/src/data-row.js
```diff
@@ -15,10 +15,18 @@ const DataRow = (props) => {
...
+    } else if (!rowHasTruncatedColumn) {
+      collapseRow(false);
+    }
+  }, [rowHasTruncatedColumn]);
+
   const shouldRenderCollapseButton = (totalColumnsLength, currentColumnIndex) =>
-    totalColumnsLength - 1 === currentColumnIndex &&
-    ((isRowCollapsed && rowHasTruncatedColumn) ||
-      (rowHasTruncatedColumn && !isRowCollapsed));
+    totalColumnsLength - 1 === currentColumnIndex && isRowCollapsed;
```
jonnybel, line 29: This is unrelated to the PR but I felt really dumb when I saw that we had this in the code 😆

NickDevG:
Thanks for working on this @jonnybel

PR #1426, opened 2020-07-10
fix(data-table): footer horizontal scroll

### LINKED ISSUE #1417: User needs to scroll horizontally to see the rest of the DataTable's footer.
**Describe the bug**
When the Datatable has many columns and its footer has content on both ends (e.g at the left end of the footer there is info about page, and on the right end there are the prev/next action buttons), the user has to scroll horizontally find the action buttons

**To Reproduce**

1. Go to the provided sandbox link
2. Scroll down to the footer area
3. You'll have to scroll horizontally to reach the prev/next action buttons.

https://codesandbox.io/s/data-table-demo-7869876fdz-kg7fl

**Expected behavior**
The footer should be fully displayed without needing to scroll.

**Screenshots**

**Additional context**
If we render the footer separately, we'll not have this issue as demonstrated in the example sandbox.

jonnybel:
#### Summary

Fixes #1417.

#### Description

After discussing with the design team, we decided to keep the footer always "sticky" on the bottom, removing the `disableFooterStickiness` prop, because we don't expect any reasonable scenario where it would be useful for a table to have a non-sticky footer.

Because of this, it is more beneficial to remove the Footer out of the Table layout and just render it right below the `table` element, as a block, removing the need to actually use the `sticky` style, which had a few caveats (for instance, this solves an issue in Safari where the footer could become abnormally tall when the table is resized under certain circumstances).

The footer is now not actually a `tfoot` element, but just a normal div. Due to this, to improve testing experience, I added the attribute `data-testid='footer'` to the element, in the same fashion as the header and cell table elements.
Not using a `tfoot` element doesn't cause any problems, as it's merely a semantical indicator with no practical effect for our use cases.


packages/components/data-table/src/cell.styles.js
```diff
@@ -104,7 +104,10 @@ const BaseCell = styled.td`
   position: relative;
   display: flex;
   background-color: ${vars.colorSurface};
-  border-bottom: 1px solid ${vars.colorNeutral90};
+  border-bottom: ${(props) =>
+    props.shouldRenderBottomBorder
```
jonnybel, line 108: This is to make sure we don't have a double border on the bottom of the table (the one from the last row and the top one from the footer), replacing the negative 1px margin "trick".

packages/components/data-table/src/data-table.js
```diff
@@ -48,67 +60,68 @@ const DataTable = (props) => {
...
+                  key={column.key}
+                  shouldWrap={props.wrapHeaderLabels}
+                  isCondensed={props.isCondensed}
+                  iconComponent={column.headerIcon}
+                  disableResizing={column.disableResizing}
+                  verticalCellAlignment={props.verticalCellAlignment}
+                  horizontalCellAlignment={
+                    column.align ? column.align : props.horizontalCellAlignment
+                  }
+                  disableHeaderStickiness={props.disableHeaderStickiness}
```
tdeekens, line 84: This can go away, right?

tdeekens: Ah my mad. Double checked. Footer not header.

PR #1427, opened 2020-07-10
chore(table): update data-table 'beta' status and warn about table deprecating

jonnybel:
#### Summary

* Removes `DataTable` "Beta" warning.
* Adds warning to `Table` component deprecating soon and removes one of its storybook examples.
* Places both table component stories on the same section of Storybook examples.


packages/components/data-table/src/data-table.story.js
```diff
@@ -324,7 +324,7 @@ ColumnConfigForm.propTypes = {
   }),
 };
 
-storiesOf('Components|Table (NEW)', module)
+storiesOf('Components|Table', module)
```
emmenko, line 327:
```suggestion
storiesOf('Components|DataTable', module)
```

packages/components/table/README.md
```diff
@@ -1,5 +1,7 @@
 # Table
 
+> Note: This component will soon be deprecated in favour of `DataTable`, and it is encouraged to use it instead of this one.
```
emmenko, line 3: What do you mean by soon?

What needs to be deprecated is actually the NPM package. If we don't plan to update this component anymore, we should just remove the code.

jonnybel: We're still using this component in the MC in many places, so we can't remove it until we migrate all those cases to use the new DataTable.

Do you mean removing the Table package code entirely or just from Storybook (for now)?

For now, I would just like to discourage the usage of this component, without removing it, as it is still needed. Then when possible removing it entirely with a new major version of ui-kit.


We could also make the new major release (without the old Table) sooner if we change the MC to use the individual table package where needed 🤔

emmenko: If this component is used or not, it doesn't matter at the end.
If we don't plan to work on it anymore, we can just remove the code theoretically.
The NPM package still exists, but we should mark it as deprecated.

jonnybel: Ah, and the ui-kit preset (the one with all packages) will still have this deprecated package even though it won't be present in the src code anymore?

emmenko: Correct, the package dependency will be resolved from npm instead of the monorepo.

I mean, you can still keep the code around if you want, but I don't see a need to do that if we're going to deprecate the package.
The NPM package will still exist, we just don't have and maintain the code anymore.

Does it make sense?

emmenko: Maybe @tdeekens can double check my assumption, just to be sure.

jonnybel: I didn't know that 😄 Cool! 

We're definitely not maintaining the old table anymore, so let's 🧹🧹

tdeekens: To me it seems like we're trying to find a path for something we will not need a path for in 2-4 weeks. If all tables are migrated in the MC we don't have any issue either or. We should push for that and currently I am a lot in all applications. 

I don't see exactly what the anticipated future cost of maintenance of the Table code is we are trying to avoid given it was zero the last 8-10 months if I am not mistaken.

What I would like to avoid is:

1. Locking the MC into an old version of ui-kit just cause 1-2 apps use the table a bit longer
2. Adding yet another layer of tooling and preset/package structure to remove the table code

I also don't see how, if we remove the code, any preset should contain the table artefacts given their not build and then published to npm in newer versions/releases.

emmenko: Hmm I guess you're right, the table version will still be bumped whenever its dependencies are updated.

Alright, then let's keep it as long as we still need it, then we can drop it.

Sorry for the confusion 🙃

tdeekens: No probs. Let's just keep our lives simple :)

jonnybel:
FYI I'm just waiting for a green light from the Product Team to only merge this in sync with updating the release notes with the new URL for the DataTable Storybook page.

PR #1434, opened 2020-07-14
feat(link): add intlMessage prop to Link in ui-kit

fuchodeveloper:
#### Summary
This PR adds a intlMessage prop to Link component, which can replace the children prop for translated text, in the same fashion as the Text components.

### DIFF
--- a/.changeset/ten-suns-count.md
+++ b/.changeset/ten-suns-count.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/link': minor
+---
+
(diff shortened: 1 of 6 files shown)


tdeekens:
Approved the VRTs. They show that you added something :)

fuchodeveloper:
> Approved the VRTs. They show that you added something :)

Thanks for clarifying. 
I was checking why it was marked in red.

tdeekens:
Yeah. They're marked as red anytime something visually changed. As they're visual regression tests. As you added a new visual spec it needs to be approved too.

PR #1449, opened 2020-07-27
fix(data-table): set default height for table rows

fuchodeveloper:
#### Summary
With the implementation of the new Data Table component, all rows should have a default height of 50px.

#### Description

Without setting a default height, if an icon is part of a table row, it pushes the height of the row above 50px.


fuchodeveloper:
Hi Abel, the update for the row height is based on design recommendations from Sven for the data tables available in the discounts app. Considering that this might not be requested across all implementations of the data table, I'll be closing this PR and making the required updates to the discount app data tables specifically.

PR #1450, opened 2020-07-28
chore(packages/spacings): move spacings to packages

mohib0306:
#### Summary

- Move code from `src/components/spacings` to `packages/components/spacings`


mohib0306:
> The packages/components/spacings/index.js should not be necessary anymore.

If we remove the `packages/components/spacings/index.js` then the `Inline` and `Stack` spacings used in the `stamp.story.js` should be imported either like 
```
import SpacingsStack from '../../spacings/stack';
import SpacingsInline from '../../spacings/inline';
```
or we add the `@commercetools-uikit/spacings` as a dependency , which i don't think would be a good idea. Right?

jonnybel:
@mohib0306 We can go for the relative paths 👍

emmenko:
From stories, we can use relative paths yes

packages/components/spacings/inline/README.md
```diff
@@ -1,10 +1,12 @@
...
+
 If you want to learn more about the spacing components take a look at
 [this](https://medium.com/eightshapes-llc/space-in-design-systems-188bcbae0d62)
 article.
 
-## Example
+#### Usage
 
 ```jsx
 import { Spacings } from '@commercetools-frontend/ui-kit';
```
emmenko, line 12: The import should be `import SpacingsInline from '@commercetools-uikit/spacings-inline';`

Same for the other ones.

PS: we can also mention to use the `spacings` preset

mohib0306: oh no.

packages/components/spacings/inline/README.md
```diff
@@ -1,10 +1,12 @@
 # Inline
 
+#### Description
```
emmenko, line 3: Use `##` instead of `####`. See other packages, for example https://raw.githubusercontent.com/commercetools/ui-kit/96fcc4fd1e8ea90b14c93270cc6c0f2ab1f16e80/packages/components/avatar/README.md

mohib0306: Some other has `####` instead of `##` i.e. https://raw.githubusercontent.com/commercetools/ui-kit/master/packages/components/inputs/toggle-input/README.md
https://raw.githubusercontent.com/commercetools/ui-kit/master/packages/components/inputs/radio-input/README.md

emmenko: Yeah, we need to fix all those READMEs.

---

Background: from the docs team I learned that the markdown and content structure is important and needs to follow certain rules. In this case, the main heading is H1, then the following headings should be H2, not H4. There is no point is skipping H2 and H3.

So we should do this:

```
#
##
###
####
```

instead of this:


```
#
####
```

PR #1451, opened 2020-07-28
chore(packages/notifications): move notifications to packages

mohib0306:
#### Summary

- Move `src/components/notification` to `packages/components/notification`


packages/components/notifications/content-notification/index.js
```diff
@@ -0,0 +1,6 @@
+// This file exists  because we want jest to use our non-compiled code to run tests
+// if this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) then jest will fail if the bundle is not yet built.
+// all apps should export all their named exports from their root index.js
+export { default } from './src';
+export * from './src';
```
mohib0306, line 6: `content-notification` code is inside a `src` folder

packages/components/notifications/index.js
```diff
@@ -1,4 +1,6 @@
-/* eslint-disable import/prefer-default-export */
-export { default as ContentNotification } from '../../../src/components/notifications/content-notification';
-
-export { default as version } from './version';
+// This file exists  because we want jest to use our non-compiled code to run tests
+// if this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) then jest will fail if the bundle is not yet built.
+// all apps should export all their named exports from their root index.js
+export { default as ContentNotification } from './content-notification';
+export * from './content-notification';
```
mohib0306, line 6: Left this `index.js` file so that the component can be imported from its own package

packages/components/notifications/content-notification/README.md
```diff
@@ -1,20 +1,20 @@
 # ContentNotification
```
mohib0306, line 1: The `README` is located under `packages/components/notifications/content-notification`. I think we don't a README under `packages/components/notifications`

packages/components/notifications/content-notification/index.js
emmenko: The folder structure is a bit weird. Instead of `notifications/content-notification/src` I would expect `notifications/src/content-notification`.

The package is `notifications`, and each package has a `src` folder.

mohib0306: I thought of it the other way around. Similar to the `packages/components/fields` or `packages/components/inputs` where the main folder holds the other smaller components which has their own `src` folders

emmenko: Yes but it's different. The `fields` folder or the `inputs` folder are not packages, they are just folders.

Here you have the `notifications` folder being the package

image

emmenko: Also, we have presets for fields and inputs packages, but for notifications I don't see a need for that.

So I would just keep the `notifications` being the package (as you have it now) but with a `src` folder.
You can still keep the `src/content-notification` structure if you want.

packages/components/notifications/src/content-notification/index.js
```diff
@@ -1 +1,3 @@
 export { default } from './content-notification';
+
+export { default as version } from './version';
```
emmenko, line 3: The version should be part of the `notifications` package, not the single component.

emmenko: The version should go into the `src` folder. I pushed a commit: https://github.com/commercetools/ui-kit/pull/1451/commits/dd8a0d44da8f116d632a2c1ade0ff1f40f9861cc

PR #1455, opened 2020-07-30
feat(data-table): allow css value strings for maxHeight and maxWidth

jonnybel:
#### Summary

`DataTable`: This PR allows the props `maxHeight` and `maxWidth` to accept css value strings (such as `'100%`' and calc() functions) in addition to numeric values.

This is particularly useful for setting dynamic heights, such as using the available viewport height (`vh`) for the height of the table.
For instance, the following example makes the table take 90% of the viewport height and subtracts 100px to it (and it adjusts the height automatically if the window is resized): `maxHeight='calc(90vh -100px)'`

The Storybook example has been updated to accept strings on these props. Numeric values are still accepted as before (if a number is passed, without any unit, it is considered the value in pixels the same way as before).


packages/components/data-table/src/data-table.story.js
```diff
@@ -22,6 +16,11 @@ import NumberInput from '../../inputs/number-input';
 import IconButton from '../../buttons/icon-button';
 import { InformationIcon } from '../../icons';
 
+// aux function for the maxHeight and maxWidth prop knobs
+// which can accept Number values but the text knob always returns String
+const castToNumberIfPossible = (value) =>
```
tdeekens, line 21: Should we maybe throw some unit tests at this?

jonnybel: That fn is just for the storybook example, but I should add for this one:
https://github.com/commercetools/ui-kit/pull/1455/files#diff-bd664be2fdc62a0525a6bec77cb25841R6

In fact, I think I'll move it to another file (as an util) and test it individually.

tdeekens: That sounds good. Thanks.

PR #1457, opened 2020-07-31
fix: ui-kit preset export and version bump for data-table-manager

jonnybel:
#### Summary

* Fixes UI-Kit preset to include export of DataTableManager
* Bumps version (patch) of data-table-manager package to add fix from [#1454](https://github.com/commercetools/ui-kit/pull/1454)
* ~Adds single VRT to DataTableManager (more should be added later).~


packages/components/data-table-manager/package.json
```diff
@@ -37,7 +37,7 @@
     "@emotion/core": "10.0.28",
     "@emotion/styled": "10.0.27",
     "debounce-promise": "^3.1.2",
-    "lodash.differencewith": "4.5.0",
```
emmenko, line 40: You have to fix the import to `lodash/differenceWith`

jonnybel: Yeah, just saw that as the deployment failed 😅

emmenko:
The percy snapshot looks weird. Is this how it's supposed to be? Can we also take snapshots when we interact with the component?

jonnybel:
Yeah sorry, I'll need to go a bit deeper with the VRTs... and yes we can add some interaction tests.

But I've removed the VRT again, because what we need for now is fix the export of the component as soon as possible. The VRTs will remain a follow-up.

PR #1462, opened 2020-08-03
chore(ui-kit): add changeset for updated dependencies

jonnybel:


.changeset/few-mice-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/ui-kit': patch
```
emmenko, line 2:
```suggestion
'@commercetools-uikit/table': patch
```

Only the table package got updated

jonnybel: Sorry, updated. (also 'visual-testing-app')

PR #1464, opened 2020-08-04
Expose hasWarning prop on select fields

mohib0306:
#### Summary

- `AsyncSelectInput` has `hasWarning` prop to indicate that the field has a warning. This prop was not passed down from the `AsyncSelectField` and this PR passes down via the `hasWarning` field.

### DIFF
--- a/packages/components/fields/creatable-select-field/src/creatable-select-field.js
+++ b/packages/components/fields/creatable-select-field/src/creatable-select-field.js
@@ -48,6 +48,7 @@ export default class CreatableSelectField extends React.Component {
     isOptionDisabled: PropTypes.func,
     isMulti: PropTypes.bool,
     isSearchable: PropTypes.bool,
+    hasWarning: PropTypes.bool,
     maxMenuHeight: PropTypes.number,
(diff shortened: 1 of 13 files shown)


mohib0306:
`AsyncCreatableSelectField` and `CreatableSelectField` components are also missing the same functionality, if it is agreed upon that they should have it, I can add it to this PR.

jonnybel:
@mohib0306 Yes, please add them in this PR 🙏 

After that, it just needs a changeset and it seems good to go 👍

packages/components/fields/creatable-select-field/src/creatable-select-field.js
```diff
@@ -17,11 +17,11 @@ const sequentialId = createSequentialId('creatable-select-field-');
 
 const hasErrors = (errors) => errors && Object.values(errors).some(Boolean);
 
-export default class SelectField extends React.Component {
-  static displayName = 'SelectField';
+export default class CreatableSelectField extends React.Component {
+  static displayName = 'CreatableSelectField';
```
tdeekens, line 21: Nice. Will likely make some snapshots bite us internally 😢.

PR #1465, opened 2020-08-04
feat(data-table): Add prop `disableSelfContainment`

jonnybel:
#### Summary

This PR adds the prop `disableSelfContainment`:

> Enabling this flag allows taking control of the containment of the table and doing it on a parent element. This means that the table will grow in size without adding scrollbars on itself, both vertically and horizontally -- and as a consequence, the `maxHeight` and `maxWidth` props are ignored, so if you need to enforce these constraints, you must also apply them on the parent element. Additionally, the sticky behaviour of the header will get fixed relatively to the closest parent element with `position: relative`. 

You can see how it affects the behaviour of the sticky header and size constraints on Storybook:
Kapture 2020-08-04 at 16 28 01

This allows, for instance, for the sticky header of the Table to become fixed to the top of a page.


packages/components/data-table/src/data-table.styles.js
```diff
@@ -16,9 +16,19 @@ const getClickableRowStyle = (props) => {
   return '';
 };
 
+const getDisabledSelfContainmentStyles = (props) => {
+  if (props.disableSelfContainment) {
+    return css`
+      position: unset;
+      overflow: unset;
```
tdeekens, line 23: Didn't know of `unset`.

jonnybel:
Thanks @mohib0306 for noticing a bug fixed in https://github.com/commercetools/ui-kit/pull/1465/commits/6b15b231bd0a38ca63948b47b7f89172076e8a74

PR #1466, opened 2020-08-07
fix(data-table):horizontal scrolling

jonnybel:
#### Summary

This PR fixes a bug where the user can't scroll horizontally on the DataTable when it exceeds the container's size.

### DIFF
--- a/packages/components/data-table/src/data-table.styles.js
+++ b/packages/components/data-table/src/data-table.styles.js
@@ -17,6 +17,8 @@ const getClickableRowStyle = (props) => {
 };
 
 const TableContainer = styled.div`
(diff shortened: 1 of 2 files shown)


packages/components/data-table/src/data-table.styles.js
```diff
@@ -17,6 +17,8 @@ const getClickableRowStyle = (props) => {
 };
 
 const TableContainer = styled.div`
+  overflow-x: auto;
```
jonnybel, line 20: Because we added this container (to include the footer on the maxWidth constraint), we must also enable the horizontal scrollbars here.

PR #1469, opened 2020-08-07
feat(data-table): add prop onColumnResized

jonnybel:
#### Summary

This PR adds ta new prop to the `DataTable`: `onColumnResized`:

> Callback function which is called when a column has been resized with the mouse and which receives an array of the column widths after the resizing. One can use this callback to get the resized column widths and persist them, to be used as values for the column widths when a user comes back to the same view, finding the table in the same state as before.

Added a section to the storybook example which shows the array that we get from this function after resizing a column:
image


packages/components/data-table/src/data-table.story.js
```diff
@@ -472,6 +474,15 @@ storiesOf('Components|DataTable', module)
             </label>
           </div>
         </div>
+        <hr />
+        <div>
+          <h4>Column Sizes (after manual resizing):</h4>
+          {columnSizes ? (
+            <code>{JSON.stringify(columnSizes)}</code>
```
tdeekens, line 481:
```suggestion
            <code>{JSON.stringify(columnSizes, null, 2)}</code>
```

PR #1471, opened 2020-08-07
feat: add package to generate readme files

emmenko:
This PR adds a new package for generating the README files for each component package.

This will help us keeping the documentation up-to-date and consistent.

The package implementation is inspired by the migration package we wrote for the CT docs migration. It uses AST to build the markdown file and typescript to help working with the AST.

As an example, I refactored the `avatar` package to generate the README automatically.

For now, you have to run the script manually. Once we refactored all packages to have the correct setup, we can automate the process even more.


generators/readme/src/index.ts
```diff
@@ -0,0 +1,170 @@
...
+      // Describe the component's properties
+      heading(2, 'Properties'),
+      ...parseMarkdownFragmentToAST(
+        path.join(options.packagePath, 'docs/properties.md')
+      ),
+      ...parseMarkdownFragmentToAST(
+        path.join(options.packagePath, 'docs/additional-info.md')
+      ),
+    ];
+  }
```
emmenko, line 127: This is the most important part.

emmenko:
@jonnybel @mohib0306 have you had the time/chance to look at this?

mohib0306:
I reviewed it thoroughly 3 days ago. But I will look into it again, first thing tomorrow

jonnybel:
Sorry, I only scrolled through it.

I will properly review later today or tomorrow morning 🙏

generators/readme/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+A package to generate README files for the React component packages.
+
+## Prerequisites
+
+Each package should have the following fragment files in a `docs` folder:
+
+- `docs/usage-example.js`: a JS example for using the component.
+- `docs/properties.md`: the markdown content for the describing the component's properties (usually as a table).
+- `docs/additional-info.md`: any additional information to be rendered AFTER the generated sections.
```
mohib0306, line 11: Where would the basic info go. I mean the intro/description. Keep in mind that not all of the components has this description section but of them has.

mohib0306, line 11: Where would the basic info go. I mean the intro/description. Keep in mind that not all of the components has this description section but of them has.

generators/readme/bin/generate-readme.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import mri from 'mri';
+import report from 'vfile-reporter';
+import { generate } from '../src/index';
+import { CommandFlags } from '../src/types';
+
+const flags = mri(process.argv.slice(2), {
+  alias: { help: ['h'], boolean: ['dry-run'] },
+});
+const [pathToPackage] = flags._;
```
mohib0306, line 11: `pathToPackage` isn't defined!

generators/readme/bin/generate-readme.ts
```diff
@@ -0,0 +1,38 @@
...
+const dryRun: boolean = flags['dry-run'] ? true : false;
+
+if (flags.help || !pathToPackage) {
+  console.log(`
+  Usage: generate-readme [path-to-package] [options]
+
+  Displays help information.
+
+  [path-to-package]
+    The path to the package where the README.md file should be generated.
```
mohib0306, line 22:
```suggestion
    The path to the package for which the README.md file should be generated.
```

generators/readme/src/index.ts
```diff
@@ -0,0 +1,170 @@
+import type { VFile, VFileCompatible } from 'vfile';
+import type { Node } from 'unist';
+import type { Parent, Heading, Text, Paragraph, Code, Root, HTML } from 'mdast';
+import type { Options as PrettierOptions } from 'prettier';
+import type {
+  CommandFlags,
+  GeneratorReadmeOptions,
+  PackgeJsonInfo,
+} from './types';
```
mohib0306, line 9: can be moved to the bottom of the imports since it is relative import

generators/readme/src/index.ts
```diff
@@ -0,0 +1,170 @@
...
+      ),
+      // The main heading is derived by the name of the folder
+      heading(1, capitalize(camelcase(path.basename(options.packagePath)))),
+      // The description is what's defined in the package.json
+      heading(2, 'Description'),
+      paragraph(packageJsonInfo.description),
+      // Describe the installation steps
+      heading(2, 'Installation'),
+      code('', `yarn add ${packageJsonInfo.name}`),
+      code('', `npm --save install ${packageJsonInfo.name}`),
```
mohib0306, line 93: the default code `lang` is empty string. Can we mention `bash` or `sh` if possible?

generators/readme/src/index.ts
mohib0306: Nice work. Just two things that cross my mind,
1-  What happens if the file is present but empty. IMO we should show that section
2- What happens if the file is not present. Wouldn't this throw errors while fetching the file

generators/readme/bin/generate-readme.ts
emmenko: What do you mean? This is array destructuring... 🤔

generators/readme/src/index.ts
emmenko: No, this is an `import type` which I prefer to keep at the top, before the normal imports.

generators/readme/src/index.ts
emmenko: We can but technically it's not really bash.

generators/readme/src/index.ts
emmenko: Which file are you referring to? The fragments?

If so, some of them should be required. But you're right, the optional ones (like the `additional-info`) can be optional, so we should check for that.
Thanks

generators/readme/bin/generate-readme.ts
mohib0306: Looking at the docs of `mri` it seems that `_` field will have the args mentioned after `--` in cli command, which does not match the usage example given on line 17.
https://www.npmjs.com/package/mri#usage

generators/readme/README.md
mohib0306: >  We can also decide to have it as a markdown fragment

IMO that would be helpful. But I am also ok with the current solution of reading it from the `package.json`

generators/readme/src/index.ts
mohib0306: > If so, some of them should be required

do we show warning or throw error incase they are missing, i haven't noticed, sorry.
It is easy to forget adding the required files but showing a warning or error would be helpful

generators/readme/bin/generate-readme.ts
emmenko: The docs are a bit misleading. The `_` contains all arguments that are NOT CLI options. For example

```
my-script foo bar baz --hello
```

The `_` field contains `['foo', 'bar', 'baz']`.

generators/readme/src/index.ts
emmenko: Yes sure. There is no validation built-in yet but we can add it for sure 👌

generators/readme/src/index.ts
```diff
@@ -56,6 +93,74 @@ const parseMarkdownFragmentToAST = (filePath: string) => {
...
+        case 'enum':
+          propType = propInfo.type.name;
+          propDefaultValue = (propInfo.type.value as { value: string }[])
+            .map((enumValue) => enumValue.value)
+            .join(', ');
+          break;
+        default:
+          propType = propInfo.type.value
+            ? String(propInfo.type.value)
+            : propInfo.type.name;
```
mohib0306, line 146: shouldn't we update `propDefaultValue` here as well? Most of the times the props of type `boolean` has some sort of default value

generators/readme/src/index.ts
```diff
@@ -86,7 +93,11 @@ function readmeTransformer(options: GeneratorReadmeOptions) {
       heading(1, capitalize(camelcase(path.basename(options.packagePath)))),
       // The description is what's defined in the package.json
       heading(2, 'Description'),
-      paragraph(packageJsonInfo.description),
+      ...(fs.existsSync(paths.description)
+        ? parseMarkdownFragmentToAST(paths.description)
+        : // Fall back to the package.json description field.
+          [paragraph(packageJsonInfo.description)]),
```
mohib0306, line 209: nice

emmenko:
Alright, the table props now renders nested structures as multiple table rows, with specific syntax notations

- `.` object
- `[]` array
- `<>` union
- `[].` array of objects
- `[]<>` array of unions
- `[]<>.` array of union object

and so on.

image

generators/readme/README.md
```diff
@@ -0,0 +1,18 @@
+# Readme generator
+
+A package to generate README files for the React component packages.
+
+## Prerequisites
+
+Each package should have the following fragment files in a `docs` folder:
+
+- `docs/usage-example.js`: a JS example for using the component.
+- `docs/additional-info.md`: any additional information to be rendered AFTER the generated sections.
```
mohib0306, line 10: can you mention `docs/description.md` as well? also the fallback to `description` on `package.json` incase that file is missing?

generators/readme/test/fixtures/avenger/src/avenger.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Avenger = (props) => <div>{`Avenger: ${props.name}`}</div>;
+Avenger.defaultProps = {
```
mohib0306, line 5: can we add a default value for a prop boolean type? Just to have most of the cases covered

generators/readme/README.md
emmenko: Ah yes, I forgot about that.

generators/readme/test/fixtures/avenger/src/avenger.js
emmenko: Sure I can do that.

PS: we can also increase the "test coverage" as we migrate packages and bump into "unsupported" use cases.

PR #1475, opened 2020-08-11
fix(data-table): pass column key to onRowClick callback

jonnybel:
#### Summary

This PR fixes the DataTable `onRowClick` callback to pass the column key as an argument of the method.

On the "old" table, this information was passed to the method, but on the new one, it was removed as an `onClick` method was available per-column. However, the `onClick` method was later [also removed](https://github.com/commercetools/ui-kit/pull/1314) with no logical replacement, which made it hard to define per-column specific behaviours for `onClick` events on the Cells.


packages/components/data-table/src/data-row.js
```diff
@@ -53,6 +47,13 @@ const DataRow = (props) => {
             props.columns.length,
             columnIndex
           )}
+          onCellClick={
```
jonnybel, line 50: The `onClick` event handler had to be moved from the Row to each of the DataCells here in order to have the `column` information available.

packages/components/data-table/src/data-row.js
```diff
@@ -53,6 +47,13 @@ const DataRow = (props) => {
...
+            props.onRowClick && !column.shouldIgnoreRowClick
+              ? () =>
+                  props.onRowClick &&
+                  props.onRowClick(props.row, props.rowIndex, column.key)
```
tdeekens, line 54:
```suggestion
                  props.onRowClick?.(props.row, props.rowIndex, column.key)
```

Should work with optional chaining 🤔 . Syntax looks really odd but I think it was it.

jonnybel: Ah nice! I'm committing this and let the tests be the judge :)

jonnybel:
@adnasa I believe this would help to solve a regression issue that you mentioned in the MC.

adnasa:
please ping me when this is in and I'll update MC-fe

Issue #1491, opened 2020-08-14
Cannot find module '@emotion/styled-base'


emmenko:
**Describe the bug**
Installing latest dependencies causes build/lint errors.

```
Cannot find module '@emotion/styled-base' from 'node_modules/@commercetools-uikit/accessible-button/dist/accessible-button.cjs.js'
```

<img width="661" alt="image" src="https://user-images.githubusercontent.com/1110551/90223225-f2f05800-de0d-11ea-8992-161115a605df.png">

**To Reproduce**
https://github.com/commercetools/merchant-center-application-kit/pull/1673/checks?check_run_id=983630152

**Expected behavior**
The `@emotion/styled-base` is not a peer dependency and shouldn't be required to be installed by consumers.

**Screenshots**
<img width="1236" alt="image" src="https://user-images.githubusercontent.com/1110551/90223092-b3c20700-de0d-11ea-84a6-678c4930475d.png">

**Additional context**
Maybe something related to this?
https://github.com/commercetools/ui-kit/blob/2db4063679813da5929b92e64a75582a8bc5e0bf/rollup.config.js#L38-L40

alochowski:
Hi,
unfortunately, I have still the same problem with this module during production build, I've got:
`Module not found: Error: Can't resolve '@emotion/styled-base' in '/app/node_modules/@commercetools-uikit/accessible-button/dist'`

is there a chance to fix this somehow?

emmenko:
Which versions of the uikit packages are you using?

alochowski:
I'm using several packages from uikit, most of them are in version `10.30.1`

emmenko:
This issue with emotion was fixed in version `10.30.1` (https://github.com/commercetools/ui-kit/pull/1483). Please make sure that all your packages are on the latest version.

alochowski:
Here is my dependencies list for uikit :
```
    "@commercetools-uikit/card": "10.30.1",
    "@commercetools-uikit/collapsible-panel": "10.30.1",
    "@commercetools-uikit/data-table": "10.30.1",
    "@commercetools-uikit/flat-button": "10.30.1",
    "@commercetools-uikit/grid": "10.30.1",
    "@commercetools-uikit/icons": "10.30.1",
    "@commercetools-uikit/label": "10.30.1",
    "@commercetools-uikit/link-button": "10.30.1",
    "@commercetools-uikit/multiline-text-field": "10.30.1",
[...]

PR #1492, opened 2020-08-14
fix: use ranged versions for emotion dependencies

emmenko:
This is an attempt to fix problems possibly related to #1491 

By using ranged versions, for consumers that also use emotion dependencies, the resulting installed versions of emotion should follow the consumer's version.


emmenko:
@jonnybel @mohib0306  anything we can do about this?

```
TestingLibraryElementError: Found a label with the text of: Date, however the element associated with this label (<input />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <input />, you can use aria-label or aria-labelledby instead.
```

jonnybel:
> @jonnybel @mohib0306 anything we can do about this?
> 
> ```
> TestingLibraryElementError: Found a label with the text of: Date, however the element associated with this label (<input />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <input />, you can use aria-label or aria-labelledby instead.
> ```

It's the same problem as here: https://github.com/commercetools/ui-kit/pull/1448#issuecomment-667991635

I was unable to reproduce the bug on Codesandbox with the exact same markup.

And now I just went through the testing-library issues again and found what could be our problem:
https://github.com/testing-library/dom-testing-library/issues/737#issuecomment-671738415

Looks like one of our dependencies is using an older version of `jsdom` which we need to update (https://github.com/testing-library/dom-testing-library/issues/737#issuecomment-672250807)

emmenko:
Still the same issue...I feel like this is also related to downshift...

packages/components/inputs/date-range-input/src/date-range-input.js
```diff
@@ -335,6 +335,9 @@ class DateRangeCalendar extends React.Component {
                 <CalendarBody
                   inputRef={this.inputRef}
                   inputProps={getInputProps({
+                    // Unset the aria-labelledby as it interfers with the link
+                    // between the <label for> and the <input id>.
+                    'aria-labelledby': undefined,
```
emmenko, line 340: @jonnybel this fixes the issue with RTL and the a11y selector.

jonnybel: Oh, so downshift no longer adds the attribute if we set it to undefined? Great find!

emmenko: The other way around: downshift sets the `aria-labelledby`, but the `input` ends up with having both the `id` and the `aria-labelledby`, which I suppose confuses RTL.

emmenko:
Ok, I guess I also need to fix the other components... 🙃

packages/components/inputs/date-range-input/src/date-range-input.js
jonnybel: Yeah I see... But it's still weird that when I tried to reproduce that rendered markup in a codesandbox, RTL didn't complain, which made me believe it wasn't the problem 🤷‍♂️

emmenko: Did you also use downshift?

jonnybel: No, I used the markup rendered by it. That was the mistake, I guess.

emmenko: I opened a separate PR with the fix: https://github.com/commercetools/ui-kit/pull/1497

PR #1494, opened 2020-08-14
chore: update react-intl and formatjs deps, add eslint plugin for formatjs

emmenko:
Just found out about this: https://formatjs.io/docs/tooling/cli#builtin-formatters

Also Transifex supports structured JSON: https://docs.transifex.com/formats/json/structured-json


.eslintrc.yaml
```diff
@@ -19,6 +19,7 @@ env:
   node: true
   mocha: false
 plugins:
+  - 'formatjs'
```
emmenko, line 22: They also provide an (experimental) eslint plugin

packages/components/fields/money-field/src/messages.js
```diff
@@ -1,9 +0,0 @@
-import { defineMessages } from 'react-intl';
-
-export default defineMessages({
-  highPrecision: {
```
emmenko, line 4: This is duplicated in the `money-input` component, and is not used in the money field.

```
warning [FormatJS CLI] Duplicate message id: "UIKit.MoneyInput.highPrecision", but the `description` and/or `defaultMessage` are different.
```

emmenko:
We'll see 🙂

emmenko:
We needed to change the file format in the tx integration:

image

tdeekens:
> We'll see 🙂

UIKit is a small project and as a result a good candidate. We have a backup in the pulled things we could merge into the busted things I guess. Should we inform the translation team that we're trying this?

tdeekens:
> We'll see 🙂

UIKit is a small project and as a result a good candidate. We have a backup in the pulled things we could merge into the busted things I guess. Should we inform the translation team that we're trying this?

emmenko:
> Should we inform the translation team that we're trying this?

Yep, already did

> UIKit is a small project and as a result a good candidate. 

I agree, we can try this here and see how it goes.

emmenko:
It might be that we need to adjust how we load the translated messages, depending on how transifex converts them. And also compile the messages: https://formatjs.io/docs/getting-started/message-distribution#translation-management-system-tms-integration

emmenko:
This is how the compiled output should look like (I think):

image

emmenko:
FYI: I changed it back to `KEYVALUEJSON` for now until @ikumiot can get the feedback if this is useful or not.

PR #1495, opened 2020-08-14
chore: remove enzyme to fix jsdom dependency

emmenko:
Let's see if this helps.

See https://github.com/commercetools/ui-kit/pull/1492#issuecomment-673989492


emmenko:
Yeah I'll take a look at that in the other PR about formatjs/react-intl updates #1494

PR #1496, opened 2020-08-14
fix(data-table): container not adjusting to table size when shrank

jonnybel:
#### Summary

`DataTable`: This PR fixed a bug where the table container is not adjusting to table size when columns are shrinked, which is noticeable when there is a footer present:

image

### DIFF
--- a/packages/components/data-table/src/data-table.styles.js
+++ b/packages/components/data-table/src/data-table.styles.js
@@ -28,7 +28,7 @@ const getDisabledSelfContainmentStyles = (props) => {
 };
 
 const TableContainer = styled.div`
-  width: min-content;
+  width: fit-content;
   overflow-x: auto;
 
   ${(props) =>
--- a/packages/components/data-table/src/data-table.styles.js
+++ b/packages/components/data-table/src/data-table.styles.js
@@ -28,6 +28,8 @@ const getDisabledSelfContainmentStyles = (props) => {
 };
 
 const TableContainer = styled.div`
+  position: relative;
+  z-index: 0;
   overflow-x: auto;
 
   ${(props) =>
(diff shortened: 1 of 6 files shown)


packages/components/data-table/src/data-table.styles.js
```diff
@@ -28,6 +28,7 @@ const getDisabledSelfContainmentStyles = (props) => {
 };
 
 const TableContainer = styled.div`
+  width: fix-content;
```
Rombelirk, line 31: wow, i didn't know that css property exists

jonnybel: I think it's relatively new (among with `max-content` and `min-content`), but it's supported by all major browsers 🙂 

Really handy for this use-case.

jonnybel:
OK, I changed it to `min-content` so that if the Footer contains text longer than the table, it will wrap in different lines instead of letting it grow larger than the table.

emmenko:
Btw: do we have VRTs for these kind of fixes/bugs?

jonnybel:
> Btw: do we have VRTs for these kind of fixes/bugs?

Not for the column resizing, I didn't figure out an effective way to simulate the mouse dragging...

jonnybel:
Note: `fit-content` actually breaks the responsive behaviour (of the table taking all the available width).

I'm working on an alternative.

jonnybel:
Alright, so simple css tricks didn't work, and the container doesn't readjust it's `auto` width when the table is manually resized, so we have to also set it manually.

jonnybel:
> Some level of testing would be nice. Either the `getTotalResizedTableWidth`, rtl or a VRT. What do you think?

You are right, I'm gonna try adding tests for the reducer with `@testing-library/react-hooks`

tdeekens (approved):
Thanks for the tests.

PR #1497, opened 2020-08-14
fix: a11y label

emmenko:
This should fix the issue we've been having lately regarding RTL not being able to "find" the element associated with the label. For example:

```
TestingLibraryElementError: Found a label with the text of: XXX, however the element associated with this label (<div />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <div />, you can use aria-label or aria-labelledby instead.
```


packages/components/inputs/radio-input/src/radio-group.js
```diff
@@ -49,15 +49,15 @@ const Group = (props) => {
   });
   if (props.direction === 'inline') {
     return (
-      <div id={props.id}>
+      <div aria-labelledby={props.id}>
```
emmenko, line 52: To make the relationship work with label + non-form elements, we have to use `<label id>` and `<div aria-labelledby>`

packages/components/field-label/README.md
```diff
@@ -55,7 +55,8 @@ The `hintIcon` also accepts a custom `color` while defaulting to `warning` in th
...
-| `htmlFor`              | `string`           |    -     | -                            | -       | ID of the labeled input                                                                                                                                                                                                                                               |
+| `htmlFor`              | `string`           |    -     | -                            | -       | ID of the labeled input, used for the `for` attribute                                                                                                                                                                                                                 |
+| `id`                   | `string`           |    -     | -                            | -       | ID of the labeled input                                                                                                                                                                                                                                               |
```
mohib0306, line 59: This is the ID of the label not the input. Right?

emmenko: yes for the label. Should we name it differently?

mohib0306: I am mentioning because the `description` indicates that it is the ID of the input.

mohib0306: > Should we name it differently?

Name is correct just the description was missleading. May be updating the description would be enough

mohib0306:
Also on other thing, shouldn't we be updating the usage to `FieldLabel` to also pass the `id` prop. i.e. https://github.com/commercetools/ui-kit/blob/master/packages/components/fields/date-time-field/src/date-time-field.js#L75

emmenko:
The id is only useful if the target element is not a form element. The only case I found so far was for the radio group.

emmenko:
C'mon richt text editor, don't let me down!!

packages/components/field-label/src/field-label.js
```diff
@@ -18,6 +18,10 @@ export const FieldLabel = (props) => {
       `ui-kit/FieldLabel: setting an hintIcon size is not supported.`
     );
   }
+  const a11yLabelProps = {
+    ...(props.id ? { id: props.id } : {}),
+    ...(props.htmlFor ? { htmlFor: props.htmlFor } : {}),
+  };
```
tdeekens, line 24: Is this capture needed? Can't we just pass the props below? 🤔

emmenko: Hmm probably not. You can't pass both, that's why I added the invariant. I think I forgot to remove this.
Thanks for the reminder.

packages/components/inputs/rich-text-input/src/rich-text-input.visualspec.js
```diff
@@ -375,5 +375,5 @@ describe('RichTextInput', () => {
 
     await input.type('Okay dokey');
     await wait(() => getByText(doc, 'Okay dokey'));
-  });
+  }, 20000);
```
emmenko, line 378: This seems to help with the visual tests for the rich text editor (hopefully)

tdeekens: The default timeout of rtl is quite low. Maybe globally increase it as in app-kit and then the MC. They changed it at some point to a lower value.

emmenko: Yeah thanks. The rich text input seems to be kind of flaky recently (for example: https://github.com/commercetools/ui-kit/runs/986022468?check_suite_focus=true), I think we need to find a way to make it more stable first.
Increasing the timeout was just an attempt.

PR #1498, opened 2020-08-14
refactor(i18n): compile i18n data to allow consumers to improve runtime performance

emmenko:
See https://formatjs.io/docs/getting-started/message-distribution


emmenko:
Update: the `compile-data` does not need to be committed, we can generate it on-the-fly both locally and on CI.

Issue #1476, opened 2020-08-12
improve how ContentNotification handles its children


YahiaElTai:
**Describe the bug**
`ContentNotification` uses `display:flex` for its children and that sometimes desirable but could also creates issues in other cases where for example: when displaying 2 HTML text elements as one inline paragraph

**Suggested solution**
One solution that comes to mind immediately is to add support for `display: block`

**Screenshots**
This screenshot shows how the "customer support" link is being displayed with `display: flex`
![image](https://user-images.githubusercontent.com/30247059/89981781-cce38000-dc74-11ea-91a8-d410d8fcf749.png)

This screenshot shows how the "customer support" link is being displayed when wrapping the 2 HTML elements in a container div (so the `ContentNotification` treat it as HTML element), so as you can see adding a wrapping container solves the issue but it might be helpful to have this handled by the uikit component itself. UiKit components should not enforce any layout on their children as I understand.
![image](https://user-images.githubusercontent.com/30247059/89981979-264baf00-dc75-11ea-8c7a-642ee59bc020.png)

jonnybel:
> UiKit components should not enforce any layout on their children as I understand

I agree, except for the obvious layout components, which isn't the case of this one.

The only good way I think we could solve this is by adding a prop like `containerType` which accepts the values `flex` and `block`, with the default being `flex` so that we won't be breaking any existing layouts.

YahiaElTai:
> > UiKit components should not enforce any layout on their children as I understand
> 
> I agree, except for the obvious layout components, which isn't the case of this one.
> 
> The only good way I think we could solve this is by adding a prop like `containerType` which accepts the values `flex` and `block`, with the default being `flex` so that we won't be breaking any existing layouts.

Sounds good to me! 🚀

YahiaElTai:
@commercetools/shield-team @tdeekens any feedback on the proposed solution?

tdeekens:
I would have liked to alternate between `props` so `intlMessage` does what we do now. `children` gives control. Then the consumer can use anything (e.g. a `Spacing`). 

That would however be breaking I assume.

I wonder if anybody wants to current behaviour/layout?

YahiaElTai:
> I would have liked to alternate between `props` so `intlMessage` does what we do now. `children` gives control. Then the consumer can use anything (e.g. a `Spacing`).
> 
> That would however be breaking I assume.
> 
> I wonder if anybody wants to current behaviour/layout?

[...]

PR #1501, opened 2020-08-16
chore: to wrap intlMessage in a div in ContentNotification

YahiaElTai:
This PR wraps IntlMessqge in a div in ContentNotification to have a `block` layout for intl message, whereas for children the layout will stay as-is `flex`.
This PR is a result of this issue #1476


packages/components/notifications/src/content-notification/README.md
```diff
@@ -16,10 +16,11 @@ import { ContentNotification } from '@commercetools-uikit/notifications';
...
-| `type`        | `string`       |    ✅    | `success`, `info`, `warning`, `error` | -       | Type of notification, determining its appearance.                    |
-| `children`    | `node`         | ✅ (\*)  | -                                     | -       | Message to be shown inside a notification panel.                     |
-| `intlMessage` | `intl message` | ✅ (\*)  | -                                     | -       | An intl message object that will be rendered with `FormattedMessage` |
+| Props           | Type           | Required | Values                                | Default | Description                                                          |
+| --------------- | -------------- | :------: | ------------------------------------- | ------- | -------------------------------------------------------------------- |
+| `type`          | `string`       |    ✅    | `success`, `info`, `warning`, `error` | -       | Type of notification, determining its appearance.                    |
+| `containerType` | `string`       |          | `flex`, `block`                       | `flex`  | Type of container for children, determining its layout.              |
```
tdeekens, line 22: MDN calls this a "display type". The CSS spec seems to differentiate between `display-inside` and `display-outside` values. `block` being a `display-outside` value and `flex` a `display-inside` value.

That brings me to two ideas:

1. `displayType` might be a better name
2. Are we just wanting to control row or column layout in the end?

I wonder if anybody ever assumed that you can layout two children inside the notification. So changing anything here isn't much breaking more of a bug fix and using `flex` and direction of sorts would be fine? Sorry for discussing it here again.

YahiaElTai: from my side, `displayType` sounds more appropriate and we are trying to control only the row layout.

YahiaElTai:
> Hmm shouldn't we simply wrap the `FormattedMessage` or `children` into a `div`?
> The current flex container can stay as it is, as long as the container rendering the message is not a flex box container...right? 🤔

This is the current solution we have on the MC but we thought maybe this should be also fixed on UiKit. fix is the wrong word here maybe make the component a bit more flexible and remove the need to add `divs` around

tdeekens:
> Hmm shouldn't we simply wrap the FormattedMessage or children into a div?

That's what I wonder now. We're trying to work around a breaking change. That's cool. However, I wonder if the current behaviour is not rather a bug. How intends to layout two children in a block? Maybe I am missing something 🦆.

emmenko:
My point was to question if we need this new prop instead of just "fixing" the rendered markup, for example by wrapping the rendered children into a `div`.

YahiaElTai:
> > Hmm shouldn't we simply wrap the FormattedMessage or children into a div?
> 
> That's what I wonder now. We're trying to work around a breaking change. That's cool. However, I wonder if the current behaviour is not rather a bug. How intends to layout two children in a block? Maybe I am missing something 🦆.

It's definitely not a bug, but more like asking too much of the component to handle our use cases. as we said before maybe the component shouldn't enforce layouts on its text children but since this is the case, the only options now are to either wrap a div around the children or add a new prop here, I'm fine with both as long as we agree on one solution.

YahiaElTai:
```javascript
<ContentNotification type="error">
  <FormattedMessage
     {...messages[resource]}
     values={{ limit, customersupport: getCustomerSupportLink }}
   >
      {(message) => <div>{message}</div>}
   </FormattedMessage>
</ContentNotification>
```

This was actually my solution to handle this case on the MC.

emmenko:
What I'm suggesting is to wrap this ⬇️ into a `div`

https://github.com/commercetools/ui-kit/blob/5734a2507bb7d5bc1a1eb32aab166aff43e8fc32/packages/components/notifications/src/content-notification/content-notification.js#L133-L137

Which is the same essentially of doing:

```js
<ContentNotification type="error">
  <div>
    Hello there
  </div>
</ContentNotification>
```

but it's done in the component itself, which also covers the case when passing the `intlMessage` prop.

YahiaElTai:
> What I'm suggesting is to wrap this ⬇️ into a `div`
> 
> https://github.com/commercetools/ui-kit/blob/5734a2507bb7d5bc1a1eb32aab166aff43e8fc32/packages/components/notifications/src/content-notification/content-notification.js#L133-L137
> 
> Which is the same essentially of doing:
> 
> ```js
> <ContentNotification type="error">
>   <div>
>     Hello there
>   </div>
> </ContentNotification>
> ```
> 
> but it's done in the component itself, which also covers the case when passing the `intlMessage` prop.

Yeah I think this is a good solution and works well for us. if everyone agrees on the solution, I will close the PR 🚀  @tdeekens @jonnybel

emmenko:
PS: You can do the changes in the PR, no need to open a new one.

YahiaElTai:
> PS: You can do the changes in the PR, no need to open a new one.

ah sorry I misunderstood your solution, if we add a `div` into the component itself that will be a breaking change because it will change the layout from `flex` to `block`, I thought you were speaking about adding a div into the MC when using the component as a consumer.

emmenko:
Well yes and no. I think what we can do at least is to wrap the rendered `FormattedMessage` in a `<div>`, when the `intlMessage` prop is provided. I think this can be considered a **fix**.

As for the `children`, it could potentially be a breaking change yes.
It's very unlikely that someone used two children, thus having a flex behavior.
We can either:
- keep it as-is, then consumers should pass the `<div>` on their own (if necessary).
- wrap it also in a `<div>`, and sell it as bug fix.

So I would suggest maybe to do as following:
- for the `intlMessage` case, we wrap it in a `<div>`.
- for the `children` case, we keep it as-is, but we document the need to pass a `<div>` in case the flex behavior is unwanted.
- add a TODO to fix the wrapping container in the next major release.

tdeekens:
Thanks @emmenko that's what I felt a bit too. That we sell it more as a bug fix as it feels odd to rely on the current layout which feels mostly broken. I think keeping the `children` case as is is good. Maybe when @YahiaElTai adopts this in the MC he can already check if/how many cases we have if any where we rely on it.

.changeset/pretty-onions-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/notifications': patch
+---
+
+chore: wrap intlMessage prop in a div to have a block layout instead of flex
```
emmenko, line 5:
```suggestion
fix: wrap `intlMessage` prop in a `<div>` to have a block layout instead of flex
```

PR #1506, opened 2020-08-18
feat(data-table): pass sizes as array of {key, width} to onColumnResized

mohib0306:
#### Summary

- Pass down the sizes of a table's columns as `Array<{key, width}>` so that the mapping of sizes to specific columns is clear and not index dependent.


mohib0306:
This is just a proof of concept and open for discussion and changes.

packages/components/data-table/src/use-manual-column-resizing-reducer.js
```diff
@@ -83,9 +83,13 @@ const useManualColumnResizing = (tableRef) => {
   // if the table element has been rendered and we haven't yet measured the columns
   if (state.tableRef.current && !state.sizes) {
     const renderedColumnMeasurements = [];
-    state.tableRef.current.querySelectorAll('th').forEach((header, index) => {
-      renderedColumnMeasurements[index] = header.getBoundingClientRect().width;
+    state.tableRef.current.querySelectorAll('th').forEach((header) => {
+      renderedColumnMeasurements.push({
+        key: header.getAttribute('data-testid').split(/-(.+)/)[1],
```
jonnybel, line 88: I'm wondering if we should add the id as a different data-attribute (`data-id` or `data-item-id`) since the `data-testid` is specifically for testing purposes and I'm not sure if we should mix things, even if it's kinda convenient here.

mohib0306: I also thought of having `id (data-id)` on the element level, and I think it makes sense. It would also relieve us of having this `str.split()` logic and the `testid`s can be changed in the future and won't affect the rest of the implementation

packages/components/data-table/src/data-table.spec.js
```diff
@@ -45,6 +45,16 @@ describe('DataTable', () => {
...
+  it('should be able to find headers by the data-id', () => {
+    const rendered = render(<DataTable {...baseProps} />);
+    expect(
+      rendered.container.querySelector('[data-id="title"]')
+    ).toBeInTheDocument();
+    expect(
+      rendered.container.querySelector('[data-id="year"]')
+    ).toBeInTheDocument();
+  });
+
```
mohib0306, line 57: I added these test just to make sure that `data-id` prop is working and is assigned to an element

jonnybel (commented):
Looks good, just needs a changeset now.

jonnybel (approved):
Nice ✨

PR #1508, opened 2020-08-18
fix(data-table): check for columns prop on runtime

emmenko:
Check for required `columns` prop for `DataTable` on runtime. This is important to allow compound components such as `DataTableManager` to inject the `columns` prop, as otherwise the prop types validation would fail.


emmenko:
Almost yeah. I additionally have a fix for the readme

PR #1517, opened 2020-08-20
fix(data-table): last column not being clipped when resizing

jonnybel:
#### Summary

<del>Currently, we're adding `overflow: hidden` to the cells only after columns are resized, but that logic got broken _somewhere on the way_ and since we couldn't really rely on the cells not having hidden overflow, because a user could resize the table at any time, might as well just **keep the rule always there**, to keep things simpler and the behaviour more consistent.</del>
<del>Possible bugs with clipped content inside of the cells will also be easier to spot instead of only happening after the table is resized.</del>

> https://github.com/commercetools/ui-kit/pull/1517#issuecomment-677656950

This PR fixes the logic to start clipping the cell contents, which was not triggering when starting to resize a column, which causes column cells to overflow out of the table container.

### DIFF
--- a/packages/components/data-table/src/cell.styles.js
+++ b/packages/components/data-table/src/cell.styles.js
@@ -97,7 +97,6 @@ const CellInner = styled.div`
 
   ${getPaddingStyle}
   ${getCellInnerStyles}
-  ${(props) => (props.shouldClipContent ? 'overflow: hidden;' : '')}
 `;
 
 const BaseCell = styled.td`
@@ -108,7 +107,7 @@ const BaseCell = styled.td`
     props.shouldRenderBottomBorder
       ? `1px solid ${vars.colorNeutral90};`
       : 'none'};
-  ${(props) => (props.shouldClipContent ? 'overflow: hidden;' : '')}
+  overflow: hidden;
 `;
 
 const BaseFooterCell = styled.td`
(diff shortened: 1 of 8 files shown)


Rombelirk (approved):

tdeekens (approved):

jonnybel:
Actually, I remembered why I was doing it that way: with `overflow:hidden` always there, the css grid will alocate space to the cells differently, allowing them to get clipped if there’s not enough horizontal space, instead of keeping the minimum needed space for them and adding a scrollbar to the whole table container.

When the table is resized, we start using fixed widths and its no longer the css grid that controls them, and we must then allow clipping the cells if the user wants to shrink the columns enough.

That being said, I'm gonna rework the PR to fix the logic 🙃

packages/components/data-table/src/cell.styles.js
```diff
@@ -97,6 +97,11 @@ const CellInner = styled.div`
 
   ${getPaddingStyle}
   ${getCellInnerStyles}
+
+  /* The following overflow rule should only enabled upon manual column resizing
```
emmenko, line 101:
```suggestion
  /* The following overflow rule should only be enabled upon manual column resizing
```

emmenko (approved):

PR #1526, opened 2020-08-24
feat(select-input): add support for icons on the left of the input value and placeholder

jonnybel:
#### Summary

This PR adds props to allow adding a big icon to the left of the placeholder or value of the `SelectInput`, for single-value inputs.

We have four variations of SelectInputs, and another four respective Fields, but it's the same change added to every one of them. Keep that in mind when reviewing.


packages/components/inputs/select-input/README.md
```diff
@@ -57,6 +57,8 @@ import SelectInput from '@commercetools-uikit/select-input';
...
 | `showOptionGroupDivider` | `boolean`          |    -     | -                            | `false` | Determines if option groups will be separated by a divider                                                                                                                                                                                                                                  |
+| `iconLeft`               | `node`             |    -     | -                            |         | Icon to display on the left of the value, when an option is selected                                                                                                                                                                                                                        |
+| `iconRight`              | `node`             |    -     | -                            |         | Icon to display on the left of the value, when an option is selected                                                                                                                                                                                                                        |
```
adnasa, line 61:
```suggestion
| `iconRight`              | `node`             |    -     | -                            |         | Icon to display on the right of the value, when an option is selected                                                                                                                                                                                                                        |
```

src/components/internals/create-select-styles.js
```diff
@@ -217,6 +217,9 @@ const valueContainerStyles = () => (base) => ({
   padding: '0',
   backgroundColor: 'none',
   overflow: 'hidden',
+  '& svg': {
+    flexShrink: 0,
```
jonnybel, line 221: This is to prevent the SVGs from shrinking when there's not enough horizontal space because they're inside a `Spacings.Inline`

src/components/internals/wrapper-with-icon/wrapper-with-icon.js
```diff
@@ -0,0 +1,50 @@
...
+  }),
+};
+WrapperWithIcon.displayName = 'WrapperWithIcon';
+
+export default WrapperWithIcon;
+
+/* eslint-disable react/display-name */
+const customComponents = {
+  SingleValue: (props) => <WrapperWithIcon {...props} type="singleValue" />,
+  Placeholder: (props) => <WrapperWithIcon {...props} type="placeholder" />,
```
mohib0306, line 46: Is this here just incase we show an icon along side with placeholder?

mohib0306 (commented):
Changes look good. Left a few questions

src/components/internals/wrapper-with-icon/wrapper-with-icon.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+const WrapperWithIcon = ({ children, ...props }) => {
+  const Component = getDefaultComponent(props.type);
+
+  return (
+    <React.Fragment>
+      {props.selectProps.iconLeft &&
+        React.cloneElement(props.selectProps.iconLeft, {
+          size: 'big',
```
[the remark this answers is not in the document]
jonnybel: Yep, the Icon is supposed to have always that size :)

src/components/internals/wrapper-with-icon/wrapper-with-icon.js
jonnybel: Yes, the icon should be shown even before selecting an option, and so it should appear on the placeholder too.

src/components/internals/wrapper-with-icon/wrapper-with-icon.js
```diff
@@ -0,0 +1,50 @@
...
+        })}
+      <span
+        css={css`
+          margin: ${customProperties.spacingXl + customProperties.spacingXs};
```
tdeekens, line 24: Does it mean placement is sort of out of rhythm of our spacings in general?

jonnybel: Kinda, but it's because `react-select` has it's own layout system and uses absolute positioning to place the value inside the input.
The Icon always has 24px (= SpacingsXL), and we want a margin of `spacingsXs` between the icon and the label.

Perhaps I should just use `24px` and add a comment, to be clear?

tdeekens: I think the math is fine. Maybe just add a comment why it's needed :)

tdeekens (approved):

adnasa (approved):
looking good

PR #1527, opened 2020-08-25
Add generate readmes for stamp, loading spinner and tag.

adnasa:
#### Summary

- uses `generate-readme` for stamp, loading spinner and tag.


packages/components/loading-spinner/README.md
```diff
@@ -1,24 +1,46 @@
...
 ```
 
 ## Properties
 
-| Props      | Type     | Required | Values   | Default | Description                                         |
-| ---------- | -------- | :------: | -------- | ------- | --------------------------------------------------- |
-| `scale`    | `string` |          | 's', 'l' | `l`     | Defines the size of the loading spinner's container |
-| `children` | `string` |          | -        | -       | Used as the text for the loader                     |
-
-Main Functions and use cases are:
```
tdeekens, line 22: Are we supporting this section somehow in the generation?

packages/components/tag/src/tag.js
```diff
@@ -231,15 +231,39 @@ const Tag = (props) => {
 };
 
 Tag.propTypes = {
+  /**
+   * Indicates color scheme of the tag.
+   */
   type: PropTypes.oneOf(['normal', 'warning']),
+  /**
+   * Styles object spread onto the tag body..
```
tdeekens, line 239:
```suggestion
   * Styles object spread onto the tag body.
```

packages/components/tag/src/tag.js
```diff
@@ -231,15 +231,39 @@ const Tag = (props) => {
...
+   */
   styles: PropTypes.shape({
     body: PropTypes.object,
   }).isRequired,
+  /**
+   * Location the tag links to when enabled.
```
tdeekens, line 245:
```suggestion
   * Link of the tag when not disabled.
```

packages/components/tag/src/tag.js
```diff
@@ -231,15 +231,39 @@ const Tag = (props) => {
...
+   */
   linkTo: PropTypes.string,
+  /**
+   * Disable the tag element along with the option to remove it.
+   */
   isDisabled: PropTypes.bool,
+  /**
+   * Called when remove button is clicked.
```
tdeekens, line 253:
```suggestion
   * Called when tag is removed.
```

packages/components/loading-spinner/src/loading-spinner.js
```diff
@@ -72,7 +72,13 @@ LoadingSpinner.defaultProps = {
   scale: 'l',
 };
 LoadingSpinner.propTypes = {
-  scale: PropTypes.oneOf(['s', 'l']).isRequired,
+  /**
+   * Set the size of the loading spinner.
+   */
+  scale: PropTypes.oneOf(scale).isRequired,
```
emmenko, line 78: 👌

packages/components/loading-spinner/README.md
adnasa: This could be part of the description. good catch

packages/components/loading-spinner/src/loading-spinner.js
```diff
@@ -72,7 +72,13 @@ LoadingSpinner.defaultProps = {
...
+  /**
+   * -
+   */
```
emmenko, line 81: You can just omit this

packages/components/loading-spinner/README.md
adnasa: the **description** section is clear enough, so I'm thinking this information is a bit redundant.
I'll see what I can do, since the description is weird.

packages/components/tag/src/tag.js
```diff
@@ -231,15 +231,39 @@ const Tag = (props) => {
...
+   */
   onRemove: PropTypes.func,
+  /**
+   * Called when tag element is clicked. This is not called with remove button is clicked.
```
emmenko, line 257:
```suggestion
   * Called when tag element is clicked. This is not called with remove button is clicked.
   * <br>
   * Signature: `(event) => void`
```

For example: https://github.com/commercetools/ui-kit/blob/c20ae492210482697c7c5e181e983e561ff86654/packages/components/data-table/src/data-table.js#L203-L208

packages/components/tag/src/tag.js
```diff
@@ -231,15 +231,39 @@ const Tag = (props) => {
...
+  /**
+   * Indicates color scheme of the tag.
+   */
   type: PropTypes.oneOf(['normal', 'warning']),
+  /**
+   * Styles object spread onto the tag body..
+   */
   styles: PropTypes.shape({
     body: PropTypes.object,
   }).isRequired,
```
emmenko, line 243: This shouldn't be required

packages/components/tag/README.md
```diff
@@ -1,43 +1,60 @@
...
+## Properties
 
-- Showing a set _example: List of categories a product is in_
+| Props                  | Type                                                                         | Required | Default         | Description                                                                           |
+| ---------------------- | ---------------------------------------------------------------------------- | :------: | --------------- | ------------------------------------------------------------------------------------- |
+| `type`                 | `enum`<br>Possible values:<br>`'normal' \| 'warning'`                        |          | `'normal'`      | Indicates color scheme of the tag.                                                    |
+| `styles`               | `object`                                                                     |          | `{ body: {}, }` | Styles object spread onto the tag body..                                              |
+| `styles.body`          | `object`                                                                     |          | `{ body: {}, }` |                                                                                       |
```
emmenko, line 54: I think we can remove the `styles: { body: {} }` from the default props.

emmenko (commented):
Thanks!

Just a bunch of suggestions, otherwise looks good 👌

packages/components/tag/src/tag.js
adnasa: hmm..
the thing is, is that it's the consumer that executes the remove action as a result of clicking the remove button.

The wording you propose, suggests that `onRemove` is called after the UI-kit removes the tag.

tdeekens: Ok. Thanks.

packages/components/tag/README.md
jonnybel: Yep, and this should "fix" the Percy diff

packages/components/loading-spinner/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@commercetools-uikit/loading-spinner",
   "version": "10.30.1",
-  "description": "",
+  "description": "A spinner animation to indicating pending requests for the user.",
```
jonnybel, line 4:
```suggestion
  "description": "A spinner animation to indicate pending requests for the user.",
```

packages/components/tag/src/tag.js
```diff
@@ -231,22 +231,46 @@ const Tag = (props) => {
...
+   * Content rendered within the tag
+   */
   children: PropTypes.node.isRequired,
 };
 Tag.defaultProps = {
   type: 'normal',
-  styles: {
-    body: {},
-  },
+  styles: {},
```
jonnybel, line 273: This default can be removed.

jonnybel (commented):
Thanks for these 😌

packages/components/tag/src/tag.js
```diff
@@ -231,22 +231,45 @@ const Tag = (props) => {
...
+   */
   linkTo: PropTypes.string,
+  /**
+   * Disable the tag element along with the option to remove it.
+   */
   isDisabled: PropTypes.bool,
+  /**
+   * Called when remove button is clicked.
+   */
   onRemove: PropTypes.func,
```
emmenko, line 255: Provide the function signature here as well please

jonnybel (approved):

tdeekens (approved):

emmenko (approved):
Thanks!! 💯

tdeekens:
CleanShot 2020-08-25 at 18 17 08@2x

/cc @adnasa

adnasa:
@tdeekens my bad. I'll push a commit now

adnasa:
the build was failing because `<Tag />` contained `<TagBody />` where the props were undocumented.
Since we don't export this component, it didn't make sense to document it for the sake of it. What I did instead was extract the `<TagBody />` and have that imported into `<Tag />`

packages/components/tag/src/tag.js
```diff
@@ -234,7 +114,7 @@ Tag.propTypes = {
   type: PropTypes.oneOf(['normal', 'warning']),
```
emmenko, line 114: What happened to the JS docs of the `<Tag>` component?

adnasa: oops 😅

PR #1528, opened 2020-08-25
fix(data-table): table column sizes not being properly reset

jonnybel:
#### Summary
`DataTable`
This PR fixes a bug where the table column resizing could break after changing the number of columns, because the column measurements were not being properly reset.

### DIFF
--- a/packages/components/data-table/src/data-table.js
+++ b/packages/components/data-table/src/data-table.js
@@ -40,30 +40,25 @@ const DataTable = (props) => {
   const tableRef = React.useRef();
   const columnResizingReducer = useManualColumnResizing(tableRef);
 
-  // if the table has been manually resized
+  // if the table columns have been measured
   // and if the list of columns, their width field, or the isCondensed prop has changed
   // then we need to reset the resized column widths
-  const hasTableBeenResized = columnResizingReducer.getHasTableBeenResized();
   const columnsInfo = getColumnsLayoutInfo(props.columns);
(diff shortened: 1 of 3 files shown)


packages/components/data-table/src/data-table.js
```diff
@@ -40,30 +40,25 @@ const DataTable = (props) => {
   const tableRef = React.useRef();
   const columnResizingReducer = useManualColumnResizing(tableRef);
 
-  // if the table has been manually resized
+  // if the table columns have been measured
```
jonnybel, line 43: Resetting the column measurements (to perform a new one) should be done not only after a table has been resized, but every time the something that affects the columns layout is changed.
Otherwise, the next resize might be using outdated measurements.

packages/components/data-table/src/data-table.story.js
```diff
@@ -403,7 +403,9 @@ storiesOf('Components|DataTable', module)
               : null
           }
           maxHeight={castToNumberIfPossible(
-            text('maxHeight', 'calc(100vh - 200px)')
+            // in this example, the table has a dynamic height equal to the viewport height minus 200px
+            // floored to a minimum value of 400px
+            text('maxHeight', 'max(400px, calc(100vh - 200px))')
```
jonnybel, line 408: Unrelated, but added this just to improve the quality of the example and show what's possible to do.

mohib0306 (approved):
💯

PR #1529, opened 2020-08-25
fix(data-table): Cell clickable area

jonnybel:
#### Summary

`DataTable`
Fixed bug where clicking near the edge of a cell wouldn't trigger `onRowClick` if the cell was substantially taller than its own content.

Kapture 2020-08-25 at 15 02 09


mohib0306 (approved):
Thanks 💯

PR #1531, opened 2020-08-26
fix(data-table): header icon vertical placement

jonnybel:
#### Summary

This PR fixes a bug in `DataTable` where, if a column header has an icon but has an empty label, it becomes badly positioned vertically, like so:
image


tdeekens (commented):
You could add a VRT for it to ensure no regressions.

tdeekens (approved):

PR #1532, opened 2020-08-26
fix(github): action to use recent puppeteer action

tdeekens:
An attempt to solve this error.

CleanShot 2020-08-26 at 17 24 38@2x

Alternatively we might `sudo apt-get install libxss1`.


emmenko (approved):

adnasa (approved):

jonnybel (approved):
🤗

PR #1541, opened 2020-09-02
Chore: Group select inputs and fields in the storybook

mohib0306:
#### Summary
- `SelectInput`s and `SelectField`s are group under their own sub-menus in the storybook to ease the find. 
- Also fixed a bug in the options filtering in the `async-select-input` and `async-createable-select-input`


jonnybel (approved):
I like it 😄

PR #1542, opened 2020-09-02
Introduce SearchSelectInput 

mohib0306:
#### Summary

- `SearchSelectInput` is being added to the select inputs family. It is built upon the `AsyncSelectInput` component with some pre-defined and non-changeable props. i.e. `defaultOptions` etc.


packages/components/inputs/search-select-input/src/search-select-input-option.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
jonnybel, line 1: Are these just being used in the Examples for Storybook?
If so, I think we should put "example" on the filename & component name and/or leave a comment about it, just so it's clearer that this is not actually part of the component itself 🙂

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    <AsyncSelectInput
+      horizontalConstraint={props.horizontalConstraint}
+      // react-select uses "id" (for the container) and "inputId" (for the input),
+      // but we use "id" (for the input) and "containerId" (for the container)
+      // instead.
+      // This makes it easier to less confusing to use with <label />s.
+      id={props.containerId}
+      inputId={props.id}
+      aria-label={props['aria-label']}
```
jonnybel, line 32: We could use `{...filterAriaAttributes(props)}` for these.

I know the other SelectInputs don't do it but they should, so let's already simplify this one 😄

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+    : intl.formatMessage(messages.noOptionsMessageWithInputValue);
+
+  return (
+    <AsyncSelectInput
+      horizontalConstraint={props.horizontalConstraint}
+      // react-select uses "id" (for the container) and "inputId" (for the input),
+      // but we use "id" (for the input) and "containerId" (for the container)
+      // instead.
+      // This makes it easier to less confusing to use with <label />s.
+      id={props.containerId}
```
jonnybel, line 30: Isn't this done by the `AsyncSelectInput` itself? It's expecting a `containerId` and an `id`, so we should just forward those props as they are.

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+      name={props.name}
+      placeholder={props.placeholder}
+      tabIndex={props.tabIndex}
+      iconLeft={props.iconLeft}
+      inputValue={props.inputValue}
+      value={props.value}
+      backspaceRemovesValue={props.backspaceRemovesValue}
+      hasWarning={props.hasWarning}
+      hasError={props.hasError}
+      isDisabled={props.isDisabled || props.isReadOnly}
```
jonnybel, line 43: This is also done by the `AsyncSelectInput`, so we can just forward them.

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+      isClearable={props.isClearable}
+      isOptionDisabled={props.isOptionDisabled}
+      isMulti={props.isMulti}
+      isSearchable={true}
```
jonnybel, line 47: I'm thinking we could improve dev experience here by separating which props are we forwarding and which ones are we setting specifically for this SearchSelectInput. At the very least, we should sort them separately and add comments about them.

Perhaps we could either have a list of the `AsyncSelectInput` props to filter and forward directly, or just spread all remaining props, since they are already "filtered" and type-checked by the `AsyncSelectInput`.
This way, maintaining this component would be much easier when changes are made to the props of `AsyncSelectInput`, while also making the code of this component more "focused" on the value it brings by itself (adding a comment about the checking the implementation of the underlying AsyncSelectInput if needed)

Furthermore, for the README for this component, we could just document props that are new or have defaults, and refer to the documentation of the `AsyncSelectInput`, without duplicating everything,

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+import AsyncSelectInput from '@commercetools-uikit/async-select-input';
+import messages from '../../../../../src/components/internals/messages/select';
+import createSelectStyles from '../../../../../src/components/internals/create-select-styles';
+
+const useTheme = () => useContext(ThemeContext);
+
+const SearchSelectInput = (props) => {
+  const intl = useIntl();
+  const theme = useTheme();
+  const noOptionsMessageWithOutInputValue = props.initialSuggestionMessage
```
jonnybel, line 16: Just to keep the camelCasing consistent 😁 

```suggestion
  const noOptionsMessageWithoutInputValue = props.initialSuggestionMessage
```

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,151 @@
...
+        DropdownIndicator: SearchIconDropdownIndicator,
+      }}
+      noOptionsMessage={({ inputValue }) =>
+        inputValue === ''
+          ? noOptionsMessageWithOutInputValue
+          : noOptionsMessageWithInputValue
+      }
+      styles={createSelectStyles(
+        {
+          menuPortalZIndex: props.menuPortalZIndex,
```
jonnybel, line 60: If this needed if we pass the `menuPortalZIndex` prop and let the underlying component do it itself?

src/components/internals/search-icon-dropdown-indicator/search-icon-dropdown-indicator.js
```diff
@@ -0,0 +1,36 @@
...
+
+const handleMouseDown = (event, props) => {
+  event.stopPropagation();
+  event.preventDefault();
+  return props.selectProps.onSearchIconClick(props.selectProps.inputValue);
+};
+
+const SearchDropdownIndicator = (props) => {
+  return (
+    <SearchIconWrapper
```
jonnybel, line 14: We should make sure this is accessible.

You can use the `AccessibleButton` component as the wrapper and use `onClick` instead of onMouseDown 🙂

packages/components/inputs/search-select-input/src/search-select-input.story.js
```diff
@@ -0,0 +1,137 @@
...
+                  isDisabled={boolean('isDisabled', false)}
+                  isReadOnly={boolean('isReadOnly', false)}
+                  isMulti={isMulti}
+                  noOptionsMessage={() =>
+                    text('No Option Message (non-empty input)', 'No options')
+                  }
+                  initialSuggestionMessage={() =>
+                    text(
+                      'Initial suggestion message (empty input)',
+                      'Please something to load options'
```
jonnybel, line 103:
```suggestion
                      'Please type something to load options'
```

jonnybel (commented):
Looks promising :) 

I made some remarks regarding the code 🤓

packages/components/inputs/search-select-input/src/search-select-input-option.js
mohib0306: I want to showcase it to the designers first that the `Option` component is extendable and not sure if we need to provide the option of choosing between which `Option` component to use, since the use-cases differs a lot. I will move to either the story itself or rename the file after the demo.

jonnybel: Sounds good 😄 

It could actually be nice to export them as pre-made styled options :)

src/components/internals/search-icon-dropdown-indicator/search-icon-dropdown-indicator.js
mohib0306: Used that but the the custom handle for the `onClick` of `AccessibleButton` does not stop the event propagation, even though if you use `event.stopPropagation(); event.preventDefault();`

jonnybel: Ah, I think react-select uses the onMouseDown event handler internally.
Can we stopPropagation & preventDefault for that event and still use the onClick?

packages/components/inputs/async-select-input/src/async-select-input.js
```diff
@@ -197,7 +202,10 @@ AsyncSelectInput.propTypes = {
   onBlur: PropTypes.func,
   onChange: PropTypes.func.isRequired,
   onFocus: PropTypes.func,
+  openMenuOnFocus: PropTypes.bool,
```
tdeekens, line 205: Maybe we can define a shared naming here as also defaulting. We often have disable* which I somewhat subjectivity like more than enable*. It can still default to true. Then all these behavior flags could be disableOpenMenuOnClick etc.

emmenko:
PS: the README is missing 🙂

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,201 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useIntl } from 'react-intl';
+import {
+  SafeHTMLElement,
+  filterAriaAttributes,
+} from '@commercetools-uikit/utils';
+import SearchIconDropdownIndicator from '../../../../../src/components/internals/search-icon-dropdown-indicator';
+import AsyncSelectInput from '../../async-select-input';
```
mohib0306, line 9: As you can notice that a relative import is used instead of using the package import. The reason is that there are some changes introduced into the `select-utils` and `async-select-input` components, however they are not yet published and hence the import of the current version will lack those changes. I see two ways forward.

1. Let the UX/UI review finish, which will make certain that we do not need any more changes in the underlying components. Publish this component with with relative imports. After it is published (the changes for the other components will also be published), create a new PR for this component, replacing the relative import with the package imports.
2. Let the UX/UI review finish, which will make certain that we do not need any more changes in the underlying components. Put this PR on hold. Create a new PR for the underlying components (`async-select-input and select-utils`), publish them. Then resume this PR and with the packages import and then publish this component

packages/components/inputs/search-select-input/src/search-select-input-option.js
mohib0306: Right now they are not exported outside the component, but we could put it in the `select-utils` and export it there and then use it back here. 
p.s. follow this conversation as well https://github.com/commercetools/ui-kit/pull/1542/files#r484823827

packages/components/inputs/search-select-input/src/search-select-input.story.js
```diff
@@ -0,0 +1,150 @@
...
+                    </b>{' '}
+                    and{' '}
+                    <b>
+                      <i>label</i>
+                    </b>{' '}
+                    fields with 2 seconds of delay
+                  </p>
+                  <b>Data used:</b>
+                  <pre>{JSON.stringify(colourOptions, undefined, 2)}</pre>
+                </div>
```
mohib0306, line 132: This was added to give a clue what data is available

packages/components/inputs/search-select-input/src/messages.js
```diff
@@ -0,0 +1,15 @@
+import { defineMessages } from 'react-intl';
```
mohib0306, line 1: The `search-select-input` will have different default message to those that are being used by the rest of the select inputs/fields. Therefore they were added here

packages/components/inputs/search-select-input/docs/additional-info.md
```diff
@@ -0,0 +1,11 @@
+## Additional information
```
adnasa, line 1: I would omit heading this for now

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,202 @@
...
+   */
+  menuShouldBlockScroll: PropTypes.bool,
+  /**
+   * Handle blur events on the control
+   */
+  onBlur: PropTypes.func,
+  /**
+   * Called with a fake event when value changes. The event's `target.name` will be the `name` supplied in props. The event's `target.value` will hold the value. The value will be the selected option, or an array of options in case `isMulti` is `true`.
+   */
+  onChange: PropTypes.func.isRequired,
```
adnasa, line 157: please add the signature

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Handle focus events on the control
+   */
+  onFocus: PropTypes.func,
+  /**
+   * Handle change events on the input
+   */
+  onInputChange: PropTypes.func,
```
adnasa, line 165: add the signature for this as well.

mohib0306:
Rebased to master to have the latest changes from the readmes

packages/components/inputs/search-select-input/docs/additional-info.md
```diff
@@ -0,0 +1,8 @@
+The underlying `@commercetools-uikit/async-select-input` is built on top of [`react-select`](https://github.com/JedWatson/react-select) v3. `@commercetools-uikit/async-select-input` supports mostly same properties as `react-select` with some minor changes in the behaviour of some of the props. The `@commercetools-uikit/search-select-input` which is built on top `@commercetools-uikit/async-select-input` has predefined values for some the props. The props that has predefined values in `@commercetools-uikit/search-select-input` are as follows:
```
jonnybel, line 1:
```suggestion
The underlying `@commercetools-uikit/async-select-input` is built on top of [`react-select`](https://github.com/JedWatson/react-select) v3. `@commercetools-uikit/async-select-input` supports mostly the same properties as `react-select` with some minor changes in the behaviour of some of the props. The `@commercetools-uikit/search-select-input` which is built on top `@commercetools-uikit/async-select-input` has predefined values for some the props. The props that have predefined values in `@commercetools-uikit/search-select-input` are as follows:
```

packages/components/inputs/search-select-input/docs/additional-info.md
```diff
@@ -0,0 +1,8 @@
...
+
+- `components.DropdownIndicator`: Default dropdown indicator is replaced with search icon indicator
+- `components.Option`: The shown in the dropdown menu can be on of `singled-lined`, `brief-detailed`, or `extended-detailed` types. However if someone still wants to have a different option layout, they can still pass their own `components.Option` to `@commercetools-uikit/search-select-input`
```
jonnybel, line 4:
```suggestion
- `components.Option`: The option shown in the dropdown menu can be one of `singled-lined`, `brief-detailed`, or `extended-detailed` types. However, if someone still wants to have a different option layout, they can still pass their own `components.Option` to `@commercetools-uikit/search-select-input`
```

packages/components/inputs/search-select-input/docs/additional-info.md
```diff
@@ -0,0 +1,8 @@
...
+- `isSearchable`: `true`
+- `defaultOptions`: `false`
+
+See the [official documentation](https://react-select.com/components) for more information about the props they receive.
```
jonnybel, line 8:
```suggestion
See the [official documentation](https://react-select.com/components) for more information about the available props.
```

packages/components/inputs/search-select-input/src/search-select-input-option.js
```diff
@@ -0,0 +1,93 @@
...
+import { components } from 'react-select';
+import { NO_VALUE_FALLBACK, SELECT_DROPDOWN_OPTION_TYPES } from './constants';
+
+export const FullDetailedSearchSelectInputOption = (props) => {
+  const { data } = props;
+  return (
+    <components.Option {...props}>
+      <Spacings.Stack scale="xs">
+        <Text.Detail isBold={true}>
+          {data.label || NO_VALUE_FALLBACK}
```
jonnybel, line 14: Should we allow setting the fallback value as a prop? 
It might be a bit restrictive to have it fixed to '----', which can easily change in the future or depend on the context.

packages/components/inputs/search-select-input/src/search-select-input.js
jonnybel: I'd say you can already move forward with 2. and create a PR for the async select-input and utils 🙂

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,213 @@
...
+    props.noOptionsMessage ||
+    (() => intl.formatMessage(messages.noOptionsMessage));
+  const loadingMessage =
+    props.loadingMessage ||
+    (() => intl.formatMessage(messages.loadingOptionsMessage));
+
+  return (
+    <AsyncSelectInput
+      {...props}
+      {...filterAriaAttributes(props)}
```
jonnybel, line 25: No point doing this if we're already spreading all props right above this line 😄 
They'll get filtered by the AsyncSelectInput, so it's fine.

packages/components/inputs/search-select-input/docs/usage-example.js
```diff
@@ -0,0 +1,27 @@
...
+      horizontalConstraint="m"
+      optionType="single-lined"
+      isAutofocussed={false}
+      backspaceRemovesValue={true}
+      isClearable={true}
+      isDisabled={false}
+      isReadOnly={false}
+      isMulti={true}
+      noOptionsMessage="No exact match found"
+      loadingMessage="loading extact matches"
```
jonnybel, line 18:
```suggestion
      loadingMessage="loading exact matches"
```

packages/components/inputs/search-select-input/src/search-select-input.story.js
```diff
@@ -0,0 +1,146 @@
...
+                  cacheOptions={boolean('cacheOptions', false)}
+                  iconLeft={
+                    iconLeft ? React.createElement(iconLeft) : undefined
+                  }
+                />
+                <div>
+                  <p>
+                    The options fetching uses the data (given below) to perform
+                    an exact match. It is case sensitive and it perfrom search
+                    based on{' '}
```
jonnybel, line 116:
```suggestion
                    In this example, our `loadOptions` function uses the data (given below) to perform
                    an exact match. It is case sensitive and it performs a search
                    based on{' '}
```

Just to make it clearer that this is just for the Storybook example, and not already set internally in the component.

packages/components/inputs/search-select-input/src/search-select-input-option.js
mohib0306: We can `fallback` prop and if given that will get precedence

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: I was about to ask the same but I forgot to comment on in

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: https://github.com/commercetools/ui-kit/pull/1554
https://github.com/commercetools/ui-kit/pull/1555

mohib0306:
Closing this PR and creating a new one with update `package.json` and other changes

PR #1550, opened 2020-09-10
style: fix typo on `authorized`

adnasa:
#### Summary

- fix typo on `authorized`


jonnybel (approved):
It is really a typo? Seems like it's just a regional difference and both forms are valid.

Fine by me tho, for consistency's sake 🙂

adnasa:
in `application-kit` at least, we expose hooks, hocs and others as `authorized`, so it makes sense to me to keep things consistent :)

PR #1551, opened 2020-09-10
fix(number-input): restore xs as a possible value for horizontalConstraint

jonnybel:


tdeekens (approved):

adnasa (approved):
don't forget changeset

PR #1540, opened 2020-09-02
refactor(fields/*): generate readmes.

adnasa:
#### Summary

- adds `generate-readme` for `fields/*`

#### Description

I will migrate all `fields/*` before I set this to `ship it`.
Will leave this WIP in the mean time

PR #1553, opened 2020-09-10
fix: restore horizontal constraint

adnasa:
#### Summary

- restore `horizontalConstraint` allowed values, caused by #1540


adnasa:
that ought to restore them all. I went through the entire #1540

packages/components/inputs/select-input/src/select-input.js
```diff
@@ -181,7 +181,7 @@ SelectInput.defaultProps = {
 };
 
 SelectInput.propTypes = {
-  horizontalConstraint: PropTypes.oneOf(['s', 'm', 'l', 'xl', 'scale']),
+  horizontalConstraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
```
jonnybel, line 184: I think this size was not valid for SelectInput

packages/components/fields/date-field/src/date-field.js
```diff
@@ -28,7 +28,7 @@ class DateField extends React.Component {
     /**
      * Horizontal size limit of the input fields.
      */
-    horizontalConstraint: PropTypes.oneOf(['s', 'm', 'l', 'xl', 'scale']),
+    horizontalConstraint: PropTypes.oneOf(['m', 'l', 'xl', 'scale']),
```
tdeekens, line 31: Verified with https://github.com/commercetools/ui-kit/pull/433

tdeekens (approved):

jonnybel (approved):

PR #1554, opened 2020-09-11
Add loadingMessage prop to asyc-select input and field components 

mohib0306:
#### Summary

- `AsyncSelectInput` => Added `loadingMessage` prop
- `AsyncSelectField` => Added `loadingMessage` prop

p.s. Needed for SearchSelectInput


packages/components/inputs/async-select-input/src/async-select-input.js
```diff
@@ -35,7 +35,10 @@ const AsyncSelectInput = (props) => {
 
   const placeholder =
     props.placeholder || intl.formatMessage(messages.placeholder);
-
+  const loadingMessage = () =>
+    props.loadingMessage
+      ? props.loadingMessage()
```
adnasa, line 40: not enough to have this as string? 
what benefits does it have being a function over a string?

mohib0306: The `react-select` accepts a callback. Thats what is it a function

mohib0306: We can ask user to send a string and then we change it to a callback when passing it to the react-select's component

adnasa: sounds reasonable to me

mohib0306: So I check the `noOptionsMessage` prop, which has similar behaviour as `loadingMessage`. The AsyncSelectInput expects a `func` and the users in the MC always sends a callback. I think we should keep it consistent otherwise it won't look good. But I will change this to expect a string  and hopefully in the future we move the `noOptionsMessage` to string as well

adnasa: consistent in terms of having everything a function?

mohib0306: consistent in terms of having similar to what react-select is expecting. i.e. here either both callback functions or both strings

jonnybel: Ah, in that case I think it makes sense to keep it the way out `react-select` and our other `Async` selects work.

Doesn't make sense to change the others in this PR and also it would be breaking.

mohib0306: of-course it won't be changed. It is big breaking change.

adnasa: it doesn't make sense to change the others, I agree.
if we want to keep consistent towards `react-select` API, sure.

but preferably, it would been great from a consumer side to pass just a string and on our side we forward it as a callback just as you originally suggested.
just my two cents.

jonnybel (approved):

packages/components/inputs/async-select-input/src/async-select-input.js
jonnybel: We could also, for this one, check the type of the passed prop and treat it accordingly. This way, users can keep the same expectation as they have for the other Selects, but are also free to use the "easier" way of just passing the string :)

adnasa:
hmm
looking at the signature of `loadingMessage` https://react-select.com/props
it accepts that the return value of `loadingMessage` is `null`. With our implementation, we fallback directly to `intl.formatMessage...` if the consumer returns `null` and actually wants it that way.

Sorry for the mess, but you were right. let's make this a direct map to `react-select` loadingMessage.

TLDR; let's make `loadingMessage` a function again (no pun intended)

packages/components/inputs/async-select-input/src/async-select-input.js
mohib0306: Added support for both `string` and `func` types of `loadingMessage`. This way we can give flexibility of what to use and at the same time be consistent with `react-select`

jonnybel (approved):

packages/components/inputs/async-select-input/src/async-select-input.js
```diff
@@ -35,8 +35,12 @@ const AsyncSelectInput = (props) => {
 
   const placeholder =
     props.placeholder || intl.formatMessage(messages.placeholder);
-  const loadingMessage = () =>
-    props.loadingMessage || intl.formatMessage(messages.loadingOptions);
+  const loadingMessage = () => {
```
adnasa, line 38: hm, yeah this works for me 🤷

adnasa (approved):

PR #1558, opened 2020-09-11
fix(icon): to prevent icon from shrinking inside of flex container

Rombelirk:
#### Summary

This PR prevents `Icon` component from shrinking when inside of resized flex container.

## Description

Bildschirmfoto 2020-09-08 um 14 24 34


jonnybel (approved):
Thanks!

mohib0306 (approved):

PR #1564, opened 2020-09-14
Feat: Add SearchSelectInput component

mohib0306:
#### Summary

- `SearchSelectInput` is being added to the select inputs family. It is built upon the `AsyncSelectInput` component with some pre-defined and non-changeable props. i.e. `defaultOptions` etc.


jonnybel (approved):
Alright, this version already looks good to go :) I like how simple the component itself actually became.

I've already approved the new VRTs.

packages/components/inputs/search-select-input/src/constants.js
```diff
@@ -0,0 +1,7 @@
+export const NO_VALUE_FALLBACK = '----';
```
jonnybel, line 1: Is this being used anywhere?

mohib0306: nope. I will remove that

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,210 @@
...
+    <AsyncSelectInput
+      {...props}
+      components={{
+        // eslint-disable-next-line react/display-name
+        Option: (optionInnerProps) => (
+          <CustomSelectInputOption
+            optionType={props.optionType}
+            optionInnerProps={optionInnerProps}
+          />
+        ),
```
tdeekens, line 30: I sometimes wonder if this should live outside. So `const components = { Option: OptionComponent }` and `const OptionComponent = props =>` cause this may cause quite some re-renders.

packages/components/inputs/search-select-input/src/search-select-input.spec.js
```diff
@@ -0,0 +1,320 @@
...
+});
+
+it('should have focus automatically when isAutofocussed is passed', () => {
+  const { getByLabelText } = renderInput({ isAutofocussed: true });
+  expect(getByLabelText('Fruit')).toHaveFocus();
+});
+
+it('should call onFocus when the input is focused', () => {
+  const onFocus = jest.fn();
+  const { getByLabelText } = renderInput({ onFocus });
```
tdeekens, line 83: Can't we use screen for this? I recall doing some refactorings to use it in UIKit a while back. The only cause when not to use screen is when you have custom queries not baked into the render.

packages/components/inputs/search-select-input/src/search-select-input.spec.js
```diff
@@ -0,0 +1,320 @@
...
+  it('should render a select input with preselected values', () => {
+    const { getByLabelText, getByText } = renderInput({
+      value: yellowOption,
+      loadOptions: () => Promise.resolve(groupedOptions),
+    });
+    const input = getByLabelText('Fruit');
+    expect(input).toBeInTheDocument();
+    expect(getByText(yellowOption.label)).toBeInTheDocument();
+  });
+});
```
tdeekens, line 320: Great test coverage.

packages/components/inputs/search-select-input/src/search-select-input.spec.js
mohib0306: Updated the test to use `screen`

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: added the requested changes

tdeekens (approved):

emmenko:
Changesets is missing

jonnybel (approved):

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,215 @@
...
+  const noOptionsMessage =
+    props.noOptionsMessage ||
+    (() => intl.formatMessage(messages.noOptionsMessage));
+  const loadingMessage =
+    props.loadingMessage || intl.formatMessage(messages.loadingOptionsMessage);
+  const placeholder =
+    props.placeholder || intl.formatMessage(messages.placeholderMessage);
+
+  // eslint-disable-next-line react/display-name
+  const OptionComponent = (optionInnerProps) => (
```
tdeekens, line 24: We still re-define this component at each render within `SearchSelectInput` which will mean it will mount-and unmount a lot to my understanding. Can't be just move it to L12?

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -0,0 +1,215 @@
...
+      optionType={props.optionType}
+      optionInnerProps={optionInnerProps}
+    />
+  );
+
+  return (
+    <SearchSelectInputWrapper>
+      <AsyncSelectInput
+        {...props}
+        components={{
```
tdeekens, line 35: Nit:

```js
const components = React.useMemo(() => ({
   Option: OptionComponent,
   ...props.component,
   DropdownIndicator: SearchIconDropdownIndicator
}), [props.component, OptionComponent, SearchIconDropdownIndicator])
```

under the hope that a static reference helps the underlying AsyncSelectInput.

packages/components/inputs/search-select-input/src/search-select-input.spec.js
```diff
@@ -0,0 +1,327 @@
...
+      const onChange = jest.fn();
+      renderInput({
+        onChange,
+        isMulti: true,
+        value: [],
+      });
+      const input = screen.getByLabelText('Fruit');
+      fireEvent.focus(input);
+      fireEvent.keyDown(input, { key: 'ArrowDown' });
+      fireEvent.change(input, { target: { value: 'mango' } });
```
tdeekens, line 250: I think eventually ui-kit should evaluate using `user-event` for newer tests which I think is now exposed from testing-library-react.

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: will add these changes. however it will also either wrapping the `OptionComponent` in its callback (since it is used inside memo and it has `optionType` dependency) or moving it to the `useMemo`

packages/components/inputs/search-select-input/src/search-select-input.spec.js
mohib0306: I checked the `package.json` for the repo and we use `@testing-library/react` which does not export the `user-event` helper. We have to add `@testing-library/user-event` to the dev-dependencies and the update the `test-utils` helper in the ui-kit to export those via `export { default as userEvent } from '@testing-library/user-event';`

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: https://github.com/commercetools/ui-kit/pull/1564/commits/3da8c419a6a663bd32880a5559e9685ace86751a

packages/components/inputs/search-select-input/src/search-select-input.spec.js
tdeekens: Thanks for checking. I think it's worthwhile as a follow-up then? Like add userEvent and migrate it all? I often feel - if scope permits - a good Friday afternoon task :)

mohib0306: I think we should add it to ui-kit jira board and do it after the `readme` migration. Otherwise we might loose track of things. i.e. status of readme migration and migrating to user-events

PR #1566, opened 2020-09-15
feat(data-table-manager): add prop `managerTheme`

jonnybel:
#### Summary

This PR adds a prop `settingsContainerTheme` which allows changing the background colour (theme) of the Card which contains the manager settings.


tdeekens (approved):
My main/only concern is with the name containerTheme. It sort of implies I can in the future combine that theme with something else. Like containerTheme and dropdownTheme. Which I think will never be true not meaningful (might be wrong). As a result maybe just theme is enough as a name.

packages/components/data-table-manager/src/data-table-manager.js
```diff
@@ -234,6 +235,13 @@ DataTableManager.propTypes = {
    * A React node for rendering additional information within the table manager.
    */
   topBar: PropTypes.node,
+  /**
+   * Sets the background theme of the Card that contains the settings
+   */
+  settingsContainerTheme: PropTypes.oneOf(['light', 'dark']).isRequired,
```
emmenko, line 241:
```suggestion
  settingsContainerTheme: PropTypes.oneOf(['light', 'dark']),
```

packages/components/data-table-manager/src/settings-container/settings-container.js
```diff
@@ -73,6 +69,10 @@ SettingsContainer.propTypes = {
   primaryButton: PropTypes.element,
   secondaryButton: PropTypes.element,
   children: PropTypes.node.isRequired,
+  containerTheme: PropTypes.oneOf(['light', 'dark']).isRequired,
```
emmenko, line 72:
```suggestion
  containerTheme: PropTypes.oneOf(['light', 'dark']),
```

emmenko (approved):

jonnybel:
> My main/only concern is with the name containerTheme. It sort of implies I can in the future combine that theme with something else. Like containerTheme and dropdownTheme. Which I think will never be true not meaningful (might be wrong). As a result maybe just theme is enough as a name.

How about changing it to `managerTheme`? I didn't want to call it "theme" because we're using that prop name for making components _themeable_.

tdeekens:
Yeah, managerTheme is fine. That indicates what behaviour is themed not the element.

tdeekens (approved):

PR #1572, opened 2020-09-16
chore: add data-table-manager to Changeset config

jonnybel:
#### Summary

We forgot to add `@commercetools-uikit/data-table-manager` to our Changeset config when we created the component, which basically means it didn't get any version update since it was created 🤦‍♂️ 

This PR fixes this and bumps the version to force an update with all the previous changes.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -29,6 +29,7 @@
       "@commercetools-uikit/creatable-select-field",
       "@commercetools-uikit/creatable-select-input",
       "@commercetools-uikit/data-table",
+      "@commercetools-uikit/data-table-manager",
       "@commercetools-uikit/date-field",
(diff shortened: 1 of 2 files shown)


emmenko (approved):
Good catch! 😱

emmenko:
Actually the package did get updates: https://github.com/commercetools/ui-kit/blob/master/packages/components/data-table-manager/CHANGELOG.md

The "linked" list is only to keep the package versions bumped at the same "top" version.

jonnybel:
> Actually the package did get updates: https://github.com/commercetools/ui-kit/blob/master/packages/components/data-table-manager/CHANGELOG.md
> 
> The "linked" list is only to keep the package versions bumped at the same "top" version.

Ah, ok, it's less scary then :)

PR #1575, opened 2020-09-17
feat(async-select-input): add className prop

mohib0306:
#### Summary

- `SearchSelectInput` which will be built on top of the `AsyncSelectInput` requires some styles changes than the default ones. This requires that the `AsyncSelectInput` accepts the `className` prop so that it process the emotions styles.


mohib0306:
The props is added along with warning to not be miss/over-used

tdeekens (commented):
Sorry, but can we think of any other way to allow the variation needed here? A prop, a container component? Anything but not classnames they are just so leaky as to what they enable.

mohib0306:
The other way would be that we add the `styles` prop to the `AsyncSelectInput`, which is responsible for styling the inner components. However this will result in overriding some of the default styles. i.e. with styles props, i have to specify all the styles (default ones) + new styles for an inner component

jonnybel:
Let's try to avoid introducing `classnames` here, even though the alternative (a styled wrapper with CSS selectors) isn't the most elegant...

We should refactor the SelectInputs at some point to use a shared base component with more flexibility to customize it and its styles and then use them to implement and export the actual SelectInput components (currently four, with the SearchSelectInput upcoming) with their curent APIs with a lesser degree of customization.

mohib0306:
As discussed above, `classnames` wouldn't be a good idea here, therefore closing this PR

PR #1589, opened 2020-09-22
test(data-table-manager): improve tests

emmenko:
Just so that we don't forget to finish this topic


adnasa (approved):
just the changeset missing, I guess

tdeekens (approved):

emmenko:
In this case the changeset is not super important, the changes are only related to the tests so there is no immediate need to  include that in a release.

jonnybel (approved):
Thanks!

PR #1590, opened 2020-09-23
fix: localized inputs to avoid re-mounts when localization control is toggled

jonnybel:
#### Summary

This PR was initially for fixing an apparently simple problem with the Localized inputs where the localization control (the button for hiding/showing all languages) would lose the focus when triggered. This particularly affected keyboard users, and it happened because it was being re-mounted in a different place in the tree.

To fix this, I changed where the control is rendered, and also refactored the code to re-use the same component, which was duplicated for every Localized input and in some cases with small, unjustified differences. 

**Additional fix:** During the refactoring, I found that the LocalizedMoneyInput had different spacings between all inputs compared to the other Localized inputs, and the design team confirmed that this was not intentional, so a fix for this is also included in the PR. This shows in the percy diffs.


packages/components/inputs/localized-money-input/src/localized-money-input.js
```diff
@@ -157,83 +143,66 @@ const LocalizedMoneyInput = (props) => {
...
-              onChange={props.onChange}
-              currency={currency}
-              placeholder={
-                props.placeholder ? props.placeholder[currency] : undefined
-              }
-              onBlur={props.onBlur}
-              onFocus={props.onFocus}
-              isDisabled={props.isDisabled}
-              isReadOnly={props.isReadOnly}
-              currenciesControl={(() => {
```
jonnybel, line 195: No need for this anymore, as well as the flags declared above within this .map function.

packages/components/inputs/localized-money-input/src/localized-money-input.js
```diff
@@ -157,83 +143,66 @@ const LocalizedMoneyInput = (props) => {
...
-              warning={props.warnings && props.warnings[currency]}
-              error={props.errors && props.errors[currency]}
-              {...createLocalizedDataAttributes(props, currency)}
-            />
-          );
-        })}
+          })}
+        </Stack>
+        {shouldRenderCurrencyControl && (
+          <LocalizedInputToggle
```
jonnybel, line 188: This is the "currenciesControl" now. It is always rendered in the same place, instead of rendering conditionally with one of the LocalizedInput components

packages/components/inputs/localized-rich-text-input/src/editor.js
```diff
@@ -75,7 +81,12 @@ const Editor = (props) => {
     onToggle();
   }
 
-  const languagesControl = props.languagesControl();
+  const shouldToggleButtonTakeSpace =
+    // if hasLanguagesControl and there are no errors/warnings to display
+    // then the toggleButton is absolutely positioned
```
jonnybel, line 86: This is because the toggle button is placed next to the LanguagesControl (now called LocalizedInputToggle) without being siblings in the DOM.

If there is a error or warning showing, then it can be placed statically because it will then be a sibling to the error/warning message, with the LocalizedInputToggle being pushed down.

src/components/internals/localized-input-toggle/localized-input-toggle.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
jonnybel, line 1: This is the component that replaces the ones duplicated for the 4 localized inputs affected in this PR.

src/components/internals/localized-input-toggle/localized-input-toggle.js
```diff
@@ -0,0 +1,53 @@
...
+  defaultMessage: PropTypes.string.isRequired,
+});
+
+LocalizedInputToggle.displayName = 'LocalizedInputToggle';
+LocalizedInputToggle.propTypes = {
+  icon: PropTypes.node.isRequired,
+  isOpen: PropTypes.bool,
+  onClick: PropTypes.func.isRequired,
+  isDisabled: PropTypes.bool,
+  showMessage: PropTypes.oneOfType([PropTypes.string, intlMessageShape])
```
jonnybel, line 41: We allow passing the pre-translated string because it's useful for the LocalizedMoneyInput (because the variable is for "currencies" instead of "languages). All the other Localized inputs can keep the default.

jonnybel (commented):
The 4 Localized input components affected had a very similar refactoring process, so the changes should be almost identical.
LocalizedMoneyInput is a bit different because of the icon and messages referring to currencies instead of languages, but that's it.
LocalizedRichTextInput has a few differences in terms of styling/positioning.

src/components/internals/localized-input-toggle/localized-input-toggle.js
```diff
@@ -0,0 +1,53 @@
...
+    <FlatButton
+      icon={props.icon}
+      label={label}
+      onClick={props.onClick}
+      isDisabled={props.isDisabled}
+    />
+  );
+};
+
+const intlMessageShape = PropTypes.shape({
```
adnasa, line 29: I think `react-intl` exposes a prop-type shape
https://github.com/formatjs/formatjs/blob/8a206d0c2c5d519cc40baa3435e4681f43e12be8/website/docs/guides/distribute-libraries.md

tdeekens: Yes but didn't we say we don't want to use shared prop-types shapes as ESLint won't be able to lint them and as we wan't to be "clear" about the part of the shape we need/use in the component?

adnasa: > Yes but didn't we say we don't want to use shared prop-types shapes as ESLint

honestly, I don't recall. discard my comment

packages/components/inputs/localized-rich-text-input/src/editor.js
```diff
@@ -75,7 +81,18 @@ const Editor = (props) => {
...
 
-  const languagesControl = props.languagesControl();
+  const shouldToggleButtonTakeSpace =
+    /*
+      - if hasLanguagesControl and there are no errors/warnings to display
+      - then the toggleButton is absolutely positioned
+      This is because the toggle button is placed next to the LocalizedInputToggle without being siblings in the DOM.
+      If there is a error or warning showing,
+      then it can be placed statically because it will then be a sibling to the error/warning message
+      and LocalizedInputToggle is placed below the errors/warnings.
```
adnasa, line 91: I'm wondering if it is worth mentioning in a README about this.
so that there won't be any surprise

src/components/internals/localized-input-toggle/localized-input-toggle.js
```diff
@@ -0,0 +1,53 @@
...
+import PropTypes from 'prop-types';
+import { useIntl } from 'react-intl';
+import FlatButton from '@commercetools-uikit/flat-button';
+import { WorldIcon } from '@commercetools-uikit/icons';
+import messages from '../messages/localized-input/messages';
+
+const LocalizedInputToggle = (props) => {
+  const intl = useIntl();
+
+  const labelMessage = props.isOpen ? props.hideMessage : props.showMessage;
```
adnasa, line 11: why not inline this into L15?

adnasa (approved):

src/components/internals/localized-input-toggle/localized-input-toggle.js
```diff
@@ -0,0 +1,53 @@
...
+  onClick: PropTypes.func.isRequired,
+  isDisabled: PropTypes.bool,
+  showMessage: PropTypes.oneOfType([PropTypes.string, intlMessageShape])
+    .isRequired,
+  hideMessage: PropTypes.oneOfType([PropTypes.string, intlMessageShape])
+    .isRequired,
+  remainingLocalizations: PropTypes.number,
+};
+LocalizedInputToggle.defaultProps = {
+  icon: <WorldIcon />,
```
emmenko, line 48: I don't think we can do this here. Just do the fall back inline where you need to pass the prop.

emmenko (changes requested):
Looks good, just one thing about the prop types

packages/components/inputs/localized-rich-text-input/src/editor.js
jonnybel: Hmm, I don't think it's necessary, because this isn't relevant to consumers of the package.

The positioning nuance is all dealt with internally, and the comment is just explaining how it's being done, for the sake of maintainability.

src/components/internals/localized-input-toggle/localized-input-toggle.js
jonnybel: For readability, and because it's used both on L14 and L15.

src/components/internals/localized-input-toggle/localized-input-toggle.js
jonnybel: Ok, is this what you meant https://github.com/commercetools/ui-kit/pull/1590/commits/79a01bba21942801ab2aa1644195f40cebe8379d?

tdeekens: Hm. why can't we do it? It's actually an interesting question :)

emmenko (approved):
Thanks!

PR #1591, opened 2020-09-23
chore(data-table-manager): update docs to be consistent with translated copy

jonnybel:
#### Summary

This PR simply updates the docs of the DataTableManager to be consistent with the copy of the translations updated in: https://github.com/commercetools/ui-kit/pull/1580

### DIFF
--- a/packages/components/data-table-manager/src/data-table-settings/messages.js
+++ b/packages/components/data-table-manager/src/data-table-settings/messages.js
@@ -9,7 +9,7 @@ export default defineMessages({
   displaySettingsOption: {
     id: 'UIKit.DataTableManager.TableSettings.displaySettingsOption',
     description: 'Display settings dropdown option label.',
-    defaultMessage: 'Display settings',
+    defaultMessage: 'Layout settings',
   },
   columnManagerOption: {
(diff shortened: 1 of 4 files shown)


emmenko (approved):

packages/components/data-table-manager/src/data-table-settings/messages.js
```diff
@@ -9,7 +9,7 @@ export default defineMessages({
   displaySettingsOption: {
     id: 'UIKit.DataTableManager.TableSettings.displaySettingsOption',
     description: 'Display settings dropdown option label.',
-    defaultMessage: 'Display settings',
+    defaultMessage: 'Layout settings',
```
emmenko, line 12: Btw, this will cause the translations to be removed, since the source file changed.

emmenko (changes requested):

packages/components/data-table-manager/src/data-table-settings/messages.js
jonnybel: Ah, I see. I'll just leave these untouched then.

emmenko (approved):

PR #1594, opened 2020-09-25
fix(time-input): to add proper hover style and remove leftovers

jonnybel:
#### Summary

Quick style fix to the hover status of TimeInput component, as well as focus outline of the clear indicator, which was slightly different from the other inputs:

Kapture 2020-09-25 at 17 10 35


mohib0306 (approved):
Looks good

PR #1597, opened 2020-09-29
fix(checkbox-input): type of label element to allow more nesting

jonnybel:
There are situations where a consumer of a CheckboxInput might want to display specially formatted content as the label (such as Icons, truncated text, etc), which is passed with the `children` prop.

However, because the CheckboxInput is currently using a `Text.Body` component for wrapping the label, which renders a `p` element, passing it any `div` elements will cause the warning:
```
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
```

This ultimately discourages the consumer from using the `children` prop, choosing to render the intended label next to the checkbox. This has problems such as the label not being clickable (not toggling the checkbox) and accessibility issues because of the checkbox ending up without a label at all.

### Summary:
This PR fixes the above problem by rendering the Text.Body label as a `span` element instead of a `p`, so the warning does not occur anymore.

In this PR I also added a few notes to the documentation of the component and integrated it with the `generate-readme` script, and encouraging the usage of `aria-label` for situations where one doesn't want to display a label.


packages/components/inputs/checkbox-input/src/checkbox-input.js
```diff
@@ -126,6 +158,7 @@ class CheckboxInput extends React.PureComponent {
             tabIndex={this.props.isReadOnly ? 0 : -1}
           >
             <Text.Body
+              as={'span'}
```
jonnybel, line 161: This is the impactful change. The rest is `generate-readme` stuff.

tdeekens, line 161:
```suggestion
              as="span"
```

emmenko (approved):
💯

tdeekens (approved):

adnasa (approved):

jonnybel:
On Hold until #1598 is merged.

PR #1598, opened 2020-09-29
fix(text): text-body not applying theme when rendered as "span"

jonnybel:
#### Summary
Self-explanatory.

Quick fix to a bug found in #1597 where the checkbox label lost the ability to be themed. The problem was traced back to the Text.Body component when passed an `as` prop.

Added a VRT for this particular case.


tdeekens (approved):

adnasa (approved):
👍

emmenko (approved):

PR #1600, opened 2020-09-30
Add SearchSelectInput to the input presets list

mohib0306:
#### Summary

- `presets/inputs`: Add SearchSelectInput to the input preset list


tdeekens (approved):

.changeset/sweet-pants-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/inputs': major
```
jonnybel, line 2: Should be `minor` as it does not include breaking changes, but new features.

jonnybel (changes requested):

jonnybel:
Add the package to the list `linked` on `.changeset/config.json`, so the version stays in sync with the rest of the packages.

.changeset/sweet-pants-tell.md
tdeekens: Shit. Should have spotted that.

mohib0306: I thought that since a new package is added to the presets, it should be a major change for presets package. But changeset would consider that a major change for the whole repo. Sorry for the confusion

jonnybel (approved):

PR #1603, opened 2020-09-30
fix(data-table-manager): remove width constraint

jonnybel:
#### Summary

This PR removes the horizontal constraint from the `DataTableManager`, as it was found that it looks odd on screens wider than `768px` and should instead be allowed to take all available space.


packages/components/data-table-manager/src/settings-container/settings-container.js
```diff
@@ -23,32 +22,30 @@ const SettingsContainer = (props) => {
     <CollapsibleMotion isDefaultClosed={false}>
       {({ registerContentNode, containerStyles }) => (
         <Card type="flat" theme={props.containerTheme}>
-          <Constraints.Horizontal constraint="xl">
```
jonnybel, line 26: This is the only change, the rest is indentation.

tdeekens (approved):

mohib0306 (approved):

PR #1604, opened 2020-10-01
Introduce SearchSelectField to the field components

mohib0306:
#### Summary

- `SearchSelectField`: introduce `SearchSelectField` built on top of the `SearchSelectInput` and `FieldLabel` to the available field components


.changeset/nervous-moose-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/search-select-field': minor
```
mohib0306, line 2: Marked as minor change (learned from the releases last few days). Is this right?

.changeset/config.json
```diff
@@ -93,7 +93,8 @@
       "@commercetools-uikit/toggle-input",
       "@commercetools-uikit/tooltip",
       "@commercetools-uikit/utils",
-      "@commercetools-uikit/search-select-input"
+      "@commercetools-uikit/search-select-input",
+      "@commercetools-uikit/search-select-field"
```
mohib0306, line 97: Also added to the linked changeset list for version bumping. Not sure if it should be done now or after its first release.

packages/components/fields/search-select-field/src/search-select-field.js
```diff
@@ -0,0 +1,278 @@
...
+} from '@commercetools-uikit/utils';
+import Constraints from '@commercetools-uikit/constraints';
+import Spacings from '@commercetools-uikit/spacings';
+import FieldLabel from '@commercetools-uikit/field-label';
+import SearchSelectInput from '@commercetools-uikit/search-select-input';
+import FieldErrors from '@commercetools-uikit/field-errors';
+
+const hasErrors = (errors) => errors && Object.values(errors).some(Boolean);
+const searchSelectInputProps = Object.keys(SearchSelectInput.propTypes);
+const filterSerachSelectFieldProps = (props) => {
```
mohib0306, line 16: Just a helper function to filter out wrong (invalid) props. It checks if the prop is allowed by `SearchSelectInput` and if yes, then adds it to the `filteredProps` list (valid list of props).

packages/components/fields/search-select-field/src/search-select-field.spec.js
```diff
@@ -0,0 +1,197 @@
...
+
+it('should call onChange when changing the value', async () => {
+  const onChange = jest.fn();
+  renderSearchSelectField({
+    onChange,
+  });
+  const input = screen.getByLabelText('SearchSelectField');
+  fireEvent.focus(input);
+  fireEvent.change(input, { target: { value: 'mango' } });
+  await waitFor(() => screen.getByText('Mango'));
```
tdeekens, line 117: I think this could be `findByText`?

tdeekens (commented):
Looks like somebody used a generator :)

packages/components/fields/search-select-field/src/search-select-field.js
jonnybel, line 16:
```suggestion
const filterSearchSelectFieldProps = (props) => {
```

jonnybel: Could you add this as a comment on the code? 🙂

jonnybel:
Upon looking at the Percy snapshots (and then on the storybook), seems like field isn't properly displaying the different states (disabled, readonly, error, warning) 🤔

mohib0306:
I will have a loot at it now

jonnybel (approved):

tdeekens (approved):

mohib0306:
UX review passed. Merging this

PR #1605, opened 2020-10-01
fix(date-inputs): to not open date picker as the input gets focused by the keyboard

jonnybel:
Another keyboard navigation tweak.

With this PR, the date picker of the Date Input components will not automatically open as soon as the input is focused by the keyboard, but only if the user starts typing in it (or pressing `ArrowDown` key).
This is to avoid the visual distraction when the user is just navigating through a form and doesn't necessarily want to open this input.

It still opens immediately if the input is clicked with the mouse.

Added a few tests for the opening of the date pickers.


packages/components/inputs/date-input/src/date-input.js
```diff
@@ -234,7 +234,6 @@ const DateInput = (props) => {
                   },
                   // we only do this for readOnly because the input
                   // doesn't ignore these events, unlike when its disabled
-                  onFocus: props.isReadOnly ? undefined : openMenu,
```
jonnybel, line 237: Here is the only line that had to be removed (per each date input) to change the behaviour.

mohib0306 (approved):
Thanks for updating the tests.

adnasa (approved):

packages/components/inputs/date-input/src/date-input.js
```diff
@@ -125,12 +125,6 @@ const DateInput = (props) => {
           if (changes.hasOwnProperty('highlightedIndex')) {
             setHighlightedIndex(changes.highlightedIndex);
           }
-
-          // ensure calendar always opens on selected item, or on current
-          // month when there is no selected item
-          if (changes.hasOwnProperty('isOpen') && changes.isOpen) {
-            setCalendarDate(props.value === '' ? getToday() : props.value);
-          }
```
jonnybel, line 133: Removed this block because the `calendarDate` state is already initialized with `props.value || getToday()` so it is not needed.
Furthermore, this was overwriting the state that may have been set in L116, causing inconsistent behaviour.

PR #1612, opened 2020-10-05
fix(search-select-input): fix isDisabled and isReadOnly states

mohib0306:
#### Summary

- `SearchSelectInput`: The design team reported an issue with the is disabled mode of the `search-select-input` that it does not show the proper cursor. This PR fixes that and shows the parent cursor while it is disabled or in read-only mode.


mohib0306:
UX review passed

jonnybel (approved):

packages/components/inputs/search-select-input/src/search-select-input.styles.js
```diff
@@ -1,28 +1,22 @@
...
+          cursor: text;
+        }
+        div[class$='indicatorContainer' i] {
+          cursor: pointer;
+          svg * {
+            fill: ${vars.colorSolid};
+          }
+        }
+      }`
+      : ''}
```
mohib0306, line 19: @jonnybel I just updated the PR to simplify the styles logic like this

PR #1615, opened 2020-10-06
Improve readmes

emmenko:


package.json
```diff
@@ -21,7 +21,6 @@
     "design-tokens:build": "yarn --cwd design-system build:tokens",
     "design-tokens:build:watch": "yarn --cwd design-system build:tokens:watch",
     "prebuild": "yarn generate-icons && yarn design-tokens:build && yarn i18n:compile && yarn generate-readmes",
-    "prestorybook:start": "yarn prebuild",
```
emmenko, line 24: As suggested by @jonnybel.
Didn't want to open a separate PR for this.

.eslintrc.yaml
```diff
@@ -34,6 +34,7 @@ overrides:
   - files: ['*.spec.js', '**/docs/*.js']
     rules:
       'react/display-name': 0
+      'react/prop-types': 0
```
adnasa, line 37: why?

adnasa (approved):

.eslintrc.yaml
emmenko: In test files and usage example files I find it's not really necessary to have, it's just more boilerplate.

tdeekens: I agree. It's about the test cases less about always being complete. Plus testing for incomplete `prop-types` has a propose.

package.json
jonnybel: Thanks!

jonnybel (approved):

PR #1632, opened 2020-10-20
feat(design-system): add new horizontal constraints and update classic ones

jonnybel:
# Context

The Horizontal Constraints of out Design System are getting a overhaul:
> The existing constraints don't really follow a specific growing pattern and some of them are even rarely used. 

The new values will follow a new system of 8 columns and their "halves" (so there are actually 16 values)
image

The old values are also being mapped to their closest approximates.

#### Max Horizontal Constraints from 1 to 16:

- 1= 42px (former XS)
- 2= 84px
- 3 = 142px (former S)
- 4 = 184px
- 5 = 242px
- 6 = 284px
- 7 = 342px (former M)
- 8 = 384px
- 9 = 442px
- 10 = 484px (former L)
- 11 = 542px
- 12 = 584px
- 13 = 642px
- 14 = 684px
- 15 = 742px
- 16 = 784px (former XL)
- scale: 100%
- auto

# PR Summary

This PR updates the design tokens according to the above.

As follow-ups, I am planning on making additional PRs to add a new prop to the HorizontalConstraint component which uses the new values, and update the codebase to do the same thing for every instance of the "classic" constraint tokens.

The plan is to eventually deprecate the "classic" tokens and remove them from the codebase.


design-system/materials/internals/definition.yaml
```diff
@@ -98,11 +98,27 @@ choiceGroups:
     prefix: constraint
     choices:
       constraint-scale: 100%
-      constraint-xs: 50px
-      constraint-s: 132px
-      constraint-m: 355px
+      constraint-05: 42px
+      constraint-xs: 42px
```
emmenko, line 102: Is there a way to mark those tokens as deprecated?

emmenko (approved):
💯

design-system/materials/internals/definition.yaml
tdeekens: Without a breaking change? Maybe options are

1. Document they are deprecated
2. Mark them as `@deprecated` in TypeScript (if we had it)
3. Rename them to be `*-deprecated` but that's also breaking
   - We could add one which is `contraint-05-deprecated` and ask people to use it
   - Similar to `UNSAFE_`

tdeekens (approved):
Cool!

design-system/materials/internals/definition.yaml
jonnybel: Renaming them to `*-deprecated` seems reasonable (tho breaking) as they should be easy to migrate with a find&replace on the codebases which use the tokens directly. 

If we decide to go for the breaking change, we can also add a deprecation warning to the HorizontalConstraint's `constraint` prop (the plan is for it to have a new prop called `max` which uses the numeric constraints).

emmenko: I wouldn't introduce a breaking change now. It should be enough to mark them as deprecated in the docs.

tdeekens: I didn't mean to rename only. We could add a `*-deprecated`. It's like when `componentWillReceiveProps` continued to work but you should've updated to `UNSAFE_componentWillReceiveProps`. So thenw e can refactor the MC to use `*-deprecated`. Then the code base becomes easy to grep for it and to refactor. While your job is not to do that. That's up to teams.

emmenko: Yes we can do that, but we can do it as a follow up.
We planned to make this change fully backwards compatible, so I'd rather introduce some migration steps as a separate task/release.

jonnybel: I've added a marker for the deprecated tokens.

I also updated the new tokens, after discussing with Maria, to use a simpler, more intuitive naming scheme (from 1 to 16).

emmenko: Should we also update the design screenshot then? Is this scale also going to be used by the design team? cc @mariabarrena

jonnybel: The values are the same, it's just the language that we use for the variables. The values refer to 8 columns and their "half" values, so they're actually 16.

Since it's not very intuitive to convey "0.5" values in a custom-property used by the developers, I asked if it was OK if the language for the devs would be 1-16, and the Design team could communicate in that language with the devs, while still using the concept of 8 columns and their halves in their process. This should be documented too.

I must admit that this was only my own perception of "intuitiveness" though 😄 
If you think it's a better language for the devs who will consume the tokens with the previous pattern of multiples of `05`, I can revert. I don't have a very strong preference.

emmenko: As long as we document the scale and use the same scale values in designs (common language), I think it's fine then.

PR #1638, opened 2020-10-27
Add isCondensed prop to the Stamp component

islam3zzat:
#### Summary

Currently, the stamp component has too much padding for some use-cases, the smaller version is important for use-cases that need the stamps after the text. With the small padding from the top and bottom, the stamp will be highlighted but not too much.

image

### DIFF
--- a/packages/components/stamp/src/stamp.js
+++ b/packages/components/stamp/src/stamp.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
-import InsetSquish from '@commercetools-uikit/spacings-inset-squish';
 
 export const availableTones = [
   'critical',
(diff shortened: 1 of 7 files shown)


Shecki:
visual it's fine for me!
Thanks

packages/components/stamp/src/stamp.visualroute.js
```diff
@@ -24,5 +24,10 @@ export const component = () => (
     <Spec label="when secondary">
       <Stamp tone="secondary">Secondary</Stamp>
     </Spec>
+    <Spec label="when condensed">
```
tdeekens, line 27: 💯

tdeekens (approved):
Code looks good. Would kindly request also @jonnybel to have a look :)

jonnybel (approved):
Looks good 👌

packages/components/stamp/src/stamp.js
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
-import InsetSquish from '@commercetools-uikit/spacings-inset-squish';
```
jonnybel, line 5: Can we remove it from the package dependencies?

islam3zzat: Thanks. Done https://github.com/commercetools/ui-kit/pull/1638/commits/331a4b4f288b4ef8d1f693681d43fec176e4359f

PR #1639, opened 2020-10-27
feat(constraints/horizontal): add new prop `max` which uses new constraints scale

jonnybel:
#### Summary

This PR is a continuation of #1632 and adds a new prop `max` to the Horizontal constraints component which uses the new constraints scale system.

Additionally:
* Begins deprecation process for the prop `constraint`
* Integrates `generate-readme` with the constraints package
* Reworks Storybook example. I think it was a bit confusing for people to understand the scales, so I simplified so it is hopefully clearer to understand.

Note that this PR is pointed at the branch [ja-update-constraint-tokens](https://github.com/commercetools/ui-kit/tree/ja-update-constraint-tokens).


packages/components/constraints/src/horizontal/horizontal.js
```diff
@@ -4,60 +4,90 @@ import { css } from '@emotion/core';
...
+  /**
+   * DEPRECATING SOON: Please use the `max` prop instead. Determines scale of the constraint.
+   */
   constraint: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl', 'scale']),
   children: PropTypes.node.isRequired,
 };
 
 Horizontal.defaultProps = {
-  constraint: 'scale',
+  max: 'scale',
```
jonnybel, line 90: We could remove this defaultProp and the component will still have the `scale` default behaviour, since the component always has `width: 100%` (except when `auto`) even with other scale values because they use the `max-width` rule (the components always "scales" until that max value).

This would allow us to add an `invariant` warning for when the `constraint` prop is being used at the same time as `max`, which represents a collision. We can't add the `invariant` warning right now because it `max` is always defined due to the default and therefore, is a user sets the `constraint` (which is the case for the whole existing codebase) it will trigger the invariant warning even though he didn't set the `max` value.

packages/components/constraints/docs/aditional-info.md
```diff
@@ -0,0 +1,33 @@
+## `max` constraint values:
```
emmenko, line 1: The filename has a typo

packages/components/constraints/docs/aditional-info.md
```diff
@@ -0,0 +1,33 @@
...
+| 9     | `442px` |
+| 10    | `484px` |
+| 11    | `542px` |
+| 12    | `584px` |
+| 13    | `642px` |
+| 14    | `684px` |
+| 15    | `742px` |
+| 16    | `784px` |
+
+## `constraint` prop values (deprecating soon)
```
emmenko, line 24: We should specify which of the deprecated props map to the new values.

packages/components/constraints/src/horizontal/horizontal.story.js
```diff
@@ -40,20 +25,15 @@ storiesOf('Components|Constraints', module)
...
-      <Horizontal constraint={constraint}>
-        <Stack>
-          {sizes.map((size) => (
-            <Row key={size.name}>
-              <InlineColorWrapper width={size.width} />
-            </Row>
-          ))}
-        </Stack>
+      <Horizontal max={max}>
+        <ColouredRow>{max.toString()}</ColouredRow>
```
emmenko, line 36: Any reason not to keep them all listed?

packages/components/constraints/src/horizontal/horizontal.js
emmenko: Then let's remove the default prop?

packages/components/constraints/src/horizontal/horizontal.js
```diff
@@ -4,60 +4,90 @@ import { css } from '@emotion/core';
...
+  if (max === 6) constraintToken = vars.constraint6;
+  if (max === 7 || constraint === 'm') constraintToken = vars.constraint7;
+  if (max === 8) constraintToken = vars.constraint8;
+  if (max === 9) constraintToken = vars.constraint9;
+  if (max === 10 || constraint === 'l') constraintToken = vars.constraint10;
+  if (max === 12) constraintToken = vars.constraint12;
+  if (max === 13) constraintToken = vars.constraint13;
+  if (max === 14) constraintToken = vars.constraint14;
+  if (max === 15) constraintToken = vars.constraint15;
+  if (max === 16 || constraint === 'xl') constraintToken = vars.constraint16;
```
emmenko, line 24: I would put this into a separate function and do an early return, maybe also use a switch statement for better readability.

emmenko (commented):
Looking good! I left some suggestions...

packages/components/constraints/src/horizontal/horizontal.js
```diff
@@ -1,63 +1,91 @@
...
+  );
+};
 
 Horizontal.displayName = 'Horizontal';
 Horizontal.propTypes = {
+  /**
+   * Determines scale of the constraint.
+   */
+  max: PropTypes.oneOf([
+    1,
```
adnasa, line 63: this looks fragile. can this be

```js
const maxProp = [...Array.from({ length: 16 }).map((_, index) => index + 1), 'scale', 'auto']
// [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,"scale","auto"]
```

packages/components/constraints/src/horizontal/horizontal.story.js
jonnybel: All the constraint sizes? Not really, that's a good idea actually.

Previously, we were showing one constraint size (controlled by the knob) with different content sizes (width percentages), which is what was confusing, because it looked like it was all the different constraints, but it wasn't 😄

emmenko: I think we should have all the constraints listed at once, like we do with the spacings

packages/components/constraints/src/horizontal/horizontal.js
jonnybel: I tried it but unfortunately, it messes up the `generate-readme` script 😞

packages/components/constraints/src/helpers.js
```diff
@@ -0,0 +1,35 @@
+import { customProperties as vars } from '@commercetools-uikit/design-system';
+
+const getMaxPropTokenValue = (max) => {
+  if (max === 1) return vars.constraint1;
```
adnasa, line 4: I think this will be tedious to update later.
can this be 

```js
// ensure null, not undefined
return vars[`constraint${max}`] ?? null
```

it's just an idea. I don't mind that we keep this as is.

packages/components/constraints/src/horizontal/horizontal.story.js
```diff
@@ -1,13 +1,13 @@
...
-import { withKnobs, select } from '@storybook/addon-knobs/react';
+import { withKnobs } from '@storybook/addon-knobs/react';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 import styled from '@emotion/styled';
 import Horizontal from './horizontal';
 import Readme from '../../README.md';
 
 const ColouredRow = styled.div`
-  height: 100px;
+  height: 30px;
```
adnasa, line 10: maybe document (code comment) why this is 30px?

jonnybel: It's just for the storybook example, so it's a totally arbitrary value.
In this case, I just changed it to allow all sizes to fit on a standard Mac screen.

packages/components/constraints/src/helpers.js
jonnybel: Yeah it works nicely for the current scale system 👍

packages/components/constraints/src/helpers.js
```diff
@@ -0,0 +1,18 @@
+import { customProperties as vars } from '@commercetools-uikit/design-system';
+
+const getMaxPropTokenValue = (max) => {
+  return vars[`constraint${max}`] ?? null;
```
emmenko, line 4: Not necessary, it would be implicitly `undefined`

```suggestion
  return vars[`constraint${max}`];
```

emmenko (commented):
Much better now thanks!

packages/components/constraints/src/helpers.js
adnasa: point was to ensure `null` as return fallback

jonnybel: Either `null` or `undefined`, it doesn't actually make a difference. So I'll simplify and leave it undefined :)

packages/components/constraints/src/index.js
```diff
@@ -1,6 +1,16 @@
...
+  getAcceptedMaxPropValues,
+} from './helpers';
 
-export default { Horizontal };
+export default {
+  Horizontal,
+  getMaxPropEquivalent,
+  parseHorizontalConstraintProp,
+  getAcceptedMaxPropValues,
+};
```
emmenko, line 14: Why do we need to expose all these helpers?

jonnybel: `parseHorizontalConstraintProp` will be used for components that have a `horizontalConstraint` prop which will accept both of the scales (until we deprecate the letters scale)
`getAcceptedMaxPropValues` is useful for storybook examples (and also propTypes)l
I can remove `getMaxPropEquivalent` from there as it is used by `parseHorizontalConstraintProp` and doesn't need to be exposed itself anymore.

emmenko: Ok...my main concern is "how" we expose those helpers. Given that we only need them for the "migration period", I would make it more explicit that people shouldn't use and rely on those.

Therefore, I would avoid exposing them as "normal exports". Instead we can have a new entry point that other components can use a deep import. Then it's easier to remove it without (breaking) changing the official component exports.

The new entry point can be for example a `helpers.js` file, or even a `experimental-helpers.js`.

The alternative is to have those exports but then we need to be aware that when we remove them it's going to be a breaking change.

jonnybel: Ah, that's right. We shouldn't expose them externally.

packages/components/constraints/src/helpers.js
```diff
@@ -24,8 +24,18 @@ const parseHorizontalConstraintProp = (prop) => {
   return getMaxPropEquivalent(prop);
 };
 
+// Generates an array of accepted values for the max prop, given a min and max
+const getAcceptedMaxPropValues = (min = 1, max = 16) => {
```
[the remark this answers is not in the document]
jonnybel: Mainly yes, though it could potentially be used for other things (like generating prop types)

jonnybel:
@emmenko ; @adnasa 
If I may, I'll merge this branch into `ja-update-constraint-tokens`

adnasa:
all right.
nice work!

PR #1644, opened 2020-10-30
feat: add support for new horizontal constraint prop values

jonnybel:
#### Summary

This PR follows-up #1639 and updates all UI-kit components which use the Horizontal Constraint component and accept a `horizontalConstraint` prop to accept the values of the new scale.

* Components support both the old or new scale values, on the same prop
* Default values were updated to the new scale
* Storybook examples only show the new scales on the prop, even though the old values are still supported. This is to discourage using the old scale for new implementations, and because it's confusing to have both scale values on the same select.


packages/components/field-label/src/field-label.js
```diff
@@ -96,7 +101,28 @@ FieldLabel.propTypes = {
   hasRequiredIndicator: PropTypes.bool,
   htmlFor: PropTypes.string,
   id: PropTypes.string,
-  horizontalConstraint: PropTypes.oneOf(['s', 'm', 'l', 'xl', 'scale']),
+  horizontalConstraint: PropTypes.oneOf([
+    's',
+    'm',
+    'l',
+    'xl',
+    3,
```
emmenko, line 109: I suppose we don't allow sizes below 3, right?

packages/components/tooltip/src/tooltip.js
```diff
@@ -226,8 +226,31 @@ Tooltip.propTypes = {
...
+    10,
+    11,
+    12,
+    13,
+    14,
+    15,
+    16,
+    'scale',
+    'auto',
+  ]).isRequired,
```
emmenko, line 253: Is this required?

emmenko (approved):
💯

packages/components/field-label/src/field-label.js
jonnybel: Yeah, I kept the same restrictions as before for the smallest constraints allowed for all of the components.

In this case, it was `s`, so the equivalent is `3`.

packages/components/tooltip/src/tooltip.js
jonnybel: I guess it doesn't need to be, since the style defaults to the `scale` constraint anyway.

tdeekens (approved):
💯

adnasa (approved):
I reviewed the prop-type validation only this time 😛

jonnybel:
The latest commit (https://github.com/commercetools/ui-kit/pull/1644/commits/33caa3913cc3733abbba7fb75ffe45ca7e9a20b5) added the value `6` to the list of allowed constraint values on the components:

- LocTextField,
- DateField, 
- DateRangeField, 
- DateTimeField, 
- LocalizedMultilineTextField, 
- MultilineTextField, 
- DateInput, 
- DateRangeInput, 
- DateTimeInput, 
- LocalizedMultilineTextInput, 
- MultilineTextInput

And adds values `3` to `6` to LocalizedMoneyInput.

It was a request by the design team.

PR #1645, opened 2020-11-02
Allow NumberInput and NumberField to have the step prop set to "any"

islam3zzat:
#### Summary

In this PR we allow the NumberInput and NumberField to have a step of `any`

## Description

https://codesandbox.io/s/pedantic-williamson-vfwi5
(I've added a red border to invalid inputs)

You'll find that the input is initially invalid, even though the value is between the min and the max.
Adding a step of `0.0001` (`step={0.0001}`) will resolve the issue, but if the user added a more precise value  (e.g 0.00005) it'll be invalid again.
Setting the step to "any" will fix the issue.

In cases like this setting step to `any` makes sense as we don't care about the step.

> A string value of any means that no stepping is implied, and any value is allowed (barring other constraints, such as min and max).
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#attr-step


packages/components/fields/number-field/README.md
```diff
@@ -67,7 +67,7 @@ export default Example;
...
 | `max`                  | `number`                                                             |          |           | Value is used as `max` property on input field                                                                                                                                                                                                                        |
-| `step`                 | `number`                                                             |          |           | Value is used as `step` property on input field                                                                                                                                                                                                                       |
+| `step`                 | `<number\|enum>`                                                     |          |           | Value is used as `step` property on input field                                                                                                                                                                                                                       |
```
islam3zzat, line 70: Is there a way to indicate that we support numbers and "any"?

jonnybel: Not currently. It's automated by the `generate-readme` script.
But you could add a suggestion as a second line to the description of the prop stating something like "Use the value `any` for inputs which accept an unpredictable amount of decimals."

jonnybel (approved):
Thanks!

TIL about the `any` value 😄

PR #1650, opened 2020-11-04
fix(rollup): resolve only main entry point when bundling CJS modules 

emmenko:
The docs say that 

> Specifies the properties to scan within a package.json, used to determine the bundle entry point. The order of property names is significant, as the first-found property is used as the resolved entry point.

So we need to use the `main` entry point only when bundling for CJS, otherwise we risk of bundling ES modules into a CJS bundle.

PS: we have discovered this issue in appkit (https://github.com/commercetools/merchant-center-application-kit/pull/1846)

I also triggered a version bump of all packages so that we bundle them properly (just to be sure).


adnasa (approved):

tdeekens (approved):
JS bundling is just such a... fun ride.

jonnybel (approved):

PR #1642, opened 2020-10-28
feat: add generator package for updating package.json files consistently

emmenko:
Closes #1641

PR #1652, opened 2020-11-04
fix: spacings pkg names

emmenko:
I accidentally introduced a bug when implementing the package.json generator #1642 

The package names are derived by their folder name, so the spacings components were missing the `spacings-` prefix in the name because the folders didn't have them.


design-system/package.json
```diff
@@ -9,12 +9,7 @@
     "directory": "design-system"
   },
   "homepage": "https://uikit.commercetools.com",
-  "keywords": [
-    "javascript",
-    "design system",
-    "react",
-    "uikit"
-  ],
+  "keywords": ["javascript", "design system", "react", "uikit"],
```
adnasa, line 12: this again? 😄

tdeekens (approved):
Ah I guess it's the `spacings-*` thing. Looks good.

jonnybel (approved):

adnasa (approved):

design-system/package.json
emmenko: It's changeset...I'll look into how they run the format, maybe we need to fix it there.

emmenko: I think this is the option that is causing the issue: https://github.com/atlassian/changesets/blob/f9ccfa3e9911718fcd95eff398ece64c8e9eef9b/packages/apply-release-plan/src/index.ts#L116

image

PR #1656, opened 2020-11-06
fix(utils): remove passthrough util and use filterInvalidAttributes instead

emmenko:
I noticed this as I was investigating a separate issue.


tdeekens (approved):

jonnybel (approved):

mohib0306:
Thanks for these fixes ❤️

emmenko:
Btw, just to be clear. Strictly speaking this is a breaking change, but the usage of the `passThroughProps` was kind of "wrong" so I consider this more of a fix, given also that the `utils` package is most likely used internally only.

PR #1662, opened 2020-11-09
refactor: explicitly get emotion theme from context 

### LINKED ISSUE #1660: Emotion theming does not work properly anymore
**Describe the bug**
When using the `ThemeProvider` to overwrite certain custom properties, components using the `css((theme) => ...)` callback function to get the theme options do not seem to work anymore, as an empty object is returned instead.

We had a similar issue last week in our docs-kit repo: https://github.com/commercetools/commercetools-docs-kit/pull/745

**Expected behavior**
Theme options are correctly read from the components.

emmenko:
Fixes #1660


packages/components/buttons/flat-button/src/flat-button.js
```diff
@@ -67,47 +68,45 @@ export const FlatButton = (props) => {
     // we fall back to `isDisabled`
     disabled: props.isDisabled,
   };
-
+  const theme = useTheme();
+  const overwrittenVars = {
+    ...vars,
+    ...theme,
```
adnasa, line 74: so consumer aren't allow to override the theme at all?

adnasa (approved):

packages/components/buttons/flat-button/src/flat-button.js
emmenko: They are yes, with the `ThemeProvider`.

This change here is only about how we access the `theme` object internally by being explicit.

jonnybel (approved):

tdeekens (approved):

PR #1663, opened 2020-11-09
Enable defaultOptions in SearchSelectInput

mohib0306:
#### Summary

- `SearchSelectOption` => In order to pre-load the selected options in the input, we need to enable the `defaultOptions` prop. This prop was previously disabled but while using it in a form, it was noticed that we need to pre-load options which are pre-selected (`value`). This PR enables the user to use the `defaultOptions` prop to do so.


packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -160,6 +159,17 @@ SearchSelectInput.propTypes = {
    * Determines if option groups will be separated by a divider
    */
   showOptionGroupDivider: PropTypes.bool,
+  /**
+   * The default set of options to show before the user starts searching. When set to true, the results for loadOptions('') will be autoloaded.
```
emmenko, line 163:
```suggestion
   * The default set of options to show before the user starts searching. When set to `true`, the results for `loadOptions('')` will be autoloaded.
```

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -160,6 +159,17 @@ SearchSelectInput.propTypes = {
...
+   */
+  defaultOptions: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        value: PropTypes.string.isRequired,
```
emmenko, line 169: Is this the only required field? Are there any other fields that should be provided?

.changeset/shy-buckets-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/search-select-input': patch
+---
+
+Enable defaultOptions on SearchSelectInput to preload selected options
```
emmenko, line 5:
```suggestion
Enable `defaultOptions` on `SearchSelectInput` to preload selected options
```

packages/components/inputs/search-select-input/src/search-select-input.js
mohib0306: Value is required, label is kinda optional. Since if label is not provided, it will fallback to empty string but it will not break the component

emmenko: Ok but we can still include it (the `label`) so it's clear what kind of object you would need to pass.

mohib0306: I have updated the search-select-input, however if we need the same change in other `select-input/fields`, we need to update their props as well as they also only require `value` as of now.

packages/components/inputs/search-select-input/src/search-select-input.js
```diff
@@ -160,6 +159,18 @@ SearchSelectInput.propTypes = {
...
   showOptionGroupDivider: PropTypes.bool,
+  /**
+   * The default set of options to show before the user starts searching. When set to `true`, the results for `loadOptions('')` will be autoloaded.
+   */
+  defaultOptions: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        value: PropTypes.string.isRequired,
+        label: PropTypes.string.isRequired,
```
emmenko, line 170: I meant if it's optional, keep it optional but just include the prop in the object, so it's also documented in the readme.

```suggestion
        label: PropTypes.string,
```

emmenko (approved):
Thanks! 🙏

PR #1668, opened 2020-11-13
chore(utils): remove is variation

adnasa:
#### Summary

- removes `utils/is-variation`


.changeset/thin-poems-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/utils': major
```
emmenko, line 2: Let's not bump a major.

We assume that the utils package is only used internally.

emmenko (changes requested):

.changeset/thin-poems-drop.md
adnasa: ah.
the assumption was that it was used externally since it was an exported package.
but sure.

adnasa:
I generated a new changeset and removed the old one.
please review

jonnybel (approved):

.changeset/strange-windows-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/utils': patch
+---
+
+remove `is-variation`
```
emmenko, line 5: What's the reason for removing it? Simply that it's not used anywhere?

adnasa: exactly. there was no use for it, neither in ui-kit or mc-frontend

adnasa:
Is there anything else on your side @emmenko ?
i'm ready to merge this.

emmenko (approved):

tdeekens (approved):

PR #1670, opened 2020-11-15
chore: update emotion to v11

emmenko:
https://emotion.sh/docs/emotion-11


tdeekens (commented):
👍 Just wanted to start doing the same 😄.

adnasa (approved):
LGTM

emmenko:
Hmm getting some strange errors 🧐

image

emmenko:
Tests are passing now! 🙌

jonnybel (approved):
Really nice work refactoring so many style functions 🥇 

And this is one of those times Percy comes rly handy 🙏

emmenko:
Yeah, absolutely.

Maybe you can go through storybook and make sure the interactions are fine, in case you have some particular special cases in mind. Just to have an extra pair of eyes.

jonnybel (changes requested):
Just found this problem with the DateInputs when opened and hovered:
image

emmenko:
Good catch, I'll take a look

packages/calendar-utils/src/calendar-body/calendar-body.styles.js
```diff
@@ -26,20 +26,58 @@ const getClearSectionStyles = (props) => {
...
     ...customProperties,
     ...props.theme,
   };
 
-  const baseIconStyles = css`
+  return css`
     align-items: center;
     box-sizing: border-box;
-    background: none;
+    background-color: ${props.isDisabled
```
jonnybel, line 71: This should be removed (`none` or `transparent`), because the background is given by the parent container. Otherwise, the background "paints" on top of the box-shadow focus indicator.

packages/calendar-utils/src/calendar-body/calendar-body.styles.js
```diff
@@ -127,66 +154,17 @@ const getInputContainerStyles = (props, state) => {
...
       box-shadow: inset 0 0 0 2px ${vars.borderColorForInputWhenFocused};
     }
+    &:focus,
+    &:hover {
+      border-color: ${props.isDisabled ||
+      props.hasError ||
+      props.hasWarning ||
+      props.isReadOnly ||
+      ((props.isOpen || state.isFocused) && !props.isReadOnly)
+        ? 'inherit'
```
jonnybel, line 164: This (`inherit`) is what causes the black border that appears when hovering with the calendar open (as seen in https://github.com/commercetools/ui-kit/pull/1670#pullrequestreview-533600128)

Leaving it as an empty string for the truthy condition seems to fix the problem.

jonnybel (approved):
I think that's it. I went through Storybook and didn't find any other problems with the styles ✨

.changeset/lovely-llamas-retire.md
```diff
@@ -0,0 +1,84 @@
...
+'@commercetools-uikit/stamp': minor
+'@commercetools-uikit/table': minor
+'@commercetools-uikit/tag': minor
+'@commercetools-uikit/text': minor
+'@commercetools-uikit/tooltip': minor
+'@commercetools-uikit/utils': minor
+'visual-testing-app': minor
+---
+
+Migrate to emotion v11. https://emotion.sh/docs/emotion-11
```
emmenko, line 84: I don't think we need to bump a major version here. The emotion packages are also not peer dependencies, so there should be no need for a major.

emmenko:
@adnasa can you take a look at this whenever you have time?

adnasa (approved):
LGTM from what I can tell.
let's go :)

tdeekens (approved):
💯

PR #1671, opened 2020-11-15
refactor: compile and bundle monorepo using preconstruct cli

emmenko:
Just experimenting a bit with `@preconstruct/cli` to see how it can simplify the setup in the monorepo.

There are still some open questions and a bunch of things to fix.

---

I saw that changesets monorepo is using `@preconstruct/cli` to manage compiling and bundling monorepos, which means:
- no more rollup (precostruct does that)
- no more ts config files in each package (TS setup will become easier)


emmenko:
Alright so we have to fix all the imports from the "internal" packages (finally!), as we shouldn't use relative imports from outside of a package.
We can do that as a separate PR as a prerequisite for this one.

adnasa:
> no more ts config files in each package (TS setup will become easier)

Are the ts config replaced with something else or are they managed directly from `preconstruct`?
In the other PR, you mentioned that we should finish this PR, then to find this PR as an experiment so I am confused.

Are we commiting to this or are we just "experimenting"?

We're getting ahead of ourselves, I think.

emmenko:
I actually wanted to fix the TS setup here first, but something was not working fine.
That's when I discovered preconstruct and found it useful.

Let's have a chat about it.

emmenko:
Previous build time: `141.35s`

New build time with preconstruct: `16.60s`

tdeekens:
> Previous build time: `>140s`
> 
> New build time with preconstruct: `~16s`

Impressive. Any clue where this time is being saved?

emmenko:
I think because previously we were literally executing at least one command per package, with preconstruct is just one command for all.

tdeekens (approved):

adnasa (approved):

jonnybel (approved):

PR #1669, opened 2020-11-13
refactor(hooks,util): typescript

adnasa:
#### Summary

- updates eslintrc
- adds tsconfig to `packages/utils`, `packages/hooks`
- converts `packages/hooks` and most of `packages/utils` to typescript.

I'll keep it as Draft, since the build fails.
I welcome help on updating the `eslintrc` for further changes needed there.

PR #1672, opened 2020-11-15
refactor: move internal stuff into utils packages

emmenko:
This PR unblocks #1669 and #1671 

This was kind of a leftover from the migration of the source code into packages. Packages should not rely on source files located outside of the package itself, using relative paths.

All the remaining internal code has now been moved into different utils packages:
- select-utils
- input-utils (NEW)
- rich-text-utils (NEW)
- calendar-utils

~~There were also some "internal" icons, which are not part of the normal icons, as I don't see a reason why they shouldn't.~~

Internal icons have been moved into the packages that require them: checkbox-input, radio-input, rich-text-utils.


adnasa (approved):
LGTM. sorry for being late on this

jonnybel (approved):
Good stuff, thanks!

PR #1677, opened 2020-11-16
fix: generated icons imports

emmenko:
Leftover of #1672


packages/components/inputs/rich-text-utils/src/rich-text-body/icons/components/unordered-list.js
```diff
@@ -6,10 +6,7 @@ This file is created by the 'scripts/generate-icon-exports.js' script.
 
 import React from 'react';
 import { useTheme } from 'emotion-theming';
-import {
-  getIconStyles,
-  iconPropTypes,
-} from '../../../../../../icons/src/create-styled-icon';
```
emmenko, line 12: Internal icons should import from `@commercetools-uikit/icons`.

packages/components/icons/src/index.js
```diff
@@ -4,6 +4,7 @@
   This file is created by the 'scripts/generate-icon-exports.js' script.
 */
 
+export * from './create-styled-icon';
```
emmenko, line 7: To make it available to other packages.

tdeekens (approved):

PR #1680, opened 2020-11-17
refactor(contributing): updated instructions for adding a new icon

nbryant-commercetools:
#### Summary:
Recently a new icon was added, and the type generation step in `merchant-application-kit` was missed. The changes here update the documented step for adding new icons to include the missed step.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,6 +225,7 @@ These are informal steps we suggest you to follow when adding a new component.
 
 1. Add the raw SVG of the new icon to `packages/components/icons/src/svg`.
 2. Run `yarn generate-icons` to generate icon files and changes to `packages/components/icons/src/index.js`. This file should never be edited manually.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -225,6 +225,7 @@ These are informal steps we suggest you to follow when adding a new component.
 
 1. Add the raw SVG of the new icon to `packages/components/icons/src/svg`.
 2. Run `yarn generate-icons` to generate icon files and changes to `packages/components/icons/src/index.js`. This file should never be edited manually.
+3. Manually update the typings for icons in `merchant-center-application-kit` to include the new icon [here](https://github.com/commercetools/merchant-center-application-kit/blob/master/%40types/commercetools-uikit__icons/index.d.ts).
```
adnasa, line 228: why is this documented here?

adnasa: having to add declaration files in application-kit is necessarily not a concern of UI-kit.

emmenko (changes requested):
Thanks but this is not the right place to document this.

We plan to migrate uikit to TS, so the manual type declarations are not needed anymore in appkit.

nbryant-commercetools:
> Thanks but this is not the right place to document this.
> 
> We plan to migrate uikit to TS, so the manual type declarations are not needed anymore in appkit.

Thanks for providing some additional context.

Issue #1490, opened 2020-08-14
Upgrade popper.js to v2


emmenko:
Just noticed this in the installation logs:

```
popper.js@1.15.0: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1
```

https://www.npmjs.com/package/@popperjs/core

PR #1684, opened 2020-11-19
feat(tooltip): upgrade to popper v2

jonnybel:
### Summary

Upgrades the Tooltip component to use [Popper v2](https://popper.js.org/docs/v2/).

Fixes https://github.com/commercetools/ui-kit/issues/1490


packages/components/tooltip/src/tooltip.styles.js
```diff
@@ -2,20 +2,6 @@ import styled from '@emotion/styled';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 import Constraints from '@commercetools-uikit/constraints';
 
-const getOffsetMargin = ({ placement }) => {
```
jonnybel, line 5: This is no longer needed because the margin offset is now set directly through the proper popper `offset` modifier.

packages/components/tooltip/src/tooltip.js
```diff
@@ -53,10 +71,14 @@ const Tooltip = (props) => {
...
   }, []);
 
-  const { reference, popper } = usePopper({
+  const [referenceElement, setReferenceElement] = React.useState(null);
+  const [popperElement, setPopperElement] = React.useState(null);
+
+  const popper = usePopper(referenceElement, popperElement, {
     placement: props.placement,
-    modifiers: props.modifiers,
+    modifiers: [...(popperDefaultModifiers || props.modifiers)],
```
emmenko, line 79: Hmm I think we need to do a deep merge, as otherwise we might lose the default modifiers.

packages/components/tooltip/src/tooltip.js
```diff
@@ -53,10 +71,14 @@ const Tooltip = (props) => {
     };
   }, []);
 
-  const { reference, popper } = usePopper({
+  const [referenceElement, setReferenceElement] = React.useState(null);
+  const [popperElement, setPopperElement] = React.useState(null);
```
emmenko, line 75: Shouldn't we use `useRef`? 🧐

emmenko (commented):
Thanks for looking into it!

packages/components/tooltip/src/tooltip.js
```diff
@@ -321,84 +342,9 @@ Tooltip.propTypes = {
     WrapperComponent: PropTypes.elementType,
   }),
   /**
-   * Provides a way to fine-tune an appearance of underlying Popper tooltip element. For more information, please check [Popper.js documentation](https://popper.js.org/popper-documentation.html#modifiers).
+   * Provides a way to fine-tune an appearance of underlying Popper tooltip element. For more information, please check [Popper.js documentation](https://popper.js.org/docs/v2/modifiers/).
    */
-  modifiers: PropTypes.shape({
```
tdeekens, line 326: The breaking changes in propper didn't include these right? I remember @pa3 added them back then to use them. Knowing that we bind all of UIKit to a breaking change if popper changes anything here potentially.

packages/components/tooltip/src/tooltip.js
jonnybel: It does seem odd, but that's how it's done in the official documentation:
https://popper.js.org/react-popper/v2/#example

packages/components/tooltip/src/tooltip.js
jonnybel: Ah I did a mistake here.

packages/components/tooltip/src/tooltip.js
jonnybel: I've checked and at least the modifiers that he needed still work the same as before, but a few of the previously supported ones do not exist anymore in v2 😕 

So now, I guess it is considered a breaking change, because we had exposed all the modifiers in the API 😬

As for the propTypes, I just removed their explicit shape because I think we don't need to fill the props table of our Readme with the same things that already exists in Popper's docs.

jonnybel:
Working on an issue detected in the VRT diff 👀

jonnybel:
The VRTs don't show this problem when run locally, so I suspect the problem is due to puppeteer's viewport size and Popper's preventOverflow mechanism 🤔

emmenko:
@jonnybel what's the status here?

jonnybel:
> @jonnybel what's the status here?

I tried to fix that VRT until I ran out of ideas and took a break 😬

PR #1691, opened 2020-11-22
fix: nohoist @storybook/theming

### LINKED ISSUE #1690: Broken storybook
**Describe the bug**
Storybook does not work properly anymore. This seems to be related to having migrated to Emotion v11, while Storybook still relies on older versions.

We would probably need to do some patching with Yarn workspaces (if possible).

**Expected behavior**
Storybook should work without errors.

**Screenshots**
image

emmenko:
Fixes #1690

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,11 @@
       "presets/*",
       "visual-testing-app"
     ],
-    "nohoist": ["react-virtualized/dom-helpers"]
+    "nohoist": [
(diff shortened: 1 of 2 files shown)


tdeekens (approved):

emmenko:
The prod build still has the issue bu locally the problem is fixed. 🙃

emmenko:
Ok, the preview seems to work now.

PR #1696, opened 2020-11-23
chore: setup eslint for typescript, fix lint errors

emmenko:


tdeekens (commented):
Looks good. To you plan to use any TypeScript preset like xo or will continue to pick rules yourself. Couldn't spot anything in the TS conf.

emmenko:
This is the "same" config as we use in appkit. For now I would go for consistency.

If we find a more "zero config" approach, I'm happy to switch over (in both repos).

tdeekens (approved):

tdeekens:
Sounds good. I had good and bad experiences with xo. At times rules were annoyingly strict, sometimes a great learning and other times silly subjective.

adnasa (approved):
nice. thanks for doing this <2 (deliberate typo)

PR #1699, opened 2020-11-24
refactor: generate icons using svgr cli

emmenko:
Instead of using a custom script to generate the icons, we can use the `@svgr/cli` which supports custom templates.

https://react-svgr.com/docs/cli/


svgr.config.js
```diff
@@ -0,0 +1,78 @@
...
+    const exportName = `${basename.replace(/react/i, 'Icon')}`;
+    return `export { default as ${exportName} } from './${basename}'`;
+  });
+  return [
+    `// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+// This file is created with 'yarn generate-icons'.`,
+  ]
+    .concat(exportEntries)
+    .join('\n');
+};
```
emmenko, line 22: This generates the `index.js` with all the exports.

svgr.config.js
```diff
@@ -0,0 +1,78 @@
...
+const Component = (props) => {
+  const theme = useTheme();
+  return <${opts.state.componentName} {...props} css={getIconStyles(props, theme)} />
+};
+Component.propTypes = iconPropTypes;
+
+export default Component;
+  `;
+  return template.smart(code, babelOptions)(templateOptions);
+};
```
emmenko, line 62: This generates the actual React component file.

svgr.config.js
```diff
@@ -0,0 +1,78 @@
...
+  const babelOptions = {
+    preserveComments: true,
+    plugins: ['jsx', opts.typescript && 'typescript'].filter(Boolean),
+  };
+  const templateOptions = { JSX: jsx };
+
+  const code = `// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+// This file is created with 'yarn generate-icons'.
+// Original SVG file: ${opts.state.filePath}
+${templateCreateStyledIcon}
```
emmenko, line 48: The styling functions are injected here from the `templates/icons.styles.js`.

templates/create-styled-icon.js
```diff
@@ -82,18 +78,3 @@ export const iconPropTypes = {
   ]),
   size: PropTypes.oneOf(['small', 'medium', 'big', 'scale']),
 };
-
-export default function createStyledIcon(Component, displayName) {
```
emmenko: This function wasn't used apparently.

templates/icon.styles.js
```diff
@@ -82,18 +78,3 @@ export const iconPropTypes = {
   ]),
   size: PropTypes.oneOf(['small', 'medium', 'big', 'scale']),
 };
-
-export default function createStyledIcon(Component, displayName) {
```
emmenko, line 86: This function wasn't used apparently.

package.json
```diff
@@ -17,7 +17,7 @@
     "auth": "npm_config_registry=https://registry.npmjs.org npm whoami",
     "clean": "lerna exec 'rm -rf dist'",
     "generate-readmes": "yarn generate-readme --all-workspace-packages",
-    "generate-icons": "node ./scripts/generate-icon-exports.js",
+    "generate-icons": "lerna run generate-icons && prettier --write '**/icons/components/*.js'",
```
tdeekens, line 20: 🤔 do we need lerna here or could we use yarn workspaces to run it?

adnasa (approved):
👍

package.json
emmenko: yarn workspaces fails if a package does not have the script defined. Lerna is more smart about it and only runs the command for the packages that contains the given script.

jonnybel (approved):

package.json
tdeekens: Oki doks. I was just hoping we could get rid of lerna by e.g. having a `exit x` or something in all scripts.

emmenko: We could but it's still not really better because yarn workspaces will run in all packages anyway, while lerna only runs to the few packages that have the command.

What we can do is to have our own "workspace" script, using the `manypkg` API to inspect the packages.

PR #1703, opened 2020-11-30
chore: to add horizontal constraint to collapsible panel

YahiaElTai:
#### Summary

This PR adds a horizontal constraint to the collapsible panel with a default value of `scale`.
This change will not be a breaking change since the value is `scale` by default but will allow us to remove all usages of nested `Constraints.Horizontal` inside `CollapsiblePanel` in the MC.

**Note**: I have updated the `min-width` value on the `CollapsiblePanel` from `550px` to `284px` to match the new column system we have for the constraints but also to help with a use case we have in the merchant center where we need the panel to go lower than `550px`
The design team doesn't know why the value `550px` was assigned in the first place, and it seems like lowering it won't affect anything.

All changes in this PR have been approved by the design team.


YahiaElTai:
I have updated percy snapshots, only one new prop added.

packages/components/collapsible-panel/src/collapsible-panel.styles.js
```diff
@@ -56,7 +56,7 @@ const getHeaderContainerStyles = (props, isOpen) => {
 
 const baseContainerStyles = css`
   position: relative;
-  min-width: 550px;
+  min-width: 284px;
```
jonnybel, line 59: I suppose this is intended to be the value of the smallest supported constraint

```suggestion
  min-width: ${vars.constraint6};
```

.changeset/chilled-chairs-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools-uikit/collapsible-panel": patch
```
jonnybel, line 2: I think we should consider these `feat`, since we're adding a new prop to the component 🙂

jonnybel (approved):
Looks good! Thanks :)

packages/components/collapsible-panel/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@babel/runtime": "7.12.5",
     "@babel/runtime-corejs3": "7.12.5",
+    "@commercetools-uikit/constraints": "10.42.2",
```
emmenko, line 28: Nit: can you sort the dep alphabetically?

packages/components/collapsible-panel/src/collapsible-panel.js
```diff
@@ -211,6 +218,27 @@ CollapsiblePanel.propTypes = {
     // uncontrolled component does not have `onToggle` so no validation needed.
     return null;
   },
+  /**
+   * Horizontal size limit of the input fields.
+   */
+  horizontalConstraint: PropTypes.oneOf([
+    'm',
+    'l',
+    'xl',
```
emmenko, line 227: I would leave them out, as this is a new prop and we don't need to be backwards compatible.

packages/components/collapsible-panel/src/collapsible-panel.styles.js
```diff
@@ -56,7 +56,7 @@ const getHeaderContainerStyles = (props, isOpen) => {
 
 const baseContainerStyles = css`
   position: relative;
-  min-width: 550px;
+  min-width: ${vars.constraint6};
```
emmenko, line 59: 🎉

emmenko (changes requested):
We should provide at least a couple of VRTs for the constraint sizes.

emmenko:
The VRTs show some weird behavior:

Screenshot 2020-12-03 at 14 09 09

@YahiaElTai please don't approve the VRTs by yourself. One of the reviewer should do that. Thanks.

YahiaElTai:
> The VRTs show some weird behavior:
> 
> Screenshot 2020-12-03 at 14 09 09
> 
> @YahiaElTai please don't approve the VRTs by yourself. One of the reviewer should do that. Thanks.

Ah so these snapshots are to be approved by the reviewers not me, sorry xD should have asked for help. I will take a look Thanks for the details

packages/components/collapsible-panel/src/collapsible-panel.js
```diff
@@ -38,77 +39,83 @@ const CollapsiblePanel = (props) => {
...
-        <div
-          css={[baseContainerStyles, getThemeStyle(props.theme)]}
-          // Allow to override the styles by passing a `className` prop.
-          // Custom styles can also be passed using the `css` prop from emotion.
-          // https://emotion.sh/docs/css-prop#style-precedence
-          className={props.className}
+        <Constraints.Horizontal
+          max={Constraints.parseHorizontalConstraintProp(
+            props.horizontalConstraint
+          )}
```
emmenko, line 45:
```suggestion
          max={props.horizontalConstraint}
```

The `parseHorizontalConstraintProp` is only used for backwards compatibility. We can simply require the new prop now as-is.

emmenko (commented):
Thanks for the fix!

emmenko (approved):
Thanks! 🎉

jonnybel:
The VRTs seemed to be good now 👍 The diffs are only due to the new prop and added specs.

emmenko:
Let's go! Thanks again 👌

PR #1718, opened 2020-12-21
feat(pagination): add new pagination component

jonnybel:
#### Summary

This PR adds a new component: Pagination.
> A component that provides an interface for navigating through pages of items


packages/components/pagination/src/index.js
```diff
@@ -0,0 +1,3 @@
+export { default } from './pagination';
```
emmenko, line 1: Early feedback. I would split this into multiple components and expose them like

* `TablePagination` (the full combined components)
image

* `PaginationInput` (only the left/right page part)
image

* `PerPageInput` (only the selector part)
image

---

Naming inspiration from [material ui](https://material-ui.com/components/pagination/#table-pagination)

jonnybel: Splitting makes sense 👍 

I was more inclined towards `Pagination` instead of `TablePagination` because I can see this component being used for other things than just tables.

packages/components/pagination/src/page-navigator/page-navigator.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
```
adnasa, line 1: Why not TS first?

packages/components/pagination/src/page-size-selector/page-size-selector.js
```diff
@@ -0,0 +1,58 @@
...
+    props.onPageSizeChange(Number(event.target.value));
+  };
+
+  const pageSizeSelectorId = sequentialId();
+
+  return (
+    <Spacings.Inline alignItems="center">
+      <Constraints.Horizontal max={2}>
+        <SelectInput
+          id={pageSizeSelectorId}
```
adnasa, line 27: why do we need `sequentialId` for this?

packages/components/pagination/src/pagination.js
```diff
@@ -0,0 +1,68 @@
...
+      />
+    </Container>
+  );
+}
+Pagination.displayName = 'Pagination';
+Pagination.propTypes = {
+  /**
+   * Number of items per page
+   */
+  pageSize: PropTypes.number.isRequired,
```
adnasa, line 41: I would use `oneOf` here.
At least React will set a warn on the log if the wrong value is provided

packages/components/pagination/src/pagination.js
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Total number of items across all pages
+   */
+  totalItems: PropTypes.number.isRequired,
+  /**
+   * The currently selected page
+   */
+  currentPage: PropTypes.number.isRequired,
+  /**
+   * Called when the page is changed
```
adnasa, line 51: Missing `Signature`?

packages/components/pagination/src/pagination.js
```diff
@@ -0,0 +1,68 @@
...
+   */
+  onPageChange: PropTypes.func.isRequired,
+  /**
+   * Options to display as pageSizes
+   */
+  pageSizeOptions: PropTypes.arrayOf(PropTypes.number).isRequired,
+  /**
+   * Called when pageSize is changed
```
adnasa, line 59: Same here, `Signature: ...` is missing

packages/components/pagination/src/page-navigator/utils.js
```diff
@@ -0,0 +1,16 @@
+export const isValid = (page, totalPages) => {
```
emmenko, line 1: Isn't this `isInvalidPage`? 🤔

packages/components/pagination/src/page-navigator/utils.js
```diff
@@ -0,0 +1,16 @@
...
+  if (page > totalPages) return true;
+  if (page <= 0) return true;
+
+  return false;
+};
+
+export const normalizeValue = (value, totalPages) => {
```
emmenko, line 8:
```suggestion
export const normalizePageValue = (value, totalPages) => {
```

packages/components/pagination/src/pagination.js
```diff
@@ -0,0 +1,68 @@
...
+function Pagination(props) {
+  const totalPages = Math.ceil(props.totalItems / props.pageSize);
+
+  const currentPageItems =
+    props.currentPage === totalPages
+      ? props.totalItems - props.pageSize * (props.currentPage - 1)
+      : props.pageSize;
+
+  return (
+    <Container>
```
emmenko, line 21: Can't we use `SpacingsInline` here? 🤔

packages/components/pagination/src/page-size-selector/page-size-selector.js
jonnybel: We just need an internal id actually

packages/components/pagination/src/page-navigator/page-navigator.js
```diff
@@ -0,0 +1,106 @@
...
+    event.preventDefault();
+
+    setPage(normalizedValue);
+
+    props.onPageChange(normalizedValue);
+  };
+
+  const onBlurNormalize = () => setPage(normalizedValue);
+
+  const isDisabled = props.totalPages === 0;
```
tdeekens: Nit: Maybe have constants first then fns.

packages/components/pagination/src/page-navigator/page-navigator.js
emmenko: There are dependent packages that we need to migrate first.

packages/components/pagination/src/page-navigator/page-navigator.js
```diff
@@ -0,0 +1,106 @@
...
+    if (nextPage > props.totalPages) return null;
+
+    setPage(nextPage);
+    props.onPageChange(nextPage);
+  };
+
+  const isPreviousDisabled = page <= 1;
+  const isNextDisabled = page >= props.totalPages;
+
+  const pageNumberInputId = sequentialId();
```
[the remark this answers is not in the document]
jonnybel: I changed the id generation strategy. We don't actually need sequentialId

jonnybel:
The build is failing on the `preconstruct dev` step for a bunch of packages like so:
```
🎁 error @commercetools-uikit/accessible-button packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/constraints packages must have at least one entrypoint, this package has no entrypoints
🎁 info If want to learn more about the above error, check https://preconstruct.tools/errors
🎁 error @commercetools-uikit/design-system packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/flat-button packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/icons packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/input-utils packages must have at least one entrypoint, this package has no entrypoints
🎁 info If the error is not there and you want to learn more about it, open an issue at https://github.com/preconstruct/preconstruct/issues/new
🎁 error @commercetools-uikit/label packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/number-input packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/secondary-icon-button packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/select-input packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/select-utils packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/spacings packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/spacings-inline packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/spacings-inset packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/spacings-inset-squish packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/spacings-stack packages must have at least one entrypoint, this package has no entrypoints
🎁 error @commercetools-uikit/text packages must have at least one entrypoint, this package has no entrypoints
error Command failed with exit code 1.
```
I haven't made any changes to these packages so I'm not sure what am I missing.

packages/components/pagination/src/pagination.js
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * A callback function, called when the page is changed.
+   * <br>
+   * Signature: `(page: number) => void`
+   */
+  onPageChange: PropTypes.func.isRequired,
+  /**
+   * Options to display as pageSizes
+   */
+  pageSizeOptions: PropTypes.arrayOf(PropTypes.number).isRequired,
```
emmenko, line 54: @mariabarrena I don't we remember what we decided here. Do we provide specific options of can you pass whatever value you want?

presets/ui-kit/src/index.js
```diff
@@ -28,6 +28,11 @@ export { default as Grid } from '@commercetools-uikit/grid';
 export { default as Label } from '@commercetools-uikit/label';
 export { default as LoadingSpinner } from '@commercetools-uikit/loading-spinner';
 export { ErrorMessage, WarningMessage } from '@commercetools-uikit/messages';
+export {
+  PageNavigator,
+  PageSizeSelector,
+  Pagination,
+} from '@commercetools-uikit/pagination';
```
emmenko, line 35: You have to specify this as a dependency in `package.json`

jonnybel: Hmm, can I do it already on this PR though, since the package is only released after this gets merged?

I think we've been always adding new components to the preset on a follow-up PR.

packages/components/pagination/src/pagination.js
mariabarrena: All properties needed are specified in the checklist: https://wiki.commercetools.com/display/shield/Pagination%3A+component+checklist

hasLimit (Boolean):
 if yes, the limit section is displayed. Default is "yes"

LimitValues:
Allows consumers to choose the type of limit between:
• option 20/50: when querying many items is expensive (with possibility to have a third breakpoint of 100 items). Default: 20
• option 200/500: when querying many items does not harm the performance of your application/feature. Default: 200

presets/ui-kit/src/index.js
emmenko: It's a monorepo 🙂
So yes, you should add it now.

packages/components/pagination/src/pagination.js
emmenko: Hmm that doesn't really work.

If we go for pre-defined ranges, maybe we should provide a different prop:

```js
pageSizeRange: 'small' | 'large'
```

Where `small` maps to `20,50,100` and `large` maps to `200,500`

adnasa:
What are presets for and where to read about it?
I tried finding resource about this

emmenko:
@adnasa presets are packages that re-export other packages. To reduce the number of packages that you would need to install.

emmenko:
I'll take a look at the preconstruct error

emmenko:
@jonnybel the last commit fixes the issue (you were using outdated versions in the package.json dependencies).

I also rebased with master.

packages/components/pagination/src/page-size-selector/page-size-selector.js
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  return (
+    <Spacings.Inline alignItems="center">
+      <Constraints.Horizontal max={2}>
+        <SelectInput
+          id={pageSizeSelectorId}
+          name="page-size-selector"
+          value={props.pageSize.toString()}
+          options={options.map((option) => ({
```
tdeekens, line 54: Nit: in some other parts we use a lot of `useCallback` etc. If we want to optimise for that we should `useMemo` here I think.

tdeekens (approved):

adnasa (approved):

PR #1719, opened 2020-12-22
fix(design-system): correctly bundle and export type declarations

emmenko:


emmenko:
The VRTs show some subtle font changes but it still seems fine to me.

@jonnybel can you double check as well?

jonnybel:
Looks like a lot of components have also lost their styles on things like borders and colors.
For instance, SelectInput:
image

PR #1721, opened 2020-12-30
feat(generators/readme): add basic support for TS types

emmenko:
As we start migrating components to TS, we also need to continue being able to generate the props table.
The `react-docgen` library does support parsing TS types, but the structure is a bit different, so we need to map those use cases separately.

This PR adds let's say 80% support for TS types. Special use cases will have to be tackled once we bump into those.


generators/readme/test/generate-readme.spec.ts
```diff
@@ -91,3 +96,117 @@ describe('generate README', () => {
...
+        /**
+         * The title of the movie.
+         */
+        title: string;
+        /**
+         * The release date of the movie (ISO).
+         */
+        releaseDate: string;
+      }
+      \`\`\`
```
emmenko, line 208: Note that we "embed" the signature types like this, and link to them from the props table.
This way we keep the table simple, while having the type signatures properly documented.

generators/readme/src/index.ts
```diff
@@ -40,6 +42,10 @@ import parse from 'remark-parse';
 import mdx from 'remark-mdx';
 import gfm from 'remark-gfm';
 import stringify from 'remark-stringify';
+// @ts-ignore
+import * as reactDocgen from 'react-docgen';
+// @ts-ignore
+import setParamsTypeDefinitionFromFunctionType from 'typescript-react-function-component-props-handler';
```
emmenko, line 48: Can be a follow up to provide the "correct" types.

generators/readme/package.json
```diff
@@ -33,6 +33,7 @@
     "remark-stringify": "9.0.1",
     "shelljs": "0.8.4",
     "to-vfile": "6.1.0",
+    "typescript-react-function-component-props-handler": "1.1.1",
```
emmenko, line 36: This dependency is used to "fix" an issue with `react-docgen` when using `FC<>`. See https://github.com/reactjs/react-docgen/issues/387

PR #1732, opened 2021-01-12
refactor(utils): migrate to typescript

emmenko:


packages/utils/src/add-static-fields.ts
```diff
@@ -0,0 +1,15 @@
+import type { FC } from 'react';
+
+type StaticFields = Record<string, unknown>;
+
+export default function addStaticFields<
+  Props extends {},
```
tdeekens, line 6: Should we prefix types with T or not. I am open to either way. It helped me to have them prefixed but that is subjective.

emmenko: Yeah I also don't have strong opinions. I think I tend to prefix them if the types are exported.

adnasa: I used to have the opinion of prefixing them, but their role/purpose is becoming less meaningful since we have intellisense.

Since this is open source, I'd add the prefix to increase readability, especially for those who don't have intellisense (for whatever reason). The same principle applies to why we suffix function types with `*Fn` or lists with `*List` (list is not a type, but the suffix indicates a set or array)

what do you think? 

---

note: I think we should drop this component eventually
it is used in a few instance and none on MC if I recall correctly. best to double check

adnasa (approved):
🙌🏽

packages/utils/src/add-static-fields.ts
tdeekens: Yeah. It's not like I made up that T-prefix. I saw it in Apollo and react-ui where it helped me reading the code knowing later what a type is and what a variable.

adnasa: now that you mentioned it; another "beauty" with adding the prefix is that you can identify a type/enum/whateverz when browsing in github. that's a bonus.

adnasa: pushing a commit to fix this so we can merge sooner 😝

adnasa (approved):

tdeekens (approved):
TTTTTTTTTT TypeScript!

adnasa:
🔥

PR #1734, opened 2021-01-13
refactor(localized-utils): use typescript

adnasa:
#### Summary

- migrate `localized-utils` to typescript


packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,164 @@
...
+
+// splits the languages into two groups:
+//  - the first group starts with the same tag as the selected language
+//  - the second group starts with a different tag
+export const splitLanguages = (
+  selectedLanguage: string,
+  languages: string[]
+): TSplittedLanguageGroups => {
+  const primaryLanguage = getPrimaryLanguage(selectedLanguage);
+  const related = languages.filter(
```
tdeekens, line 40: Can't this be one for loop?

packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,164 @@
...
+    switch (key) {
+      case 'data-track-component':
+      case 'data-testid':
+      case 'data-test':
+        acc[key] = `${value}-${language}`;
+        break;
+      default:
+        acc[key] = value;
+    }
+    return acc;
```
tdeekens, line 84: Nit: maybe name acc to what it is.

packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,164 @@
...
+  }, {} as TLocalizedDataAttributes);
+};
+
+export const getHasErrorOnRemainingLanguages = <TErrors extends {}>(
+  errors?: TErrors,
+  selectedLanguage?: string
+): boolean => {
+  if (errors && Object.keys(errors).length > 0 && selectedLanguage) {
+    return (
+      Object.keys(errors).filter((language) => language !== selectedLanguage)
```
tdeekens, line 94: I think a .some would be more expressive.

packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,163 @@
...
+};
+
+// sorts the languages with the following priority:
+// - The selected language is placed first (e.g pt-BR)
+// - All languages using the same primary language as the selected language
+//   follow (e.g. pt, pt-PT). They are sorted alphabetically.
+// - All other languages follow, sorted alphabetically as well
+export const sortLanguages = (
+  selectedLanguage: string,
+  allLanguages: string[]
```
tdeekens, line 60: Just sharing something I struggled with in the past. I always wonder when I should give primitive types names and if that helps. So here `type TLanguage = string` and `TLanguages= TLanguage[]`. It felt silly at first but then I liked it. I don't know. Would love to hear your thoughts.

tdeekens (approved):
Thanks for the improvements.

packages/localized-utils/src/localized.ts
adnasa: I wanted to do the exact same thing, but decided to skip it.
It felt silly at first, and it still is silly.

Developers should be aware on the context of the variable they work with. The context is given through proper variable naming and accumulated domain knowledge. Because of this, it is redundant to extract type inference into a specified type. (e.g `languages = TLanguages[]` vs `languages = string[]`.

My internal guideline is:

1. Be meticulous about naming. Because of this, specifying primitives should be straight forward (simply `languages = string[]`). Given the reader of the code has some domain knowledge, `languages` must be `string[]`.
2. Specifying type should be considered bonus, and they don't apply on primitives.
3. Do you need to extract the primitive into a specified type?
Unless it is a field on, then you have the problem of mixing concerns.

packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export const omitEmptyTranslations = <TTranslations extends TLocalizedString>(
+  localizedString?: TTranslations
+): TLocalizedString => {
+  invariant(
+    typeof localizedString === 'object',
+    'omitEmptyTranslations must be called with an object'
+  );
+
+  return Object.entries(localizedString).reduce(
```
emmenko, line 144:
```suggestion
  return Object.entries(localizedString).reduce<TLocalizedString>(
```

emmenko (commented):
👌

packages/localized-utils/src/localized.ts
tdeekens: Thanks for taking the discussion. I agree with your points!

packages/localized-utils/src/localized.ts
```diff
@@ -0,0 +1,164 @@
+import invariant from 'tiny-invariant';
```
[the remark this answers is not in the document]
adnasa: oh my bad :)

tdeekens (approved):

emmenko (approved):

PR #1735, opened 2021-01-13
refactor(horizontal-constraint): use typescript

adnasa:
#### Summary 

- migrate `packages/components/constraints` to typescript


packages/components/constraints/package.json
```diff
@@ -36,6 +36,6 @@
     "react": ">= 16.8.0"
   },
   "readme": {
-    "componentPath": "./src/horizontal/horizontal.js"
+    "componentPath": "./src/horizontal/horizontal.tsx"
```
adnasa, line 39: I'm not sure about this part.

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -1,11 +1,40 @@
...
+  | 11
+  | 12
+  | 13
+  | 14
+  | 15
+  | 16
+  | 'scale'
+  | 'auto';
+
+type TConstraintProp = 'xs' | 's' | 'm' | 'l' | 'xl' | 'scale';
```
adnasa, line 27: To be deprecated and remove in the next major release. Keeping this for now to ease the friction.
If we remove this now, it becomes a breaking change and I'm not sure for the moment if I want to bake this into the typescript migration.

adnasa:
It would be great to prioritize a review/merge on this PR, then we can migrate inputs for example.

packages/components/constraints/src/helpers.ts
```diff
@@ -1,12 +1,15 @@
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 
-const getMaxPropTokenValue = (max) => {
+// `null` is derived from `getMaxPropEquivalent`
+const getMaxPropTokenValue = (max: number | string | null) => {
+  if (!max) return null;
+  // @ts-expect-error
```
tdeekens, line 6: What is the error?

emmenko: I think it's computed property name.
To be more type safe we would need to have a switch statement or something.
It can also be a follow up.

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -16,7 +45,7 @@ function getConstraintStyles(maxProp, constraintProp) {
   `;
 }
 
-const Horizontal = (props) => {
+const Horizontal = (props: TProps): JSX.Element => {
```
emmenko, line 48: I started using the `FC` type util. Let's go with that

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -46,36 +75,4 @@ const Horizontal = (props) => {
   );
 };
 
-Horizontal.displayName = 'Horizontal';
-Horizontal.propTypes = {
-  /**
-   * Determines scale of the constraint.
-   */
-  max: PropTypes.oneOf([
```
emmenko, line 54: Please keep these comments

packages/components/constraints/package.json
emmenko: It's for the readme generation. You should also run it

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -1,11 +1,40 @@
 import React from 'react';
-import PropTypes from 'prop-types';
```
emmenko, line 2: We should remove this dep from `package.json` (I probably forgot to do it in the other packages I migrated as well).

emmenko (commented):
Looking good!

packages/components/constraints/src/horizontal/horizontal.tsx
adnasa: okay, will do :=

packages/components/constraints/src/helpers.ts
adnasa: it's the computed property. I described in L3 where it comes from ;)

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -1,11 +1,46 @@
...
+  | 16
+  | 'scale'
+  | 'auto';
+
+type TConstraintProp = 'xs' | 's' | 'm' | 'l' | 'xl' | 'scale';
+type TProps = {
+  /**
+   * Determines scale of the constraint.
+   */
+  max: TMaxProp;
```
emmenko, line 32: This field is optional

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -1,11 +1,46 @@
...
+  /**
+   * @deprecated: Please use the `max` prop instead. Determines scale of the constraint.
+   */
+  constraint: TConstraintProp;
```
emmenko, line 36: This field is optional

packages/components/constraints/src/horizontal/horizontal.tsx
```diff
@@ -31,15 +31,15 @@ type TProps = {
    */
   max?: TMaxProp;
   /**
-   * @deprecated: Please use the `max` prop instead. Determines scale of the constraint.
```
tdeekens: This can actually be quite useful I think. In the furture :D https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#deprecated-support

adnasa: I love it
image

tdeekens: Yup. It's about that. TypeScript basically does a ~value~ in the type's IntelliSense.

tdeekens (approved):

packages/components/constraints/README.md
```diff
@@ -44,11 +44,11 @@ import Constraints from '@commercetools-uikit/constraints';
...
-| ------------ | ------------------------------------------------------------------------------------------------------ | :------: | ------- | ---------------------------------------------------------------------------------------- |
-| `max`        | `enum`<br>Possible values:<br>`1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 'scale', 'auto'` |          |         | Determines scale of the constraint.                                                      |
-| `constraint` | `enum`<br>Possible values:<br>`'xs', 's', 'm', 'l', 'xl', 'scale'`                                     |          |         | DEPRECATING SOON: Please use the `max` prop instead. Determines scale of the constraint. |
-| `children`   | `node`                                                                                                 |    ✅    |         |                                                                                          |
+| Props        | Type                                                                                                      | Required | Default | Description                         |
```
emmenko, line 47: Regenerate the readme

emmenko (commented):
Regenerate the readme, otherwise looks good.

packages/components/constraints/package.json
```diff
@@ -29,13 +29,12 @@
     "@commercetools-uikit/utils": "10.42.2",
     "@emotion/react": "^11.1.1",
     "@emotion/styled": "^11.0.0",
-    "prop-types": "15.7.2",
```
emmenko, line 32: Actually thinking about that, I think we should keep it.
The reason is because the bundles do contain prop types (generated), and the dependency is thus required.

Maybe in the future we can make it an optional peer dependency.

emmenko (changes requested):

packages/components/constraints/package.json
tdeekens: If it was an optional peer dependency and somebody wouldn't install it then the bundles would be useless in dev mode? 🤔

emmenko: Good question. Anyway, to avoid issues let's keep it.

adnasa:
I addressed all comments, @emmenko 
please have a look :)

emmenko (approved):
🚀

PR #1736, opened 2021-01-14
refactor(spacings): to typescript

adnasa:
#### Summary

- migrates `spacings`

Note that we can not remove the dep `prop-types`.
To do this, we have to convert `Text` to typescript first. The other option would be declaration file which is not what I want.


packages/components/spacings/spacings-inline/src/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  | 'space-around'
+  | 'space-evenly';
+
+export type TScale = 'xs' | 's' | 'm' | 'l' | 'xl';
+
+export type TProps = {
+  scale?: TScale;
+  alignItems?: TAlignItem;
+  justifyContent?: TJustifyContent;
+  children?: React.ReactNode;
```
tdeekens, line 26: Strange that children are optional but seem sit was in the prop-types too.

tdeekens (approved):
💯 Thanks. Can't spot anything 💯

packages/components/spacings/spacings-inline/src/types.ts
adnasa: It somehow makes sense, if we want to be flexible in nullified node instance

```jsx
<Spacings.Stack>{isAuthorized ? <div />: null }</Spacings.Stack>
```

packages/components/spacings/spacings-inline/src/types.ts
```diff
@@ -0,0 +1,27 @@
+// @TODO extract this into a shared file
```
adnasa, line 1: This is something we can do now or follow up. it's not priority IMO.
My concern with `spacings` is that we have redundant requirements.

packages/components/spacings/spacings-inline/src/types.ts
tdeekens: Fair point. Just wondered if then `<Spacings.Stack>{isAuthorized ? <div />: </> }</Spacings.Stack>` would be more apt. I don't know if fragment counted as a React.ReactNode I just can't remember React's type inheritance.

Anyway, thanks for clarifying.

.changeset/yellow-rats-heal.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-uikit/spacings-inline': minor
+'@commercetools-uikit/spacings-inset': minor
+'@commercetools-uikit/spacings-inset-squish': minor
+'@commercetools-uikit/spacings-stack': minor
+---
+
+Convert @commercetools-uikit/spacings to typescript
```
emmenko, line 8:
```suggestion
Migrate spacing components to typescript
```

packages/components/spacings/spacings-inline/src/inline.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { filterDataAttributes } from '@commercetools-uikit/utils';
+import getStyles from './inline.styles';
+import type { TProps } from './types';
```
emmenko, line 4: I prefer to have the `import type` imports before the normal imports

packages/components/spacings/spacings-inline/src/types.ts
emmenko: We should have the types within the same component file, otherwise the prop types generation won't work (babel doesn't know about other files when compiling this file).

packages/components/spacings/spacings-inline/src/types.ts
emmenko: `null` should be a valid `ReactNode` actually. I would keep `children` as required.

emmenko (changes requested):

packages/components/spacings/spacings-inline/src/inline.tsx
adnasa: The thing about this is that this is also needed in `inline.styles` file.
Should I then import it there from the this component file?

packages/components/spacings/spacings-inline/src/types.ts
adnasa: The thing about this is that this is also needed in `inline.styles` file.
Should I then import it there from the this component file?

emmenko: Yes

emmenko (approved):

PR #1739, opened 2021-01-18
chore(renovate): preserve semver ranges

emmenko:
* Moves renovate config into `.github` folder
* Uses `preserveSemverRanges` instead of `pinOnlyDevDependencies`

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,19 +1,10 @@
 {
-  "extends": [
-    "config:base",
-    ":pinOnlyDevDependencies",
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,29 +1,18 @@
 {
-  "extends": [
-    "config:base",
-    ":pinOnlyDevDependencies",
```
emmenko, line 4: As we saw in appkit, this could interfere with the `manypkg` validation.

adnasa (approved):
👍

PR #1742, opened 2021-01-21
fix(types): avoid using `React.FC` as it does not properly work with `defaultProps`.

emmenko:
I forgot why we weren't using `FC` in appkit. It's because it does not properly work with `defaultProps`.

This PR fixes that.

PS: additionally, I also migrated the `spacings` preset to TS, as otherwise the spacing component types are not available when using the preset.


tdeekens (approved):
Maybe ADRs for ui-kit could help to document such decisions?

emmenko:
Yes we should document how to use/write TS components.

adnasa (approved):
so don't use `React.FC` then? 😆

emmenko:
Yep 🙂

PR #1744, opened 2021-01-21
fix(spacings): be explicit in typing Spacings object

emmenko:
Follow up of #1742 

Letting TS inferring the exported type somehow leads to weird imports such as `import("@commercetools-uikit/spacings-inline/src/inline")`, whichis an invalid path.


tdeekens (approved):
😍

emmenko:
Merging and releasing to unblock me.

PS: we can think if there is a "simpler" way to do that, especially for other bigger presets such as buttons or uikit.

PR #1746, opened 2021-01-22
refactor(components/text): to typescript

adnasa:
#### Summary

- Migrates `text` to typescript

#### Small changes in behavior

- on Headline/Subheadline: if `as` nor `elementType` are specified, we render a plain text and set a warning on the console. 

I think we should drop the `elementType` prop now. Since this is a breaking change, it makes sense to start working on that topic but I welcome other suggestions on the release strategies around this.


packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+);
+Text.displayName = 'Text';
+
+type THeadlineProps = {
+  as?: 'h1' | 'h2' | 'h3';
+  // @deprecated
+  elementType?: 'h1' | 'h2' | 'h3';
+  title?: string | null;
+  truncate?: boolean;
+} & TTextProps;
```
adnasa, line 68: I'm not sure if this is a good idea anymore.
feedback welcome

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+};
+
+type TDeprecatedProps = {
+  // @deprecated
+  elementType: string;
+};
+
+const warnIfNoTextProps = (props: TTextProps, componentName) => {
+  if (!props.intlMessage && !React.Children.count(props.children)) {
+    console.error(
```
adnasa, line 34: I should be using `tiny-invariant`. update coming up

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+const Headline = (props: THeadlineProps) => {
+  const theme = useTheme();
+
+  if (props.elementType) {
+    warnDeprecatedProp(
+      'elementType',
+      'TextHeadline',
+      `\n \`elementType\` is deprecated. \n Please use "as" prop instead.`
+    );
+  }
```
adnasa, line 79: I wanted at first to extract this into a `warnWithDeprecatedProps`, but was unsure if it made sense since it affects a small number of components and not the entire suit, so I kept it here

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const HeadlineElement = props.as || props.elementType;
+
+  // however, if none of the prop is specified,
+  // we render plain text and set a warning on the log.
+  if (!HeadlineElement) {
+    console.error(
+      'ui-kit/Text: You attempt to render a TextHeadline without specifying `as` prop.'
+    );
+    return <Text intlMessage={props.intlMessage}>{props.children}</Text>;
+  }
```
adnasa, line 94: It makes sense to me to just render plain text, rather than falling back to e.g a `h1`

other suggestions welcome.
note, that at some point, we will remove `elementType` and make `as` required, so this shouldn't be needed.
my question is if we accept this as an intermediate solution.

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+type TTextProps = {
+  intlMessage?: MessageDescriptor;
+  children?: React.ReactNode;
+};
+
+type TDeprecatedProps = {
+  // @deprecated
+  elementType: string;
+};
```
emmenko, line 30: This seems unused. Remove.

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+    ) : (
+      props.children
+    )}
+  </>
+);
+Text.displayName = 'Text';
+
+type THeadlineProps = {
+  as?: 'h1' | 'h2' | 'h3';
+  // @deprecated
```
emmenko, line 64:
```suggestion
  // @deprecated: use `as` prop instead.
```

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+  elementType?: 'h1' | 'h2' | 'h3';
+  title?: string | null;
```
emmenko, line 66: Maybe extract this into `TNonEmptyString`? 🤔

packages/components/text/src/text.tsx
emmenko: That's ok. I would maybe rename the type `TBasicTextProps`.

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+  intlMessage?: MessageDescriptor;
+  children?: React.ReactNode;
+};
+
+type TDeprecatedProps = {
+  // @deprecated
+  elementType: string;
+};
+
+const warnIfNoTextProps = (props: TTextProps, componentName) => {
```
emmenko, line 32:
```suggestion
const warnIfMissingContent = (props: TTextProps, componentName) => {
```

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+  warnIfNoTextProps(props, 'TextHeadline');
+
+  // For backwards compatibility
+  // we allow both `as` and `elementType` to be optional.
+  const HeadlineElement = props.as || props.elementType;
+
+  // however, if none of the prop is specified,
+  // we render plain text and set a warning on the log.
+  if (!HeadlineElement) {
+    console.error(
```
emmenko, line 90: Use `invariant`

packages/components/text/src/text.tsx
emmenko: I think it's fine yes. Thanks!

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+    </HeadlineElement>
+  );
+};
+Headline.displayName = 'TextHeadline';
+
+type TSubheadlineProps = {
+  as?: 'h4' | 'h5';
+  // @deprecated
+  elementType?: 'h4' | 'h5';
+  title?: string;
```
emmenko, line 112: Use the "non empty string" type, or add `| null`

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,248 @@
...
+    </div>
+  );
+};
+Wrap.displayName = 'TextWrap';
+
+type TBodyProps = {
+  as: 'span' | 'p';
+  isBold?: boolean;
+  isItalic?: boolean;
+  // @deprecated
```
emmenko, line 172:
```suggestion
  // @deprecated: use `as="span"` instead
```

emmenko (commented):
Whoop whoop 🙌

packages/components/text/src/text.tsx
adnasa: what do want to accomplish with `TNonEmptyString`?
This could just be `title: string` actually, because otherwise the approach is to emit an error to the log.

Or am I missing something?

tdeekens (approved):

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,284 @@
...
+    );
+  }
+};
+
+const warnIfMissingContent = (
+  props: TBasicTextProps,
+  componentName: string
+) => {
+  const hasContent =
+    getIsIntlMessage(props.intlMessage) ||
```
emmenko, line 59: Hmm I don't think we need this. Is there a particular reason for that? 
Here we only check if the `prop.intlMessage` is defined. The format of the prop is then checked by the TS type.

packages/components/text/src/text.tsx
```diff
@@ -0,0 +1,284 @@
...
+    Boolean(React.Children.count(props.children));
+
+  invariant(
+    hasContent,
+    oneLine`
```
emmenko, line 64: Maybe we can just drop this dependency (and thus the related dep type)? It feels like a bit unnecessary...

packages/components/text/src/text.tsx
adnasa: I want to make sure that it is a `MesssageDescriptor` and a string as a result of `intl.formatMessage` from the consumer side.

packages/components/text/src/text.tsx
adnasa: My preference is to deconstruct a long string into multi lines so that it is easier to read, and I value that.
I propose we keep it

packages/components/text/src/text.tsx
emmenko: Yes TS checks that. And if the user passes a wrong object, react-intl complains. I don't see a point in having to check things extra on our own.

packages/components/text/src/text.tsx
emmenko: You can simply do

```js
[
  'This is the first line',
  'Then a second',
  'and a third.'
].join('\n');
```

I just want to avoid having unnecessary dependencies.

packages/components/text/src/text.tsx
adnasa: I get that, but note that the consumer won't be fully aware the `intlMessage=MessageDescriptor` type.
Not even intellisense helps here

(example from a plain js consumer side)
image

emmenko: Consider that JS users still have the prop type, which should be an object if I'm not mistaken (you can inspect the bundles).

It's a little trade off at the end.

emmenko (approved):
Thanks, this should be good to go now 👌

PR #1770, opened 2021-02-09
feat(pagination-story): add range selector and derive a minimum page size

adnasa:
#### Summary

- adds `range` into `PageSizeSelector` (only)


packages/components/pagination/src/pagination.story.js
```diff
@@ -36,21 +36,27 @@ storiesOf('Components|Pagination', module)
...
-            pageSizeRange={'s'}
-            currentPageItems={1}
-            onPageSizeChange={onPageSizeChange}
-          />
-        )}
-      />
-    </Section>
-  ))
+  .add('PageSizeSelector', () => {
+    const range = select('range', ['s', 'm', 'l'], 's');
```
mohib0306, line 40: The `s` and `m` will have the same `20` value. Is that on purpose

mohib0306 (commented):
Looks good, just asked a question about value for `m`.

packages/components/pagination/src/pagination.story.js
adnasa: yes, since `20` is the minimum pageSize in respective range

mohib0306 (approved):

tdeekens (approved):

PR #1771, opened 2021-02-09
refactor(components/label): migrate to typescript

adnasa:
#### Summary

- migrate `Label` to typescript.


packages/components/label/src/label.tsx
```diff
@@ -6,21 +8,69 @@ import { FormattedMessage } from 'react-intl';
...
+    ].join(' ')
+  );
+  invariant(
+    hasContent,
+    [
+      'Warning: Failed prop type:',
+      `The prop \`children\` is marked as required in \`Label\``,
+      'but its value is `undefined`',
+    ].join(' ')
+  );
```
adnasa, line 64: This is the same as [`warnIfMissingContent`](https://github.com/commercetools/ui-kit/blob/main/packages/components/text/src/text.tsx#L40:L63) but I don't see a strong reason yet to extract it into a util.

packages/components/label/src/label.tsx
```diff
@@ -0,0 +1,79 @@
...
+  htmlFor?: string;
+  isBold?: boolean;
+  isRequiredIndicatorVisible?: boolean;
+  tone?: 'primary' | 'inverted';
+} & TBasicTextProps;
+
+const Label = (props: TLabelProps) => {
+  if (props.id && props.htmlFor) {
+    invariant(
+      false,
```
tdeekens, line 38: Isn't this more of how the inventors thought of it? :)

```suggestion
    invariant(
      props.id && props.htmlFor,
```

tdeekens (approved):

packages/components/label/src/label.tsx
adnasa: haha I just migrated blindy. good catch, bud  👋 (slap)

adnasa: actually the boolean checked passed in is incorrect

packages/components/label/src/label.tsx
tdeekens: Don't do it until it hurts :)

tdeekens (approved):
Thanks! 💯

packages/components/label/src/required-indicator.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+
```
mohib0306, line 2: can be removed

mohib0306 (approved):

packages/components/label/src/label.tsx
emmenko: Can't we combine the message into one?

packages/components/label/src/label.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <label id={props.id} htmlFor={props.htmlFor}>
+      <Text.Body
+        tone={props.tone}
+        isBold={props.isBold}
+        intlMessage={props.intlMessage}
+      >
+        {props.children}
+        {props.isRequiredIndicatorVisible && <RequiredIndicator />}
+      </Text.Body>
```
emmenko, line 70: Looking at this, if you pass `intlMessage`, the Text children are not rendered, which means that the required indicator does not get rendered.

Can you check that please? It might be that we have always had this "issue" ...

emmenko (changes requested):
While we're at it, can you also automate the readme generation for this package?
Thanks!

packages/components/label/src/label.tsx
adnasa: that is a great catch! will revert to previous solution where we render `intlMessage ? <FormattedMessage /> : children`

adnasa:
> While we're at it, can you also automate the readme generation for this package?

Good idea, I think it makes sense to have this run on commit.

packages/components/label/src/label.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <label id={props.id} htmlFor={props.htmlFor}>
+      <Text.Body tone={props.tone} isBold={props.isBold}>
+        {props.intlMessage ? (
+          <FormattedMessage {...props.intlMessage} />
+        ) : (
+          props.children
+        )}
+        {props.isRequiredIndicatorVisible && <RequiredIndicator />}
+        {props.isRequiredIndicatorVisible && <RequiredIndicator />}
+      </Text.Body>
```
emmenko, line 72: Can't we maybe render the text as `span` and render the indicator next to it (instead of inside)?

PS: also, you have duplicated lines here.

adnasa: 🤦

packages/components/label/src/label.tsx
adnasa: I keep them separate. I find them easier to structure

emmenko: The point is that the user needs to provide one or the other. If you show two separate warning messages, it seems like both props are required, which is not true.

adnasa:
Hm, I see.
The build failed now because now the `RequiredIndicator` is no longer bold.
It doesn't make sense to approve the build. It is preferable to keep the indicator bold.

packages/components/label/src/label.tsx
emmenko: @adnasa ⬆️?

adnasa: My bad. missed this comment.
yeah you made a good point. will address it

adnasa: 9ae1868

emmenko (approved):
💯

PR #1772, opened 2021-02-09
refactor(components/messages): migrate to typescript

adnasa:
#### Summary

- migrates `components/messages` to typescript


packages/components/messages/src/warning-message/warning-message.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type TIntlMessageDescriptor = {
+  id: string | number;
+  description?: string | object;
+  defaultMessage: string;
+};
+
+type TWarningMessageProps = {
+  intlMessage?: TIntlMessageDescriptor;
+  children?: React.ReactNode;
```
tdeekens, line 12: A `React.ReactNode` can be `null` or `undefined`. As a result this can be `children: React.ReactNode` if I am not mistaken.

- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L237

adnasa: Yes, ReactNode can be `undefined`, but I recall getting type errors when not specifying `children?` because this is encapsulated into a custom specified type `TWarningMessageProps`

tdeekens: Gotcha.

tdeekens (approved):

packages/components/messages/src/error-message/error-message.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+type TErrorMessageProps = {
+  intlMessage?: TIntlMessageDescriptor;
+  children?: React.ReactNode;
+};
+
+const ErrorMessage = (props: TErrorMessageProps) => (
+  // intlMessage is required is children is missing and vice versa
+  // we let Text.Detail set the warning if content is missing.
```
emmenko, line 18: We can remove this comment

packages/components/messages/src/warning-message/warning-message.tsx
emmenko: Yes, the "optional" prop is not the same as the value being `undefined`.

image

packages/components/messages/src/warning-message/warning-message.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const WarningMessage = (props: TWarningMessageProps) => (
+  // intlMessage is required is children is missing and vice versa
+  // we let Text.Detail set the warning if content is missing.
```
emmenko, line 17: Remove as well

emmenko (approved):
💯

PR #1774, opened 2021-02-09
feat(icons): add new list with search icon

stephsprinkle:
#### Summary

- Added ListWithSearch icon

image

## Description

UX has requested a new icon be added to the icon set. This icon will be used as the menu icon for Audit log (previously known as Change history).


emmenko:
Hmm not sure why but the VRTs somehow don't render/show the new icon 🤔

packages/components/icons/src/svg/list-with-search.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="32px" height="32px" viewBox="0 0 32 32" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Component / icon / 24px / audit-log</title>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="New-Audit-log-icon|-magnifyin-glass" transform="translate(-18.000000, -442.000000)" fill="#000000">
+            <g id="Component-/-icon-/-24px-/-audit-log" transform="translate(18.000000, 442.000000)">
+                <g id="audit-log-icon" transform="translate(4.000000, 4.000000)">
```
emmenko, line 7: @commercetools/ux-ui-design-team please check that the IDs of what you export use the correct name. Thanks.

mariabarrena: @arianavitor-ct I guess this is yours.

For a faster answer, please always approach/assign PRs to the dedicated UX/UI designer part of the vertical team owning the PR and not the entire UX/UI design team, as that's the person who will be able to help. Thanks.

emmenko: My message is directed to all designers, because this is not the first time this happens. No matter who is creating the icon, you should be aware of this as a team.

emmenko (approved):

packages/components/icons/src/svg/list-with-search.svg
mariabarrena: In that case, please document this in a place that won't be gone after the PR is closed. Thank you

stephsprinkle: Apologies, I asked Ariana if we could change the name of the icon to be less Audit log specific and did not realize that the ID should be updated as well.

PR #1775, opened 2021-02-10
fix(text): use internal invariant, do not throw on production

adnasa:


packages/utils/src/invariant.ts
```diff
@@ -0,0 +1,22 @@
...
+// Strip out error messages for production
+// > Not providing an inline default argument for message as the result is smaller
+const invariant = (condition: any, message?: string): asserts condition => {
+  if (condition) {
+    return;
+  }
+  // GIVEN product, supress the message.
+  if (isProduction) {
+    return;
+  }
```
adnasa, line 15: the original invariant throws on production
https://github.com/alexreardon/tiny-invariant/blob/master/src/tiny-invariant.ts#L18

Here we skip it.

packages/utils/src/invariant.ts
```diff
@@ -0,0 +1,22 @@
+const isProduction: boolean = process.env.NODE_ENV === 'production';
+// @TODO: allow consumer to set the prefix e.g `[ Text.Body ]`
+const prefix: string = 'Invariant failed';
+
+// Throw an error if the condition fails
+// Strip out error messages for production
+// > Not providing an inline default argument for message as the result is smaller
+const invariant = (condition: any, message?: string): asserts condition => {
```
emmenko, line 8: Ah nice, using `asserts`

packages/utils/src/invariant.ts
emmenko, line 15: Maybe we should log this instead of just ignoring it?

.changeset/rich-pandas-collect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/text': patch
+'@commercetools-uikit/utils': patch
```
emmenko, line 3: Shouldn't we migrate all packages to use this? 🤔

tdeekens: Can also be a separate step and PR once this is in.

packages/components/text/src/text.spec.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
-import invariant from 'tiny-invariant';
+import { invariant } from '@commercetools-uikit/utils';
 import { screen, render } from '../../../../test/test-utils';
 import Text from './text';
 
-jest.mock('tiny-invariant');
+jest.mock('@commercetools-uikit/utils', () => ({
+  ...jest.requireActual('@commercetools-uikit/utils'),
+  invariant: jest.fn(),
```
tdeekens, line 8: Now or later: the `invariant` could have a `__mock__`. Then you could just to `jest.mock()` maybe.

packages/utils/src/invariant.ts
tdeekens: Let's maybe ignore it for now and think about it. We could report to the console or sentry somehow.

tdeekens (approved):

.changeset/rich-pandas-collect.md
emmenko: Why migrate only the `Text` component then?

packages/utils/src/invariant.ts
emmenko: Why do you want to ignore it? I'm not sure I understand the intent here...

.changeset/rich-pandas-collect.md
adnasa: yes.
Let's just focus on `Text` test and `invariant`, then we have address the basics before migrating everything :)

packages/utils/src/invariant.ts
adnasa: > I think it's okay to first silence the error to fix the actual issue.

agree. let's get this out first, then we can refine it.

emmenko: I see, I was missing the context.

packages/components/text/src/text.spec.js
adnasa: I'll leave this as homework

adnasa:
Okay.
this is again ready for review.
once this is approved, we merge it, but not **release** it until we migrate the other packages (separate PR).
I'll start on that now

tdeekens (approved):

emmenko (approved):

PR #1777, opened 2021-02-10
refactor(packages): use ui-kit/invariant

adnasa:
#### Summary

Migrates all packages into `ui-kit/invariant`


tdeekens (approved):
Looks good. It shall be called `graceful-invariant` 😄

emmenko (approved):

PR #1786, opened 2021-02-12
refactor(accessible-button): to typescript

adnasa:
#### Summary

- migrates `accessible-button` to typescript.
- This is needed for the input-utils


emmenko:
I'll take over this

packages/utils/src/invariant.ts
```diff
@@ -1,11 +1,19 @@
...
+
 const isProduction: boolean = process.env.NODE_ENV === 'production';
 // @TODO: allow consumer to set the prefix e.g `[ Text.Body ]`
 const prefix: string = 'Invariant failed';
 
 // Throw an error if the condition fails
 // Strip out error messages for production
 // > Not providing an inline default argument for message as the result is smaller
-const invariant = (condition: unknown, message?: string): asserts condition => {
+const invariant: TInvariantFunction = (condition, message) => {
```
tdeekens, line 16: Isn't this renamed to warning by now?

tdeekens (commented):
😍

packages/utils/src/invariant.ts
emmenko: Ah yes, I've rebased the branch now.

emmenko:
FYI: notice that the "README docs" are in a `docs_disabled` folder. The reason is that the current generator setup (react-docgen) throws an error because of the usage of `forwardRef`, which the `setParamsTypeDefinitionFromFunctionType` custom handler does not support apparently.

https://github.com/commercetools/ui-kit/blob/96c0718a4c5b06ef6f3b70b78ab16fc30186fefa/generators/readme/src/index.ts#L325

I think we should try looking into another library that has better docgen support for typescript, for example https://github.com/styleguidist/react-docgen-typescript

tdeekens (approved):

adnasa:
> I think we should try looking into another library that has better docgen support for typescript, for example https://github.com/styleguidist/react-docgen-typescript


okay, good call and thanks for looking into it
I will merge this PR and follow up on the docgen.

emmenko:
Alright, so the real error is this:

image

I'll push the fix.

emmenko:
So this is tricky. For some reason the babel plugin transforms `React.ElementType` to `PropTypes.node`, instead of `PropTypes.elementType` which is what I would expect.

image

The prop type warning is caused by passing an element type like `as={Link}`, which is not a `node`.

PR #1787, opened 2021-02-12
rename invariant to warning

adnasa:
#### Summary

- renames `invariant` tool to `warning`
- update imports and usage

### DIFF
--- a/packages/utils/src/warning.ts
+++ b/packages/utils/src/warning.ts
@@ -1,16 +1,16 @@
 const isProduction: boolean = process.env.NODE_ENV === 'production';
 // @TODO: allow consumer to set the prefix e.g `[ Text.Body ]`
-const prefix: string = 'Invariant failed';
+const prefix: string = 'Warning';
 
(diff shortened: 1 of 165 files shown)


packages/utils/src/warning.ts
```diff
@@ -1,16 +1,16 @@
...
+const warning = (condition: unknown, message?: string): asserts condition => {
+  if (isProduction) {
     return;
   }
-  // GIVEN product, suppress the message.
-  if (isProduction) {
+
+  if (condition) {
     return;
   }
```
mohib0306, line 15: can we put these into single OR statement?

mohib0306 (approved):
Looks good

packages/utils/src/warning.ts
adnasa: I was tempted to do it, but I like how this is structured actually #oldschool  😅

tdeekens (approved):
💯

PR #1794, opened 2021-02-22
pagination revert prop naming

adnasa:
#### Summary

**NB**

- rename `pageSize -> perPage`
- rename `currentPage -> page` (prefixing has no value)

## Description

Refer to FE chapter discussion thread.

### DIFF
--- a/.changeset/neat-experts-admire.md
+++ b/.changeset/neat-experts-admire.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-uikit/pagination': minor
+'@commercetools-uikit/pagination': major
 ---
 
 rename pageSize -> perPage, currentPage -> page
--- a/.changeset/neat-experts-admire.md
+++ b/.changeset/neat-experts-admire.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/pagination': minor
+---
+
+rename pageSize -> perPage, currentPage -> page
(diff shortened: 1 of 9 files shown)


packages/components/pagination/src/page-size-selector/page-size-selector.js
```diff
@@ -8,8 +8,8 @@ import { warning } from '@commercetools-uikit/utils';
 import Label from '@commercetools-uikit/label';
```
adnasa, line 8: We keep this component name. only prop names has changed

packages/components/pagination/src/page-size-selector/page-size-selector.js
```diff
@@ -59,11 +58,11 @@ const PageSizeSelector = (props) => {
         />
       </Constraints.Horizontal>
       <Label
-        htmlFor={pageSizeSelectorId}
+        htmlFor={perPageSelectorId}
         intlMessage={{
           ...messages.pageSize,
```
adnasa, line 63: this is where we better keep the `pageSize` name.

.changeset/neat-experts-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/pagination': minor
```
tdeekens, line 2: Major or minor. I don't mind too much.

tdeekens (approved):
Thanks bud!

mohib0306 (approved):
These changes looks good to me. One thing that I notice is that the underlying component is still called as `PageSizeSelector`. Shouldn't that be `PageRangeSelector`?

.changeset/neat-experts-admire.md
adnasa: good call. I'd consider major on hindsight

adnasa:
> Shouldn't that be PageRangeSelector?

I already addressed this in a comment above

adnasa:
LOL. percy breaks, but it is minor because of the prop rename

PR #1804, opened 2021-03-01
refactor(packages/card): migrate Card to typescript

mohib0306:
#### Summary
- `Card` => migrate component to typescript


packages/components/card/src/card.tsx
```diff
@@ -1,10 +1,29 @@
-import PropTypes from 'prop-types';
-import React from 'react';
+// import PropTypes from 'prop-types';
```
adnasa, line 1: what about this?

mohib0306:
I see that in some ts-migrated components we have added readme block in the `package.json` i.e. https://github.com/commercetools/ui-kit/blob/9a66387ab6cc5209b98cb52c85945b0d8808e03b/packages/components/constraints/package.json#L37-39. Is that also required? What is it actually?

adnasa:
> Is that also required?

no I would not do that yet.

---

I think it is a pointer for the important when generating the docs.

adnasa (approved):

packages/components/card/src/card.tsx
```diff
@@ -1,10 +1,28 @@
...
+   */
+  theme: 'light' | 'dark';
+  /**
+   * Pass a custom CSS class, useful to override the styles.
+   * <br>
+   * NOTE: This is not recommended and should only be used for building new components
+   * that require special style adjustments.
+   */
+  className?: string;
+  children?: ReactNode;
```
mohib0306, line 22: These two props are optional, as they were in the original PropTypes definition.

packages/components/card/src/card.tsx
```diff
@@ -1,10 +1,28 @@
...
-const Card = (props) => (
+type TCardProps = {
+  /**
+   * Determines the visual effect of the card. A raised card has a box shadow while a flat card has just a border.
+   */
+  type: 'raised' | 'flat';
+  /**
+   * Determines the background color of the card.
+   */
+  theme: 'light' | 'dark';
```
mohib0306, line 14: No need to mark these as optional, as they have default values defined

tdeekens (approved):
Nice. Thanks for picking this up!

PR #1806, opened 2021-03-02
feat: add manypkg for mono-repo management

tdeekens:
#### Summary

This is similar to app-kit and internally. It will help us to manage so many packages and eventually we can replace lerna.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,6 @@
 {
   "extends": [
     "config:base",
-    "pinOnlyDevDependencies",
     ":preserveSemverRanges",
     "schedule:weekly"
   ],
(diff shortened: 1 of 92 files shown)


.github/renovate.json
```diff
@@ -1,5 +1,10 @@
 {
-  "extends": ["config:base", ":preserveSemverRanges", "schedule:weekly"],
+  "extends": [
+    "config:base",
+    "pinOnlyDevDependencies",
```
emmenko, line 4: no need for this

emmenko (commented):
👍

.github/renovate.json
tdeekens: Just wanted to keep things in sync. I don't mind. I can remove it.

emmenko: Exactly, our OSS repos don't use this anymore 😉
Thanks!

tdeekens:
Taking the 👍🏻 as an approval. Also as this type of change has been approved across various other repositories.

PR #1809, opened 2021-03-02
refactor: to remove lerna and only use manypkg

tdeekens:
#### Summary

Similar to the work on app-kit we remove the tooling infrastructure and homogenise it. We now only use `manypkg` as lerna is much more and not needed to do the few things we need here also.


package.json
```diff
@@ -15,18 +15,18 @@
   "scripts": {
     "postinstall": "manypkg check && preconstruct dev && yarn --cwd docs",
```
tdeekens, line 16: We need to `cwd` the docs as it's not a workspace.

package.json
```diff
@@ -15,18 +15,18 @@
...
     "auth": "npm_config_registry=https://registry.npmjs.org npm whoami",
-    "clean": "lerna exec 'rm -rf dist'",
+    "clean": "manypkg exec 'rm -rf dist'",
     "generate-readmes": "yarn generate-readme --all-workspace-packages",
-    "generate-icons": "lerna run generate-icons && prettier --write '**/generated/*.tsx'",
-    "design-tokens:build": "yarn --cwd design-system build:tokens",
-    "design-tokens:build:watch": "yarn --cwd design-system build:tokens:watch",
+    "generate-icons": "manypkg run icons generate-icons && manypkg run checkbox-input generate-icons && manypkg run radio-input generate-icons && manypkg run @commercetools-uikit/rich-text-input generate-icons && prettier --write '**/generated/*.tsx'",
```
tdeekens, line 20: Same as app-kit. We're explicit here and have to. There is an issue to manypkg opened which would allow globbing or running a run script based on its existence.

packages/components/icons/src/generated/AngleDownReact.tsx
```diff
@@ -70,14 +79,44 @@ const getIconStyles = (props: Props, theme: Theme) => css`
...
+            />
+          </g>
+        </g>
+      </g>
+    </g>
+  </svg>
+);
 
-SvgAngleDown.displayName = "SvgAngleDown";
+SvgAngleDown.displayName = 'SvgAngleDown';
```
tdeekens: With the lerna setup I don't think prettier ever successfully ran on these.

adnasa (approved):

adnasa:
why is the main workflow canceled?

tdeekens:
Will have to investigate why this runs into timeouts with the snapshots later this week.

PR #1811, opened 2021-03-04
feat: to add a `primary` and `inverted` tone to the `link` component

Rombelirk:
#### Summary

This PR adds `isWhite` prop to the `Link` component

## Description
Right now we can use the Link component only in one color which in some cases doesn't fit to background color: 
link
This PR adds possibility to make the link white which is supposed to cover all the cases when primary color is not visible enough.

### DIFF
--- a/packages/components/link/src/link.js
+++ b/packages/components/link/src/link.js
@@ -7,6 +7,21 @@ import { FormattedMessage } from 'react-intl';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 import { filterInvalidAttributes } from '@commercetools-uikit/utils';
 
+const getColorValue = (tone, overwrittenVars) => {
+  if (tone === 'primary') {
+    return overwrittenVars.colorPrimary;
+  }
+
+  return overwrittenVars.fontColorForTextWhenInverted;
+};
+const getActiveColorValue = (tone, overwrittenVars) => {
+  if (tone === 'primary') {
+    return overwrittenVars.colorPrimary25;
+  }
+
+  return overwrittenVars.fontColorForTextWhenInverted;
+};
+
 const getLinkStyles = (props, theme) => {
   const overwrittenVars = {
     ...vars,
@@ -15,14 +30,14 @@ const getLinkStyles = (props, theme) => {
 
   return css`
     font-family: inherit;
-    color: ${overwrittenVars.colorPrimary};
+    color: ${getColorValue(props.tone, overwrittenVars)};
     font-size: ${overwrittenVars.fontSizeDefault};
     text-decoration: ${props.hasUnderline ? 'underline' : 'none'};
 
     &:hover,
     &:focus,
     &:active {
-      color: ${overwrittenVars.colorPrimary25};
+      color: ${getActiveColorValue(props.tone, overwrittenVars)};
     }
   `;
 };
@@ -66,6 +81,7 @@ const Link = (props) => {
 Link.displayName = 'Link';
 
 Link.propTypes = {
+  tone: PropTypes.oneOf(['primary', 'inverted']),
   hasUnderline: PropTypes.bool.isRequired,
   isExternal: PropTypes.bool.isRequired,
   to: requiredIf(
@@ -92,6 +108,7 @@ Link.propTypes = {
 };
 
 Link.defaultProps = {
+  tone: 'primary',
   hasUnderline: true,
   isExternal: false,
 };
--- a/packages/components/link/src/link.spec.js
+++ b/packages/components/link/src/link.spec.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { customProperties as vars } from '@commercetools-uikit/design-system';
 import { screen, render } from '../../../../test/test-utils';
 import Link from './link';
 
--- a/packages/components/link/src/link.spec.js
+++ b/packages/components/link/src/link.spec.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { customProperties as vars } from '@commercetools-uikit/design-system';
 import { screen, render } from '../../../../test/test-utils';
 import Link from './link';
 
--- a/packages/components/link/src/link.visualroute.js
+++ b/packages/components/link/src/link.visualroute.js
@@ -35,5 +35,8 @@ export const component = () => (
     <Spec label="intlMessage">
       <Link to="/" intlMessage={intlMessage} />
     </Spec>
+    <Spec label="white">
+      <Link to="/" tone={'inverted'} />
+    </Spec>
   </Suite>
 );
--- a/packages/components/link/src/link.visualroute.js
+++ b/packages/components/link/src/link.visualroute.js
@@ -12,7 +12,7 @@ const purpleTheme = {
   colorPrimary25: 'deeppurple',
 };
 
-export const component = () => (
+export const component = ({ themes }) => (
   <Suite>
     <Spec label="regular">
       <Link to="/">A label text</Link>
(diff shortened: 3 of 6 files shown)


tdeekens:
Have you checked with ui-kit in general? I think boolean props are a bit of an anti-pattern. I think we often all this inverted here. So it could be `tone="inverted"` maybe to be inline with what we have?

https://github.com/commercetools/ui-kit/search?q=inverted

tdeekens (approved):

tdeekens:
We're somewhat blocked here by non-working VRTs. I am looking into that but can't promise to find a fix.

packages/components/link/src/link.spec.js
```diff
@@ -48,4 +49,30 @@ describe('rendering', () => {
       expect(link).toBeInTheDocument();
     });
   });
+  describe('when rendering link with a specific color', () => {
```
emmenko, line 52: Testing css styles is not necessary here, since we should have VRTs.

packages/components/link/src/link.js
```diff
@@ -8,22 +8,49 @@ import { FormattedMessage } from 'react-intl';
...
+export const tones = {
+  primary: 'primary',
+  inverted: 'inverted',
+};
+const linkStates = {
+  normal: 'normal',
+  hover: 'hover',
+};
+
+export const whiteColor = '#FFFFFF';
```
emmenko, line 20: Use the color tokens from the design system please.

packages/components/link/src/link.js
```diff
@@ -8,22 +8,49 @@ import { FormattedMessage } from 'react-intl';
...
 import { filterInvalidAttributes } from '@commercetools-uikit/utils';
 
+export const tones = {
+  primary: 'primary',
+  inverted: 'inverted',
+};
+const linkStates = {
+  normal: 'normal',
+  hover: 'hover',
+};
```
emmenko, line 18: This won't really work in TS (once we migrate this component to it) as there is no need for map/constants.
I would keep it simple and use string values.

packages/components/link/src/link.js
```diff
@@ -8,22 +8,49 @@ import { FormattedMessage } from 'react-intl';
...
+        [linkStates.hover]: overwrittenVars.colorPrimary25,
+      },
+
+      [tones.inverted]: {
+        [linkStates.normal]: whiteColor,
+        [linkStates.hover]: whiteColor,
+      },
+    };
+    return colorMap[tone][state];
+  };
```
emmenko, line 42: Hmm this looks overly complicated. We can also have two functions `getColorValue` and `getActiveColorValue`.

emmenko (changes requested):
Thanks for contributing.

Besides the comments I left, has having a tone for links been approved by UX? Just making sure we're all on the same page.

Thanks

tdeekens:
Managed to unblock CI and percy @Rombelirk. A rebase over `main` (not the branch here is not `master`) should fix this now.

Rombelirk:
> Thanks for contributing.
> 
> Besides the comments I left, has having a tone for links been approved by UX? Just making sure we're all on the same page.
> 
> Thanks

Yes, this is agreed by the UX team.

YahiaElTai (approved):
Looks good. Thanks 🤩

packages/components/link/src/link.visualroute.js
```diff
@@ -35,5 +35,8 @@ export const component = () => (
     <Spec label="intlMessage">
       <Link to="/" intlMessage={intlMessage} />
     </Spec>
+    <Spec label="white">
+      <Link to="/" tone={'inverted'} />
```
adnasa, line 39: does this have to be `{''}`? why not use double quotes directly?

tdeekens:
@Rombelirk please address the feedback from @emmenko when you find the time. I think it's valuable :)

packages/components/link/src/link.js
```diff
@@ -14,16 +14,27 @@ const getLinkStyles = (props, theme) => {
...
   };
 
+  const getColorValue = (tone) => {
+    if (tone === 'inverted')
+      return overwrittenVars.fontColorForTextWhenInverted;
+    return overwrittenVars.colorPrimary;
+  };
+  const getActiveColorValue = (tone) => {
+    if (tone === 'inverted')
+      return overwrittenVars.fontColorForTextWhenInverted;
```
emmenko, line 24: Is it correct to use the same color for the hover/active state?

emmenko (commented):
Thanks, much better now!

packages/components/link/src/link.js
Rombelirk: This is agreed with UX as well.

Rombelirk:
> @Rombelirk please address the feedback from @emmenko when you find the time. I think it's valuable :)
> 
> CleanShot 2021-03-05 at 14 15 25@2x
> 
> Also the snapshot you took needs to be inverted - as in the background. You're capturing white on white which is an always passing test and a worthless snapshot :). I think we have a way to invert the background colour too.

Yeah, that what I was worrying about. Please remind where I can view those snapshots.

packages/components/link/src/link.visualroute.js
Rombelirk: this is converted to single quotes by prettier

tdeekens:
> > @Rombelirk please address the feedback from @emmenko when you find the time. I think it's valuable :)
> > CleanShot 2021-03-05 at 14 15 25@2x
> > Also the snapshot you took needs to be inverted - as in the background. You're capturing white on white which is an always passing test and a worthless snapshot :). I think we have a way to invert the background colour too.
> 
> Yeah, that what I was worrying about. Please remind where I can view those snapshots.

[These are](https://github.com/commercetools/ui-kit/blob/c617983f8144f5f3e5e53c77864621968b77dabd/test/percy/spec.js) the components used to render the specs and suite. I am not 100% sure if the can have an inverted background. Did you check some `visual.spec.js` files for reference? Maybe we can add something if it's needed cause without the snapshot is pointless.

Rombelirk:
> > > @Rombelirk please address the feedback from @emmenko when you find the time. I think it's valuable :)
> > > CleanShot 2021-03-05 at 14 15 25@2x
> > > Also the snapshot you took needs to be inverted - as in the background. You're capturing white on white which is an always passing test and a worthless snapshot :). I think we have a way to invert the background colour too.
> > 
> > 
> > Yeah, that what I was worrying about. Please remind where I can view those snapshots.
> 
> [These are](https://github.com/commercetools/ui-kit/blob/c617983f8144f5f3e5e53c77864621968b77dabd/test/percy/spec.js) the components used to render the specs and suite. I am not 100% sure if the can have an inverted background. Did you check some `visual.spec.js` files for reference? Maybe we can add something if it's needed cause without the snapshot is pointless.

Yeah, I found this component , but how can I actually view the spec rendered locally? is this even possible?

emmenko:
You have to use the ThemeProvider to render a dark theme. Check out the VRTs for the `Text` component.

tdeekens:
@Rombelirk the last commit adds the theme provider around the spec. It was already used in the same file and in similar fashion in other visual specs.

packages/components/link/src/link.visualroute.js
```diff
@@ -35,8 +35,17 @@ export const component = () => (
     <Spec label="intlMessage">
       <Link to="/" intlMessage={intlMessage} />
     </Spec>
-    <Spec label="white">
-      <Link to="/" tone={'inverted'} />
-    </Spec>
+    <ThemeProvider theme={themes.darkTheme}>
+      <Spec label="tone - inverted - with underline">
```
tdeekens, line 39: Underlines can be achieved in different ways. E.g. with border or natively to the font. As a result I think it's beneficial to test both in a VRT for an inverted option.

tdeekens (approved):

packages/components/link/src/link.visualroute.js
```diff
@@ -35,5 +35,17 @@ export const component = () => (
     <Spec label="intlMessage">
       <Link to="/" intlMessage={intlMessage} />
     </Spec>
+    <ThemeProvider theme={themes.darkTheme}>
+      <Spec label="tone - inverted - with underline">
+        <Link to="/" tone="inverted">
+          An inverted label text
```
tdeekens, line 41: @Rombelirk you rendered no text inside the `Link` prior. I am not sure how/what you wanted to test here in terms of font colour inversion if there is no text ;)

tdeekens (approved):

packages/components/link/src/link.visualroute.js
Rombelirk: 🤦‍♂️

tdeekens:
Approved the VRTs as they look good. The respective designer can also check them :)

emmenko (approved):
Awesome, thanks again for working on this!

emmenko:
Btw, next time please try to open an issue first with whatever feature request or bug you want to work on. It helps us to follow our contribution process and to discuss potential issues beforehand.

PR #1813, opened 2021-03-04
fix: to adjust executable path via percy v2

tdeekens:


.github/workflows/main.yml
```diff
@@ -56,13 +56,12 @@ jobs:
         run: yarn visual-testing-app:build
 
       - name: Running Visual Regression Tests for UI components
-        uses: ianwalter/puppeteer-container@v4.0.0
+        run: yarn vrt:components
```
tdeekens, line 59: This docker container isn't needed.

tdeekens (commented):
Merging to unblock CI.

PR #1814, opened 2021-03-05
feat(Card): make inset spacing configurable

mohib0306:


tdeekens:
Do we have `padding` props elsewhere 🤔. Just thinking out loud as I think we should really build a design language which is consistent. 

I don't think we should mix layout and visual components. To we it's two different things. That's my we have a `Stack` for instance. So we don't have a component accept a `direction` prop or so.

Could we achieve the same by composition?

```jsx
<Card>
	<Spacings.Inset scale="m">{/* Container with padding 🎉 */}</Spacings.Inset>
</Card>
```

mohib0306:
Thanks for pointing it out. I totally missed the `Inset` component even though I looked at it a few days ago. We can wrap the children (`props.children`) of the card in an `Inset` component, however we still a way that the users could specify the how much should it padded from the border. That's why `padding` prop is added to the `card` component and it should expose it to the user

mohib0306:
The idea from UI/UX team was that the card component will itself now has the padding of `18px` (fixed value) and sometimes it is not needed or it has to be less. Therefore they wanted to not have this fixed behaviour but rather a flexible. i.e. by default it will `18px` but you can also have as `9px` by using/specifying this `padding` prop. Doing it the following way 
```
<Card>
	<Spacings.Inset scale="m">{/* Container with padding 🎉 */}</Spacings.Inset>
</Card>
```
will even more increase the padding. That will need to remove the current padding from the card component at all and might make it unusable since you to wrap the children of the card component by yourself.

Do you follow what I am pointing at? 🙂

tdeekens:
I think the UX team should also start to think in terms of composition using the building blocks of components and not treat them as singular entities mentally.

If the card component currently has a padding then it rather should not. At least from the view point of a "pure" design system composition where visual components have nothing to do with layout components and both are not mixed. 

The already existing custom padding to me (personally) is an anti-pattern to the above and hence could be removed. From there on-top of my head I see two appraoches

1. Consumers have to wrap the "actual" visual children with a layout component of their choice

```jsx
<Card>
	<Spacings.Inset scale="m">{/* Container with padding 🎉 */}</Spacings.Inset>
</Card>
```

2. Consumers have to specify the `scale`

```jsx
<Card scale="m">
	<Spacings.Inset scale={props.scale}>{/* Container with padding 🎉 */}</Spacings.Inset>
</Card>
```

The 2. encapsulates things while 1. gives more freedom. It's all a matter how things are meant to be applied and how much freedom the design system should give.

Again, I think having a `padding` prop or native padding is wrong.

mohib0306:
Thanks for the comment. I think you have raised a good point. Removing the native padding from `Card` and then letting consumer use `Inset` (or any other container) with whatever padding they want, could be more clearer but might create a bit of work in the MC (migrating current usages). I will ask the UX people to look into this and lets see what they think.

tdeekens:
> Thanks for the comment. I think you have raised a good point. Removing the native padding from `Card` and then letting consumer use `Inset` (or any other container) with whatever padding they want, could be more clearer but might create a bit of work in the MC (migrating current usages). I will ask the UX people to look into this and lets see what they think.

I think my second suggestion was not clear. It would remove a large scale refactoring.

```jsx
const Card = props => {
    <Spacings.Inset scale={props.scale}>
        {props.children}
    </Spacings.Inset>
}

Card.defaultProps = {
    scale: "m"
}
```

Would mean the padding can go away while we use our layout components and default a scale to have padding.

adnasa (dismissed):
nice one 💯

adnasa:
sorry I jumped quickly into reviewing 😅 
reading the comments above it makes more sense to wrap the component with a spacings component.
if they differ in values, this needs to be escalated.

packages/components/card/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools-uikit/card",
-  "description": "Cards are used to display content and actions on a single topic.",
+  "description": "Cards are used to display content and actions on a single topic. Only use padding 's' (or 8px) as container for single fields. When in doubt, use padding 'm' (or 16px)",
```
emmenko, line 3: Let's keep the description here short. If you need to provide more info, use a `docs/description.md` file.

packages/components/card/src/card.tsx
```diff
@@ -8,6 +8,10 @@ type TCardProps = {
    * Determines the visual effect of the card. A raised card has a box shadow while a flat card has just a border.
    */
   type: 'raised' | 'flat';
+  /**
+   * Determinse the padding that the conent should have from the card borders
+   */
+  padding: 's' | 'm';
```
emmenko, line 14: I wouldn't name this `padding` for a number of reasons.
I think something like `insetSpace` might be more appropriate, as it aligns more with the spacing component.

emmenko (changes requested):
Thanks for contributing!

packages/components/card/src/card.tsx
tdeekens: Yup. That's what I suggested above too. `scale` or `insetScale`.

emmenko:
Be aware that removing the `padding` css style and replacing it with the spacing component will likely cause breaking changes, as we do overwrite the styles in different places/repos.

If we want to change the internals we need to first investigate what the impact would be.
If you are thinking "consumers should not rely on internal", that's half true in this case because by design we made it possible for consumers to override some styles.

mohib0306:
It did a bit of searching and finding about the usage of `Card` component across some of the repos.

#### In the MC-FE `<Card>` has 56 usages in 47 files

1. Mostly without custom style or className => `<Card [theme or type or key]>/** Some other component*/</Card>`
2. 7 Usages in 6 files with className prop specified 
    1. `application-account => user-profile.js`: background color specified in the custom css
    2. `application-authentication => login.js`: one usage => padding specified in the custom css + another usage => width and radius props specified
    3. `application-fallback => covid19-message.js`: padding specifed in the custom css
    4. `application-orders => parcel-detail-panel.js`: width and margin specified in the custom css
    5. `application-products => pim-search-list-input => more-info.js`: background color specified in the custom css
    6. `application-fallback => user-research-program-card.js`: padding and margin specified in the custom css

p.s. The is two outstanding PRs which also uses `<Card` instead of `FormBox` and also wraps `Card` component in a `Inset` component. i.e. `<Spacings.Inset><Card>/** children */</Card></Spacings.Inset>`

#### In the App-Kit `<Card>` has 6 usages in 6 files

1. `application-components => dialog-container.tsx` => with custom css specifying min-height, height and some styles on child `div` element
2. Remaining 5 usages are of type `<Card [theme or type]>/** Some other component*/</Card>`

#### In the UI-Kit `<Card>` has 1 usage in 1 file
`data-table-manager => settings-container.js` => usual `<Card [theme or type]>/** Some other component*/</Card>` usage

#### In the audit-log `<Card>` has 1 usage in 1 file
`change-history-components => marketing-overlay.js` =>  usual `<Card [theme or type]>/** Some other component*/</Card>` usage

Of-course, wrapping the `Card.children` in `Inset` component might result that some component do not look as they should. I think we should update these affected components in the same PR when we are updating the uikit repo itself and then ask the teams to review the changes.

As the uikit is open-source and it will also affect external users.

tdeekens:
Thanks for that detailled analysis @mohib0306!

emmenko:
Thanks for the analysis.

Personally, I would not introduce this change here but instead we can include it in the next major, just to be on the safe side.

tdeekens:
> Thanks for the analysis.
> 
> Personally, I would not introduce this change here but instead we can include it in the next major, just to be on the safe side.

I dare to disagree. We have no defined guideline to my knowledge if a visual change is considered a breaking change. It is a discussion one can have but we have often taken the different approach in accepting visual changes as the essential desired path forward. With more adoption of ui-kit I might change my mind again. 

In that by understand is that we remove how we achieve padding to be better composed using our primitives. That this changes the actual padding can then we considered as intended. I am not sure why we have "to be on the safe side" nor for whom. We have done incredible inventory of how we use the component internally. We have no data if anybody uses it externally but "pretend" as if half of the world (forgive the irony) does. I don't think we have to be that strict here. 

I can agree if we change the behaviour of the className prop. However also here we have no idea on if anybody uses it. 

I don't want to argue for a loose SemVer application. I just urge us to have a more pragmatic, moving mindset over always assuming we have build the most widely used component library which just slows us down and complicates process on a library we struggle to continuously maintain and somebody wants to make a worthwhile contribution to. 

Happy to be proven wrong or talk about it. However a "let's be on the safe side" for a (what I understand) visual change is exaggerated to me here if we migrate carefully internal.

emmenko:
> Happy to be proven wrong or talk about it

There you go: https://github.com/commercetools/commercetools-docs-kit/blob/898af56a02a30505ad0170e482378eb17c085add/packages/gatsby-theme-docs/src/components/content-pagination.js#L52-L53

I wasn't referring to "external usages" but I know for a fact that (at least I) have been customizing the `Card` component styles in some places.

I don't understand to be honest what all this fuss is about. I'm all for it and I simply suggested to do the proposed refactoring separately since it's not blocking this PR anyway. And we already planned to do a major in the next couple of weeks, so we can include it there.

Anyway, peace and love. No hard feelings.

tdeekens:
I scheduled a meeting for us on Monday to talk it through.

tdeekens (approved):
💯

emmenko (approved):

adnasa (approved):

mariabarrena:
For reference: using the spacing design token (with sizes m and s) for this task was mentioned in the original ticket (https://jira.commercetools.com/browse/CON-535). And the prop name "padding" was only a proposal based on what we were trying to fix. You guys have a more complete overview of the terminology that has been used in props historically so I expected a better option would be suggested :)  

As @mohib0306 is off, I leave here my UX review. I guess this might be a storybook related bug but in any case: the "none" size changes depending on the previous selection of the insetScale. See gif attached. 
Kapture 2021-03-09 at 09 16 49.

In general, why do we need the "none" value? Is it because the previous default did not include any padding at all? If we need it technically speaking and it has to be the default one, can we keep the "Use M as default" sentence in the prop description? We need to make sure externals don't use the "none" option and always use M unless they are trying to emulate the detail pages and want to use them as a substitution of form-boxes (then they should use s).

emmenko:
> In general, why do we need the "none" value?

Regarding the `none` option. We currently have in a bunch of places some custom styles for the `Card` , where we set the `padding`  to `0 `. Therefore, we decided to make this a first-class option and allow people to unset the padding.
In a perfect world, we would document the UX guidelines when and where to use which values but for now we just leave it "undocumented".

> I guess this might be a storybook related bug but in any case: the "none" size changes depending on the previous selection of the insetScale.

I'll take a look, thanks!

mariabarrena:
Ok, the way that the scales are displayed now makes easier to compare for consumers. Thanks for that. Just one small request. Could we use the property name in the overviews too? So that everybody can make the connection between the property described in the readme, the insetScale, with what's being displayed in the examples. This is what I mean (on each example, of course): 
Screenshot 2021-03-10 at 11 43 04

Once that's done I don't need to do another review. You can proceed.

tdeekens (approved):

PR #1773, opened 2021-02-09
refactor(components/loading-spinner): use typescript

adnasa:
#### Summary

- migrate `LoadingSpinner` to typescript

PR #1821, opened 2021-03-12
docs(components): update readmes

adnasa:
#### Summary

- result of generating readmes from #1773


adnasa:
Closing this for now, and regenerate after v12 is released.

Issue #1737, opened 2021-01-14
Upcoming breaking changes (v12)


adnasa:
## Summary

- [x] #1827 Drop Table
`<Table />` component has been deprecated in favor of `<DataTable />`. [reference](https://jira.commercetools.com/browse/UKIT-186)
- [x] #1828 Drop deprecated prop `constraint` on `Constraints.Horizontal` in favor of `max`. [reference](https://wiki.commercetools.com/display/shield/2020/11/17/New+Constraints+Scale)
- [x] #1825 Move RichTextEdit/LocalizedRichTextEditor out of beta [reference](https://jira.commercetools.com/browse/UKIT-194) 
- [x] #1824 Drop deprecated props on `Text` component
The `Text` component has a handful of deprecated props, such as `elementType` and `isInline`
- [x] #1826 Drop `hasUnderline` from `Link` (see #1548)
- [x] #1830 Drop deprecated prop `linkTo` in `SecondayButton`
  - remove `linkTo` from `<Tag />`

tdeekens:
Controversial maybe but we could attempt (or suggest) to product to remove the `RichTextEdit`. I think no customer currently uses it. Less code is good code. Should be verified with the respective team I think.

adnasa:
> Should be verified with the respective team I think.

true. We also have to check if there are plans on using these.

tdeekens:
> > Should be verified with the respective team I think.
> 
> true. We also have to check if there are plans on using these.

Quickly reached out to the team. They're using it with more customers by now. They're also planning to roll it out in general but want to do so with receiving customer feedback in the process. As a result we should keep it. Thanks.

PR #1825, opened 2021-03-17
docs(rich-text): remove beta

adnasa:
#### Summary

- removes `beta` flag from rich text #1737

### DIFF
--- a/.changeset/eleven-meals-remain.md
+++ b/.changeset/eleven-meals-remain.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/localized-rich-text-input': patch
+'@commercetools-uikit/rich-text-input': patch
+---
+
+set rich-text out of beta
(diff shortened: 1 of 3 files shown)


tdeekens (commented):
There is a pending update of the underlying library. Should we get that in before we get it out of beta? 🤔

adnasa:
yeah let's try to update it then

.changeset/eleven-meals-remain.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/localized-rich-text-input': major
+'@commercetools-uikit/rich-text-input': major
```
emmenko, line 3: Not sure we need to bump a major here, technically nothing changed 🤔

adnasa: yeah good thinking. 
we can mark it as `patch` but is part of the v12 release anyway :)

adnasa: https://github.com/commercetools/ui-kit/pull/1825/commits/f5fe8d457275c5d8b80ef8ba9d5ff002319d13d3

emmenko (approved):

PR #1828, opened 2021-03-17
refactor(constraints): remove constraint prop

adnasa:
#### Summary

- remove `constraint` prop
- add warning on helper if consumer component (e.g `DateField`) uses an old value

ref #1737


packages/components/constraints/src/helpers.ts
```diff
@@ -24,7 +25,13 @@ const parseHorizontalConstraintProp = (prop: number | string) => {
     // prop is of type `max`
     return prop;
   }
+
+  // @TODO: remove this dependency
   // prop is of type `constraint`
+  warning(
+    false,
+    'Component is using old constraint values. Please use one of `max` constraint values.'
+  );
```
adnasa, line 34: I added this warning, incase a consumer does the following

```
// should use one of `max` values instead.
<DateField horizontalConstraint="xs" /> 
```

emmenko: We should remove any usage/reference of the old constraints and therefore we should not accept those values anymore.

emmenko (changes requested):

packages/components/constraints/src/helpers.ts
adnasa: I removed it all.
This is a real breaking change now 😄

emmenko (commented):
Let's also remove the constraints from the design tokens: https://github.com/commercetools/ui-kit/blob/6514a11a7a11ff26e4734e93aad4fa358343cb68/design-system/materials/internals/definition.yaml#L100-L121

Also, you might want to regenerate the changeset as many packages have changes here.

adnasa:
@emmenko done

packages/components/inputs/search-select-input/docs/usage-example.js
```diff
@@ -6,7 +6,7 @@ const Example = () => {
     <SearchSelectInput
       id="customers"
       name="customers"
-      horizontalConstraint="m"
+      horizontalConstraint="7"
```
emmenko, line 9: Those should be numbers not strings

emmenko (commented):
Looking good!

emmenko (approved):

tdeekens (approved):

emmenko:
@adnasa can you take a look at the remaining snapshot in percy? There is a change in the text component and I'm not sure if it relates to this PR:

image
image

adnasa:
Approving the percy build

adnasa:
@emmenko 
This is happening in the `ui-kit-v12` branch too

image


because we removed the `isInline` prop.
We actually should remove this test.

adnasa:
We removed the `isInline` prop here
https://github.com/commercetools/ui-kit/pull/1829/commits/07975e30a234f4029491c02d38bd6a9c61056030

here is the build for it
https://percy.io/commercetools-GmbH/ui-kit/builds/9459892/changed/534515259?browser=chrome&browser_ids=16&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1024&widths=1024

and this test, tests the `isInline` prop which is superfluous.
I will remove, push again and approve the build

adnasa:
it's done here ho ho ho 🎅 
https://github.com/commercetools/ui-kit/pull/1839

PR #1830, opened 2021-03-19
refactor(secondary-button): remove `linkTo` from component

adnasa:
#### Summary

- remove `linkTo` prop

reference #1737

PR #1834, opened 2021-03-22
remove `linkTo`

adnasa:
#### Summary

- ref #1830


adnasa:
This has already been approved. but now targeting v12 branch

Issue #1838, opened 2021-03-29
Change PageNavigator page number programmatically


Tormand:
When changing the page number on PageNavigator programmatically, it dosent update the viewed page number 

**To Reproduce**
`<Pagination perPageRange="m" perPage={perPage} totalItems={totalOrders} page={page} onPageChange={pageChange} onPerPageChange={perPageChange} />`

In another function, change 'page' to another value.

adnasa:
Hey @MartinTormandKruso , thank you for your message.

Can you elaborate this?
> In another function, change 'page' to another value.

If we look at the storybook, the pagination reflects the page change when you navigate between pages
https://uikit.commercetools.com/?path=/story/components-pagination--pagination

so I'm not sure how to reproduce your problem otherwise. what does  `pageChange` in `onPageChange={pageChange}` do exactly?

Thanks

MartinTormandKruso:
Hey @adnasa 

Yeah, I can see my initial post is a bit confusing.

Say I have a DataTable with the Pagination and I also have a filter option(i.e. filter orders by date). Then when I'm on page 4 and change the date filter, I will then make a new call for orders filtered by the new date and change `Page` to page 1. But when I do that in the call function it dosen't change the page number that the user sees.

My `pageChange` does the following:
```
const pageChange = (pageNumber) => {
	page = pageNumber
	getAllOrders()
}
```
In `getAllOrders()` I fetch orders with different filters:
```
const getAllOrders = () => {
	asyncDispatch(actions.fetchAllOrders(perPage, page, textSearch, status, shippingMethod, shippingCountry, store, paymentStatus, date))
			.then((result) => {
				dispatch({ type: 'ok', payload: result })
				setTotalOrders(result.total)
				setOrders(orders(result.results)
			})
			.catch((error) => {
				dispatch({ type: 'error', payload: error })
				showApiErrorNotification({ errors: error })
			})
	}
```

Then I have a SelectInput where you can filter by dates(7 days back, 1 month back etc):
```
<SelectInput
	options={dates}
	value={date}
	onChange={(e) => {
		date = e.target.value
		page = 1
		getAllOrders()
	}}
/>
```
When I set the page to 1, it dosent show in the Pagination component, but it does get orders from the correct page.

My page variable is set outside the React function:
```
var perPage = 20
var page = 1

const ViewOrders = (props) => {
	const asyncDispatch = useAsyncDispatch()
	const showApiErrorNotification = useShowApiErrorNotification()
[...]

PR #1840, opened 2021-03-30
refactor(page-navigator): use formik state

adnasa:
#### Summary

- use `formik` state with `enableReinitialize`
- reference #1838 

#### Description

In previous implementation, when the client passes a `page` it is set as the `initial` state of `PageNavigator` and from there leverage an internal state via `useState()`
When the `page` prop updates, `PageNavigator` doesn't reflect the updated changes as the prop is accept in subsequent renders.

#### Proposal

use `formik` to manage the state with `enableReinitialize`, this gives us a couple of advantages.

- previous implementation of focus and blur handle are abstracted away from the pagination. They are now handled with formik with the help of the `validateOnBlur`, `validateOnChange` and `validateOnMount` configs
- via `enableReinitialize`, the internal formik state updates

### DIFF
--- a/.changeset/unlucky-mirrors-argue.md
+++ b/.changeset/unlucky-mirrors-argue.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/pagination': minor
+---
+
+refactor page-navigator to not use internal state
--- a/.changeset/unlucky-mirrors-argue.md
+++ b/.changeset/unlucky-mirrors-argue.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-uikit/pagination': minor
+'@commercetools-uikit/pagination': patch
 ---
 
-refactor page-navigator to not use internal state
+Fix issue with `Pagination` where programmatic changes to the `page` prop are not reflected into the component internal state.
--- a/packages/components/pagination/package.json
+++ b/packages/components/pagination/package.json
@@ -37,6 +37,7 @@
     "@commercetools-uikit/utils": "11.2.1",
     "@emotion/react": "^11.1.1",
     "@emotion/styled": "^11.0.0",
+    "formik": "2.2.6",
     "lodash": "4.17.20",
     "prop-types": "15.7.2"
   },
--- a/packages/components/pagination/package.json
+++ b/packages/components/pagination/package.json
@@ -37,7 +37,7 @@
     "@commercetools-uikit/utils": "11.2.1",
     "@emotion/react": "^11.1.1",
     "@emotion/styled": "^11.0.0",
-    "formik": "2.2.6",
+    "formik": "^2.2.6",
     "lodash": "4.17.20",
     "prop-types": "15.7.2"
   },
--- a/packages/components/pagination/src/page-navigator/page-navigator.js
+++ b/packages/components/pagination/src/page-navigator/page-navigator.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useFormik } from 'formik';
 import { useIntl } from 'react-intl';
 import uniqueId from 'lodash/uniqueId';
 import { AngleLeftIcon, AngleRightIcon } from '@commercetools-uikit/icons';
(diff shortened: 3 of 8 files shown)


packages/components/pagination/package.json
```diff
@@ -37,6 +37,7 @@
     "@commercetools-uikit/utils": "11.2.1",
     "@emotion/react": "^11.1.1",
     "@emotion/styled": "^11.0.0",
+    "formik": "2.2.6",
```
emmenko, line 40: Let's use a version range, to avoid possible version conflicts with formik versions used by custom apps dependencies.

```suggestion
    "formik": "^2.2.6",
```

adnasa:
The tests are failing locally.
I think part of the issue is that `setFieldValue` is an async operation.
will get into that .

.changeset/unlucky-mirrors-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/pagination': minor
+---
+
+refactor page-navigator to not use internal state
```
emmenko, line 5:
```suggestion
---
'@commercetools-uikit/pagination': patch
---

Fix issue with `Pagination` where programmatic changes to the `page` prop are not reflected into the component internal state.
```

packages/components/pagination/src/page-navigator/page-navigator.js
```diff
@@ -14,77 +15,69 @@ import messages from './messages';
...
           label={intl.formatMessage(messages.previousPageLabel)}
-          onClick={handlePrevPage}
+          onClick={handlePrevPageChange}
           isDisabled={isPreviousDisabled || isDisabled}
           icon={<AngleLeftIcon />}
         />
         <Label htmlFor={pageNumberInputId} intlMessage={messages.page} />
         <div>
           <NumberInput
+            name="page"
```
emmenko, line 71:
```suggestion
            name="page-input"
```

packages/components/pagination/src/page-navigator/page-navigator.js
```diff
@@ -14,77 +15,69 @@ import messages from './messages';
...
             max={totalPages}
-            onBlur={onBlurNormalize}
-            onFocus={handleFocus}
-            onChange={handleChange}
+            onBlur={paginationForm.handleBlur}
+            onFocus={paginationForm.handleFocus}
+            onChange={paginationForm.handleChange}
             isDisabled={isDisabled}
-            hasWarning={!isValid(page, props.totalPages)}
+            hasWarning={paginationForm.errors.page}
```
emmenko, line 80: 👌

emmenko (commented):
Nice, I like this solution! 🙌

packages/components/pagination/src/page-navigator/page-navigator.js
adnasa: this name should reflect the name of the formik state ;)

adnasa:
oh yeah, build passed! 😉 
let's `ship it` when you approve, @emmenko

packages/components/pagination/src/pagination.spec.js
```diff
@@ -108,7 +111,7 @@ describe('page navigator interaction', () => {
 });
 
 describe('per page selector interaction', () => {
-  it('should call onPerPageChange with the selected value', async () => {
```
adnasa, line 111: this is the one that caused the `act()` warning.

packages/components/pagination/src/page-navigator/page-navigator.js
emmenko: I know, I just wanted to suggest a more specific name 😉

emmenko (approved):

tdeekens:
I like the idea. Quick question though without wanting to be a party pooper: do we want to "bind" ui-kit to formik. How much work would a barebones solution be?

emmenko:
Yes thanks for the concern. We decided first to go with this to be able to fix the issue. Then we can see how we can refactor/simplify the internal implementation.

There is also something we need to talk/clarify with UX in regards to the number input, as this is the only piece needing the form state. So eventually having a "form state" also depends on the outcome of this discussion.

adnasa:
> How much work would a barebones solution be?

if we take the barebones approach, we would use `useEffect()` as the underlying implementation of `enableReinitialize`.

The implementation is simple enough building it ourselves, but then we have to *build the whole thing ourselves* (validation on mount, validation on change, validation on `AnyEvent`), all of which are encapsulated with `enableReinitialize` config (and the `effect` that comes with it..

---

### Thought process

I can see your concern that binding to `formik` introduces dependency and adds complexity to interoperability. This implementation is still subject to change later on and not set in stone. Let's try this. The idea is proven now that `useEffect` is a valid approach here.

My real concern with `enableReinitialize` is that `set*` callbacks are `async`, but in the case of a leaf it should not be a problem. It would be a bigger problem if this was parent component N level on the render tree.

reference https://github.com/formium/formik/blob/master/packages/formik/src/Formik.tsx#L414:L434

adnasa:
>  do we want to "bind" ui-kit to formik

btw.. I also think this ship has sailed. Some of our inputs relies on very specific shapes derived from `docToForm` transofmers, which kind of leaks in formik shapes. Maybe I misunderstood something.

tdeekens:
Thanks for being open to my concern. 

I agree that we might have leaked too much formik standards into ui-kit already maybe. Not the fact that inputs operate on raw events but should the nature of some things. I was just wondering how far we wanted to go here. I was just scratching my eyebrow when I saw that we don't keep a three input compound component agnostic without a form library. 

We can also revisit this obviously. It's code. It can change. I just also remember we had an issue with "state-mirroring" of the pagination component. Where the input didn't update until we blur. We already agreed it was due to the parallel controlled and uncontrolled (mirrored) behaviour of the component. 

I will have a look at the codesandbox but don't want to block this. 

Thanks again for the effort.

adnasa:
> I was just scratching my eyebrow when I saw that we don't keep a three input compound component agnostic without a form library.

That's a good one. thanks for sharing this!
This can definitely be improved!

let's go for this for now.

PR #1842, opened 2021-03-30
Add pagination and table sorting state

adnasa:
#### Summary

- adds `pagination-state` hook
- adds `data-table-sort-state` hook
to be frank, I'm not sure if this is the correct naming or if this is the repo for it,
because it is more than likely that this hook is used to forward the state to fetchers within the MC. feedback welcome

#### Description

We need to expose these hooks so that other consumers than the Merchant Center can use this interchangeably with the Pagination component (for pagination-state) without having to re-implement it themselves.

The idea is that once this is merged, we migrate the MC to use this.

shall be included in this release https://github.com/commercetools/ui-kit/pull/1841

### DIFF
--- a/.changeset/slow-grapes-hope.md
+++ b/.changeset/slow-grapes-hope.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/hooks': minor
+---
+
+Add pagination-state and data-table-sorting-state.
--- a/.changeset/slow-grapes-hope.md
+++ b/.changeset/slow-grapes-hope.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/hooks': minor
 ---
 
-Add pagination-state and data-table-sorting-state.
+Add `usePaginationState` and `useDataTableSortingState`. These hooks help to manage pagination state and should be used together with the `<Pagination>` and/or `<DataTable>` components.
--- a/packages/hooks/package.json
+++ b/packages/hooks/package.json
@@ -26,6 +26,7 @@
     "@babel/runtime": "7.13.10",
     "@babel/runtime-corejs3": "7.13.10",
     "@commercetools-uikit/utils": "11.2.1",
+    "@testing-library/react": "11.2.5",
     "lodash": "4.17.20"
   },
   "devDependencies": {
--- a/packages/hooks/package.json
+++ b/packages/hooks/package.json
@@ -26,10 +26,10 @@
     "@babel/runtime": "7.13.10",
     "@babel/runtime-corejs3": "7.13.10",
     "@commercetools-uikit/utils": "11.2.1",
-    "@testing-library/react": "11.2.5",
     "lodash": "4.17.20"
   },
   "devDependencies": {
+    "@testing-library/react": "11.2.5",
     "react": "17.0.1"
   },
   "peerDependencies": {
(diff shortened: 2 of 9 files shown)


packages/hooks/src/use-data-table-sorting-state/use-data-table-sorting-state.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useCallback } from 'react';
+
+const defaultValues = { key: 'createdAt', order: 'desc' };
+
+const useDataTableSortingState = (initialValues = defaultValues) => {
```
tdeekens, line 5: 🤔 do we need this this way? I totally value the explicit merging. I just wonder if `initialValues = defaultValues` is needed as we spread in the `defaultValues` anyway into `mergedValues` and we could spread an `undefined` `initialValues` too?

So my question is would `const useDataTableSortingState = (initialValues) => {` not always be the same?

tdeekens (commented):
Yay!

packages/hooks/src/use-pagination-state/use-pagination-state.js
```diff
@@ -0,0 +1,44 @@
...
+  const onPageChange = useCallback(
+    (nextPage) => {
+      setPage(nextPage);
+    },
+    [setPage]
+  );
+
+  const onPerPageChange = useCallback(
+    (nextPerPage) => {
+      setPage(1);
```
emmenko, line 26: Add a comment why this is needed.

.changeset/slow-grapes-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/hooks': minor
+---
+
+Add pagination-state and data-table-sorting-state.
```
emmenko, line 5:
```suggestion
Add `usePaginationState` and `useDataTableSortingState`. These hooks help to manage pagination state and should be used together with the `<Pagination>` and/or `<DataTable>` components.
```

emmenko (commented):
Looking good

packages/hooks/src/use-pagination-state/use-pagination-state.js
tdeekens: ...and add a test for this case. At least I can't spot one :)

adnasa: good call. I got this.

adnasa:
### latest:

- removes fallback from signature
- adds comment and test
- use `screen`

tdeekens (approved):

packages/hooks/package.json
```diff
@@ -26,6 +26,7 @@
     "@babel/runtime": "7.13.10",
     "@babel/runtime-corejs3": "7.13.10",
     "@commercetools-uikit/utils": "11.2.1",
+    "@testing-library/react": "11.2.5",
```
emmenko, line 29: This should be a dev dep

packages/hooks/src/use-data-table-sorting-state/use-data-table-sorting-state.js
```diff
@@ -0,0 +1,25 @@
+import { useState, useCallback } from 'react';
+
+const defaultValues = { key: 'createdAt', order: 'desc' };
+const applyIf = (values, key) => (values[key] ? { key } : {});
+
+const useDataTableSortingState = (initialValues) => {
+  const mergedValues = {
+    ...defaultValues,
+    ...applyIf(initialValues, 'key'),
+    ...applyIf(initialValues, 'order'),
```
adnasa, line 10: I preferred to take this approach over `pick`.
My opinion was that if the consumer misses to provide a key or order, the hook is "resilient" to catch and fallback to the defined default values.

image

emmenko (approved):
💯

PR #1843, opened 2021-03-31
refactor(link): use typescript

adnasa:
#### Summary

- migrates `Link` to TS 😎


packages/components/link/src/link.tsx
```diff
@@ -0,0 +1,143 @@
...
+  `;
+};
+
+const Link = (props: TLinkProps) => {
+  const theme = useTheme();
+
+  const remainingProps = filterInvalidAttributes(props);
+  warnIfMissingContent(remainingProps as TLinkProps);
+
+  if (props.isExternal && typeof props.to === 'string') {
```
adnasa, line 107: There is probably a better way for type guarding. suggestions welcome

packages/components/link/src/link.tsx
```diff
@@ -0,0 +1,143 @@
...
+type TLinkShape = {
+  pathname: string;
+  search: string;
+  hash?: string;
+  state: unknown;
+};
+
+type TLinkProps = {
+  tone?: 'primary' | 'inverted';
+  // @TODO: deprecate?
```
adnasa, line 22: I just realised that this will be covered with V12.
Setting this PR to `on hold` in the mean time

packages/components/link/src/link.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+type TExtendedTheme = Theme & {
+  [key: string]: string;
+};
+type TLinkShape = {
+  pathname: string;
+  search: string;
+  hash?: string;
+  state: unknown;
+};
```
adnasa, line 18: I'm tempted to use the `LocationDescriptor` instead, because it makes a lot more sense than us having to define it ourselves

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/04657b253bcdd391b28709cbbf9bd8d78f81611c/types/react-router/v2/lib/Router.d.ts#L36:L41

This will mean dep on `@types/history`.

what do you think?

packages/components/link/src/link.tsx
emmenko: Hmm in this case I would only keep the `if (props.isExternal)` condition.

If the `to` prop is a location object, we can construct the URL string on our own.

adnasa: good call.
I'll have a look at that.

emmenko (approved):

PR #1854, opened 2021-04-08
refactor(text): accept `as` prop for inline render of `Text.Detail`

adnasa:
#### Summary

- caused by #1824 

Discovered when migration app-kit
https://github.com/commercetools/merchant-center-application-kit/pull/2131#discussion_r609585711

I don't think `isInline` was marked deprecated on `Text.Detail` in the first place, nor was it intended that `Text.Detail` should support `as={}` prop.

### DIFF
--- a/packages/components/text/src/text.styles.ts
+++ b/packages/components/text/src/text.styles.ts
@@ -34,6 +34,10 @@ const italic = `
   font-style: italic;
 `;
 
+const inline = `
+  display: inline-block;
+`;
+
 const getTone = (tone: string, theme: Theme) => {
   const overwrittenVars = {
     ...vars,
@@ -117,6 +121,7 @@ export const detailStyles = (props: TDetailProps, theme: Theme) => css`
   ${getBaseStyles(theme)}
   display: block;
   font-size: 0.9231rem;
+  ${props.isInline && inline}
   ${props.isBold && bold}
   ${props.isItalic && italic}
   ${props.tone && getTone(props.tone, theme)}
--- a/packages/components/text/src/text.styles.ts
+++ b/packages/components/text/src/text.styles.ts
@@ -34,10 +34,6 @@ const italic = `
   font-style: italic;
 `;
 
-const inline = `
-  display: inline-block;
-`;
-
 const getTone = (tone: string, theme: Theme) => {
   const overwrittenVars = {
     ...vars,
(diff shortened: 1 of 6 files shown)


adnasa:
Missing `changeset` was intended.
It should be part of this release.

https://github.com/commercetools/ui-kit/pull/1850

packages/components/text/src/text.styles.ts
```diff
@@ -117,6 +121,7 @@ export const detailStyles = (props: TDetailProps, theme: Theme) => css`
   ${getBaseStyles(theme)}
   display: block;
```
emmenko, line 122: Hmm I guess this is the main cause of this issue 🤔

Using `as="span"` changes the HTML semantic since we want to use `<small>` (I suppose).

I wonder why we have `display: block` in the first place though, if we're using `<small>`.

I'm thinking to actually change `small` to a `div`, then we can use `as="span"` and things are consistent. And remove `display: block`

emmenko (changes requested):

packages/components/text/src/text.styles.ts
adnasa: Hm, that would be a good proposal.
I'm leaning towards more removing just the `display` rule.
I don't think we need to change `small -> div`

adnasa: aaah dear... now I finally found a doc spec where `small` is by default an inline element

packages/components/text/src/text.tsx
```diff
@@ -203,6 +203,8 @@ Body.displayName = 'TextBody';
 export type TDetailProps = {
   isBold?: boolean;
   isItalic?: boolean;
+  // only span is supported for now.
+  as?: 'span';
```
adnasa, line 207: I realise now that we switch the default render to `div` and had `display: block` removed.
Given that `<small>` is an inline element, we can also accept it here.

emmenko: Yep

emmenko (approved):

packages/components/text/src/text.tsx
```diff
@@ -219,14 +220,27 @@ const Detail = (props: TDetailProps) => {
   const theme = useTheme();
   warnIfMissingTitle(props, 'TextDetail');
   warnIfMissingContent(props, 'TextDetail');
+  if (props.as) {
+    const TextDetailElement = props.as;
```
emmenko, line 224: Nit: do we need the if-else? Can't we simply do `const TextDetailElement = props.as || div`? 🤔

adnasa: I like this.
I added support for `small` too, as another option to inline the text, besides `span` 😉

PR #1866, opened 2021-04-15
chore(package): add `patch-package` and apply patch

adnasa:
#### Summary

- adds `patch-package` and apply patch on `@types/react-textarea-autosize` until the official package is updated.

## Description

This is an intermediate solution until https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52391 is merged.
Most of the diff on the patch file is due to indentation, but the important changes are:

1. updating signature of `onHeightChange`
2. exposing `TextareaHeightChangeMeta` (for optional use)
3. exposing `style` prop as it is defined on the original package (https://github.com/Andarist/react-textarea-autosize/blob/master/src/index.tsx#L24:L24)

## Open question

The `patch-package` has recommendations on updating the CI. not sure how to translate that into GH actions. I welcome pairing on this.


package.json
```diff
@@ -13,7 +13,7 @@
   "keywords": ["javascript", "frontend", "ui-kit", "design-system", "react"],
   "sideEffects": false,
   "scripts": {
-    "postinstall": "manypkg check && preconstruct dev && yarn --cwd docs",
+    "postinstall": "patch-package",
```
adnasa, line 16: oh wait..

tdeekens:
Can we remove the indentation somehow? I wonder if it causes issues if the library moves in for us unrelated/uninteresting lines.

emmenko (commented):
I'm confused. Doesn't the library have it's own types?
Why do we need the outdated types package? 🤔

adnasa:
> I'm confused. Doesn't the library have it's own types?
Why do we need the outdated types package? 🤔


@emmenko 
hey you're right and I managed to get around it.
Thanks!

Closing this now.

I still think that `patch-package` is a dope tool, actually. so thanks for the tip, @tdeekens

PR #1867, opened 2021-04-16
Migrate `input-utils` to TypeScript

adnasa:
#### Summary

- migrated `input-utils` to TypeScript
- blocked by `buttons` migration , so I set the target of this PR to #1856

PR #1868, opened 2021-04-17
refactor(text-input): migrate to TypeScript

adnasa:
#### Summary

- blocked by #1867
- This is cherry picked work, so I expect that the build will fail until the blocking PR is merged


packages/components/inputs/text-input/src/text-input.tsx
```diff
@@ -1,127 +1,131 @@
...
-        contentEditable={!props.isReadOnly}
-      />
-    </Constraints.Horizontal>
-  );
-};
-
-TextInput.displayName = 'TextInput';
-
-TextInput.propTypes = {
+type TTextInputProps = {
```
fuchodeveloper, line 7: Do we have a convention typing convention for ui-kit at the moment?

adnasa: thanks for this question, but is there something specific you're referring to?
I set a `T*` prefix to differentiate Types and Interface for devs who rely on editors without intellisense. Does that make sense?

fuchodeveloper: Basically, I'm asking if using a naming like, `T*` prefix, is a convention for ui-kit. This potentially could help with migrations for other files and avoids ambiguity.
Does that help?

adnasa: then yes it is

fuchodeveloper (approved):

dogayuksel (approved):

PR #1869, opened 2021-04-17
refactor(toggle-input): use TypeScript

adnasa:
#### Summary

- blocked by #1867
- This is cherry picked work, so I expect that the build will fail until the blocking PR is merged


Rombelirk (approved):

adnasa:
I've dug around the issue to why the tests are failing, but can't find a good source on how to resolve this

```
  ● should render children

    Error: Uncaught [Error: Component selectors can only be used in conjunction with @emotion/babel-plugin.]

      45 |     if (shouldNotThrowWarnings(messages)) return;
      46 |
    > 47 |     throw new Error(...messages);
         |     ^
      48 |   } else {
      49 |     log(colors.bgYellow.black(' WARN '), warning, '\n', ...messages);
      50 |   }

```

any ideas, @emmenko ?

adnasa:
Okay, it's clear now what the problem is.
In this case, this block causes an issue
https://github.com/commercetools/ui-kit/blob/adnasa-toggle-input-typescript/packages/components/inputs/toggle-input/src/toggle-input.tsx#L123:L133

jest.test.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     // - https://github.com/kulshekhar/ts-jest/blob/master/docs/user/config/isolatedModules.md
     'ts-jest': {
       isolatedModules: true,
+      babelConfig: true,
```
tdeekens, line 18: Otherwise the babel plugins won't run at all. Which means that e.g. the emotion plugin doesn't run which is needed for component selectors.

https://kulshekhar.github.io/ts-jest/docs/getting-started/options/babelConfig/

tdeekens (approved):

PR #1871, opened 2021-04-17
refactor(checkbox-input): migrate to TypeScript

adnasa:
#### Summary

- blocked by #1867
- This is cherry picked work, so i expect that the build will fail until the blocking PR is merged


bradd123 (approved):

packages/components/inputs/checkbox-input/src/checkbox.tsx
```diff
@@ -36,26 +56,10 @@ const Checkbox = (props) => {
       readOnly={props.isReadOnly}
       checked={props.isChecked && !props.isIndeterminate}
       onChange={handleChange}
-      ref={ref}
-      /* ARIA */
-      aria-readonly={props.isReadOnly}
-      aria-checked={props.isChecked}
-      {...props}
```
mohib0306, line 43: Is removing this prop spreading on purpose?

packages/components/inputs/checkbox-input/src/checkbox.tsx
```diff
@@ -1,33 +1,53 @@
...
   }
 `;
 
-const Checkbox = (props) => {
-  const ref = React.useRef();
+type TInputRef = {
+  indeterminate: boolean;
+};
+
+type TProps = Omit<TCheckboxProps, 'children'> & {
```
mohib0306, line 21: I think you need to omit `hasError` and `isHovered` props as well as they are not part of the old PropTypes.

packages/components/inputs/checkbox-input/src/checkbox.tsx
adnasa: to me, it does.
We pass in the props we accept on the component and the rest are `aria/data` attributes we forward.
then there isn't any leak from what I can tell.

mohib0306: I am thinking if someone is leveraging this props spreading in their code base, it might result in breaking change.

adnasa: and they shouldnt.

tdeekens (approved):

PR #1872, opened 2021-04-17
refactor(hook): migrate to TypeScript

adnasa:
#### Summary

😬


packages/hooks/src/use-data-table-sorting-state/use-data-table-sorting-state.ts
```diff
@@ -0,0 +1,51 @@
...
+type TSortDefinition = {
+  key: string;
+  order: 'desc' | 'asc';
+};
+
+type TDataTableSortingState = {
+  value: TSortDefinition;
+  onChange: (
+    key: TSortDefinition['key'],
+    order: TSortDefinition['order']
```
adnasa, line 13: I was tempted to update the callback signature but that would be a breaking change. I played around with Types and thought maybe this is a great use case to use function overloads and let the type describe intent with the signature

https://gist.github.com/adnasa/66611e1fb9064074caa2eaf17297562e

ahmehri (approved):

PR #1874, opened 2021-04-17
refactor(components/field-errors): migrate to TypeScript

adnasa:


adnasa:
Assigning review request by random

packages/components/field-errors/src/field-errors.tsx
```diff
@@ -1,18 +1,39 @@
...
+};
+
+const FieldErrors = (props: TFieldErrorsProps) => {
   if (!props.isVisible) return null;
   if (!isObject(props.errors)) return null;
 
   return (
     <React.Fragment>
-      {Object.entries(props.errors)
+      {Object.entries<boolean>(props.errors)
```
pa3, line 36: I guess TS is smart enough to infer `boolean` here, isn't it? Should work fine without explicit type parameter.

adnasa: oh.. I set it there because I wanted to test the `errors` object with different types.
but yes, the intellisense covers this :)

pa3 (approved):

PR #1875, opened 2021-04-17
refactor(number-input): migrate to TypeScript

adnasa:
#### Summary

- blocked by #1867
- This is cherry picked, so I expect that the build will fail until the blocking PR is merged


packages/components/inputs/number-input/src/number-input.tsx
```diff
@@ -151,13 +149,13 @@ NumberInput.toFormValue = (numberOrString) => {
...
 
-NumberInput.isEmpty = (value) => {
+NumberInput.isEmpty = (value: number | string): boolean => {
   if (typeof value === 'string') return value.trim().length === 0;
   if (typeof value === 'number') return isNaN(value);
   return true;
 };
 
-NumberInput.hasFractionDigits = (number) => {
+NumberInput.hasFractionDigits = (number: number) => {
```
dogayuksel, line 158: warning message sounds like i may be able to pass in a string type to this function, but the function signature only specifies numbers.

adnasa: that's my mistake. good catch!

adnasa: that's my mistake, and good catch on the message.

tdeekens (approved):
💯

PR #1891, opened 2021-05-04
refactor(tooltip): migrate to TypeScript

adnasa:
## Description

This needs to be merged into unblock migration of calendar/date inputs and fields


packages/components/tooltip/README.md
```diff
@@ -180,75 +180,51 @@ export {
...
+(e?: ChangeEvent | FocusEvent) => void
+```
+
+### Signature `onOpen`
+
+```ts
+(e?: ChangeEvent | FocusEvent) => void
+```
+
+### Signature `components`
```
adnasa, line 213: hmm. interesting how these got extracted 😅

packages/components/tooltip/README.md
```diff
@@ -180,75 +180,51 @@ export {
...
+| ---------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------: | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `children`             | `ReactReactElement`                                                                                                                                                               |    ✅    |           |                                                                                                                                                                                                          |
```
ahmehri, line 185: Is `ReactReactElement` a correct type?

packages/components/tooltip/README.md
```diff
@@ -180,75 +180,51 @@ export {
...
+| `components`           | `Object`<br/>[See signature.](#signature-components)                                                                                                                              |          |           | Customize the appearance of certain elements of the tooltip.                                                                                                                                             |
+| `horizontalConstraint` | `union`<br/>Possible values:<br/>`, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 'scale', 'auto'`                                                                       |          | `'scale'` | Horizontal size limit of the tooltip.                                                                                                                                                                    |
```
ahmehri, line 197:
```suggestion
| `horizontalConstraint` | `union`<br/>Possible values:<br/>` 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 'scale', 'auto'`                                                                       |          | `'scale'` | Horizontal size limit of the tooltip.                                                                                                                                                                    |
```

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -1,49 +1,153 @@
...
-  );
-  warning(
-    props.components.WrapperComponent
-      ? isValidElementType(props.components.WrapperComponent)
-      : true,
-    `ui-kit/Tooltip: the prop 'components.WrapperComponent' is not a valid React element.`
-  );
+const Tooltip = (props: TTooltipProps) => {
+  const leaveTimer = React.useRef<ReturnType<typeof setTimeout>>();
+  const childrenRef = React.useRef<ReturnType<typeof setTimeout>>();
```
ahmehri, line 132: I get the usage of `typeof setTimeout` in the previous line but is it correct also in this line?

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -1,49 +1,153 @@
...
+   */
+  off: boolean;
+  /**
+   * An identifier for the tooltip, used for `aria-describedby`.
+   */
+  id?: string;
+  /**
+   * A callback function, called when the tooltip is closing.
+   */
+  onClose?: (e?: ChangeEvent | FocusEvent) => void;
```
tdeekens, line 54: Suspecting an obvious answer but actually don't know: what makes the event optional? It means every consider has to check for the event being defined.

adnasa: see the implementation block of `handleEnter` and `handleClose` to get an insight on this.
The underlying implementation mixes concerns.

packages/components/tooltip/src/tooltip.tsx
adnasa: yes.

packages/components/tooltip/README.md
adnasa: This is generated, so I assume so?

packages/components/tooltip/README.md
adnasa: This is a generator quirk. will not fix here.

packages/components/tooltip/README.md
ahmehri: I didn't know that it's generated. I'm asking because I don't know whether there is a typo in the name so that the type is `ReactElement` instead or `ReactReactElement` is considered as a correct type.

ahmehri (approved):

packages/components/tooltip/src/tooltip.tsx
tdeekens: Thanks for the reference.

mohib0306 (approved):

packages/components/tooltip/src/tooltip.tsx
adnasa: Looking at the implementation again, I realised that it wasn't a great idea to mix concerns there.
I considered re-factoring and faking synthetic events.

tdeekens (approved):

PR #1901, opened 2021-05-10
refactor(hooks): convert entry points to TS, types in signatures for `useToggleState` and `useFieldId`

adnasa:
#### Summary

- updates signature for `useFieldId` to accept optional `id`
Some components made `props.id` optional and pass it to `useFieldId` which causes a compile error.
I set the parameter to optional so `useFieldId` can resolve this through `createFieldIdFn`
- useToggleState: remove type when assigning `useCallback`.
I found this to be superfluous and not really needed when it is inferred through the callback itself.
- updates entry points to be `*.ts`.


adnasa:
blocks #1891

packages/hooks/src/use-field-id/use-field-id.ts
```diff
@@ -3,8 +3,9 @@ import { getFieldId } from '@commercetools-uikit/utils';
 
 type CreateIdFn = () => string;
 
-const useFieldId = (id: string, createIdFn: CreateIdFn): string => {
-  const [internalId, setId] = React.useState<string>(id);
+// @ts-ignore
+const useFieldId = (id?: string, createIdFn: CreateIdFn): string => {
```
adnasa, line 7: this has to be optional without updating the signature.
`@ts-ignore` is added, since we have a requirement that a required param shouldn't follow an optional. changing this is risk to breaking change.

will update the issue for v13.

packages/hooks/src/use-toggle-state/use-toggle-state.ts
```diff
@@ -1,10 +1,12 @@
 import { useCallback, useState } from 'react';
 
-const useToggleState = (defaultValue?: boolean) => {
+const useToggleState = (
+  defaultValue?: boolean
+): [boolean, (forceIsToggled?: boolean) => void] => {
```
adnasa, line 5: it's better to be explicit here rather than moving the `ReturnType` to a declaration.
We need to do this, because TS misunderstood the return type 

output of intellisense, note the `[]`
```
const useToggleState: (defaultValue?: boolean | undefined) => (boolean | ((forceIsToggled?: boolean | undefined) => void))[] // <-- note the []
```

packages/hooks/src/use-field-id/use-field-id.ts
```diff
@@ -13,7 +14,7 @@ const useFieldId = (id: string, createIdFn: CreateIdFn): string => {
     setId(getFieldId(props, state, createIdFn));
   }, [id, internalId, setId, createIdFn]);
 
-  return internalId;
+  return internalId as string;
```
adnasa, line 17: we have to cast this, however at this point we should rely on the `internalId` being finally resolved via the `createIdFn` which is a required param

packages/components/inputs/multiline-text-input/src/multiline-text-input.tsx
```diff
@@ -7,7 +7,6 @@ import { useIntl } from 'react-intl';
 import { css } from '@emotion/react';
 import { AngleUpIcon, AngleDownIcon } from '@commercetools-uikit/icons';
 import FlatButton from '@commercetools-uikit/flat-button';
-// @ts-ignore
```
adnasa, line 10: 😬

adnasa:
oooh yess, the build passed 😎

packages/components/inputs/multiline-text-input/src/multiline-text-input.tsx
adnasa: this was just a left over from previous migration 😄

packages/hooks/src/use-field-id/use-field-id.ts
emmenko: Can you remind me why this should be optional?

adnasa: Some components made `props.id` optional, but still pass it to `useFieldId`

adnasa: sounds good, let's try that.

adnasa: https://github.com/commercetools/ui-kit/pull/1901/commits/4f03a25af6b9ceda96fb5728dd68ae770d756cc9

emmenko (approved):

PR #1904, opened 2021-05-10
chore(field-errors): export field error type

adnasa:
#### Summary

Exporting the types from `FieldErrors`.
we would need it in app-kit


packages/components/field-errors/src/index.ts
```diff
@@ -1,3 +1,10 @@
...
+import type {
+  TFieldErrorsProps as FieldErrorProps,
+  TFieldErrors as FieldErrors,
+} from './field-errors';
 
+export { default } from './field-errors';
 export { default as version } from './version';
+
+export type TFieldErrorsProps = FieldErrorProps;
+export type TFieldErrors = FieldErrors;
```
tdeekens, line 10: I guess `export type TFieldErrors from './field-errors';` won't work?

adnasa: I like this style of barrel export for `type` better :)

adnasa: it really is a preference thing from my side 😄

adnasa: I understand. I think it makes no difference in the end.

you misunderstood when _Ah that's another thing which isn't a type (FieldErrors)_
but we can change this back if this is what you want. it's not clear

tdeekens (approved):

packages/components/field-errors/src/index.ts
adnasa: agree. 
This has become a preference/habit due to the nuisance of the ts-linter.

It has come to my attention that TS compiler doesn't like this:

```
export default {...} // <-- a const must be defined. adding `as const` casts a read-only
```

tdeekens: Yes, recently eslint presets also want about it if not assigned to a const before default exporting. Which can be understandable: what are we exporting otherwise in terms of name.

emmenko: > I guess `export type TFieldErrors from './field-errors';` won't work?

Maybe it works, we can try that. I remember having issues with that in appkit with rollup, as it's babel doing the compilation and not typescript.

Maybe with preconstruct things work slightly different and this is now possible.

PR #1909, opened 2021-05-11
refactor(collapsible): migrate to TypeScript

adnasa:
#### Summary

let's go


packages/components/field-label/README.md
```diff
@@ -32,7 +32,7 @@ npm --save install react
 ```jsx
 import React from 'react';
 import { defineMessage, FormattedMessage } from 'react-intl';
-import { FlatButton } from '@commercetools-buttons/flat-button';
+import FlatButton from '@commercetools-uikit/flat-button';
```
adnasa, line 35: unrelated, but we did merge field-label yesterday so this can be included in the same release

tdeekens (approved):

PR #1930, opened 2021-06-01
fix(data-table-manager): areHiddenColumnsSearchable prop has no effect

pa3:
According to `<DataTableManager />` documentation the name of the prop controlling presence of search fields in column manager is `areHiddenColumnsSearchable`. But, it is never took in account, and `<ColumnSettingsManager/>` relies on an undocumented prop called `disableHiddenColumnSearch`. 

This PR brings searching behaviour in accordance with docs.

### DIFF
--- a/.changeset/big-bobcats-shout.md
+++ b/.changeset/big-bobcats-shout.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/data-table-manager': patch
+---
+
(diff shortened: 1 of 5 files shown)


tdeekens (approved):

jaikumar-tj (approved):

ahmehri (approved):

tdeekens:
@pa3 feel free to merge. We can release at any time. Should be good.

PR #1931, opened 2021-06-02
fix(data-table-manager): make top bar take up all available space

pa3:
`<DataTableManager />` wraps provided top bar controls in an unstyled `<div />`. That brings some troubles when styling controls being shown in a table's top bar. In general, as that `<div />` doesn't grow, it works against UI-Kit layout pattern (or let's better say MC pattern of UIKit usage) of letting controls grow up to the limit provided by the wrapping `<Constraints.Horizontal />`.

This PR solves described issue by making top bar container grow (by adding `flex-grow: 1`):

|Before|After|
|--|--|
|Screenshot from 2021-06-02 12-13-05|Screenshot from 2021-06-02 12-11-43| 

I have checked already how this change affects exiting usages of `tobBar` in MC and there are no visual changes.

### DIFF
--- a/.changeset/old-bugs-exist.md
+++ b/.changeset/old-bugs-exist.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/data-table-manager': patch
+---
+
+fix(data-table-manager): make top bar take up all available space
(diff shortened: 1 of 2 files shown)


jaikumar-tj (approved):

tdeekens (approved):

tdeekens:
Maybe UX can verify the Percy diff.

PR #1938, opened 2021-06-11
fix: disable core-js in test environment

tdeekens:


packages/components/collapsible-motion/package.json
```diff
@@ -9,12 +9,7 @@
     "directory": "packages/components/collapsible-motion"
   },
   "homepage": "https://uikit.commercetools.com",
-  "keywords": [
-    "javascript",
-    "design system",
-    "react",
-    "uikit"
-  ],
```
tdeekens, line 17: Prettier.

pa3 (approved):

PR #1941, opened 2021-06-17
feat(link): add icon when link is external

tdeekens:
#### Summary

This automatically adds an `ExternalLinkIcon` to any `Link` which has `isExternal`.

## Description

We learned that we don't consistently use the same patterns for external links and have to custom build components internally when a link is external and should have an icon to reflect that.

The discussion has open ends, for instance if the link should be underlined. As a result this is not ready. It's merely the result of a first investigation anybody could pick up.

CleanShot 2021-06-17 at 15 57 50

CleanShot 2021-06-17 at 16 48 59

## Todos

- [ ] Should the link have an underline or not
- [ ] How do we deal with active colour on the icon

### DIFF
--- a/packages/components/link/src/link.tsx
+++ b/packages/components/link/src/link.tsx
@@ -7,6 +7,8 @@ import { css, useTheme } from '@emotion/react';
 import { FormattedMessage } from 'react-intl';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 import { filterInvalidAttributes, warning } from '@commercetools-uikit/utils';
+import { ExternalLinkIcon } from '@commercetools-uikit/icons';
+import SpacingsInline from '@commercetools-uikit/spacings-inline';
 
 type TExtendedTheme = Theme & {
   [key: string]: string;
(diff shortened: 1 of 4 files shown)


packages/components/link/src/link.visualroute.js
```diff
@@ -36,14 +31,14 @@ export const component = ({ themes }) => (
       <Link to="/" intlMessage={intlMessage} />
     </Spec>
     <ThemeProvider theme={themes.darkTheme}>
-      <Spec label="tone - inverted - with underline">
```
tdeekens, line 39: We don't support removing the underline anymore. The snapshot here was flawed.

packages/components/link/src/link.tsx
```diff
@@ -98,17 +100,19 @@ const getLinkStyles = (props: TLinkProps, theme: Theme) => {
...
-    font-size: ${overwrittenVars.fontSizeDefault};
-    text-decoration: underline;
-    &:hover,
-    &:focus,
-    &:active {
-      color: ${hoverColor};
-    }
-  `;
+  return [
+    css`
```
tdeekens, line 104: Unrelated, can we reverted.

antoniolodias (approved):

PR #1964, opened 2021-07-08
feat(icons): add <InlineSvg> component

emmenko:
See README for explanation about this component and how to use it.

There is also a storybook page to play around with it: https://ui-kit-git-nm-icons-inline-svg-commercetools.vercel.app/?path=/story/components-icons--inline-svg


package.json
```diff
@@ -54,14 +54,28 @@
     "packages": [
       "design-system",
       "generators/*",
-      "packages/**/*",
+      "packages/*",
+      "packages/components/*",
+      "packages/components/buttons/*",
+      "packages/components/fields/*",
+      "packages/components/inputs/*",
+      "packages/components/spacings/*",
```
emmenko, line 62: The changes here are required (it's the only way I could find to make things work) as issues with manypkg vs preconstruct started to appear when I added the new entrypoint.

For instance, it was asking me to put a `name` field to the `package.json` of the entrypoint. But doing that caused another incompatible error.

So instead of specifying a `**/*` glob to include subfolders, I'm explicitly providing the path to nested packages.

svgr.config.js
```diff
@@ -3,7 +3,10 @@ const fs = require('fs');
 const path = require('path');
 
 const templateCreateStyledIcon = fs.readFileSync(
-  path.join(__dirname, `templates/icon.styles.tsx`),
+  path.join(
+    __dirname,
+    `packages/components/icons/src/templates/icon.styles.tsx`
```
emmenko, line 8: Moved this into the `icons` package to be able to export a bunch of stuff needed by the new component.

packages/components/icons/src/inline-svg.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+    * {
+      fill: ${fillColor};
+    }
+  </style>`;
+
+  // Render SVG within an sandboxed iframe, to prevent any sort of embedded scripts to be executed.
+  return (
+    // eslint-disable-next-line jsx-a11y/iframe-has-title
+    <iframe
```
tdeekens, line 72: Can you share some research you did driving you decision to use an `iframe` instead of sanitising the SVG using e.g. `dompurify` which is about `17.7 KB` (non GZIP)?

emmenko: I did look into various methods and libraries to sanitize the SVG (not dompurify admittedly) and most of the "guides/tutorials" were recommending those two approaches, mostly as a secure way of preventing any sort of XSS.

However, now that I look at dompurify it does indeed seem to do the job.

I'll take another look at this, thanks for the tip.

The important thing is that there is no risk of XSS.

tdeekens: Yes, that should be our goal. I was just wondering if an iframe tackles all the issues or leaves some vectors open. I know that DOMPurify is used by many large companies to do exactly this.

Generally I would think that we given a user can upload an SVG somewhere:

1. Use `dompurity` on the API to sanitize also anything we save
2. Additionally sanitize anything we show which isn't coming from us

Somehow maybe the 2. point might not be strictly needed if we don't inject anything uploaded into the DOM before not having sanitized it in the API. 🤔

packages/components/icons/src/inline-svg.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const InlineSvg = (props: InlineSvgProps) => {
+  const theme = useTheme();
+  const sanitized = DOMPurify.sanitize(props.data, {
+    USE_PROFILES: { svg: true },
+    FORBID_ATTR: [
+      // To avoid injection by using `style="filter:url(\"data:image/svg+xml,<svg`
+      'style',
+    ],
+  });
```
emmenko, line 21: Here we sanitize now. I tested many different combinations (see `fixtures/xss.js`) and no "alert" popped up.

packages/components/icons/src/inline-svg.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return <SVG src={sanitized} css={getIconStyles(props, theme)} />;
```
emmenko, line 22: We render the SVG using `react-inlinesvg`, which nicely parses the string into react/dom elements.

It also works with remote URLs, in case one is provided.

image

packages/components/icons/src/icon.story.js
```diff
@@ -25,6 +37,147 @@ const IconContainer = styled.div`
...
+                  );
+                })}
+              </Grid>
+            </div>
+          </Spacings.Stack>
+        ))}
+      </Spacings.Stack>
+    </Spacings.Stack>
+  );
+};
```
emmenko, line 179: This is the playground page to play around with any custom data.

packages/components/icons/src/inline-svg.tsx
```diff
@@ -1,97 +1,25 @@
...
-      return <InlineSvgAsImage {...props} />;
-    case 'iframe':
-      return <InlineSvgAsIFrame {...props} />;
-    default:
-      throw new Error(
-        `Wrong "as" value: "${props.as}". Expected one of "image" or "iframe".`
-      );
-  }
+  const theme = useTheme();
+  const sanitized = DOMPurify.sanitize(props.data, {
```
tdeekens, line 15: I would encapsulate this to not wastefully re-sanitise on reach render:

```js
function useSanitizedSvg(svg) {
   const sanitizedSvg = useMemo(() => , [svg])

   return sanitizedSvg;
}
```

tdeekens (approved):

emmenko:
@tdeekens FYI: I refactored things a bit more, specifically dropped `react-inlinesvg` and picked only the logic needed for our use case (5% of the library implementation). Main reason is that we don't want to support other data formats like URL, or data URI, as we're then vulnerable again to XSS.

See last commit: https://github.com/commercetools/ui-kit/pull/1964/commits/753ac50b3217e94c8e92dcf7b6b4fb8d0d55f2a9

tdeekens (approved):
Nice idea. Like using only what we need.

PR #1965, opened 2021-07-09
fix(icons): do not override fill color when fill=none is defined

emmenko:


packages/components/icons/src/templates/icon.styles.tsx
```diff
@@ -81,7 +81,7 @@ const getColor = (color: Props['color'], theme: Theme) => {
 
 // @ts-ignore
 export const getIconStyles = (props: Props, theme: Theme) => css`
-  * {
+  *:not([fill='none']) {
```
emmenko, line 84: The fix

tdeekens (approved):
Edge cases ruin our lives :-)

tdeekens (approved):

tdeekens (approved):

PR #1970, opened 2021-07-23
fix: type annotations for as props, patch prop-types babel plugin

emmenko:
See changeset


emmenko:
Turned out that the `Record` type was not supported by the babel plugin. I added another patch to convert that to `PropTypes.objectOf`.

islam3zzat (approved):
💯

PR #1990, opened 2021-08-18
PAN-1029: add delay to loading spinner

ByronDWall:
#### Summary
Adds a default 1 second delay between `LoadingSpinner` component mount and render of the loading animation.

## Description
A default timeout of 1 second has been added to reduce the overall amount of animation flickering experienced by the user if loading times are under 1 second.  For more context into this UX descision, see [this article](https://www.smashingmagazine.com/2016/12/best-practices-for-animated-progress-indicators/).

The test suite, documentation, and readme for the `LoadingSpinner` have been updated, and the readmes of a few files have been updated to display better type labels (ie from `ReactReactNode` to `ReactNode`

- Documentation

  - [x] Has `README.md` file
  - [x] Explains use-cases
  - [x] Explains props
  - [x] Has story for Storybook

- Props

- Design

  - [x] Uses design-tokens (instead of e.g. color variables)

- Testing

  - [x] Uses [@testing-library/react](https://github.com/testing-library/react-testing-library) tests
  - [x] Has visual regression tests


packages/components/loading-spinner/src/loading-spinner.visualroute.js
```diff
@@ -5,7 +5,7 @@ export const routePath = '/loading-spinner';
 
 export const component = () => (
   <Suite>
-    <Spec label='with scale "l" (default)'>
+    <Spec label='with scale "l", maxDelayDuration "1000" (default)'>
```
ByronDWall, line 8: looks like I need to do more to account for the timeout here.  I will look into this.

ByronDWall: The percy/pupeteer tests now pass w/the loading spinner showing up after adding a `waitFor` method to the visualspec.  The jest/unit tests test to ensure that the spinner does not display if the component has been mounted for less than the time period set by the `maxDelayDuration` prop, so I am unsure whether I should also test that behavior in the e2e tests.  Please let me know what the best practice would be for a case like this.

 Either way, the tests on percy will need to reviewed/approved since the new prop was added.

lint-staged.config.js
```diff
@@ -25,6 +25,10 @@ module.exports = {
     //
     // For that reason, we move the `--onlyChanged` flag next to it.
     'yarn lint:js --reporters=jest-silent-reporter --onlyChanged',
-    'tsc-files --noEmit',
+    // NOTE: lint-staged ignores tsconfig.json when called through husky hooks because husky passes filenames as an argument
+    // calling tsc using function syntax allows tsc to use tsconfig.json,
+    // causing tsc to have the same output as it would if it were called via `yarn typecheck`
+    // see https://github.com/okonet/lint-staged/issues/825#issuecomment-620018284
```
islam3zzat, line 31: 👍

islam3zzat (approved):
Thanks for the PR comments, made the review easier <3

bradd123 (approved):
Nice work 💯

lint-staged.config.js
```diff
@@ -25,6 +25,10 @@ module.exports = {
     //
     // For that reason, we move the `--onlyChanged` flag next to it.
     'yarn lint:js --reporters=jest-silent-reporter --onlyChanged',
-    'tsc-files --noEmit',
```
emmenko, line 28: This si in purpose, and it's a wrapper around `tsc`. It's meant to be used with something like lint-staged.

What's the reason for changing that?

packages/components/loading-spinner/src/loading-spinner.tsx
```diff
@@ -27,6 +27,11 @@ const sizePerScale = {
 const positionOrigin = '20px';
 
 type TLoadingSpinnerProps = {
+  /**
+   * Set the amount of time to delay the loading spinner before it renders.
+   * The default value of 1000ms has been agreed upon by the design team for all usecases.
```
emmenko, line 32: This is meant for public documentation, no need to mention the agreement thing.

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "esModuleInterop": true,
     "importHelpers": true,
     // See https://devblogs.microsoft.com/typescript/typescript-and-babel-7/
+    "noEmit": true,
```
emmenko, line 9: Why this change?

.changeset/green-moons-wonder.md
```diff
@@ -0,0 +1,14 @@
...
+'@commercetools-uikit/secondary-icon-button': patch
+'@commercetools-uikit/collapsible-panel': patch
+'@commercetools-uikit/field-errors': patch
+'@commercetools-uikit/checkbox-input': patch
+'@commercetools-uikit/link': patch
+'@commercetools-uikit/loading-spinner': patch
+'@commercetools-uikit/tooltip': patch
+---
+
+feat(LoadingSpinner): add 1 second delay between component mount and render of spinner, update loading spinner tests and docs, chore(readmes): update readmes
```
emmenko, line 14:
```suggestion
The `<LoadingSpinner>` now waits for 1s before rendering. This is meant to improve the visual perceived feedback.
```

emmenko (commented):
Can you clarify a bunch of things? See comments, thanks.

lint-staged.config.js
ByronDWall: I would have to go back and clone from scratch to recreate the issue, but this was changed because the `lint-staged` husky commit hook was failing for me after the initial install.  Changing the config to use this syntax made it so that the typecheck step ran without issue.

This issue/comment in the `lint-staged` github repo goes into more detail: https://github.com/okonet/lint-staged/issues/825#issuecomment-620018284

tsconfig.json
ByronDWall: I would have to clone and reinstall the repo to be sure, but I was having issues with babel/typescript interop, this made it so that the Babel transforms worked (in my environment at least). I added this based on MS's own ts/babel config setup:
Screen Shot 2021-09-28 at 2 36 33 PM

packages/components/loading-spinner/src/loading-spinner.tsx
ByronDWall: noted, and updated.

lint-staged.config.js
emmenko: Note that `tsc-files` is not the same as `tsc`. Please revert those changes as it was working fine before (we also have the same setup in other repos).
In any case, it's something not related with the scope of this PR, we can look at this separately.

Thanks

packages/components/loading-spinner/README.md
```diff
@@ -39,7 +39,8 @@ export default Example;
...
-| ---------- | -------------------------------------------- | :------: | ------- | ------------------------------------------------- |
-| `scale`    | `union`<br/>Possible values:<br/>`'s' , 'l'` |          | `'l'`   | Set the size of the loading spinner.              |
-| `children` | `ReactReactNode`                             |          |         | The content rendered inside the `LoadingSpinner`. |
+| Props              | Type                                         | Required | Default | Description                                                                                                                                                      |
+| ------------------ | -------------------------------------------- | :------: | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `maxDelayDuration` | `number`                                     |          | `1000`  | Set the amount of time to delay the loading spinner before it renders.&#xA;The default value of 1000ms has been agreed upon by the design team for all usecases. |
```
emmenko, line 44: I think you need to re-run the readme generator

packages/components/loading-spinner/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools-uikit/loading-spinner",
-  "description": "A spinner animation to indicate loading state (e.g pending requests) for the user.",
+  "description": "A spinner animation to indicate loading state (e.g pending requests) for the user. By default, there is a 1 second delay between when the component mounts and when it renders.",
```
emmenko, line 3: Hmm I don't think the "1s delay" info needs to be mentioned here. To provide additional information in the readme there is an "additional description" MD file that can be used for that.

tsconfig.json
emmenko: Same here, revert this please

packages/components/loading-spinner/src/loading-spinner.tsx
```diff
@@ -27,6 +27,11 @@ const sizePerScale = {
 const positionOrigin = '20px';
 
 type TLoadingSpinnerProps = {
+  /**
+   * Set the amount of time to delay the loading spinner before it renders.
+   * The default value of is 1000ms.
+   */
+  maxDelayDuration: number;
```
emmenko, line 34: This should be optional.

```suggestion
  maxDelayDuration?: number;
```

emmenko (approved):
Awesome, let's go! 🚀

PR #2019, opened 2021-11-15
refactor(data-table): Migrate to TS

Rhotimee:
#### Summary
Migrate Data Table to typescript


packages/components/data-table/src/cell.styles.tsx
```diff
@@ -2,8 +2,9 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
 import AccessibleButton from '@commercetools-uikit/accessible-button';
+import type { TDataCell } from './cell';
 
-const getPaddingStyle = (props) => {
+const getPaddingStyle = (props: Pick<TDataCell, 'isCondensed'>) => {
```
emmenko, line 7: Nit: I think this should be using `TCellInner` as props.

packages/components/data-table/src/cell.styles.tsx
```diff
@@ -13,7 +14,9 @@ const getPaddingStyle = (props) => {
   `;
 };
 
-const getHorizontalAlignmentStyle = (props) => {
+const getHorizontalAlignmentStyle = (
+  props: Pick<TDataCell, 'horizontalCellAlignment'>
```
emmenko, line 18: Same here

packages/components/data-table/src/cell.styles.tsx
```diff
@@ -32,7 +35,9 @@ const getHorizontalAlignmentStyle = (props) => {
   `;
 };
 
-const getVerticalAlignmentStyle = (props) => {
+const getVerticalAlignmentStyle = (
+  props: Pick<TDataCell, 'verticalCellAlignment'>
```
emmenko, line 39: Same here

packages/components/data-table/src/cell.styles.tsx
```diff
@@ -48,7 +53,7 @@ const getVerticalAlignmentStyle = (props) => {
   `;
 };
 
-const getTruncatedStyle = (props) => {
+const getTruncatedStyle = (props: Pick<TDataCell, 'isTruncated'>) => {
```
emmenko, line 56: Same here

packages/components/data-table/src/cell.tsx
```diff
@@ -1,18 +1,53 @@
...
+};
+
+const DataCell = (props: TDataCell) => {
+  if (props.shouldRenderCollapseButton) {
+    warning(
+      typeof props.handleRowCollapseClick === 'function',
+      'DataTable: "handleRowCollapseClick" is required when shouldRenderCollapseButton is true.'
+    );
+    warning(
+      props.isRowCollapsed,
```
emmenko, line 42:
```suggestion
      typeof props.isRowCollapsed === 'boolean',
```

otherwise it won't work when passing `false`.

packages/components/data-table/src/cell.tsx
```diff
@@ -35,7 +70,7 @@ const DataCell = (props) => {
         <RowExpandCollapseButton
           label="Expand/Collapse Row"
           onClick={(event) => {
-            props.handleRowCollapseClick();
+            props.handleRowCollapseClick && props.handleRowCollapseClick();
```
emmenko, line 73:
```suggestion
            if (props.handleRowCollapseClick) props.handleRowCollapseClick();
```

packages/components/data-table/src/data-row.tsx
```diff
@@ -52,7 +92,7 @@ const DataRow = (props) => {
               ? () => props.onRowClick?.(props.row, props.rowIndex, column.key)
               : undefined
           }
-          shouldClipContent={props.shouldClipContent}
+          // shouldClipContent={props.shouldClipContent}
```
emmenko, line 95: ? 🤔

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
```diff
@@ -1,20 +1,74 @@
...
+  columnBeingResized: undefined;
+  hasBeenResized: true;
+} & TState;
+
+type TRegisterColumnMeasurements = {
+  sizes: TRenderedColumnMeasurements[];
+} & TState;
+
+type TAction =
+  | { type: 'reset'; payload?: TPayload }
```
emmenko, line 47:
```suggestion
  | { type: 'reset' }
```

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
```diff
@@ -1,20 +1,74 @@
...
+  | { type: 'registerColumnMeasurements'; payload: TRegisterColumnMeasurements }
+  | { type: 'startResizing'; payload: TStartResizingPayload }
+  | { type: 'finishResizing'; payload?: TFinishResizingPayload }
+  | { type: undefined };
```
emmenko, line 51: Hmm this looks weird. Why do we need this?

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
```diff
@@ -1,20 +1,74 @@
...
+
+type TRegisterColumnMeasurements = {
+  sizes: TRenderedColumnMeasurements[];
+} & TState;
+
+type TAction =
+  | { type: 'reset'; payload?: TPayload }
+  | { type: 'registerColumnMeasurements'; payload: TRegisterColumnMeasurements }
+  | { type: 'startResizing'; payload: TStartResizingPayload }
+  | { type: 'finishResizing'; payload?: TFinishResizingPayload }
```
emmenko, line 50: It does not seem to my like the payload is actually used.

```suggestion
  | { type: 'finishResizing'; }
```

emmenko (commented):
Thanks!

Looks good in general. I left some comments to clean up some more things.

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
Rhotimee: We need it to properly type the action parameter in the reducer function.

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
Rhotimee: you're right. This is not used.

packages/components/data-table/src/data-row.tsx
Rhotimee: This isn't used in the implementation of the Row component. I guess the author forgot it there. 
I've removed it.

packages/components/data-table/src/use-manual-column-resizing-reducer.ts
emmenko: I'm not following. Having an action like `type: undefined` does not make much sense.

What is the error you get without this?

Rhotimee: I get this
image

emmenko: Ah yes. We don't need to throw an error. For the `default` case we should simply do `return state;`.

Rhotimee: Okay, that makes sense. 
Updated.

emmenko (approved):
Thanks 💯

emmenko:
Hi @Duaard ,

thanks for reaching out. You're right, we should make the type more flexible.

I think we can either defined the value as `unknown` or allow to pass a generic type to `DataTable`.

PR #2024, opened 2021-11-19
refactor(data-table): Make TRow more flexible

### LINKED ISSUE #2023: Do not assume the signature of `TRow` type (`DataTable`)
See https://github.com/commercetools/ui-kit/pull/2019#issuecomment-971681606

Follow up of #2019

Rhotimee:
#### Summary
Enable type TRow to be more flexible. 

fixes #2023


packages/components/data-table/src/data-table.tsx
```diff
@@ -61,9 +61,9 @@ const defaultProps: Pick<
   itemRenderer: (row, column) => row[column.key],
 };
 
+type TAny = string | number | boolean | undefined | null;
 export type TRow = {
-  id: string;
-  [key: string]: string;
```
emmenko, line 66: I would simply use `[key: string]: unknown`.

emmenko (changes requested):
Alternative is to pass a generic type.

```ts
type MyData = {
  id: string,
  name: string,
  roles: string[],
}


<DataTable<MyData>
  // ...
/>
```

The signature would be something like:

```ts
type TRow = {
  id: string
}

const DataTable = <Row extends TRow>(props: TProps) => {
  // ...
  return (
    // ....
    <DataRow<Row>
      // ...
    />
  )
}
```

.changeset/popular-mice-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/data-table': patch
+---
+
+Update type defination for TRow
```
emmenko, line 5:
````suggestion
Allow to pass a generic type for table rows shape.

Example:

```ts
type MyData = {
  id: string,
  name: string,
  roles: string[],
}


<DataTable<MyData>
  // ...
/>
```

The shape should have at least an `id` field.
````

File: packages/components/data-table/src/data-row.tsx [hunk 3]
```diff
@@ -8,11 +8,11 @@ import ColumnResizingContext from './column-resizing-context';
 export type TDataRow = {
   row: TRow;
   rowIndex: number;
-  columns: TColumn[];
+  columns: TColumn<TRow>[];
```
emmenko, line 11: We should pass the generic type to `TDataRow`

packages/components/data-table/src/data-table.tsx
```diff
@@ -21,11 +21,17 @@ import DataRow from './data-row';
 import useManualColumnResizing from './use-manual-column-resizing-reducer';
 import ColumnResizingContext from './column-resizing-context';
 
+export type TRow = {
+  id: string;
+  [key: string]: string;
+};
```
emmenko, line 27: We don't nee this anymore. The generic signature of `DataTable` should be something like

```ts
interface TRow {
  id: string
}

const DataTable = <Row extends TRow = TRow>(props: TProps<Row>)
```

All types accepting the generic argument should also implement the same signature. For example:

```ts
type TProps<Row extends TRow = TRow> = {
  // ...
}
```

packages/components/data-table/src/data-table.tsx
```diff
@@ -156,21 +157,21 @@ export type TColumn = {
   shouldIgnoreRowClick?: boolean;
 };
 
-export type TDataTable = {
+export type TDataTable<RowData> = {
```
emmenko, line 160: See my other comment above

```suggestion
export type TDataTableProps<Row extends TRow = TRow> = {
```

packages/components/data-table/src/data-table.tsx
```diff
@@ -259,7 +260,7 @@ export type TDataTable = {
   sortDirection?: 'desc' | 'asc';
 };
 
-const DataTable = (props: TDataTable) => {
+const DataTable = (props: TDataTable<TRow>) => {
```
emmenko, line 263:
```suggestion
const DataTable = <Row extends TRow = TRow>(props: TDataTableProps<Row>) => {
```

emmenko (commented):
Still not what I was looking for. See comments

packages/components/data-table/src/data-table.tsx
```diff
@@ -156,13 +157,13 @@ export type TColumn = {
   shouldIgnoreRowClick?: boolean;
 };
 
-export type TDataTable = {
+export type TDataTable<Row extends TRow = TRow> = {
```
emmenko, line 160:
```suggestion
export type TDataTableProps<Row extends TRow = TRow> = {
```

packages/components/data-table/src/data-table.tsx
```diff
@@ -259,7 +260,7 @@ export type TDataTable = {
   sortDirection?: 'desc' | 'asc';
 };
 
-const DataTable = (props: TDataTable) => {
+const DataTable = <Row extends TRow = TRow>(props: TDataTable<Row>) => {
```
emmenko, line 263:
```suggestion
const DataTable = <Row extends TRow = TRow>(props: TDataTableProps<Row>) => {
```

emmenko (approved):
Looks good now 👌

PR #2028, opened 2021-11-25
Cleanup and missing deps

emmenko:
In addition to that, I fixed a bit the Vercel setup to properly cache dependencies between builds. Now Vercel builds should be faster.


Rhotimee (approved):
💯

PR #2029, opened 2021-11-25
refactor(select-utils): Migrate to typescript

Rhotimee:
#### Summary
Migrate select utils to typescript


packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
```diff
@@ -1,11 +1,19 @@
-import PropTypes from 'prop-types';
+import { CSSProperties, LegacyRef } from 'react';
```
emmenko, line 1: When importing types, prefer to use `import type`

```suggestion
import type { CSSProperties, LegacyRef } from 'react';
```

emmenko, line 1: When importing types, prefer to use `import type`

```suggestion
import type { CSSProperties, LegacyRef } from 'react';
```

packages/components/inputs/select-utils/src/wrapper-with-icon/wrapper-with-icon.tsx
```diff
@@ -1,16 +1,29 @@
...
 import { customProperties } from '@commercetools-uikit/design-system';
-import { components as defaultComponents } from 'react-select';
+import {
+  components as defaultComponents,
+  SingleValueProps,
+  PlaceholderProps,
+} from 'react-select';
 
-const getDefaultComponent = (type) => {
+const getDefaultComponent = (type: string) => {
```
emmenko, line 11:
```suggestion
const getDefaultComponent = (type: TWrapperWithIconProps['type']) => {
```

.changeset/tough-eagles-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/select-utils': patch
+---
+
+Migrate to typescript
```
emmenko, line 5: Let's also mention that `react-select` got updated to v5

packages/components/inputs/select-utils/package.json
```diff
@@ -31,7 +31,7 @@
     "@emotion/styled": "^11.3.0",
     "lodash": "4.17.21",
     "prop-types": "15.7.2",
-    "react-select": "4.3.1"
+    "react-select": "^5.2.1"
```
emmenko, line 34: We need to update all other packages to v5 as well.

PS: we can maybe extract the upgrade to a separate PR.

Rhotimee:
@emmenko, I am currently working on upgrading react-select in this [PR](https://github.com/commercetools/ui-kit/pull/1997).  That will automatically fix the issues.

.changeset/tough-eagles-visit.md
Rhotimee: This is already done in the react-select PR

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -0,0 +1,472 @@
...
+  borderColor?: string;
+  boxShadow?: string;
+  pointerEvents?: string;
+};
+
+type TState = {
+  isFocused?: boolean;
+  isDisabled?: boolean;
+  isSelected?: boolean;
+};
```
emmenko, line 44: Is there a way to import those from react-select? For example https://github.com/JedWatson/react-select/blob/4e06d22e47bbd82a4f9f67452c7f16787c4ce529/packages/react-select/src/styles.ts#L52-L79

emmenko (commented):
Overall looks good. Just have a question for the css styles.

packages/components/inputs/select-utils/src/create-select-styles.ts
Rhotimee: For TBase, I tried using `CSSObjectWithLabel` but it caused the build to fail and I got this error.
`The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.` 

For TState, We can use `ClassNamesState` but the implementation is not explicit. So, I'm not sure it is a good idea to use it.

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
```diff
@@ -1,11 +1,21 @@
...
+type TInnerProps = {
+  ref: LegacyRef<HTMLButtonElement>;
+} & JSX.IntrinsicElements['button'];
+
+type TClearIndicator = {
+  getStyles: (s: string, props: unknown) => CSSProperties;
+  innerProps: TInnerProps;
+} & ClearIndicatorProps;
+
+const ClearIndicator = (props: TClearIndicator) => {
```
emmenko, line 18: According to this example, we should be able to use `ClearIndicatorProps` directly. Can you double check?

https://github.com/JedWatson/react-select/blob/4e06d22e47bbd82a4f9f67452c7f16787c4ce529/docs/examples/CustomClearIndicator.tsx#L8-L13

emmenko (approved):
One small comment, otherwise I think this is good to go for now. Thanks!

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
Rhotimee: Yes, we can use it directly. However, we need to update the ref type to use HTMLButtonElement since we are using a button instead of a div.
The `getStyles`  type def has been cleaned. 
Thanks.

PR #2036, opened 2021-12-02
refactor(data-table): Allow string as width type

Rhotimee:
Regression of #2019 

#### Summary
- Allow string as width type
- Cleanup THeaderCellWrapper and TGetColumnsLayoutInfo


packages/components/data-table/src/data-table.tsx
```diff
@@ -93,7 +88,7 @@ export type TColumn<Row extends TRow = TRow> = {
    *
    * @@defaultValue@@: auto
    */
-  width?: number;
+  width?: number | string;
```
emmenko, line 91: Before the prop type was only `string`. Do we also need `number`? 🤔

packages/components/data-table/src/header-cell.tsx
```diff
@@ -98,7 +100,7 @@ export type THeaderCell = {
   sortedBy?: string;
   children: ReactNode;
   columnKey: string;
-  columnWidth?: number;
+  columnWidth?: number | string;
```
emmenko, line 103: Same here. Can you double check?

emmenko (commented):
Thanks

packages/components/data-table/src/data-table.tsx
Rhotimee: type number for width was need because some calculations was done with the width in `use-manual-column-resizing-reducer.ts`. It has been updated now though.

emmenko (approved):
💯 Thanks

PR #2037, opened 2021-12-02
refactor(icons): use isomorphic-dompurify for SSR support

emmenko:

PR #2038, opened 2021-12-02
fix(icons): use better safeguard for SSR

emmenko:
The attempt to fix the issue with `DOMPurify` and SSR (#2037 ) didn't really work.

We can simply have a better safeguard check, if the DOM is available.


emmenko:
Another try

PR #2053, opened 2022-01-05
refactor(async-select-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate async-select-input to typescript


packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -226,155 +91,307 @@ AsyncSelectInput.propTypes = {
...
    */
-  iconLeft: PropTypes.node,
+  iconLeft?: ReactNode;
+};
+
+type TDefaultProps = {
+  value: TAsyncSelectInputProps['value'];
+  isSearchable: TAsyncSelectInputProps['isSearchable'];
+  menuPortalZIndex: TAsyncSelectInputProps['menuPortalZIndex'];
+};
```
emmenko, line 235: We usually use `Pick<>`.

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -226,155 +91,307 @@ AsyncSelectInput.propTypes = {
...
+                }
+              : undefined
+          }
+          onChange={(value, info) => {
+            let newValue = value;
+            if (props.isMulti && !newValue) {
+              newValue = [];
+            }
+
+            props.onChange &&
```
emmenko, line 358: This should not be necessary, as the prop is required.

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -1,12 +1,18 @@
...
 import { useIntl } from 'react-intl';
 import isEmpty from 'lodash/isEmpty';
 import { useTheme } from '@emotion/react';
-import { components as defaultComponents } from 'react-select';
+import {
+  ActionMeta,
+  components as defaultComponents,
+  type GroupBase,
+  type OptionsOrGroups,
+  type Props,
```
emmenko, line 10: Let's rename it to `ReactSelectProps`, so that it's clearer where this is coming from when used in the component props (and also what is shown in the README).

```suggestion
  type ReactSelectProps,
```

emmenko: The only small downside with selecting props from the type is that in the README the actual value is not clear.

I think it would help that we link to the react select docs (https://react-select.com/props) for all those props. The link can be included in the docs block comment for each prop (you can also use markdown syntax in the comments).

emmenko (changes requested):
Thanks, looks good.

However, I would like to change a bit how we define the react select props, mostly for making things more clear from a documentation point of view. See my comments.

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -90,84 +90,101 @@ type TAsyncSelectInputProps = {
   // See https://react-select.com/props#select-props
   /**
    * Aria label (for assistive tech)
+   * [Props from React select was used](https://react-select.com/props)
```
emmenko, line 93: Let's add a blank line (same for the other props).

```suggestion
   * Aria label (for assistive tech)
   * <br>
   * [Props from React select was used](https://react-select.com/props)
```

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -225,156 +90,328 @@ AsyncSelectInput.propTypes = {
...
+  filterOption?: ReactSelectProps['filterOption'];
   // This forwarded as react-select's "inputId"
   /**
    * The id of the search input
+   * [Props from React select was used](https://react-select.com/props)
    */
-  id: PropTypes.string,
-  inputValue: PropTypes.string,
+  id?: ReactSelectProps['inputId'];
+  inputValue?: ReactSelectProps['inputValue'];
```
emmenko, line 127: Let's add a js doc comment here as well, with the link.

emmenko (commented):
Thanks, much better now!

emmenko (approved):

PR #2054, opened 2022-01-06
refactor(creatable-select-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate create-select-input to typescript


packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -223,167 +93,164 @@ CreatableSelectInput.propTypes = {
...
   /**
    * Dom element to portal the select menu to
    */
-  menuPortalTarget: PropTypes.instanceOf(SafeHTMLElement),
+  menuPortalTarget?: Props['menuPortalTarget'];
   /**
    * z-index value for the menu portal
    */
-  menuPortalZIndex: PropTypes.number.isRequired,
+  menuPortalZIndex?: number;
```
ddouglasz, line 159: I think you missed this: prop type was initially required but it's now optional here.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -25,192 +34,62 @@ const customizedComponents = {
...
+type TEvent = {
+  target: {
+    name?: string;
+    value?: unknown;
+  };
+  persist?: () => void;
+};
 
-CreatableSelectInput.propTypes = {
+type TCreatableSelectInput = {
```
emmenko, line 56:
```suggestion
type TCreatableSelectInputProps = {
```

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -222,168 +101,213 @@ CreatableSelectInput.propTypes = {
...
    */
-  value: (props, ...rest) =>
-    props.isMulti
-      ? PropTypes.arrayOf(
-          PropTypes.shape({ value: PropTypes.string.isRequired })
-        )(props, ...rest)
-      : PropTypes.shape({ value: PropTypes.string.isRequired })(props, ...rest),
+  value: PropsValue<TValue>;
 
   // Creatable props
```
emmenko, line 275: All these props here are also from react-select right? https://react-select.com/props#creatable-props

Can you just reference them as we do with other props, e.g. `ReactSelectProps['allowCreateWhileLoading']`?

Also provide the link to the react-select docs.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -397,6 +321,158 @@ CreatableSelectInput.defaultProps = {
...
+
+CreatableSelectInput.displayName = 'CreatableSelectInput';
+
+// Both "true" and an empty array [] represent a touched state. The Boolean
+// conveniently handles both cases
+CreatableSelectInput.isTouched = (touched: unknown) => Boolean(touched);
+
+CreatableSelectInput.defaultProps = defaultProps;
+
 addStaticFields(CreatableSelectInput, {
```
emmenko, line 476: Can you check how the type declarations are after the package is built?

We might need to manually expose some of the static fields as types, so that TS can recognize them.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
Rhotimee: They are currently referenced from react-select, and links to the docs are provided.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
Rhotimee: I don't think we need to manually expose them.  See below how they are built. 
image

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
Rhotimee: updated.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
emmenko: Yes but the spreaded components are not exposed. Anyway, we can iterate on this as a follow up maybe.

For now it's fine like this.

emmenko (approved):

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
Rhotimee: Okay 👍🏽

PR #2059, opened 2022-01-12
refactor(money-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate money input to typescript


ddouglasz (approved):
LGTM

packages/components/inputs/money-input/README.md
```diff
@@ -49,29 +49,58 @@ export default Example;
...
+### Signature `value`
+
+```ts
+{
+  amount: string;
+  currencyCode: TCurrencyCode;
+}
+```
+
+### Signature `currencies`
```
CarlosCortizasCT, line 85: This signature section is blank.
The property is already explained in the table above so maybe we can remove this sub-heading.
What do you think?

packages/components/inputs/money-input/src/money-input.tsx
```diff
@@ -319,17 +360,149 @@ const formatAmount = (rawAmount, currencyCode, locale) => {
...
+   */
+  hasWarning?: boolean;
+  /**
+   * Shows high precision badge in case current value uses high precision.
+   */
+  hasHighPrecisionBadge?: boolean;
+  /**
+   * Horizontal size limit of the input fields.
+   */
+  horizontalConstraint?:
```
CarlosCortizasCT, line 464: Maybe the available values should start in 1?
I'm saying because of what I read [here](https://github.com/commercetools/ui-kit/pull/2062#discussion_r783883512).

Rhotimee: Oh.. No. It doesn't have to start from 1. 
It just needs to be consisted with the previous proptTypes. 
Btw, I actually made a mistake with the comment [here](https://github.com/commercetools/ui-kit/pull/2062#discussion_r783883512).

packages/components/inputs/money-input/README.md
Rhotimee: I just realised that types that are an array of values don't appear in the signature section. 
I'm not sure why. 
@emmenko, do you have ideas why that is happening?

emmenko: Did you re-run the readme generator?

Rhotimee: Yes, I did. 
This is also an issue in some other parts of the app.
Take a look at [data-table](https://github.com/commercetools/ui-kit/tree/main/packages/components/data-table) also

emmenko: Hmm I see...well actually `Row` is a generic type, that's why the signature is empty.

Nevertheless, we might want to adjust the readme generator to trying showing more useful information. We can also avoid linking empty signatures. The field description should already provide the basic information.

As for the `currencies` field, it's a `string[]` so I don't know why it thinks there is a need for a signature. We would need to debug that.

Rhotimee: Sure. I will create an issue for that.

packages/components/inputs/money-input/README.md
```diff
@@ -49,29 +49,58 @@ export default Example;
...
```
emmenko, line 85: Did you rebase and re-run the generator?

packages/components/inputs/money-input/src/money-input.styles.ts
```diff
@@ -16,7 +16,19 @@ const getCurrencyLabelStyles = () => css`
   box-sizing: border-box;
 `;
 
-const getAmountInputStyles = (props) => [
+type TInputProps = {
```
emmenko, line 19: I think we can import them from the component file.

packages/components/inputs/money-input/src/money-input.tsx
```diff
@@ -34,42 +35,63 @@ const TooltipWrapper = styled.div`
...
 
-const getPortalId = (id) => `portal-${id}`;
-const getPortalNode = (id) => document.querySelector(`#${getPortalId(id)}`);
+const getPortalId = (id?: string) => `portal-${id}`;
+const getPortalNode = (id?: string) =>
+  document.querySelector(`#${getPortalId(id)}`);
 
-const Portal = (props) => {
+type TLabel = {
+  id?: string;
```
emmenko, line 43: I think an `id` is always defined, right?

emmenko (approved):
Looks good, just one question/comment.

PR #2063, opened 2022-01-14
refactor(number-field): Migrate to typescript

Rhotimee:
#### Summary
Migrate number field to typescript


ddouglasz (approved):
LGTM

packages/components/fields/number-field/src/number-field.tsx
```diff
@@ -0,0 +1,259 @@
...
+   */
+  value: string | number;
+  /**
+   * Used as HTML `autocomplete` of the input component. property
+   */
+  autoComplete?: string;
+  /**
+   * Called with an event containing the new value. Required when input is not read only. Parent should pass it back as value.
+   * <br />
+   * Signature: `(event) => void`
```
emmenko, line 94: You can remove these `Signature` comments. With TS now they are provided by the generator.

emmenko (approved):

PR #2067, opened 2022-01-17
refactor(date-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate date input to typescript


packages/components/inputs/date-input/src/date-input.tsx
```diff
@@ -165,20 +256,20 @@ const DateInput = (props) => {
                   onMouseEnter: () => {
                     // we remove the highlight so that the user can use the
                     // arrow keys to move the cursor when hovering
-                    if (isOpen) setDownshiftHighlightedIndex(null);
+                    if (isOpen) setDownshiftHighlightedIndex(0);
```
emmenko, line 259: Is this correct? I think this should stay as `null`. See https://github.com/downshift-js/downshift/blob/1c7698080805e05f5180dd223cbc17932a85fd42/src/downshift.js#L80

packages/components/inputs/date-input/src/date-input.tsx
```diff
@@ -282,7 +376,7 @@ const DateInput = (props) => {
                           ),
                           item,
                           onMouseOut: () => {
-                            setDownshiftHighlightedIndex(null);
+                            setDownshiftHighlightedIndex(0);
```
emmenko, line 379: Same here.

packages/components/inputs/date-input/src/date-input.tsx
```diff
@@ -304,81 +398,6 @@ const DateInput = (props) => {
...
-  /**
-   * A minimum selectable date. Must either be an empty string or a date formatted as "YYYY-MM-DD".
-   */
-  minValue: PropTypes.string,
-  /**
-   * A maximum selectable date. Must either be an empty string or a date formatted as "YYYY-MM-DD".
-   */
-  maxValue: PropTypes.string,
-};
+DateInput.isEmpty = (value: string) => value === '';
```
emmenko, line 401: I wonder if value can be optional.

```suggestion
DateInput.isEmpty = (value?: string) => value === '';
```

emmenko (changes requested):
I found a small inconsistency, otherwise looks good.

emmenko (approved):

PR #2068, opened 2022-01-19
fix: random infinite loop initializing project

CarlosCortizasCT:
#### Summary

Infinite loop when running initialization command in some cases

## Description

It seems that sometimes there is an infinite loop when running the initialization command on a fresh cloned repository.
Running `yarn` from the root directory have this [post-install](https://github.com/commercetools/ui-kit/blob/main/scripts/postinstall.sh) process where it tries to install isolated dependencies for the documentation tool (Storybook) but it does not end. After it install those dependencies, the process starts over again.


emmenko (approved):
Thanks!

PR #2079, opened 2022-01-28
refactor(search-select-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate search select input to typescript


packages/components/inputs/search-select-input/src/search-select-input-open.visualroute.js
```diff
@@ -1,4 +1,4 @@
-import SearchSelectInput from './search-select-input';
+import { SearchSelectInput } from '@commercetools-frontend/ui-kit';
```
CarlosCortizasCT, line 1: Just wondering if we have some rule about _imports_. I ask it because it seems natural to me to rely on relative paths when the file is in fact part the the component (package) and the absolute path (module resolution) when referencing the component from the outside.

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -10,235 +10,282 @@ import {
...
+        <CustomSelectInputOption
+          {...optionInnerProps}
+          optionType={optionType}
+          optionInnerProps={optionInnerProps}
+        />
+      ),
+      ...props.components,
+      DropdownIndicator: SearchIconDropdownIndicator,
+    };
+  }, [props.components, optionType]);
```
CarlosCortizasCT, line 278: nit: Maybe using the condensed format?

```suggestion
const components = useMemo(() => ({  
  Option: (optionInnerProps: TOptionInnerProps) => (
    <CustomSelectInputOption
      {...optionInnerProps}
      optionType={optionType}
      optionInnerProps={optionInnerProps}
    />
  ),
  ...props.components,
  DropdownIndicator: SearchIconDropdownIndicator,
}), [props.components, optionType]);
```

packages/components/inputs/search-select-input/src/search-select-input-open.visualroute.js
Rhotimee: I have the same sentiment. 
However, If we use relative paths and the file we are importing is in typescript, the visualroute/visualspec file cannot resolve the imported package.

CarlosCortizasCT: Ah! Ok, I see.
Thanks for the explanation.

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -10,235 +10,282 @@ import {
...
+
+type TOptionInnerProps = {
+  data?: TOptionInnerPropsData;
+} & OptionProps;
+
+const SearchSelectInput = (props: TSearchSelectInputProps) => {
+  const intl = useIntl();
+  const noOptionsMessage =
+    props.noOptionsMessage ||
+    (() => intl.formatMessage(messages.noOptionsMessage));
```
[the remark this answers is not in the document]
Rhotimee: Because the type is what is required from [react-select](https://react-select.com/props#select-props)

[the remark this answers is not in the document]
CarlosCortizasCT: Oh, I see. 
But it is the same for the `loadingMessage` prop so it also needs to be a function, right?
image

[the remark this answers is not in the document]
Rhotimee: True. However, we are using AsyncSelectInput Component in SearchSelectInput. 
AsyncSelectInput is the component using react-select directly. And in AsyncSelectInput, we have a [loadingMessage function](https://github.com/commercetools/ui-kit/blob/4d2aa4d364a46605031600b9f0fc972942044045/packages/components/inputs/async-select-input/src/async-select-input.tsx#L301-L306) which is passed to react-select.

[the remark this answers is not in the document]
CarlosCortizasCT: Thanks for the explanation.
I was confused because of having different types for messages provided to the `AsyncSelectInput`.

ddouglasz (approved):
Looks good.

CarlosCortizasCT (approved):
LGTM

.changeset/eighty-points-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/async-select-input': patch
+---
+
+Update type defination and Include react select link
```
emmenko, line 5:
```suggestion
Update type definition and include react select links
```

emmenko (approved):

Issue #2047, opened 2021-12-10
Prop passing for the AsyncSelectInput


parthikrb:
### Request Overview
In Products, we have a [scenario](https://jira.commercetools.com/browse/PCM-2506) where we want to keep the select options OPEN even after selecting an option (for better UX).
On investigating the existing AsyncSelectInput component, we noticed that `AsyncSelect` from `react-select` is abstracted behind the implementation

We also noticed that the [react-select](https://react-select.com/props) library provides a prop called `closeMenuOnSelect` to keep the options open, but unfortunately UI-kit component is not exposing the prop.

So, as part of this request, we want to implement the prop passing for `closeMenuOnSelect`.

mmaltsev-ct:
Hi @parthikrb, thanks for the request. Also, thank you for the suggested solution. Is it still relevant or did you find a workaround at the moment?

It sounds like a good first contribution topic. Do you think you'll be comfortable creating a PR for this feature request? Otherwise, our team can help with implementing this feature.

On the other note, if we are planning to expose the `closeMenuSelect` property anyway, let's do it for all _react-select_-based components (`AsyncCreatableSelectInput`, `AsyncSelectInput`, `CreatableSelectInput`, `SearchSelectInput`, `SelectInput`).

parthikrb:
@mmaltsev-ct, It is still relevant to our feature request as there is no workaround to keep the menu open.

For the second question, yes, I would like to contribute to this and will create a PR soon.

Thanks.

PR #2084, opened 2022-01-31
feat(select-inputs): add closeMenuOnSelect props to select-inputs

parthikrb:
Summary
The aim of this PR is to accommodate this [request](https://github.com/commercetools/ui-kit/issues/2047) 

As part of this implementation, the closeMenuOnSelect prop has been exposed on the following components,
- AsyncCreatableSelectInput
- AsyncSelectInput
- CreatableSelectInput
- SelectInput

And the respective storybook component was also updated.


CarlosCortizasCT:
Hi @parthikrb,

Regarding the issue with the unit testing, after taking a closer look it seems to be related with how focus in managed in `js-dom` context when simulating events with RTL and also that testing the `react-select` is usually tricky.

If you check the unit test expected behaviour in Storybook you can see that it's working ok.
The difference is that, in the browser, after selecting an option, the focus is kept in the select input element but, in the unit tests context, the focus is moved away to the body, which causes `react-select` to close the menu automatically (it does not check the `closeMenuOnSelect` prop as that focus event is simulating like a click outside the select component so it must close itself).
I've been trying to find a different way of selecting the new element so it does not move the focus away from the input but I don't find any feasible solution.

Because of this and actually because these new tests are testing the `react-select` component instead of our code (we just pass through the prop to the library) I think it would be better to remove those new tests.

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -165,6 +165,10 @@ export type TSearchSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   menuShouldBlockScroll?: ReactSelectAsyncProps['menuShouldBlockScroll'];
+  /**
+   * whether the menu should be closed when a value is selected
```
Rhotimee, line 169: add react-select link here

Rhotimee:
```
<br>
[Props from React select was used](https://react-select.com/props)
```

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -193,6 +193,12 @@ type TAsyncSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   menuShouldBlockScroll?: ReactSelectAsyncProps['menuShouldBlockScroll'];
+  /**
+   * When set to false, the select menu will not be closed after an option is selected. By default, it is set to true.
+   * <br>
+   * [Props from React select was used](https://react-select.com/props)
+   */
+  closeMenuOnSelect: ReactSelectProps['closeMenuOnSelect'];
```
Rhotimee, line 201: I don't think this should be required. 
```suggestion
  closeMenuOnSelect?: ReactSelectProps['closeMenuOnSelect'];
```

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -200,6 +200,10 @@ type TCreatableSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props#creatable-props)
    */
   menuShouldBlockScroll?: ReactSelectCreatableProps['menuShouldBlockScroll'];
+  /**
+   * When set to false, the select menu will not be closed after an option is selected. By default, it is set to true.
+   */
+  closeMenuOnSelect: ReactSelectCreatableProps['closeMenuOnSelect'];
```
Rhotimee, line 206: This should be an optional prop

packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.js
```diff
@@ -372,6 +373,10 @@ AsyncCreatableSelectInput.propTypes = {
    * If cacheOptions is truthy, then the loaded data will be cached. The cache will remain until cacheOptions changes value.
    */
   cacheOptions: PropTypes.any,
+  /**
+   * When set to false, the select menu will not be closed after an option is selected. By default, it is set to true.
+   */
+  closeMenuOnSelect: PropTypes.bool,
```
CarlosCortizasCT, line 379: I think it would be great to also add the default value to the `defaultProps` attribute of the component:
```
AsyncCreatableSelectInput.defaultProps = {
  // Using "null" will ensure that the currently selected value disappears in
  // case "undefined" gets passed as the next value
  value: null,
  isSearchable: true,
  menuPortalZIndex: 1,
  closeMenuOnSelect: true,
};
```

packages/components/inputs/async-creatable-select-input/README.md
```diff
@@ -83,6 +83,7 @@ export default Example;
...
 | `menuShouldBlockScroll`           | `bool`                                                                                             |          |         | whether the menu should block scroll while open                                                                                                                                                                                                                                             |
+| `closeMenuOnSelect`               | `bool`                                                                                             |          | `true`  | whether the menu should be closed on option select open                                                                                                                                                                                                                                     |
```
CarlosCortizasCT, line 86: Did you edit this file by hand?
There is this `generate-readmes` yarn command that handles the generation of this files automatically so its easier to handle and descriptions are the same in code comments and in these files.

packages/components/inputs/async-select-input/src/async-select-input.spec.js
```diff
@@ -143,6 +143,7 @@ describe('in single mode', () => {
       // list should closed and not visible
       expect(queryByText('Banana')).not.toBeInTheDocument();
     });
+
```
CarlosCortizasCT, line 146: Maybe we could avoid adding a non-code change to the PR.

Rhotimee (changes requested):
Thanks for the PR.
It generally looks ok. Just some minor feedback. 

It'll be nice to write some tests for the addition also.

packages/components/inputs/async-select-input/README.md
```diff
@@ -82,6 +82,7 @@ export default Example;
...
 | `menuShouldBlockScroll`  | `AsyncProps['menuShouldBlockScroll']`                                                                 |          |         | whether the menu should block scroll while open&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                             |
+| `closeMenuOnSelect`      | `boolean`                                                                                             |          | `true`  | whether the menu should be closed on option select&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                          |
```
CarlosCortizasCT, line 85: Please, better use the `generate-readmes` yarn command to update this file.

packages/components/inputs/creatable-select-input/README.md
```diff
@@ -77,6 +77,7 @@ export default Example;
...
 | `menuShouldBlockScroll`   | `ReactSelectProps['menuShouldBlockScroll']`                                                           |          |         | whether the menu should block scroll while open&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                             |
+| `closeMenuOnSelect`                | `bool`                                                                                             |          | `true`  | whether the menu should be closed on option select open                                                                                                                                                                                                                                             |
```
CarlosCortizasCT, line 80: Please, better use the `generate-readmes` yarn command to update this file.

packages/components/inputs/search-select-input/README.md
```diff
@@ -86,6 +86,7 @@ export default Example;
...
 | `menuShouldBlockScroll`  | `AsyncProps['menuShouldBlockScroll']`                                                                 |          |         | whether the menu should block scroll while open&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                             |
+| `closeMenuOnSelect`      | `boolean`                                                                                             |          | `true`  | whether the menu should be closed on option select&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                          |
```
CarlosCortizasCT, line 89: Please, better use the `generate-readmes` yarn command to update this file.

packages/components/inputs/search-select-input/src/search-select-input.tsx
CarlosCortizasCT, line 169:
```suggestion
   * whether the menu should be closed after a value is selected
```

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -221,7 +221,12 @@ type TSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   menuShouldBlockScroll?: ReactSelectProps['menuShouldBlockScroll'];
-  // menuShouldScrollIntoView: PropTypes.bool,
+  /**
+   * whether the menu should close when a value is selected
```
CarlosCortizasCT, line 225:
```suggestion
   * whether the menu should close after a value is selected
```

packages/components/inputs/async-select-input/src/async-select-input.tsx
CarlosCortizasCT: Shouldn't we mark this prop as optional?

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -362,6 +368,7 @@ const AsyncSelectInput = (props: TAsyncSelectInputProps) => {
           maxMenuHeight={props.maxMenuHeight}
           menuPortalTarget={props.menuPortalTarget}
           menuShouldBlockScroll={props.menuShouldBlockScroll}
+          closeMenuOnSelect={props.closeMenuOnSelect}
```
CarlosCortizasCT, line 371: I think it would be great to also add the default value to the `defaultProps` attribute of the component.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -391,6 +395,7 @@ const CreatableSelectInput = (props: TCreatableSelectInputProps) => {
           maxMenuHeight={props.maxMenuHeight}
           menuPortalTarget={props.menuPortalTarget}
           menuShouldBlockScroll={props.menuShouldBlockScroll}
+          closeMenuOnSelect={props.closeMenuOnSelect}
```
CarlosCortizasCT, line 398: I think it would be great to also add the default value to the `defaultProps` attribute of the component.

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
CarlosCortizasCT, line 206: I guess this prop should be declared as optional?

packages/components/inputs/search-select-input/src/search-select-input.spec.js
```diff
@@ -160,6 +160,7 @@ describe('in single mode', () => {
       // list should closed and not visible
       expect(screen.queryByText('Banana')).not.toBeInTheDocument();
     });
+
```
CarlosCortizasCT, line 163: Please, could we avoid adding non-code changes to the PR?

packages/components/inputs/select-input/src/select-input.spec.js
```diff
@@ -156,6 +156,7 @@ describe('in single mode', () => {
       // list should closed and not visible
       expect(queryByText('Banana')).not.toBeInTheDocument();
     });
+
```
CarlosCortizasCT, line 159: Please, could we avoid adding non-code changes to the PR?

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -291,6 +295,7 @@ const SearchSelectInput = (props: TSearchSelectInputProps) => {
         loadingMessage={loadingMessage}
         noOptionsMessage={noOptionsMessage}
         isSearchable={true}
+        closeMenuOnSelect={props.closeMenuOnSelect}
```
CarlosCortizasCT, line 298: I think it would be great to also add the default value to the `defaultProps` attribute of the component.

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -413,6 +418,7 @@ const SelectInput = (props: TSelectInputProps) => {
           maxMenuHeight={props.maxMenuHeight}
           menuPortalTarget={props.menuPortalTarget}
           menuShouldBlockScroll={props.menuShouldBlockScroll}
+          closeMenuOnSelect={props.closeMenuOnSelect}
```
CarlosCortizasCT, line 421: I think it would be great to also add the default value to the defaultProps attribute of the component

packages/components/inputs/select-input/README.md
```diff
@@ -83,6 +83,7 @@ export default Example;
...
 | `menuShouldBlockScroll`  | `ReactSelectProps['menuShouldBlockScroll']`                                                           |          |         | whether the menu should block scroll while open&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                                                                                                |
+| `closeMenuOnSelect`      | `boolean`                                                                                             |          | `true`  | whether the menu should be closed on option select&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                                          |
```
CarlosCortizasCT, line 86: Did you edit this file by hand?
There is this `generate-readmes` yarn command that handles the generation of this files automatically so its easier to handle and descriptions are the same in code comments and in these files.

CarlosCortizasCT (changes requested):

packages/i18n/data/core.json
```diff
@@ -1,11 +1,4 @@
 {
-  "UIKit.CalendarHeader.nextMonthTooltip": "Next month",
-  "UIKit.CalendarHeader.nextYearTooltip": "Next year",
-  "UIKit.CalendarHeader.previousMonthTooltip": "Previous month",
-  "UIKit.CalendarHeader.previousYearTooltip": "Previous year",
-  "UIKit.CalendarHeader.todayTooltip": "Today",
-  "UIKit.ClearButton.clearButtonLabel": "Clear",
-  "UIKit.CreatableSelectInput.createLabel": "Create \"{inputValue}\"",
```
emmenko, line 8: What are these changes for? 🤔

packages/components/inputs/select-input/src/select-input.tsx
emmenko: Any particular reason for having a default value? Since we forward this prop to react-select, I would just let react-select figure out the default value, as we don't want to change that behavior.

CarlosCortizasCT: My reasoning was based on that, up until now, the uses cases where we use these components were relying on the menu actually closing upon selection and, given that now we can select that behaviour, we would expect it to keep it that way by default so we don't break anything. I mean, it seems to me that our default is closing the menu upon selection.
If we don't use our own default and rely on `react-select` and the library's author decides to change its default behaviour, our own use cases would change as well but not because it was our intention but the library's author.

Anyway, that was the way I thought about this but it's ok from my side if we prefer not to use this default.

packages/i18n/data/core.json
parthikrb: It's not part of my change but came when I pulled the latest translations.

packages/components/inputs/select-input/src/select-input.tsx
parthikrb: It makes sense from @CarlosCortizasCT point of view to have a default value, however, should I keep or remove the default value? @emmenko

emmenko: I understand but we would need to potentially do this for every prop that we forward, which becomes a bit unmaintainable.

If react-select changes the default behavior, it would be a breaking change so we can then adjust things accordingly from our side.

So I would prefer to not have default values from our side for now.

packages/i18n/data/core.json
emmenko: What do you mean pulled? Which command did you run to get this?

Anyway, please revert these changes as they are not relevant with this PR. Thanks

.changeset/nine-laws-sit.md
```diff
@@ -0,0 +1,10 @@
+---
+'@commercetools-uikit/async-creatable-select-input': minor
+'@commercetools-uikit/async-select-input': minor
+'@commercetools-uikit/creatable-select-input': minor
+'@commercetools-uikit/search-select-input': minor
+'@commercetools-uikit/select-input': minor
+'@commercetools-uikit/i18n': minor
```
emmenko, line 7: Remove this line

.changeset/nine-laws-sit.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+expose closeMenuOnSelect prop on all select input components which uses react-select
```
emmenko, line 10:
```suggestion
Expose `closeMenuOnSelect` prop on all select input components. By default it closes the menu. When set to `false`, the menu stays open when options are selected.
```

packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.tsx
```diff
@@ -202,6 +202,12 @@ type TAsyncCreatableSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   menuShouldBlockScroll?: ReactSelectAsyncCreatableProps['menuShouldBlockScroll'];
+  /**
+   * whether the menu should be closed after a value is selected
```
emmenko, line 206:
```suggestion
   * Whether the menu should be closed after a value is selected. Defaults to `true`.
```

emmenko (changes requested):

CarlosCortizasCT (approved):

Rhotimee (approved):

packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
```diff
@@ -200,6 +200,10 @@ type TCreatableSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props#creatable-props)
    */
   menuShouldBlockScroll?: ReactSelectCreatableProps['menuShouldBlockScroll'];
+  /**
+   * When set to false, the select menu will not be closed after an option is selected. By default, it is set to true.
```
emmenko, line 204:
```suggestion
   * When set to false, the select menu will not be closed after an option is selected. By default, it is set to true.
   * <br>
   * [Props from React select was used](https://react-select.com/props)
```

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -221,7 +221,12 @@ type TSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   menuShouldBlockScroll?: ReactSelectProps['menuShouldBlockScroll'];
-  // menuShouldScrollIntoView: PropTypes.bool,
+  /**
+   * whether the menu should close after a value is selected
```
emmenko, line 225:
```suggestion
   * Whether the menu should close after a value is selected. Defaults to `true`
   * <br>
   * [Props from React select was used](https://react-select.com/props)
```

emmenko (commented):
Sorry to be picky but we should be consistent with the prop description between components.

emmenko (approved):
Thank you, all good now! 🙌

PR #2085, opened 2022-02-02
refactor(radio-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate Radio Input to typescript


packages/components/inputs/radio-input/src/radio-option.tsx
```diff
@@ -36,6 +85,7 @@ const Option = (props) => {
         css={accessibleHiddenInputStyles}
         id={props.id}
         name={props.name}
+        // @ts-ignore
```
CarlosCortizasCT, line 88: Why do we need to ignore next line?

CarlosCortizasCT: Could we use somehting like this instead?

Rhotimee: Good idea. Thanks.

CarlosCortizasCT (approved):

packages/components/inputs/radio-input/src/radio-group.tsx
```diff
@@ -11,11 +17,55 @@ const directionWrapper = {
...
+    | 12
+    | 13
+    | 14
+    | 15
+    | 16
+    | 'scale'
+    | 'auto';
+  direction?: keyof typeof directionWrapper;
+  directionProps?: object;
+  children: ReactElement | ReactElement[];
```
emmenko, line 46: Usually we use `ReactNode`. Any reason for not using that?

emmenko (commented):
Looks good, just one small question.

packages/components/inputs/radio-input/src/radio-group.tsx
Rhotimee: This is because we need to clone the element and pass some props to it. 
ReactNode could be null, undefined, boolean etc which will not work for our use case. 
See line 86.

emmenko: React has a function `isValidElement`. I think we can use that to validate if the prop is an element.

https://reactjs.org/docs/react-api.html#isvalidelement

Rhotimee: Okay, that makes sense. 
Updated

emmenko (approved):
Thanks!

PR #2086, opened 2022-02-04
refactor(async-creatable-select-field): Migrate to typescript

Rhotimee:
#### Summary
Migrate async-creatable-select-field to typescript


packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -0,0 +1,480 @@
+import {
+  Component,
+  isValidElement,
+  type ReactElement,
+  type ReactNode,
+} from 'react';
+import { type GroupBase, type ActionMeta } from 'react-select';
```
CarlosCortizasCT, line 7: Maybe we could standardise and use the `type` directive outside the brackets if all what we're importing are types (just like in the next line).
This way, we would only need to repeat in in case we're importing several types from a package where we also import components (like in the previous line).

What do you think?

Rhotimee: I completely agree

CarlosCortizasCT (approved):

emmenko (approved):

PR #2091, opened 2022-02-08
Fix Tag component color when disabled

### LINKED ISSUE #2090: Tag component don't use the right color for the text when disabled
**Describe the bug**
When using the [Tag component](https://uikit.commercetools.com/?path=/story/components-tags--tag) and providing the **isDisabled** prop, visualized text should have a lighter color but it is using the same color as the non-disabled state.

**To Reproduce**

1. Go to the [component storybook page](https://uikit.commercetools.com/?path=/story/components-tags--tag)
2. Select the "knobs" tab
3. Check the isDisabled checkbox
4. You will see no change in the component but the text color should be lighter

**Expected behavior**
Text color of the component should be lighter when disabled.
We should be using the **secondary** _tone_ for the text in that context.

**Screenshots**
Current visualization when disabled:
tag_ko

Expected visualization when disabled:
tag_ok

CarlosCortizasCT:
fixes #2090

### DIFF
--- a/.changeset/tender-elephants-turn.md
+++ b/.changeset/tender-elephants-turn.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tag': patch
+---
+
+Fix Tag component text color when disabled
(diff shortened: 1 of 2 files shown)


emmenko (commented):
Thanks, looks good. Remember to add a changeset.

Rhotimee (approved):
💯

ddouglasz (approved):

emmenko (approved):

PR #2092, opened 2022-02-08
refactor(date-time-input): Migrate to typescript

Rhotimee:
#### Summary
 Migrate date-time-input to typescript


packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -39,99 +46,137 @@ const activationTypes = [
...
+    | 'scale'
+    | 'auto';
+  /**
+   * The selected date, must either be an empty string or a date formatted in ISO 8601 (e.g. "2018-10-04T09:00:00.000Z").
+   */
+  value: string;
+  /**
+   * Called when the date changes. Called with an event containing an empty string (no value) or a string in this format: "YYYY-MM-DD".
+   * <br />
+   * Signature: `(event) => void`
```
CarlosCortizasCT, line 118: The signature is added automatically when generating the README, right?

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -39,99 +46,137 @@ const activationTypes = [
...
+  /**
+   * Called when the date input gains focus.
+   */
+  onFocus?: FocusEventHandler;
+  /**
+   * Called when the date input loses focus.
+   */
+  onBlur?: (event: TEvent) => void;
+  /**
+   * Specifies the time zone in which the calendar and selected values are shown. It also influences how entered dates and times are parsed.
```
CarlosCortizasCT, line 130: Maybe we could add an example of the time zone possible value to better understand its format.

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -157,11 +202,11 @@ class DateTimeInput extends Component {
       (prevState) => ({
         calendarDate: today,
         highlightedIndex:
-          prevState.suggestedItems.length +
+          (prevState.suggestedItems as string[]).length +
```
CarlosCortizasCT, line 205: Just curious about why we need to explicit use the type here because we already typed the component state in its declaration.

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -342,20 +388,27 @@ class DateTimeInput extends Component {
                     onMouseEnter: () => {
                       // we remove the highlight so that the user can use the
                       // arrow keys to move the cursor when hovering
+                      // @ts-ignore
                       if (isOpen) setHighlightedIndex(null);
                     },
-                    onKeyDown: (event) => {
+                    onKeyDown: (
+                      event: KeyboardEvent<HTMLButtonElement | HTMLInputElement>
```
CarlosCortizasCT, line 395: could we type the _event_ with `TPreventDownshiftDefaultEvent` so we don't need to cast later?

Rhotimee: I separated createBlurHandler and preventDownshiftDefault events. That solved the problem.

packages/components/inputs/date-time-input/src/date-time-input.tsx
Rhotimee: True, but, it could be undefined. I updated this to use an empty array if suggestedItems is undefined.

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -128,6 +129,7 @@ export type TDateTimeInputProps = {
   onBlur?: (event: TEvent) => void;
   /**
    * Specifies the time zone in which the calendar and selected values are shown. It also influences how entered dates and times are parsed.
+   * Get list of timezone with `moment.tz.names()` [See moment docs](https://momentjs.com/timezone/docs/#/data-loading/getting-zone-names/)
```
CarlosCortizasCT, line 132: 🔝

CarlosCortizasCT (approved):
LGTM

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -342,9 +377,10 @@ class DateTimeInput extends Component {
                     onMouseEnter: () => {
                       // we remove the highlight so that the user can use the
                       // arrow keys to move the cursor when hovering
+                      // @ts-ignore
```
emmenko, line 380: What's the issue here?

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -215,17 +249,18 @@ class DateTimeInput extends Component {
           selectedItem={this.props.value === '' ? null : this.props.value}
           highlightedIndex={this.state.highlightedIndex}
           onChange={this.emit}
-          stateReducer={(state, changes) => {
+          stateReducer={(_, changes) => {
+            // @ts-ignore
```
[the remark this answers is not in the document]
Rhotimee: I've fixed it now.

packages/components/inputs/date-time-input/src/date-time-input.tsx
Rhotimee: setHighlightedIndex is expecting a number for its index [docs](https://github.com/downshift-js/downshift#actions).

packages/components/inputs/date-time-input/src/time-input.tsx
```diff
@@ -1,6 +1,8 @@
-import PropTypes from 'prop-types';
+/// <reference types="@emotion/react/types/css-prop" />
+import type { KeyboardEventHandler, RefObject } from 'react';
 import { css } from '@emotion/react';
 import { customProperties as vars } from '@commercetools-uikit/design-system';
+import type { TDateTimeInputProps } from './date-time-input';
```
CarlosCortizasCT, line 5: Not sure about this one as we're creating a circular dependency here.
I think we should invert the types dependencies. I mean, `DateTimeInput` should depend on `TimeInput` but the other way.

I guess `isDisabled`, `onChange`, `value` and `placeholder` should be defined here and imported from `DateTimeInput`.

CarlosCortizasCT (changes requested):

packages/components/inputs/date-time-input/src/date-time-input.tsx
CarlosCortizasCT: Could we use the value `-1`?
If not, maybe we can add a comment in the code so we know in the future why we needed to use that `@ts-ignore`.

CarlosCortizasCT (changes requested):

packages/components/inputs/date-time-input/src/time-input.tsx
Rhotimee: I think import type help solve this problem.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

CarlosCortizasCT: Thanks for the link, @Rhotimee.

I guess `import type` is for helping when dealing with types when you use a compiler like `Babel` which works one-file-at-a-time and you have the **isolatedModules** option set to `true` (this is our case) in your `tsconfig` file.

I realized this is not going to be a problem at build time because types are actually removed (they don't exist at runtime). However, I think conceptually `TimeInput` is the building block for `DateTimeInput` so the dependency should only go from the latter to the former. I think it's ok for the `DateTimeInput` to be coupled to `TimeInput` but I don't think the same for the opposite.
For me `TimeInput` is a building block and should not depend on other components.
([reference](https://bradfrost.com/blog/post/atomic-web-design/))

Actually, the props in the `DateTimeInput` have a different purpose so, overall, I don't think we should share them just because they match and this component should have its own (I wouldn't say the same if `DateTimeInput` had that props only to pass them through to the `TimeInput`).

Anyway, this is not a blocker.

CarlosCortizasCT: We had a meeting to discuss this and I realized I was confused about this file.
@Rhotimee explained the rationale so now I understand why it's done like this.

CarlosCortizasCT (approved):

ddouglasz (approved):
Looks good

PR #2100, opened 2022-02-10
Migrate Pagination components to Typescript

CarlosCortizasCT:
#### Summary

Migrate pagination components to Typescript:
* PageSizeSelector
* PageNavigator
* Pagination

## Description

There are a couple of changes outside the scope of _pagination_ components that were needed in order to complete this task:
* [SelectInput](https://uikit.commercetools.com/?path=/story/components-inputs-selectinputs--selectinput) component had its `inputValue` property marked as required when it actually is not (that is mainly needed in the [SearchSelectInput](https://uikit.commercetools.com/?path=/story/components-inputs-selectinputs--searchselectinput) component)
* There were several components accepting an `intlMessage` prop for localization that were typed using the [MessageDescriptor](https://github.com/formatjs/formatjs/blob/main/packages/intl/src/types.ts#L211) type. The problem is that type is not suitable for parameterized messages (ex: "There are {count} products in this page"). So we had to extend the type definition in those components to also accept the `values` property to handle this scenario. 
Some components defined a similar type instead of using the one from `react-intl`. I just updated those instead of using the third-party one so we don't need to add a new dependency to those packages.


packages/components/link/README.md
```diff
@@ -47,7 +47,7 @@ export default Example;
...
 | ------------- | -------------------------------------------------------------- | :------: | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `children`    | `ReactNode`                                                    |          |             | Value of the link.&#xA;<br />&#xA;Required if `intlMessage` is not provided.                                                                                         |
-| `intlMessage` | `MessageDescriptor`                                            |          |             | An `intl` message object that will be rendered with `FormattedMessage`.&#xA;<br />&#xA;Required if `children` is not provided.                                       |
+| `intlMessage` | `intersection`                                                 |          |             | An `intl` message object that will be rendered with `FormattedMessage`.&#xA;<br />&#xA;Required if `children` is not provided.                                       |
```
CarlosCortizasCT, line 50: After extending the `MessageDescriptor` type in the definition, now we have this not-so-meaningful value in the READMEs.
Could use some help to figure out the better way to address this.

packages/components/pagination/src/page-size-selector/page-size-selector.tsx
```diff
@@ -23,11 +56,13 @@ const mapRangeToListOfOptions = (perPageRange) => {
   }
 };
 
-const PageSizeSelector = (props) => {
+const PageSizeSelector = (props: TPageSizeSelectorProps) => {
   const [perPageSelectorId] = useState(uniqueId('per-page-selector-'));
   const options = mapRangeToListOfOptions(props.perPageRange);
   const hasValidPerPageOption = options.includes(props.perPage);
 
+  console.log('PageSizeSelector:', { props });
```
emmenko, line 64: 😅

packages/components/pagination/src/pagination.tsx
```diff
@@ -1,17 +1,59 @@
...
+
+const Pagination = (props: TPaginationProps) => {
   const totalPages = Math.ceil(props.totalItems / props.perPage);
 
   const pageItems =
     props.page === totalPages
       ? props.totalItems - props.perPage * (props.page - 1)
       : props.perPage;
 
+  console.log(
```
emmenko, line 52: Same here

emmenko (dismissed):
Thanks, looks good overall. Just the thing with the enums

packages/components/link/README.md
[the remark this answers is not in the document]
emmenko: @CarlosCortizasCT FYI: the logic to fix this should be in this place

https://github.com/commercetools/ui-kit/blob/3704da64487f52d50dd349799def35407ecadbff/generators/readme/src/index.ts#L363-L494

packages/components/pagination/src/page-size-selector/page-size-selector.tsx
CarlosCortizasCT: Removed 22086a8c755996baef84bc6c7723ffa9aa37e1e9

packages/components/pagination/src/page-size-selector/page-size-selector.tsx
```diff
@@ -8,7 +7,37 @@ import { warning } from '@commercetools-uikit/utils';
 import Label from '@commercetools-uikit/label';
 import messages from './messages';
 
-const mapRangeToListOfOptions = (perPageRange) => {
+export type PAGE_RANGE_SIZE = 's' | 'm' | 'l';
```
emmenko, line 10:
```suggestion
export type TPageRangeSize = 's' | 'm' | 'l';
```

CarlosCortizasCT: Updated in 0930a7f385c5b35c8fe6f8120674d8f005ce7e8a

Thanks 👍

.changeset/thirty-years-drop.md
```diff
@@ -0,0 +1,11 @@
...
+'@commercetools-uikit/select-input': patch
+'@commercetools-uikit/label': patch
+'@commercetools-uikit/link': patch
+'@commercetools-uikit/messages': patch
+'@commercetools-uikit/notifications': patch
+'@commercetools-uikit/pagination': patch
+'@commercetools-uikit/text': patch
+---
+
+Pagination component migrated to Typescript
```
Rhotimee, line 11: I think it'll be nice to have separate changesets for the packages and give information on what was changed for each package.

CarlosCortizasCT: Updated in this commit: cd354fe3d2e0fbdca6a6512866851f3ec6469a72

Thanks for the advice 👍

CarlosCortizasCT:
@Rhotimee, @ddouglasz as we agreed, I disabled the auto-generation of README files for the components which use intersection types until we update the generator so I think all comments have already been addressed.
Please do your final review so we can merge this PR.

Rhotimee (approved):

Rhotimee (approved):
Looks good. Just minor feedback.

packages/components/pagination/src/pagination.tsx
```diff
@@ -34,48 +71,15 @@ const Pagination = (props) => {
...
-  /**
-   * A callback function, called when `perPage` is changed.
-   * <br/>
-   * Signature: `(nextPerPage: number) => void`
-   */
-  onPerPageChange: PropTypes.func.isRequired,
-};
-Pagination.defaultProps = {
+
+const defaultProps = {
```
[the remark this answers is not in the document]
CarlosCortizasCT: Fixed here: 639ec81706b070d5726c07ae936d72f3bb1bebe4

Thanks

packages/components/label/src/label.tsx
```diff
@@ -47,7 +47,9 @@ type TLabelProps = {
    * <br />
    * This is required when `children` is `undefined` and vice versa
    */
-  intlMessage?: MessageDescriptor;
+  intlMessage?: MessageDescriptor & {
+    values?: Record<string, React.ReactNode>;
```
ddouglasz, line 51: ReactNode seem to be already imported from react in this file, same as other places with similar feedback. Maybe we just need to add `type` to all places where they have been imported. 
Suggestion for taking modifying the `React.ReactNode` syntax is just for consistency with other files. 
```suggestion
    values?: Record<string, ReactNode>;
```

packages/components/notifications/src/content-notification/content-notification.tsx
```diff
@@ -23,7 +23,9 @@ type TContentNotificationProps = {
    * <br />
    * Required if `children` is not provided.
    */
-  intlMessage?: MessageDescriptor;
+  intlMessage?: MessageDescriptor & {
+    values?: Record<string, React.ReactNode>;
```
ddouglasz, line 27:
```suggestion
    values?: Record<string, ReactNode>;
```

packages/components/link/src/link.tsx
```diff
@@ -26,7 +26,9 @@ type TLinkProps = {
    * <br />
    * Required if `children` is not provided.
    */
-  intlMessage?: MessageDescriptor;
+  intlMessage?: MessageDescriptor & {
+    values?: Record<string, React.ReactNode>;
```
ddouglasz, line 30:
```suggestion
    values?: Record<string, ReactNode>;
```

ddouglasz (changes requested):

packages/components/label/src/label.tsx
CarlosCortizasCT: Yes, you're right but I this (also `ContentNotification` and `Link`) were out of the scope of this PR. I just modified its `MessageDescriptor` to allow parameterized messages and din't look into them any further.

Anyway, I can also make these changes now that I'm modifying those files for the other reason.

Thanks 👍

ddouglasz: Ah, I see. Thanks for clarifying.

CarlosCortizasCT: @ddouglasz I standarized `ReactNode` usage in this commit: 7f0fc7356c510a2b6754da069cd5b6489114d228

ddouglasz: Great!

ddouglasz (approved):
LGTM

PR #2072, opened 2022-01-24
Migrate rich text utils to TS

ddouglasz:
#### Summary

- Migrate rich-text-util to typescript

PR #2102, opened 2022-02-11
Migrate DataTableManager component to Typescript

CarlosCortizasCT:
## Summary

Migrate [DataTypeManager](https://uikit.commercetools.com/?path=/story/components-datatable--datatablemanager) and all of its subcomponent to Typescript.

Will need to wait for #2100 PR to be merged because it depends on a fix in a shared component (`SelectInput`) which is implemented over there.


packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -0,0 +1,192 @@
...
+   * Any React node. Usually you want to render the `<DataTable>` component.
+   * <br>
+   * Note that the child component will implicitly receive the props `columns` and `isCondensed` from the `<DataTableManager>`.
+   */
+  children: JSX.Element;
+
+  /**
+   * The managed display settings of the table.
+   */
+  displaySettings?: TDataTableSettingsProps['displaySettings'];
```
CarlosCortizasCT, line 128: Not sure the best way to approach this.
Both `displaySettings` and `columnManager` props were just reflected from the private `DataTableSettings` component so here I'm just using the definitions in it but this will not generate a meaningful README.
Maybe I should type that objects here as well so we get a right README or should we go for extending the readme generation script to handle this situation (not sure about its complexity because it would involve parsing a different file that the one parsed when generating the documentation).

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -19,20 +25,54 @@ import DroppablePanel from '../droppable-panel';
...
+type TDroppableContainerProps = {
+  isDragging: boolean;
+};
+
+// TODO: Very hacky solution.
+// searchHiddenColums is only required if areHiddenColumnsSearchable is true
+// but that callback is used in AsyncSelectInput.loadOptions which is required
+// That property (loadOptions) is passed to react-select AsyncSelect where it's
+// declared as optional
+const noSearch = () => Promise.resolve();
```
CarlosCortizasCT, line 60: Not sure if this is the best solution...

`searchHiddenColums` is only required if `areHiddenColumnsSearchable` is true but that callback is used in `AsyncSelectInput.loadOptions` which is required.
That property (`loadOptions`) is passed to `react-select` `AsyncSelect` where it's declared as optional

Rhotimee:
@CarlosCortizasCT, I noticed that JSX.Element was used  as opposed to ReactElement. Is there a reason for that?
Since ReactElement was used throughout the codebase, I feel it will be nice to maintain that consistency.

packages/components/data-table-manager/src/data-table-manager.tsx
Rhotimee: I'm not sure as well, another option is to add the link to the docs for the TDataTableSettingsProps just like we did for react-select.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
Rhotimee: We can trigger a warning instead if areHiddenColumnsSearchable is true and searchHiddenColums is not a function.

Rhotimee: Btw, whenever, requiredIf is used in prop-types we use [warning](https://github.com/commercetools/ui-kit/blob/main/packages/utils/src/warning.ts) instead when migrating.

CarlosCortizasCT:
> @CarlosCortizasCT, I noticed that JSX.Element was used as opposed to ReactElement. Is there a reason for that? Since ReactElement was used throughout the codebase, I feel it will be nice to maintain that consistency.

I saw in [another PR](https://github.com/commercetools/ui-kit/pull/2072#discussion_r803830611) that type used so I thought I should use it as well but if we prefer to use `ReactElement` type, I can apply the changes with no problem.

packages/components/data-table-manager/src/data-table-manager.tsx
CarlosCortizasCT: Yes, but the problem is that `DataTableSettingsProps` is a private component so no README for it exists in the documentation where to link to.

ddouglasz (changes requested):

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
CarlosCortizasCT: @Rhotimee, @ddouglasz Do you mind if I look for some time in your calendars and schedule a meeting to discuss this one?

Rhotimee: sure, we can have a call tomorrow to discuss this.

ddouglasz: Sure

packages/components/data-table-manager/src/data-table-manager.tsx
CarlosCortizasCT: We will need to update the READMEs generator to take this cases into account.
Until them, I disabled the auto-generation of this component README by renaming the `docs` folder (the generator will ignore it).

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
CarlosCortizasCT: As discussed, we agreed in using the `@ts-ignore` directive as it is safe for this use case and we avoid creating a dummy function.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -19,20 +26,47 @@ import DroppablePanel from '../droppable-panel';
...
+type TDroppableContainerProps = {
+  isDragging: boolean;
+};
 
 const DroppableContainer = styled.div`
   width: 100%;
   position: relative;
   max-width: ${vars.constraint10};
-  cursor: ${(props) => (props.isDragging ? 'grabbing' : 'auto')};
+  cursor: ${(props: TDroppableContainerProps) =>
```
kark, line 60: Perhaps instead of this we might pass the type like that:
`const DroppableContainer = styled.div<TDroppableContainerProps>`
but it's up to you

packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * A callback function, called when the header cell is clicked.
+   */
+  onClick?: (event: MouseEventHandler) => void;
+
+  /**
+   * A callback function to render the content of cells under this column, overriding
+   * the default `itemRenderer` prop of the table.
+   */
+  renderItem?: (row: ReactNode, isRowCollapsed: boolean) => ReactNode;
```
kark, line 57: I found another place where we type this and perhaps Rotimi's approach in [here](https://github.com/commercetools/ui-kit/blob/23fd9e859ac86dbe778100ddb10cb9464de53d0c/packages/components/data-table/src/data-table.tsx#L110) is worth following

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -1,7 +1,6 @@
-import PropTypes from 'prop-types';
-import { useState } from 'react';
+import { useState, type ReactElement, type ReactNode } from 'react';
 import { warning } from '@commercetools-uikit/utils';
-import { useIntl } from 'react-intl';
+import { MessageDescriptor, useIntl } from 'react-intl';
```
kark, line 3:
```suggestion
import { useIntl, type MessageDescriptor } from 'react-intl'
```

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -50,19 +159,21 @@ export const getDropdownOptions = ({
     : []),
 ];
 
-export const getMappedColumns = (columns = []) =>
+export const getMappedColumns = (columns = [] as TColumnData[]) =>
   columns.reduce(
```
kark, line 163: I have seen this pattern used in the codebase `columns.reduce<MappedColumns>`

kark (commented):
This is great work 🙌 A fantastic source of knowledge for me personally. I left some comments which are mostly nitpicks

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
CarlosCortizasCT: Thanks for the suggestion, Kacper.
I'm not sure if we have a policy on this but I would argue that sometimes generics are not that readable to me. I mean, when you see a generic value applied you don't immediately know what that type is used for so you need to search for its purpose.

I know there can be done both ways and maybe it makes more sense if you use props several times in a component.
Anyway, I would like to know @Rhotimee and @ddouglasz opinion about this so we try to standarize.

packages/components/data-table-manager/src/data-table-manager.tsx
CarlosCortizasCT: Thanks Kacper.
I'm not sure about this. Linked code is using and internal type which is a styled component but that type is not exposed.
I can't see how should I apply that pattern over here.
Could you provide an example of how you would change this?

kark: Thanks Carlos.
I am not sure about it, but my point is that it seems we are changing the signature here from `(row: object, isRowCollapsed: boolean) => React.Node`.
And there was an [issue raised](https://github.com/commercetools/ui-kit/issues/2023) followed by a [PR](https://github.com/commercetools/ui-kit/pull/2024) in the past regarding using non-flexible type of `row`.
Although I am not sure if that applies in this context.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
ddouglasz: I think it can be done both ways and it might be more handy to pass as generics if props were used in multiple places in a component.
I see the generics pattern in other parts of the codebase. Maybe we can just do same for uniformity since `TDroppableContainerProps` already seems defined, I don't think in this case, it would be difficult to read the type. Either ways, I see no harm.

Rhotimee: Either is fine with me.  However, I see that we use the pattern @kark suggested more in the codebase though.

CarlosCortizasCT: So we have a consensus.
I'll update to use the generic pattern.

Thank you all 👍

CarlosCortizasCT: Updated in this commit 4d6e6f47f43f5fa062f50edc499ca50fd205e37e

packages/components/data-table-manager/package.json
```diff
@@ -49,6 +49,8 @@
     "react-required-if": "1.0.3"
```
emmenko, line 49: This dependency can be removed now.

emmenko (commented):
Looking good 👌

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -163,6 +204,11 @@ export const ColumnSettingsManager = (props) => {
                   // loadOptions is used instead of onInputChange
                   // because the loading indicator is displayed
                   // only when loadOptions is invoked
+                  // ts-ignore directive is used because 'searchHiddenColumns' only
+                  // exists when 'areHiddenColumnsSearchable' prop is true (otherwise
+                  // this block is not rendered) but the AsyncSelectInput component
+                  // requires the 'loadOptions' property
+                  // @ts-ignore
```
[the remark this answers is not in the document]
CarlosCortizasCT: That was my first idea but then we talked about it and [decided](https://github.com/commercetools/ui-kit/pull/2102#discussion_r805638258) to just ignore so we didn't need to create a new empty function.

Anyway, if we prefer to have that _noop_ function I change it.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -19,20 +26,47 @@ import DroppablePanel from '../droppable-panel';
 import SettingsContainer from '../settings-container';
 import messages from './messages';
 import { HIDDEN_COLUMNS_PANEL, SELECTED_COLUMNS_PANEL } from './constants';
+import { warning } from '@commercetools-uikit/utils';
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Sets a width for this column. Accepts the same values as the ones specified for
+   * individual [grid-template-columns](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns).
+   * <br>
+   * For example, using `minmax` pairs (e.g. `minmax(200px, 400px)`), a combinations of
+   * fraction values (`1fr`/`2fr`/etc), or fixed values such as `200px`.
+   * By default, the column grows according to the content and respecting the total table available width.
+   *
+   * @@defaultValue@@: auto
```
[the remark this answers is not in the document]
CarlosCortizasCT: Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -0,0 +1,197 @@
...
+   * instead of breaking into multiple lines.
+   * <br>
+   * NOTE: when using this option, it is recommended to specify a `width` for the column, because
+   * if the table doesn't have enough space for all columns, it will start clipping the columns
+   * with _truncated_ content, and if no `width` is set (or the value is set `auto` -- the default)
+   * it can shrink until the column disappears completely.
+   * By enforcing a minimum width for these columns, the table will respect them and grow horizontally,
+   * adding scrollbars if needed.
+   *
+   * @@defaultValue@@: false
```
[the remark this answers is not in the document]
CarlosCortizasCT: Remove in this commit Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -16,6 +15,112 @@ import DisplaySettingsManager, {
...
+  label: ReactNode;
+};
+
+type MappedColumns = Record<string, TColumnData>;
+
+export type TDisplaySettingsProps = {
+  /**
+   * Set this flag to `false` to show the display settings panel option.
+   *
+   * @@defaultValue@@: true
```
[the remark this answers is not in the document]
CarlosCortizasCT: Removed in this commit Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -50,19 +159,21 @@ export const getDropdownOptions = ({
     : []),
 ];
 
-export const getMappedColumns = (columns = []) =>
-  columns.reduce(
+export const getMappedColumns = (columns = [] as TColumnData[]) =>
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -50,19 +159,21 @@ export const getDropdownOptions = ({
...
       [column.key]: column,
     }),
     {}
   );
 
-export const getSelectedColumns = (mappedColumns, visibleColumnsKeys = []) =>
-  visibleColumnsKeys.map((columnKey) => mappedColumns[columnKey]);
+export const getSelectedColumns = (
+  mappedColumns: MappedColumns,
+  visibleColumnsKeys = [] as string[]
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/draggable-tag/draggable-tag.tsx
```diff
@@ -1,12 +1,24 @@
...
+
+type TDraggableTagProps = {
+  column: TColumnData;
+  index: number;
+  isDisabled?: boolean;
+  onRemove?: (index: number) => void;
+};
+
+const DraggableTag = (props: TDraggableTagProps) => {
+  const handleRemoveColumn = () => props.onRemove!(props.index);
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit Removed in this commit 009bda4b8398d4a322afdb932062594e0ee9772f

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit 5e96bcc6ea998b07a3e1f106b68b7c6c82922bc8

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit 5e96bcc6ea998b07a3e1f106b68b7c6c82922bc8

Let's see if it its acceptable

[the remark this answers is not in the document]
emmenko: Ah sorry, I didn't know about that discussion. I think I would still prefer to pass an empty function, as `ts-ignore` might mask other potential problems.

packages/components/data-table-manager/package.json
CarlosCortizasCT: Removed in this commit: 35ac3cf400bd274b73b64acbefb86aa0cf1efc17

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -106,7 +147,7 @@ export const ColumnSettingsManager = (props) => {
   );
 
   const handleDragEnd = useCallback(
-    (dragResult) =>
+    (dragResult: DropResult) =>
```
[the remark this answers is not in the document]
CarlosCortizasCT: For what I see in the types definition of [`useCallback`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1112) the actual signature would have to be like this:
```
const handleDragEnd = useCallback<(dragResult: DropResult) => void>((dragResult) => {...})
```

It feels boilerplate to me, but is just my opinion. If we prefer that shape, I'll change it.

packages/components/data-table-manager/src/draggable-tag/draggable-tag.tsx
```diff
@@ -32,7 +44,6 @@ const DraggableTag = (props) => {
                 </Spacings.Inline>
               </Tag>
             </DraggingContainer>
-            {provided.placeholder}
```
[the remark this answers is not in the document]
CarlosCortizasCT: `provided` prop comes from [`Draggable`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-beautiful-dnd/index.d.ts#L693) component [render prop](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-beautiful-dnd/index.d.ts#L678).

That [`provided`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fb785106d6264285d452e2e7efb5c68c0639fbd8/types/react-beautiful-dnd/index.d.ts#L658) prop does not have a `placeholder` attribute and I also experimented in storybook with the component and didn't find any difference when removing it so it seems it never had any reason to be there.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -61,7 +95,7 @@ export const handleColumnsUpdate = (
     const columns = isSwap ? selectedColumns : availableColumns;
     const draggedColumn = columns.find(
       (col) => col.key === dragResult.draggableId
-    );
+    )!;
```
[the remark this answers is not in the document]
CarlosCortizasCT: For what I understood, here we're searching in the available columns for the one dragged by `id` so I think we should always find it.
Once it is found, it call the `onUpdateColumns` callback to set the new columns order.

In case we would want to be more defensive for any reason I might not be seeing, I think we could ignore the update if we don't find the column.
What do you think?
```
const draggedColumn = columns.find(
  (col) => col.key === dragResult.draggableId
)!;

// push the column in the new position if draggedColumn is found
if (draggedColumn) {
    onUpdateColumns([
      ...items.slice(0, dragResult.destination.index),
      draggedColumn,
      ...items.slice(dragResult.destination.index),
    ]);
}
```

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -77,23 +188,32 @@ const DataTableSettings = (props) => {
...
+  const handleDropdownChange = (event: SelectChangeEvent) =>
+    setOpenedPanelId(
+      Array.isArray(event.target.value)
+        ? event.target.value[0]
+        : event.target.value
+    );
 
   const mappedColumns = getMappedColumns(
-    areColumnSettingsEnabled ? props.columnManager.hideableColumns : undefined
+    areColumnSettingsEnabled ? props.columnManager!.hideableColumns : undefined
```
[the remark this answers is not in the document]
CarlosCortizasCT: This component defines `columnManager` prop as optional and then it has a constant called `areColumnSettingsEnable` which holds whether that prop has been received or not.

In this line we first check that constant to be true so it is safe to use the `!` if it is true.

packages/components/data-table-manager/src/data-table-settings/data-table-settings.tsx
```diff
@@ -77,23 +188,32 @@ const DataTableSettings = (props) => {
...
   );
 
   const selectedColumns = getSelectedColumns(
     mappedColumns,
-    areColumnSettingsEnabled ? props.columnManager.visibleColumnKeys : undefined
+    areColumnSettingsEnabled
+      ? props.columnManager!.visibleColumnKeys
```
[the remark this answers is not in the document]
CarlosCortizasCT: This component defines `columnManager` prop as optional and then it has a constant called `areColumnSettingsEnable` which holds whether that prop has been received or not.

In this line we first check that constant to be true so it is safe to use the `!` if it is true.

packages/components/data-table-manager/src/draggable-tag/draggable-tag.tsx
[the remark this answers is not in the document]
emmenko: Nice, thanks

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
emmenko: Yes I think we usually do that. Thanks

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
emmenko: If it's not too much effort, I think we can maybe be more on the safe side?

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
emmenko: Btw, if we prefer to type these things differently, I'm happy to discuss it and eventually make it a recommendation, as long as it does the job.

We can follow up on this maybe.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Sure.

Updated in this commit c3696bfee231a5ffc8eaaf06d4b1ec0034c72d86

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -118,9 +159,9 @@ export const ColumnSettingsManager = (props) => {
   );
 
   const handleInputChange = useCallback(
-    (input) =>
+    (inputValue: string) =>
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit c3696bfee231a5ffc8eaaf06d4b1ec0034c72d86

emmenko (approved):

ddouglasz (approved):

kark (approved):

PR #2103, opened 2022-02-11
refactor(radio-field): Migrate to typescript

Rhotimee:
#### Summary
Migrate radio-field to typescript


packages/components/fields/radio-field/src/radio-field.tsx
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Provides a description for the title.
+   */
+  description?: string | ReactNode;
+  /**
+   * Function called when info button is pressed.
+   * <br />
+   * Info button will only be visible when this prop is passed.
+   */
+  onInfoButtonClick?: () => void;
```
CarlosCortizasCT, line 133: Since this is forwarded to the `FieldLabel` component, maybe we should type the handler like it is over there:
```
onInfoButtonClick?: (
  ) => void;
```
or something like:
```
onInfoButtonClick?: (
    event: MouseEventHandler | KeyboardEventHandler
  ) => void;
```

What do you think?

Rhotimee: You're right. It's been updated. 
Thanks.

packages/components/fields/radio-field/src/radio-field.tsx
```diff
@@ -0,0 +1,229 @@
...
+   */
+  isReadOnly?: boolean;
+  /**
+   * Rendering direction of the radio `RadioInput.Option`s
+   */
+  direction?: 'stack' | 'inline';
+  /**
+   * Passes props of the `Spacings.Stack` or `Spacings.Inline`, dependeing on the chosen direction
+   */
+  directionProps?: object;
```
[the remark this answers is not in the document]
Rhotimee: That makes sense. 
I exposed the props from spacing then used it here.

CarlosCortizasCT (approved):

CarlosCortizasCT:
great job!

packages/components/fields/radio-field/src/radio-field.tsx
```diff
@@ -0,0 +1,233 @@
...
+    id: getFieldId(props, state, sequentialId),
+  });
+
+  render() {
+    const hasError = this.props.touched && hasErrors(this.props.errors);
+
+    if (!this.props.isReadOnly) {
+      warning(
+        typeof this.props.onChange === 'function',
+        'RadioField: `onChange` is required when is not read only.'
```
ddouglasz, line 180:
```suggestion
        'RadioField: `onChange` is required when it's not read only.'
```

ddouglasz (changes requested):

ddouglasz (approved):
LGTM

PR #2104, opened 2022-02-11
Migrate localized text input to types…

ddouglasz:
Migrate localized text input to typescript


packages/components/inputs/localized-text-input/src/localized-text-input.styles.ts
```diff
@@ -17,7 +19,10 @@ const getLocalizedInputStyles = () => [
   `,
 ];
 
-const getLanguageLabelStyles = (props, theme) => {
+const getLanguageLabelStyles = (
+  _props: LocalizedTextInputStylesProps,
```
CarlosCortizasCT, line 23: I'm wondering if we could use the `unknown` type instead of creating an empty one 🤔

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -1,7 +1,4 @@
-import { useCallback } from 'react';
-import PropTypes from 'prop-types';
-import requiredIf from 'react-required-if';
-import { oneLine } from 'common-tags';
+import { ReactNode, useCallback } from 'react';
```
CarlosCortizasCT, line 1:
```suggestion
import { 
  useCallback,
  type ReactNode,
} from 'react';
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -31,9 +28,130 @@ import {
...
+    name?: string;
+    value?: unknown;
+  };
+};
+
+type TLocalizedTextInputProps = {
+  id?: string;
+  name?: string;
+  autoComplete?: string;
+  // then input doesn't accept a "languages" prop, instead all possible
```
CarlosCortizasCT, line 42: Should this comment be formatted as JSDoc so it is reflected in the README?

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -191,6 +303,7 @@ const LocalizedTextInput = (props) => {
         {shouldRenderLanguagesButton && (
           <LocalizedInputToggle
             isOpen={areLanguagesExpanded}
+            // @ts-ignore
             onClick={toggleLanguages}
```
CarlosCortizasCT, line 307: Not sure about this one but we're providing a function (`toggleLanguages`) which has this signature:
```
toggleLanguages(forceIsToggled?: boolean) => void
```
to a prop which has this signature:
```
onClick: (
  ) => void;
```

When dealing with this situations I think is is not completely safe to use functions with different signatures since they can stop working as expected in the future.
Let's imagine the `LocalizedInputToggle.onClick` changes its implementation (yes, I know it's not very probable in this context) and its first param becomes a _boolean_ (`true`): the languages toggle would will ever be `true` so no toggle anymore.

I believe is safer to use an intermediate function in this situations that is tailored for its use case:
```
const onLocalizedInputToggle = () => toggleLanguages();
```

Maybe this is complicate things more than needed, but I wanted to bring this thought up.

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -149,6 +247,20 @@ const LocalizedTextInput = (props) => {
   const shouldRenderLanguagesButton =
     languages.length > 1 && !props.hideLanguageExpansionControls;
 
+  if (!props.isReadOnly) {
+    warning(
+      typeof props.onChange === 'function',
+      'LocaliszedTextInput: "onChange" is required when isReadOnly is not true'
```
CarlosCortizasCT, line 253:
```suggestion
      'LocalizedTextInput: "onChange" is required when isReadOnly is not true'
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -149,6 +247,20 @@ const LocalizedTextInput = (props) => {
...
+    );
+  }
+
+  if (props.hideLanguageExpansionControls) {
+    warning(
+      typeof props.hideLanguageExpansionControls === 'boolean',
+      'LocaliszedTextInput: "defaultExpandLanguages" does not have any effect when "hideLanguageExpansionControls" is set.'
```
CarlosCortizasCT, line 260:
```suggestion
      'LocalizedTextInput: "defaultExpandLanguages" does not have any effect when "hideLanguageExpansionControls" is set.'
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -149,6 +247,20 @@ const LocalizedTextInput = (props) => {
...
+    );
+  }
```
Rhotimee, line 262: I don't think this is correct. 
I think it's supposed to be something like this;

```suggestion
  if (props.hideLanguageExpansionControls) {
    warning(
      typeof props.defaultExpandLanguages !== 'boolean',
      'LocaliszedTextInput: "defaultExpandLanguages" does not have any effect when "hideLanguageExpansionControls" is set.'
    );
  }
```

Please confirm this.

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -31,9 +28,130 @@ import {
...
+   */
+  onChange?: (event: TEvent) => void;
+  /**
+   * Specifies which language will be shown in case the `LocalizedTextInput` is collapsed.
+   */
+  selectedLanguage: string;
+  /**
+   * Called when any field is blurred. Is called with the `event` of that field.
+   */
+  onBlur?: () => void;
```
Rhotimee, line 59: Is there a reason FocusEventHandler isn't used here instead?
```suggestion
  onBlur?: FocusEventHandler;
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -31,9 +28,130 @@ import {
...
+  /**
+   * Called when any field is focussed. Is called with the `event` of that field.
+   */
+  onFocus?: () => void;
```
Rhotimee, line 63:
```suggestion
  onFocus?: FocusEventHandler;
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -31,9 +28,130 @@ import {
...
+  // then input doesn't accept a "languages" prop, instead all possible
+  // languages have to exist (with empty or filled strings) on the value:
+  //   { en: 'foo', de: '', es: '' }
+  value: {
+    [key: string]: string;
+  };
+  /**
+   * Gets called when any input is changed. Is called with the change event of the changed input.
+   */
+  onChange?: (event: TEvent) => void;
```
Rhotimee, line 51: Why not use ChangeEventHandler instead?
```suggestion
  onChange?: ChangeEventHandler;
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -31,9 +28,130 @@ import {
...
+   */
+  autoComplete?: string;
+  value: string;
+  /**
+   * Gets called when any input is changed. Is called with the change event of the changed input.
+   */
+  onChange?: (event: TEvent) => void;
+  language: string;
+  onBlur?: () => void;
+  onFocus?: () => void;
```
Rhotimee, line 144:
```suggestion
  onChange?: ChangeEventHandler;
  language: string;
  onBlur?: FocusEventHandler;
  onFocus?: FocusEventHandler;
```

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -50,12 +168,19 @@ const LocalizedInput = (props) => {
       //
       // eslint-disable-next-line no-param-reassign
```
Rhotimee, line 169: It'll be nice to remove this eslint comment

Rhotimee (changes requested):

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
ddouglasz: Exactly, I actually thought about this and felt we should have a talk about this particular prop as it appears to be a bit confusing  on how to handle it.

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
ddouglasz: I am not so sure because this was my initial implementation but it breaks a test and I assume we are not meant to change any tests that already satisfies a criteria right?

PR #2113, opened 2022-02-16
Fix invalid hint prop validation

### LINKED ISSUE #2112: Invalid check of hint prop in several components
**Describe the bug**
The validation introduced in the components migrated to Typescript which can receive `hint` and `hintIcon` properties for checking the former is incorrect which causes the components to break in a production environment.

**To Reproduce**
Providing such a component both `hint` and `hintIcon` props.

Example:
```
const ProjectDeleteFormKeyField = (props) => (
  <TextField
    name="key"
    title={<FormattedMessage {...messages.keyLabel} />}
    value={props.formik.values.key}
    isRequired={true}
    isDisabled={props.formik.isSubmitting}
    errors={props.formik.errors.key}
    renderError={renderError}
    touched={props.formik.touched.key}
    onChange={props.formik.handleChange}
    onBlur={props.formik.handleBlur}
--> hint={
      <FormattedMessage
        {...messages.keyHasToBeMatchingInformationMessage}
        values={{ projectKey: props.projectKeyPendingDeletion }}
      />
    }
--> hintIcon={<InfoIcon color="warning" />}
    data-testid="field-key"
  />
);
```

This would raise an error with this message (in the browser console):
```
'TextField: `hint` is required to be string or ReactNode if hintIcon is present'
```

**Expected behavior**
Component should render normally as provided props are actually valid.

The origin of the problem is the way the `hint` prop is being check to be a valid React element.
This is the current validation pattern:
```
// WRONG
if (this.props.hintIcon) {
  warning(
    typeof this.props.hint === 'string' ||
      isValidElement(typeof this.props.hint),
    'TextField: `hint` is required to be string or ReactNode if hintIcon is present'
  );
}
```
Here we're passing the `React.isValidElement` function a string representing the type of the `hint` property where we should be passing the value of that property itself.

This is the expected validation pattern
```
// VALID
if (this.props.hintIcon) {
  warning(
    typeof this.props.hint === 'string' ||
      isValidElement( this.props.hint),      // <--- REMOVE THE typeof KEYWORD
    'TextField: `hint` is required to be string or ReactNode if hintIcon is present'
  );
}
```

**Additional context**
This is the list of the affected components :
* [async-creatable-select-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx#L394)
* [async-select-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/async-select-field/src/async-select-field.tsx#L341)
* [creatable-select-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/creatable-select-field/src/creatable-select-field.tsx#L371)
* [date-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/date-field/src/date-field.tsx#L188)
* [date-range-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/date-range-field/src/date-range-field.tsx#L184)
* [number-field-tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/number-field/src/number-field.tsx#L205)
* [radio-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/radio-field/src/radio-field.tsx#L187)
* [text-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/text-field/src/text-field.tsx#L179)
* [password-field.tsx](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/password-field/src/password-field.tsx#L194)

CarlosCortizasCT:
#### Summary

fixes #2112


packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -390,8 +390,7 @@ export default class AsyncCreatableSelectField extends Component<
 
     if (this.props.hintIcon) {
       warning(
-        typeof this.props.hint === 'string' ||
-          isValidElement(typeof this.props.hint),
+        typeof this.props.hint === 'string' || isValidElement(this.props.hint),
```
tdeekens, line 393: Makes sense. Thanks.

I presume `isValidElement` comes from [here](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/packages/react/src/jsx/ReactJSXElementValidator.js#L65). There is also a `isValidElementType` defined [here](https://github.com/facebook/react/blob/main/packages/shared/isValidElementType.js#L39). 

I just wanted to verify that the need the `typeof ''` check in [this sandbox](https://codesandbox.io/s/dreamy-edison-2xkz0n?file=/src/App.js) cause `isValidElementType` looked like it covers strings.

tdeekens (approved):
Many thanks for the quick fix.

Rhotimee (approved):

PR #2117, opened 2022-02-18
fix(multiline-input): change in the default line break behavior to per-word-break

### LINKED ISSUE #2050: MultiLine attributes wrapping by letter instead of by word
A UI/UX improvement opportunity for MC

Product descriptions and any custom product attributes with `inputHint: "MultiLine"` are wrapping at the individual character level rather than at the word level. It isn't saving this way, just rendering on the screen this way. This is coming from `word-break: break-all` in the CSS.

{{SUPPORT-14505}}

kark:
fixes #2050

#### Summary

Change in the default line break behavior of MultilineInput component from per-character-break to per-word-break

**Before:**
break-all
break-all-2

**After:**

break-word
break-word-2


CarlosCortizasCT:
Please, remember to update this branch with _main_ to allow it to be merged.

Rhotimee (approved):

PR #2118, opened 2022-02-18
Fixes MoneyInput issue when using auto-generated IDs

CarlosCortizasCT:
### Summary

During the [update of the MC-frontend](https://github.com/commercetools/merchant-center-frontend/pull/11847) to the newest version of ui-kit (v13) there have been found some errors from the applications using the `MoneyInput` component when no `id` is provided in its _props_.

### Description

The problem is that `MoneyInput` is not handling correctly the auto-generation of IDs for the HTML labels and inputs it uses and this causes problems and also tests fail because its inner component `CurrencyLabel` has its _id_ prop set as required but the main component is not creating one when it's not received in its props.

I also updated the `useFieldId`: this is used from React functional components which need to generate an ID. You provide an optional ID and a generator function and it will return the ID received if it exists; otherwise, it will create a new one using the received generator.
The problem is that generation only happens after the component is first rendered because of the `useEffect` usage so components using this, will have an undefined ID in its first render which can cause tests to fail (as mentioned with the `CurrencyLabel` ID requirement).
The solution I implemented is to initialize the hook's initial internal state with the generated ID in case it doesn't receive one so it always have a value from the very beginning.


packages/hooks/src/use-field-id/use-field-id.ts
```diff
@@ -4,7 +4,7 @@ import { getFieldId } from '@commercetools-uikit/utils';
 type CreateIdFn = () => string;
 
 const useFieldId = (id: string | undefined, createIdFn: CreateIdFn): string => {
```
tdeekens, line 6: 🤔 Can't one argument be `type TCreateIDFn = () => string | undefined`? 

Feels like we add this additional argument to not have changes spread out too much in the code base and be able to pass the return of `createSequentialId` in, right?

tdeekens: Disregard, I get it now. We pass the ID or use the return of the incrementing fn.

tdeekens (approved):

Rhotimee (approved):
Looks good. Thanks

PR #2121, opened 2022-02-21
Fix SelectField props validation

CarlosCortizasCT:
### Summary

During the [update of the MC-frontend](https://github.com/commercetools/merchant-center-frontend/pull/11847) to the newest version of ui-kit (v13) there have been found some errors from the applications using the `SelectField` component when `isMulti` property is _true_.

### Description

The problem relies in that the component perform some runtime validations on properties `value` and `touched` for them to be arrays in case `isMulti` is _true_.

Both those properties are declares as optional, but when using `isMulti` _true_ we're expecting (incorrectly) for them to exist.


packages/components/fields/select-field/src/select-field.tsx
```diff
@@ -308,12 +308,14 @@ export default class SelectField extends Component<TSelectFieldProps> {
 
     if (this.props.isMulti) {
       warning(
-        Array.isArray(this.props.value),
+        Array.isArray(this.props.value) ||
+          typeof this.props.value === 'undefined',
         'SelectField: `value` is expected to be an array of string when isMulti is true'
```
emmenko, line 313:
```suggestion
        `SelectField: "value" is expected to be an array of string when "isMulti" is true, instead got ${this.props.value}.`
```

packages/components/fields/select-field/src/select-field.tsx
```diff
@@ -308,12 +308,14 @@ export default class SelectField extends Component<TSelectFieldProps> {
...
       );
 
       warning(
-        Array.isArray(this.props.touched),
+        Array.isArray(this.props.touched) ||
+          typeof this.props.touched === 'undefined',
         'SelectField: `touched` is expected to be an array of boolean when isMulti is true'
```
emmenko, line 319:
```suggestion
        `SelectField: "touched" is expected to be an array of boolean when "isMulti" is true, instead got ${this.props.touched}.`
```

.changeset/slow-cameras-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/select-field': patch
```
emmenko, line 2:
```suggestion
'@commercetools-uikit/select-field': patch
'@commercetools-uikit/select-input': patch
```

packages/components/fields/select-field/src/select-field.tsx
CarlosCortizasCT: Updated in 60d2ee841c61658c25251cd1b69ce397c929b0de

Thanks!

packages/components/fields/select-field/src/select-field.tsx
CarlosCortizasCT: Updated in 60d2ee841c61658c25251cd1b69ce397c929b0de

Thanks!

.changeset/slow-cameras-brush.md
CarlosCortizasCT: Added another changeset file for `select-input` so we have different messages per package: 6ad608a0c3884dec23057a35bda03005c368b6c9

Thanks~

emmenko (approved):
💯

Issue #2077, opened 2022-01-27
Add ARIA attributes to Inputs


codeshifu:
**Is your feature request related to a problem? Please describe.**
We want to be able to grab/access invalid input elements in the DOM and (automatically) scroll the input into view.

**Describe the solution you'd like**
Add `aria-errormessage` and `aria-invalid` attributes to input elements. Such that if inputs are invalid we have
`aria-invalid="true"` otherwise `aria-invalid="false"`

**Describe alternatives you've considered**
A [previous approach](https://github.com/commercetools/merchant-center-frontend/blob/04e62314ff29f78cdcd2da32e40243153af2f9a8/packages-shared/core/components/error-field-focus/error-field-focus.js#L8-L12) was to look for element/class names which could easily break the implementation if either of those gets updated

**Additional context**
This is in relation to this [PR](https://github.com/commercetools/merchant-center-frontend/pull/11316) which is to automatically scroll to error fields when users submits an invalid form.

tdeekens:
Thanks for the issue. Upon first glimpse I guess we could add aria labels based on the [hasError](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/text-input/src/text-input.tsx#L55) prop. We however don't seem to have a error message at hand in the component so we could allow forwarding it.

emmenko:
I think the aria attributes should be added to the field components, not inputs.

tdeekens:
Shouldn't at at least be forwarded to the inputs and hence also directly applied?

codeshifu:
I was gonna ask the same. @emmenko do we forward to the inputs or adding to just the fields will suffice?

emmenko:
Yes sorry I wasn't so clear.

The aria attributes will go to the input but also the error field message should have the correct identifier to be referenced by the `aria-errormessage` attribute.

tdeekens:
Ah, good point yes, it should affect the interplay of fields, errors and inputs.

emmenko:
FYI: these aria attributes should also be forwarded to react-select

<img width="616" alt="image" src="https://user-images.githubusercontent.com/1110551/151824188-87d063b5-e434-468d-beb7-db133282f13e.png">

codeshifu:
got it

emmenko:
It's been released.

codeshifu:
thanks

PR #2122, opened 2022-02-21
Add support for `aria-invalid` and `aria-errormessage` to fields and inputs components

codeshifu:
#### Summary
Add *ARIA* attributes to fields & inputs.

## Description
#2077


packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -387,6 +388,7 @@ export default class AsyncCreatableSelectField extends Component<
     const hasError =
       AsyncCreatableSelectInput.isTouched(this.props.touched) &&
       hasErrors(this.props.errors);
+    const errorFieldId = randomString();
```
emmenko, line 391: I would suggest to use the `createSequentialId` instead of using a new function.

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -466,8 +468,12 @@ export default class AsyncCreatableSelectField extends Component<
             showOptionGroupDivider={this.props.showOptionGroupDivider}
             iconLeft={this.props.iconLeft}
             {...filterDataAttributes(this.props)}
+            /* ARIA */
+            aria-invalid={hasError}
+            aria-errormessage={errorFieldId}
```
emmenko, line 473: Can you define them near the other two aria props? (L428-429)

packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.tsx
```diff
@@ -108,6 +108,14 @@ type TAsyncCreatableSelectInputProps = {
...
    */
   'aria-labelledby'?: ReactSelectAsyncCreatableProps['aria-labelledby'];
+  /**
+   * a11y attribute to determine if input is valid/not
+   */
+  'aria-invalid'?: boolean;
+  /**
+   * a11y attribute identifier that points to a field with error message
+   */
+  'aria-errormessage'?: string;
```
emmenko, line 118: Please use the same prop descriptions style as in the other aria props.

```suggestion
  /**
   * Indicate if the value entered in the input is invalid.
   * <br>
   * [Props from React select was used](https://react-select.com/props)
   */
  'aria-invalid'?: boolean;
  /**
   * HTML ID of an element containing an error message related to the input.
   * <br>
   * [Props from React select was used](https://react-select.com/props)
   */
  'aria-errormessage'?: string;
```

Same for the other components (except for the link to react-select, which is only needed for "select" components).

packages/utils/src/random-string.ts
```diff
@@ -0,0 +1,5 @@
+import { nanoid } from 'nanoid';
+
+const randomString = (): string => nanoid().replace(/[_-]/g, '').toLowerCase();
+
+export default randomString;
```
emmenko, line 5: Remove, as suggested in the other comment.

emmenko (changes requested):
Thanks for giving this a shot.

Overall looks good, I left some comments/suggestions to improve/simplify things a bit.

packages/utils/src/random-string.ts
codeshifu: Will do

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -388,7 +387,7 @@ export default class AsyncCreatableSelectField extends Component<
     const hasError =
       AsyncCreatableSelectInput.isTouched(this.props.touched) &&
       hasErrors(this.props.errors);
-    const errorFieldId = randomString();
+    const errorFieldId = sequentialId();
```
emmenko, line 390: I would prefer to create a separate function for errors

```js
const sequentialId = createSequentialId('async-creatable-select-field-');

const sequentialErrorsId = createSequentialId('async-creatable-select-field-error-');
```

Same for the other components

packages/components/fields/async-select-field/src/async-select-field.js
```diff
@@ -0,0 +1,360 @@
+import { Component } from 'react';
```
emmenko, line 1: You might want to correctly rebase your branch with `main`, as these components have already been migrated to TS.

packages/calendar-utils/src/calendar-body/calendar-body.tsx
```diff
@@ -44,11 +44,15 @@ ClearSection.displayName = 'ClearSection';
 
 export type TInputProps = {
   /**
-   * a11y attribute to determine if input is valid/not
+   * Indicate if the value entered in the input is invalid.
+   * <br>
+   * [Props from React select was used](https://react-select.com/props)
```
emmenko, line 49: Sorry maybe I wasn't super clear in the previous comments. The link to react-select should only be added in case the prop is indeed passed to a react-select component.

emmenko (changes requested):
Thanks, looking better now. There are still some bad merge conflicts to fix and some other minor things (see comments).

Thanks

packages/components/fields/search-select-field/src/search-select-field.tsx
```diff
@@ -34,6 +34,7 @@ type TErrorRenderer = (key: string, error?: boolean) => ReactNode;
 const hasErrors = (errors?: TFieldErrors) =>
   errors && Object.values(errors).some(Boolean);
 const sequentialId = createSequentialId('async-select-field-');
```
emmenko, line 36: This was wrong before. Can you fix this as well?

```suggestion
const sequentialId = createSequentialId('search-select-field-');
```

packages/components/fields/search-select-field/src/search-select-field.tsx
```diff
@@ -34,6 +34,7 @@ type TErrorRenderer = (key: string, error?: boolean) => ReactNode;
...
+const sequentialErrorsId = createSequentialId('async-select-field-error-')();
```
emmenko, line 37:
```suggestion
const sequentialErrorsId = createSequentialId('searc-select-field-error-')();
```

packages/components/inputs/radio-input/src/radio-group.tsx
```diff
@@ -104,15 +112,23 @@ const Group = (props: TGroupProps) => {
   });
   if (props.direction === 'inline') {
     return (
-      <div aria-labelledby={props.id}>
+      <div
+        aria-labelledby={props.id}
+        aria-invalid={props['aria-invalid']}
+        aria-errormessage={props['aria-errormessage']}
```
emmenko, line 118: I think these aria attributes can only be used in `input` elements.

emmenko (commented):
Much better now, thanks!

Please remember to also add a changeset.

packages/components/inputs/radio-input/src/radio-group.tsx
codeshifu: From MDN 
```
The aria-invalid attribute can be used with any typical HTML form element, and is not limited to elements that have an ARIA role assigned.
```
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid

If you still want me to remove it, please let me know

emmenko: Ah thanks, I missed that info. All good then 👌

emmenko (approved):
I added the changeset. From my side this is good to go now, thanks!!

Rhotimee (commented):
Good job @codeshifu. 
Please run `yarn generate-readmes` to update the docs. 
Thanks.

CarlosCortizasCT (approved):
awesome job @codeshifu 👍

PR #2124, opened 2022-02-22
chore(LinkButton): add story with a deprecation note

### LINKED ISSUE #1848: <LinkButton /> deprecation is not documented
**Describe the bug**
`<LinkButton />` deprecation is not documented, and no alternative is proposed.

**To Reproduce**

1. Go to https://uikit.commercetools.com/?path=/story/components-buttons--linkbutton

**Expected behavior**
The deprecation is not documented. I was aware of it via a console warning and couldn't know the alternative until checking https://github.com/commercetools/ui-kit/pull/1169.

**Screenshots**

Screenshot 2021-04-06 at 12 35 14

kark:
closes #1848

Comments:
- I assumed that README with full API this component should not be added
- I am not changing the component by adding `as` prop as suggested [in the comment](https://github.com/commercetools/ui-kit/issues/1848#issuecomment-815696476)
- I am assuming no `changeset` is required
- Apparently, with `LinkButton` implementation as it is now, there is an error while passing `isDisabled` prop to `styled.a` emotion component:
 Screenshot 2022-02-21 at 17 45 59
The solution I figured out is changing the implementation of the `StyledExternalLink` component from:

```jsx
const StyledExternalLink = styled.a<
  Pick<TLinkButtonProps, 'to' | 'isDisabled'>
>`
   ...
`
```

to:

```jsx
import isPropValid from '@emotion/is-prop-valid';

const StyledExternalLink = styled('a', {
  shouldForwardProp: isPropValid,
})<Pick<TLinkButtonProps, 'to' | 'isDisabled'>>`
   ...
`
```
However, considering that this is a deprecated component and this change would be out of the scope of this PR I decided not to introduce that change. If in your opinion this is something worth pursuing I'll be happy to add that.


packages/components/buttons/link-button/src/link-button.story.js
```diff
@@ -0,0 +1,33 @@
...
+          to={text('to', '/foo/bar')}
+          iconLeft={createElement(
+            icons[select('icon', iconNames, iconNames[0])]
+          )}
+          isDisabled={boolean('isDisabled', false)}
+          isExternal={boolean('isExternal', false)}
+        />
+      </Section>
+    </Router>
+  ));
```
emmenko, line 33: I don't think we need this in the story. The readme page is enough.

emmenko (commented):
Thanks

packages/components/buttons/link-button/src/link-button.story.js
kark: I see, I used `noop` instead

emmenko (approved):
Thanks!

PR #2125, opened 2022-02-22
refactor(i18n): migrate to TypeScript

emmenko:


packages/i18n/src/index.js
```diff
@@ -1,10 +0,0 @@
-import en from '../data/en.json';
-import de from '../data/de.json';
-import es from '../data/es.json';
-import frFR from '../data/fr-FR.json';
-import zhCN from '../data/zh-CN.json';
-import ja from '../data/ja.json';
-
-export { en, de, es, frFR, zhCN, ja };
```
emmenko, line 8: We can removed them because we always import the JSON files directly and never through the package entry point.

Rhotimee:
Isn't there a need to also migrate `transifex-transformer.js` to typescript?

Rhotimee:
Thanks for the explanation @CarlosCortizasCT

Rhotimee (approved):

CarlosCortizasCT (approved):
LGTM

PR #2128, opened 2022-02-23
chore: add `react-intl` as a missing dependency

### LINKED ISSUE #2101: Minor, dependencies:  loading-spinner or text do not provide react-intl
**Describe the bug**
Getting a warning in yarn 3 with pnp:
`@commercetools-uikit/loading-spinner@npm:12.2.9 [2ef48] doesn't provide react-intl (p61d0f), requested by @commercetools-uikit/text`

**To Reproduce**

1. Have a `package.json` indirectly or directly using the above components
2. `yarn install`

**Expected behavior**
No warning. 

**Additional context**
Not a priority, just tuning

kark:
Closes #2101

In my PR I am attempting to cover all cases where `react-intl` is a missing peer dependency originating from using `@commercetools-uikit/text` package.
 
In commit 4ecd50a0 I am adding `react-intl` to all packages that are using `@commercetools-uikit/text` package **directly** as a dependency.
This is the list of packages:
```
@commercetools-uikit/flat-button
@commercetools-uikit/icon-button
@commercetools-uikit/primary-button
@commercetools-uikit/secondary-icon-button
@commercetools-uikit/collapsible-panel
@commercetools-uikit/field-label
@commercetools-uikit/checkbox-input
@commercetools-uikit/loading-spinner
@commercetools-uikit/messages
@commercetools-uikit/primary-action-dropdown
@commercetools-uikit/tag
```
In commit c8ae3efd I am adding `react-intl` to all packages that are using `@commercetools-uikit/text` package **indirectly** as a dependency.
This is the list of packages:
```
@commercetools-uikit/data-table
@commercetools-uikit/async-creatable-select-field
@commercetools-uikit/async-select-field
@commercetools-uikit/creatable-select-field
@commercetools-uikit/date-field
@commercetools-uikit/date-range-field
@commercetools-uikit/date-time-field
@commercetools-uikit/localized-multiline-text-field
@commercetools-uikit/localized-text-field
@commercetools-uikit/money-field
@commercetools-uikit/multiline-text-field
@commercetools-uikit/number-field
@commercetools-uikit/radio-field
@commercetools-uikit/search-select-field
@commercetools-uikit/select-field
@commercetools-uikit/text-field
@commercetools-uikit/time-field
```


packages/components/buttons/icon-button/package.json
```diff
@@ -33,9 +33,11 @@
     "prop-types": "15.8.1"
   },
   "devDependencies": {
-    "react": "17.0.2"
+    "react": "17.0.2",
+    "react-intl": "5.24.6"
```
CarlosCortizasCT, line 37: I'm confused because I don't understand why we need the dependency both as _dev_ and _peer_. I mean, if we declare the dependency in the _dev_ section, we wouldn't need to add it in the _peer_ as the package would already include it on its own.

In the other hand, I guess we talked about declaring the `react-intl` package in the _peer_ section. This way, we're delegating in the consumers to provide its own version of that package instead of include it in all of the ui-kit packages.
This way we would be following the intent of the original ui-kit package (eg: `text`) by not including `react/intl` in the ui-kit packages.

Maybe I missed something along the way and there is a reason for current code proposal.

kark: Sorry, I could have used author's comment for that. I understand your confusion, I am confused too 😅
However, upon adding a sheer peer dependency like:
`yarn workspace @commercetools-uikit/icon-button add --peer react-intl@5.x`
I get an error:
```bash
error @commercetools-uikit/icon-button has a peerDependency on react-intl but it is not also specified in devDependencies, please add it there.
```
Therefore, (as the error msg suggests) I added it as a dev dependency too (following the pattern in all other places where `react-inlt` has already been added)

CarlosCortizasCT: I'm sorry but I'm still confused.
I think having `react-intl` both in _dev_ and _peer_ dependencies is not what we want.

When running the command from the package directory it doesn't log that error:
```
local:~/Development/ui-kit/packages/components/buttons/icon-button (cm-ts-presets) $ yarn add --peer react-intl@5.x
```

I'm not very familiar yet with yarn and monorepos so maybe @emmenko could help us understand this one.

emmenko: Hey guys, I thought a bit about this and I think (at least for now) it would be enough to treat these peer dependency requirements as normal dependencies.

So for example the `icon-button` package will have `react-intl` as a normal dependency (using `^` caret version for flexibility). Same for all other packages that for instance require the `text` package which has `react-intl` as a peer dependency.

kark: @emmenko I'm sorry, I'm having issues again 😕
```bash
yarn workspace @commercetools-uikit/flat-button add react-intl@^5.24.6
```
results in
```bash
☔️ error @commercetools-uikit/flat-button has a dependency on react-intl@^5.24.6 but the most common range in the repo is 5.24.6, the range should be set to 5.24.6
```
Would you like me rather to update all current packages using `react-intl` to use `^5.24.6` or should I simply drop the caret in the newly updated packages?

emmenko: Use `^` everywhere, thanks

CarlosCortizasCT: Checking again the package.json files I realized we have some packages where `react-intl` in defined in the `dependencies` group and others where it is in the `devDependencies` group and I was wondering why we have that difference but don't seem to find an explanation.

@emmenko Do you have any info about this? Should we standarize where that dependency should belong to or is it ok to have it like this?

.changeset/cold-cougars-rest.md
```diff
@@ -0,0 +1,32 @@
+---
```
emmenko, line 1: Why do we have 2 changelogs?

emmenko (commented):
Thanks, looks good now. Just fix the changelogs and we're good to go.

.changeset/cold-cougars-rest.md
kark: One reflects upgrade in all components to `^` version of the package, the other one reflects addition of `react-intl` wherever needed. I thought that was a common pattern to differentiate changes in changesets 🤔

packages/components/buttons/icon-button/package.json
emmenko: When it's a devDependency it's because the package is defined as a peerDependency.

.changeset/cold-cougars-rest.md
emmenko: I see. Well in this case it's fine to have 1 changeset, as it's the same context for the changes.

Multiple changesets are useful to differentiate unrelated changes in different packages.

packages/components/buttons/icon-button/package.json
CarlosCortizasCT: Ok. I was confused because `react-intl` will be anyway a _runtime dependency_ as it is needed by the component but I get it.

Thanks for the explanation.

emmenko (approved):
🎉

CarlosCortizasCT (approved):

Issue #2120, opened 2022-02-21
Upgrade `slate-*` to latest versions


emmenko:
Currently we use the following versions of [Slate](https://docs.slatejs.org):

* `slate`: `0.47.9`
* `slate-html-serializer`: `0.8.13`
* `slate-react`: `0.22.10`

These packages haven't been updated for a long time and they are currently ignored from Renovate updates (I can't remember exactly why).

The latest versions of these packages are:

* `slate`: `0.72.8`
* `slate-html-serializer`: `0.8.13`
* `slate-react`: `0.72.9`

According to the [changelog](https://docs.slatejs.org/general/changelog) many `minor` version updates contain some breaking changes, so upgrading is not that easy.

Their docs also state: https://docs.slatejs.org/concepts/xx-migrating

> Migrating from earlier versions of Slate to the `0.50.x` versions is not a simple task. The entire framework was re-considered from the ground up.

Question is: how much effort do we want to put into this? Staying on older versions might cause issues in the future...

PR #2129, opened 2022-02-24
chore(rich-text-input): migrate to TypeScript

kark:
#### Summary

Migrate rich-text-input to TypeScript
Make `style` prop of rich-text-editor-body optional


@types/slate-react/index.d.ts
```diff
@@ -0,0 +1,68 @@
...
+    name: unknown;
+    onFocus: unknown;
+    onBlur: unknown;
+    disabled: unknown;
+    readOnly: unknown;
+    value: unknown;
+    options: unknown;
+    onChange: unknown;
+    plugins: unknown;
+    renderEditor: unknown;
```
kark, line 20: I’ve decided to leave these types as they are. I guess copy-pasting them here won’t give us any more type safety, will bloat this declaration file and we have a task for updating this package anyway (#2120)

packages/components/inputs/rich-text-input/README.md
```diff
@@ -1,49 +1,95 @@
...
+
+```
+npm --save install react react-intl
+```
 
 ## Usage
 
-```js
+```jsx
+import React from 'react';
```
kark, line 33: I don’t know how the other files were committed but `eslint` prevented me from committing without adding this line

packages/components/inputs/rich-text-input/src/editor.tsx
```diff
@@ -56,23 +106,21 @@ const Editor = (props) => {
...
-                isDisabled={props.isDisabled}
-                isReadOnly={props.isReadOnly}
-              >
+              <EditorWrapper isDisabled={props.isDisabled}>
                 <RichTextBody
-                  ref={{
-                    containerRef: ref,
-                    registerContentNode,
-                  }}
+                  ref={refObj as unknown as Ref<TRichtTextEditorBodyRef>}
```
kark, line 118: I'm sorry for using the assertion but I don’t know how to make it work otherwise

packages/components/inputs/rich-text-input/src/editor.tsx
```diff
@@ -56,23 +106,21 @@ const Editor = (props) => {
...
+        const refObj: TRichtTextEditorBodyRef = {
+          containerRef: ref,
+          registerContentNode,
+        };
         return (
           <Constraints.Horizontal max={props.horizontalConstraint}>
             <Stack scale="xs" alignItems="flex-end">
-              <EditorWrapper
-                isDisabled={props.isDisabled}
-                isReadOnly={props.isReadOnly}
```
kark, line 64: this prop turned out to be unused

packages/components/inputs/rich-text-input/docs/usage-example.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
kark, line 1: `eslint` prevented me from committing without adding this line

packages/components/inputs/rich-text-input/src/editor.tsx
```diff
@@ -56,23 +106,21 @@ const Editor = (props) => {
...
-                    containerRef: ref,
-                    registerContentNode,
-                  }}
+                  ref={refObj as unknown as Ref<TRichtTextEditorBodyRef>}
                   hasError={props.hasError}
                   isDisabled={props.isDisabled}
                   hasWarning={props.hasWarning}
-                  isReadOnly={props.isReadOnly}
-                  showExpandIcon={props.showExpandIcon}
+                  isReadOnly={Boolean(props.isReadOnly)}
```
CarlosCortizasCT, line 122: I'm curious about these castings as both props are already defined as boolean.

packages/components/inputs/rich-text-input/src/rich-text-input.tsx
```diff
@@ -0,0 +1,254 @@
...
+  > = {
+    defaultExpandMultilineText: false,
+    horizontalConstraint: 'scale',
+    placeholder: '',
+    showExpandIcon: false,
+  };
+
+  static displayName = 'RichTextInput';
+  static isEmpty = isEmpty;
+  static isTouched = (touched?: boolean) => Boolean(touched);
```
CarlosCortizasCT, line 122: Maybe we should need to allow both `boolean` and `boolean[]` types for the way this function is used from formik state.
I saw that pattern in other PRs lile [this one](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/select-input/src/select-input.tsx#L497).

packages/components/inputs/rich-text-input/src/editor.tsx
kark: Thanks, I had to do that since both props are optional (so the type is `boolean | undefined`)

CarlosCortizasCT: Ah, ok! They are optional props of the `editor.tsx` but are mandatory props in the `RichTextBody`. Now I see.

Thanks.

@types/slate-react/index.d.ts
CarlosCortizasCT: Since the `Editor` component is using the `RichTextBody` component internally and we already have defined [the types](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/editor.types.ts) in that package, I was wondering if we import those types here instead of creating this file for consistency.
_(we would need to export them from the `rich-text-utils` package)_

packages/components/inputs/rich-text-input/src/editor.tsx
emmenko: Actually I think we need to fix that, as according to the localized-richt-text-input we do use it

https://github.com/commercetools/ui-kit/blob/c485b1899cc57a053f924008d7ec688534b08330/packages/components/inputs/localized-rich-text-input/src/editor.styles.js#L56

Can you fix it for this component as well? Thanks

packages/components/inputs/rich-text-input/src/editor.tsx
emmenko: Yeah that's fine, this `ref` is kind of a mess anyway.

packages/components/inputs/rich-text-input/src/editor.tsx
```diff
@@ -107,8 +153,24 @@ const Editor = (props) => {
   );
 };
 
-// eslint-disable-next-line react/display-name
-const renderEditor = (props, editor, next) => {
+type TEditorOptionsProps = {
+  options: {
```
emmenko, line 157: Nit: it's a bit weird to type it like this. I would expect the `options` property to not be part of this type, then do

```ts
props: TEditorProps & { options: TEditorOptionsProps }
```

@types/slate-react/index.d.ts
kark: I'm wondering if I understand correctly, what you are suggesting is copy-pasting `TEditor` type from `rich-text-utils` to `slate-react` type declaration file and finally importing this type from `slate-react` package in both `rich-text-utils` and `rich-text-input`?

@types/slate-react/index.d.ts
```diff
@@ -0,0 +1,68 @@
+declare module 'slate-react' {
```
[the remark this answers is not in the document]
kark: I'm a bit lost here 😅 I'm not sure if I understand.
We are using `slate-react` [here](https://github.com/commercetools/ui-kit/blob/8ac71c4ef1ea27566a82fd3cafbca3ad7ddda871/packages/components/inputs/rich-text-input/src/rich-text-input.tsx#L2) and that was one of the ways of getting type declarations. Considering that we have recently typed [`TEditor`](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/editor.types.ts) on our own that was the way I thought of. Of course only as a temporary solution easily removable once we migrate to the latest (TS-based) version of `slate-`. The other option was using `// @ts-ignore` but I assumed that was out of the question 😅
I am wondering if you don't approve of that at all? In that case I would need a suggestion what would be the appropriate way

packages/components/inputs/rich-text-input/src/editor.tsx
kark: true 🙂 updated

@types/slate-react/index.d.ts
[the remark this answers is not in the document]
emmenko: FYI: @kark and me took a deeper look at this and decided to remove the type declarations and have a wrapper (typed) component for the `Editor` component (`slate-react`).

This should be all we need, as the `TEditor` object is not the same as the component and it can stay in the utils package.

packages/components/inputs/rich-text-input/src/editor.types.ts
```diff
@@ -0,0 +1,111 @@
...
+  wrapBlock: (blockTag: string) => void;
+};
+
+type TNode = {
+  texts: (nodeParam?: Record<string, string>) => TNodeTextsResult[];
+};
+
+type TNodeTextsResult = [{ key: string; text: string }, string];
+
+export type TEditor = {
```
Rhotimee, line 69: I feel it will be better to expose `TEditor` and some of the other types in this file from  `rich-text-utils` instead of duplicating the types here.

kark: Having in mind it's only a temporary solution and I took advantage of @emmenko's [freedom of choice](https://github.com/commercetools/ui-kit/pull/2129#discussion_r814594227) 😅

CarlosCortizasCT: I understand @Rhotimee concern but I also agree this is a good solution as well.

If we wanted to use `rich-test-utils` types we would need to modify it to expose them, but then we would have to remove that exposure when the [`react-slate` update](https://github.com/commercetools/ui-kit/issues/2120) is finished.

So I think this time a little bit of duplication is a better choice.

emmenko: Good point about removing them again. Ok let's just keep them duplicated for now.

packages/components/inputs/rich-text-input/src/editor.styles.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable import/prefer-default-export */
```
Rhotimee, line 1: Is this eslint disable comment necessary? It will be nice to remove it if it is not.

packages/components/inputs/rich-text-input/src/editor.styles.ts
```diff
@@ -0,0 +1,12 @@
+import styled from '@emotion/styled';
+import type { TEditorProps } from './editor';
+
+type TEditorWrapperProps = Pick<TEditorProps, 'isDisabled' | 'isReadOnly'>;
+
+const EditorWrapper = styled.div<TEditorWrapperProps>`
+  align-self: stretch;
+  cursor: ${(props) =>
+    props.isDisabled || props.isReadOnly ? 'not-allowed' : 'inherit'};
```
CarlosCortizasCT, line 9: In the previous version we were not using the `not-allowed` cursor for the `isReadOnly` state so I was wondering if this is an intended change.
I don't see we use it like that in other components (eg [TextInput](https://uikit.commercetools.com/?path=/story/components-inputs--textinput))

emmenko: I asked to add it here as well for consistency. https://github.com/commercetools/ui-kit/pull/2129#discussion_r814589790

CarlosCortizasCT: Oh! Ok. Sorry.

CarlosCortizasCT (approved):

Rhotimee (approved):

emmenko (approved):
Thanks!

PR #2130, opened 2022-02-24
Adjust radio option layout

ddouglasz:
Provision for additional text in radio options component such that the radio button always aligns with the label of the content.

image


packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -18,6 +24,15 @@ const LabelTextWrapper = styled.div<TStylesProps>`
       : customProperties.fontColorForInput};
 `;
 
+const RadioOptionsWrapper = styled.div<TStylesProps>`
+  display: grid;
+  grid-template-columns: auto;
+  grid-template-rows: auto;
+  grid-template-areas:
+    'radio label'
+    '. content';
```
emmenko, line 33: I'm not sure it makes sense to use named areas here. For instance, if the same component is rendered multiple times in a page, wouldn't there be conflicts with the names?

I believe it should be possible to achieve the same with only `grid-template-columns` and `grid-template-rows`.

packages/components/inputs/radio-input/src/radio-option.tsx
```diff
@@ -34,6 +36,8 @@ export type TOptionProps = {
   value: string | boolean;
   children: string | ReactElement | (() => ReactElement);
   components?: TComponents;
+  additionalContent?: ReactNode;
+  gridLabel?: string;
```
emmenko, line 40: I think this can be removed.

.changeset/serious-geckos-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/radio-input': patch
+---
+
+adjust radio options layout
```
emmenko, line 5: Please explain a bit more the changes and how to use the new `additionalContent` prop.

emmenko (commented):
Looking good in general, just some minor things to improve.

.changeset/serious-geckos-grab.md
ddouglasz: Thanks for all the feedback. I will update soon.

CarlosCortizasCT:
Maybe we could extend the Storybook story to show an example of the visualization of the new capability.

What do you think?

ddouglasz:
> Maybe we could extend the Storybook story to show an example of the visualization of the new capability.
> 
> What do you think?

I think that makes sense. I would be a great way to display how to take advantage of the new capability.

FilPob:
Hi guys, Can I still ask for one thing? To always include a 4px margin between those two containers? Otherwise they look to be too close to each other. 
image

packages/components/inputs/radio-input/src/radio-input.story.js
```diff
@@ -101,6 +101,7 @@ storiesOf('Components|Inputs', module)
                 radioOption3
               )}
               isHovered={boolean('isHovered #3', false, radioOption3)}
+              additionalContent="Lorem ipsum dolor sit amet, consetetur sadipscing elitr"
```
CarlosCortizasCT, line 104: I think it would be great to also add a new use case to the [visualroute tests](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/radio-input/src/radio-input.visualroute.js) so we can have a regression test for this new type of visualization.

packages/components/inputs/radio-input/src/radio-option.tsx
```diff
@@ -17,7 +18,10 @@ import {
   getLabelStyles,
   getContainerStyles,
   LabelTextWrapper,
+  RadioOptionsWrapper,
+  AdditionalTextWrapper,
 } from './radio-option.styles';
+import Spacings from '../../../../../presets/spacings';
```
CarlosCortizasCT, line 24: We need an absolute `import` statement here.

```suggestion
import Spacings from '@commercetools-uikit/spacings';
```

.changeset/serious-geckos-grab.md
emmenko: There are duplicated changesets. Remove one, thanks

.changeset/hungry-ties-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/radio-input': patch
+---
+
+adjust radio options layout such that it can take a prop when there is additional content to be rendered. Pass a prop `additionalContent` to the `RadioInput.Option` component. It can be a string or react component.
```
emmenko, line 5:
````suggestion
In case you need to render additional information to a radio option label, we recommend now to use a new prop `additionalContext` instead of rendering everything in the `<RadioOption>`'s `children`. This ensures that the radio input and the label are correctly aligned.

// Before
```js
<RadioInput.Option>
  <Spacings.Stack>
    <Text.Body>The label<Text.Body>
    <Text.Detail>Additional information</Text.Detail>
  </Spacings.Stack>
</RadioInput.Option>
```

// After
```js
<RadioInput.Option additionalContent={<Text.Detail>Additional information</Text.Detail>}>
  <Text.Body>The label<Text.Body>
</RadioInput.Option>
```

````

emmenko (commented):
Besides a couple of minor things, add a VRT and then we are good to go. Thanks

CarlosCortizasCT (approved):

PR #2133, opened 2022-02-28
refactor(localized-rich-text-input): Migrate to typescript

Rhotimee:
#### Summary
Migrate localized-rich-text-input to typescript


packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -239,13 +319,13 @@ const renderEditor = (props, editor, next) => {
...
 
-  const isFocused = props.editor.value.selection.isFocused;
+  const isFocused = props.editor?.value.selection.isFocused;
 
   return (
     <Editor editor={editor} {...passedProps}>
       {children}
       <HiddenInput
-        isFocused={isFocused}
+        isFocused={Boolean(isFocused)}
```
ddouglasz, line 328: `isFocused` seems like it's already a boolean from `editor` props but it's been casted here.

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
```diff
@@ -1,14 +1,60 @@
-/* eslint-disable react/prop-types */
-import { PureComponent } from 'react';
-import PropTypes from 'prop-types';
+import { PureComponent, ReactNode } from 'react';
```
CarlosCortizasCT, line 1: Maybe we could add `type` keyword to `ReactNode`.

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
```diff
@@ -162,6 +247,7 @@ const LocalizedRichTextInput = (props) => {
           >
             <LocalizedInputToggle
               isOpen={areLanguagesOpened}
+              // @ts-ignore
```
ddouglasz, line 250: Is this still useful here? If yes, would be nice to have a little comment for having it there.

ddouglasz (changes requested):

packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -44,10 +54,45 @@ const Row = styled.div`
...
-const Editor = (props) => {
+export type TEditorProps = {
+  children: ReactNode;
+  id?: string;
+  isOpen: boolean;
+  isDisabled?: boolean;
+  isReadOnly?: boolean;
+  hasWarning?: boolean;
+  hasError?: boolean;
+  editor?: TEditor;
```
kark, line 65: perhaps it's worth marking this prop as required. It would help out with e.g. `editor={props.editor as TEditor}` in line 185 and seems this is what `RichTextBody` is expecting really

packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -17,6 +26,7 @@ import { messagesMultilineInput } from '@commercetools-uikit/input-utils';
 import {
   RichTextBody,
   HiddenInput,
+  type TEditor,
```
kark, line 29: following the comment from Carlos I guess [this approach](https://github.com/commercetools/ui-kit/pull/2129#discussion_r814795761) will make us less coupled with `rich-text-utils` later on

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
```diff
@@ -17,10 +14,80 @@ import {
...
+type TEvent = {
+  target: {
+    id?: string;
+    name?: string;
+    language?: string;
+    value?: string;
+  };
+};
+
+type TLocalizedRichTextInputProps = {
```
kark, line 39: I'm wondering in which cases we need to add JSDoc comments? If this is not the case we need them, please don't bother

packages/components/inputs/localized-rich-text-input/src/editor.tsx
Rhotimee: `isFocused` is required on HiddenInput on optional on Editor. That is why casting is necessary there.

ddouglasz: Oh, thanks!

packages/components/inputs/localized-rich-text-input/src/editor.tsx
Rhotimee: That means every other component dependent on the Editor component will have the editor prop as required also. 
I don't think that is the intended functionality.

packages/components/inputs/localized-rich-text-input/src/editor.tsx
Rhotimee: I've implemented the approach. Thanks.

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
Rhotimee: updated.

emmenko: Generally it's for generating the component docs in the readme.

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
```diff
@@ -1,14 +1,57 @@
-/* eslint-disable react/prop-types */
-import { PureComponent } from 'react';
-import PropTypes from 'prop-types';
+import { PureComponent, type ReactNode } from 'react';
 import pick from 'lodash/pick';
-import requiredIf from 'react-required-if';
+import type { Value } from 'slate';
+// TODO: remove after upgrade of `slate-react` to the latest version
+// @ts-ignore
 import { Editor } from 'slate-react';
```
emmenko, line 6: The wrapper component also needs to be defined. See https://github.com/commercetools/ui-kit/pull/2129/files#diff-c5851815d5a8e6ae99dfe4d2298aa2c3c9a45f8642a3c9d43b09afed31e59ccfR33-R37

emmenko (commented):
Looks good overall, just one leftover

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
emmenko: I'll take care of it and merge the PR afterwards

Rhotimee: Alright, thanks 👍🏽

emmenko (approved):

CarlosCortizasCT (approved):

PR #2136, opened 2022-03-03
chore: prepare v14 release

emmenko:
This is a proposal to realign all package versions given that we had an accidental major release (v13), resulting in some packages still being at version v12.

Additionally, given that we migrated all packages to TypeScript, I think it would also make sense to "mark" this step with a new release.

Let me know what you think.


design-system/package.json
```diff
@@ -23,7 +23,7 @@
     "build:tokens:watch": "nodemon -e yaml --quiet --watch materials/internals --exec 'yarn build:tokens'"
   },
   "dependencies": {
-    "@babel/runtime": "7.17.2",
-    "@babel/runtime-corejs3": "7.17.2"
+    "@babel/runtime": "^7.17.2",
+    "@babel/runtime-corejs3": "^7.17.2"
```
emmenko, line 27: I took the chance also to "loosen up" the version range a bit, to avoid possible version conflicts from consumers.

.changeset/tiny-suits-grow.md
```diff
@@ -0,0 +1,96 @@
...
+'@commercetools-uikit/utils': major
+'@commercetools-uikit/buttons': major
+'@commercetools-uikit/fields': major
+'@commercetools-uikit/inputs': major
+'@commercetools-uikit/spacings': major
+'@commercetools-frontend/ui-kit': major
+'visual-testing-app': major
+---
+
+This release does not introduce any breaking changes but marks a new starting point with all packages having been migrated to TypeScript.
```
tdeekens, line 96:
```suggestion
This release does not introduce any breaking changes but intends to homogenize versions across all packages with the migration to TypeScript being completed. Having similar versions across packages should ease maintenance and updating from now on.
```

CarlosCortizasCT:
It seems reasonable to me. 👍

Rhotimee (approved):
💯

CarlosCortizasCT (approved):

ddouglasz (approved):

PR #2145, opened 2022-03-10
Add CI job for stale bot

mmaltsev-ct:
Added action to mark issues as stale after 14 days of inactivity and close them after 14 more days

### DIFF
--- a/.github/workflows/close-inactive-issues.yml
+++ b/.github/workflows/close-inactive-issues.yml
@@ -12,7 +12,8 @@ jobs:
       - uses: actions/stale@v3
         with:
           any-of-labels: "Status: Waiting for Customer"
-          start-date: "2022-04-06 00:00:00.000",
+          start-date: "2022-04-06 00:00:00.000"
+          debug-only: true
           days-before-issue-stale: 14
           days-before-issue-close: 14
           stale-issue-label: "stale"
--- a/.github/workflows/close-inactive-issues.yml
+++ b/.github/workflows/close-inactive-issues.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          any-of-labels: "Status: Waiting for Customer"
+          any-of-labels: "⌛ Status: Waiting for Customer"
           start-date: "2022-04-06 00:00:00.000"
           debug-only: true
           days-before-issue-stale: 14
--- a/.github/workflows/close-inactive-issues.yml
+++ b/.github/workflows/close-inactive-issues.yml
@@ -1,0 +1,23 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          any-of-labels: "Status: Waiting for Customer"
+          start-date: "2022-04-06 00:00:00.000",
+          days-before-issue-stale: 14
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue has been marked as 'stale' due to its inactivity for the last 14 days. Please, add a new comment to keep the conversation going; otherwise, the issue will be closed in two weeks."
+          close-issue-message: "This issue has been closed due to its inactivity for the last month. Please, feel free to reopen it and add a new comment in case you think the problem has not been resolved."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,22 @@ on:
   pull_request:
 
 jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          any-of-labels: "Status: Waiting for Customer"
+          days-before-issue-stale: 14
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 14 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
   build_lint_and_test:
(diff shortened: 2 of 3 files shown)


tdeekens (approved):
Nice.

.github/workflows/main.yml
```diff
@@ -11,6 +11,22 @@ on:
   pull_request:
 
 jobs:
+  close-issues:
```
emmenko, line 14: I would define this is a separate workflow and use a schedule cron trigger like in the github example.

.github/workflows/main.yml
```diff
@@ -11,6 +11,22 @@ on:
...
+      issues: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          any-of-labels: "Status: Waiting for Customer"
+          days-before-issue-stale: 14
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 14 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
```
emmenko, line 26: Nit: I think we should provide a bit more info in the message.

emmenko (commented):
Thanks for looking into setting this up.

.github/workflows/main.yml
CarlosCortizasCT: So here are my proposals:

*stale-issue-message*
> This issue has been marked as "stale" due to its inactivity for the last 14 days.
Please, add a new comment to keep the conversation ongoing; otherwise, the issue will be closed in two weeks.

*close-issue-message*
> This issue has been closed due to its inactivity for the last month.
Please, feel free to reopen it and add a new comment in case you think the problem has not been solved.

.github/workflows/main.yml
```diff
@@ -11,6 +11,22 @@ on:
...
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
```
CarlosCortizasCT, line 28: I wouldn't activate this action for pull requests but only for issues.

.github/workflows/main.yml
```diff
@@ -11,6 +11,22 @@ on:
...
 
 jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          any-of-labels: "Status: Waiting for Customer"
```
CarlosCortizasCT, line 21: Maybe we could add the [`start-date`](https://github.com/actions/stale#start-date) parameter if we don't want current issues to be analyzed by this action.

.github/workflows/main.yml
mmaltsev-ct: Agree, updated in the new commit.

.github/workflows/main.yml
mmaltsev-ct: @CarlosCortizasCT thanks for your suggestion. Updated it just a little bit, added to the new commit.

.github/workflows/main.yml
mmaltsev-ct: Added 'start-date' property. Not sure about the necessity of 'debug-only', since we anyway need to add a specific label for the flow to be triggered.

CarlosCortizasCT: Yes, the `debug-only` would be to just play around before actually changing issues status.
You could add the label and have this new workflow log what it would do so we could check that it would be what is expected.
Anyway, it was just an idea.

.github/workflows/main.yml
CarlosCortizasCT, line 14: Now that we added a new workflow in `close-inactive-issues.yml` I guess we should remove the configuration in this file (`main.yml`).

emmenko, line 14: FYI: this job can be removed now from the main workflow

mmaltsev-ct: done, thanks for a nice catch!

mmaltsev-ct: done, thanks for a nice catch!

.github/workflows/main.yml
mmaltsev-ct: added the property, let's see how it works

.github/workflows/close-inactive-issues.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          any-of-labels: "Status: Waiting for Customer"
```
emmenko, line 14: FYI: I assume this label must exist.

For managing labels, we use this config file: https://github.com/commercetools/ui-kit/blob/main/.github-labelsrc.json
I guess we can go with the black color

https://github.com/commercetools/ui-kit/blob/25e3e2f7e3d4b231705e89592ef6bbf306864701/.github-labelsrc.json#L160-L161

There is a command `yarn labels:sync` that creates the github labels based on the config. You need to have a `AUTH_TOKEN` environment variable to make the command work. See https://github.com/commercetools/github-labels/

emmenko (commented):
Thanks, looks good overall!

I left a suggestion to create the new github label.

.github/workflows/close-inactive-issues.yml
mmaltsev-ct: Yeap, thanks, the label should be present for sure. Don't you think, creating it [directly in the repo](https://github.com/commercetools/ui-kit/issues/labels) is enough?

mmaltsev-ct: Added new label

CarlosCortizasCT, line 14: I think we might have to add the emoji also here because, from what I can tell after taking a look at [the action code](https://github.com/actions/stale/blob/57708ae6dd2006b5642c4e1e88bac7efb3d9382e/src/classes/issues-processor.ts#L353), it will search for a label which have the same exact name

CarlosCortizasCT: @mmaltsev-ct did you find this change is not needed?

CarlosCortizasCT (commented):
There is a comment it would be great if it can be reviewed.

.github/workflows/close-inactive-issues.yml
mmaltsev-ct: @CarlosCortizasCT no, I think I just somehow missed your comment 🤷‍♂️. Nice catch, thank you!

emmenko (approved):
🚀 

PS: remember that this is enabled in "dry run" mode at the moment.

CarlosCortizasCT (approved):
💯

PR #2147, opened 2022-03-14
refactor: refine event handler types and other minor inconsistencies

emmenko:
This is a follow up of the recent TS migration.

I noticed there were some minor inconsistencies or leftovers across several components so I went through them to clean things up.


packages/components/inputs/date-input/src/date-input.tsx
```diff
@@ -279,7 +284,7 @@ const DateInput = (props: TDateInput) => {
                       clearSelection();
                     }
                     // ArrowDown
-                    if (event.keyCode === 40) {
+                    if (event.key === 'ArrowDown') {
```
CarlosCortizasCT, line 287: I was wondering about our browsers version support policy as I read this [in MDN](https://devdocs.io/dom/keyboardevent/key/key_values):
image

Do we publicly state which browsers & versions we support?

packages/components/inputs/number-input/src/number-input.tsx
```diff
@@ -1,4 +1,4 @@
-import { ChangeEventHandler, FocusEventHandler } from 'react';
+import { type ChangeEventHandler, type FocusEventHandler } from 'react';
```
CarlosCortizasCT, line 1: nit:
```suggestion
import type { ChangeEventHandler, FocusEventHandler } from 'react';
```

packages/components/inputs/password-input/src/password-input.tsx
```diff
@@ -1,4 +1,4 @@
-import { ChangeEventHandler, FocusEventHandler } from 'react';
+import { type ChangeEventHandler, type FocusEventHandler } from 'react';
```
CarlosCortizasCT, line 1: nit:
```suggestion
import type { ChangeEventHandler, FocusEventHandler } from 'react';
```

packages/components/inputs/text-input/src/text-input.tsx
```diff
@@ -1,4 +1,4 @@
-import { FocusEventHandler, ChangeEventHandler } from 'react';
+import { type FocusEventHandler, type ChangeEventHandler } from 'react';
```
CarlosCortizasCT, line 1: nit:
```suggestion
import type { FocusEventHandler, ChangeEventHandler } from 'react';
```

packages/components/inputs/date-input/src/date-input.tsx
emmenko: We have some info here

https://docs.commercetools.com/custom-applications/help-needed #browser-support

So it should be fine, as those mentioned there are pretty old versions

ddouglasz (approved):
LGTM

PR #2149, opened 2022-03-18
Export Typescript types related to data tables

CarlosCortizasCT:
### Summary

when trying to build a ui to display some data table information with typescript I found some useful types are not exported.

## Description

Some of the types regarding `DataTable` and `DataTableManager` components and also for `useDataTableSortingState' hook are needed to seamlessly integrate those in a new view so the goal of this PR would be to have them publicly exposed for consumers to use them.

### DIFF
--- a/packages/hooks/src/index.ts
+++ b/packages/hooks/src/index.ts
@@ -5,6 +5,6 @@ export { default as useRowSelection } from './use-row-selection';
 export { default as useSorting } from './use-sorting';
 export { default as usePaginationState } from './use-pagination-state';
 export { default as useDataTableSortingState } from './use-data-table-sorting-state';
-export * from './use-data-table-sorting-state';
+export * from './use-data-table-sorting-state/export-types';
 
 export { default as version } from './version';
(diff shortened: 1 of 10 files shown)


ddouglasz (approved):

packages/hooks/src/index.ts
```diff
@@ -5,5 +5,6 @@ export { default as useRowSelection } from './use-row-selection';
 export { default as useSorting } from './use-sorting';
 export { default as usePaginationState } from './use-pagination-state';
 export { default as useDataTableSortingState } from './use-data-table-sorting-state';
+export * from './use-data-table-sorting-state';
```
emmenko, line 8: Hmm I would be more explicit maybe if this is intended to export the types

```suggestion
export * from './use-data-table-sorting-state/export-types';
```

emmenko (commented):
Thanks

packages/hooks/src/index.ts
CarlosCortizasCT: Thanks for the suggestions.

Updated with this commit: b31956abfaa3855ece61e57a532205774607271d

emmenko (approved):

PR #2159, opened 2022-03-30
[Breaking] Upgrade `slate-*` libraries for rich-text inputs

kark:
#### Summary

The aim of this PR was updating all `slate-*` packages to the newest versions and migrating the ui-kit components to use new packages


packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
```diff
@@ -1,5 +1,20 @@
...
+
+beforeEach(async () => {
+  browser = await puppeteer.launch({
+    slowMo: 10,
+  });
+  page = await browser.newPage();
+});
+afterEach(async () => {
+  await browser.close();
+});
```
kark, line 17: This is hardly a solution, but running each of the tests in slow motion makes them pass 🤯 And e.g. with value `5` they don't pass. Seems to be some kind of a race condition but any help with that would be welcome

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
```diff
@@ -40,6 +55,8 @@ describe('LocalizedRichTextInput', () => {
     const doc = await getDocument(page);
     let input = await queries.findByTestId(doc, 'rich-text-data-test-en');
 
+    await input.focus();
```
kark, line 58: There is a change in the default behavior of the slate editor. The way it works now is that editor without any prior focus assumes that point [0,0] is selected. We might toggle the marks but the results won't be visible until the editor gets the focus.
slate-default-selection

After the changes the editor has to explicitly receive focus to apply the initial selection.

packages/components/inputs/rich-text-input/src/rich-text-input.tsx
```diff
@@ -49,25 +21,11 @@ type TChangeEvent = {
...
-  value: ReturnType<typeof html.deserialize>;
-};
-type TOnChangeFn = (change: TOnChangeParam) => unknown;
-
 export type TRichTextInputProps = {
   /**
    * Focus the control when it is mounted
    */
-  isAutofocussed?: boolean;
+  isAutofocused?: TEditorProps['isAutofocused'];
```
CarlosCortizasCT, line 28: It seems this would also be a breaking change since all consumers using the old property name would see their behaviour modified.

packages/components/inputs/rich-text-input/src/rich-text-input.visualspec.js
```diff
@@ -1,5 +1,20 @@
 import percySnapshot from '@percy/puppeteer';
+import puppeteer from 'puppeteer';
 import { getDocument, queries } from 'pptr-testing-library';
+let browser;
+let page;
+
+jest.setTimeout(20000);
```
CarlosCortizasCT, line 7: nit: you could also specify a custom timeout for specific tests: https://jestjs.io/docs/api#testname-fn-timeout

CarlosCortizasCT (commented):
Great job 👍 

I left some comments upon my first review.

packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
```diff
@@ -0,0 +1,240 @@
...
+  const newProperties: Partial<CustomElement> = {
+    type: isActive ? BLOCK_TAGS.p : isList ? BLOCK_TAGS.li : format,
+  };
+  Transforms.setNodes<SlateElement>(editor, newProperties);
+
+  if (!isActive && isList) {
+    const block: CustomElement = { type: format, children: [] };
+    Transforms.wrapNodes(editor, block);
+  }
+};
```
kark, line 200: Please mind that that the block and mark toggling logic comes directly from [slate's own implementation of rich text editor](https://github.com/ianstormtaylor/slate/blob/main/site/examples/richtext.tsx#L67:L199)

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
```diff
@@ -1,68 +1,29 @@
...
-  placeholder: string;
-  isDisabled?: boolean;
-  isReadOnly?: boolean;
-  onChange?: (event: TCustomEvent) => void;
-  onBlur?: (event: TCustomEvent) => void;
-  onFocus?: (event: TCustomEvent) => void;
+  placeholder: TEditorProps['placeholder'];
+  isDisabled?: TEditorProps['isDisabled'];
+  isReadOnly?: TEditorProps['isReadOnly'];
+  onChange?: (state: string) => void;
```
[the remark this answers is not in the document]
kark: Actually this is only an internal component. The exposed one is <LocalizedRichTextInput> and it's signature is unchanged
```ts
onChange?: (event: TCustomEvent) => void;
````

File: packages/components/inputs/rich-text-input/src/rich-text-input.tsx (hunk 3 above)
Comment (kark) [reply packages/components/inputs/rich-text-input/src/rich-text-input.tsx:28]: good point, I will mention that in the changeset

**Review by kark — COMMENTED** [review]
Here are my 2 cents to hopefully make reviewing of this PR easier 🙂

File: packages/components/inputs/rich-text-input/package.json [hunk 7]
```diff
@@ -40,9 +40,9 @@
     "is-hotkey": "0.2.0",
     "lodash": "4.17.21",
     "prop-types": "15.8.1",
-    "slate": "0.47.9",
-    "slate-html-serializer": "0.8.13",
-    "slate-react": "0.22.10"
+    "slate": "^0.75.0",
+    "slate-history": "^0.66.0",
+    "slate-react": "^0.75.0"
```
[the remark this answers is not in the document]
kark: I was not sure. But you are definitely right that we shouldn't 👍

packages/components/inputs/rich-text-input/src/rich-text-input.visualspec.js
kark: Thanks. Yes, I started with that, but then I switched to defining it in one place for brevity. I thought this fine tuning of those values might be done as the last step if we don't come up with any better solution for fixing the visual tests.

kark:
As discussed, in my last commit I used `useImperativeHandler` hook for resetting the editor. However, I had to manually change the READMEs, because `generate-readme` script seems not to work with components that use `forwardRef`

packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
```diff
@@ -0,0 +1,241 @@
...
+        <ol style={style} {...attributes}>
+          {children}
+        </ol>
+      );
+    case BLOCK_TAGS.p:
+      return (
+        <p style={style} {...attributes}>
+          {children}
+        </p>
+      );
```
emmenko, line 114: Nit: this can be omitted if it's meant to be the default case.

packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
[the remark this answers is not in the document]
emmenko: Let's leave a comment with the link about this.

emmenko (commented):
Thanks again for picking up this monster of migration. Rich text components are always tricky and complex to manage.

So far looks all good. I tested the component in storybook preview and it all behaves correctly.

Only the reset logic is not working as expected (see gif).

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
```diff
@@ -127,11 +129,21 @@ type TLocalizedRichTextInputProps = {
...
-  showExpandIcon: boolean;
+  showExpandIcon: TRichTextInputProps['showExpandIcon'];
   /**
    * Called when the `expand` button is clicked
    */
-  onClickExpand?: () => boolean;
+  onClickExpand?: TRichTextInputProps['onClickExpand'];
+};
+
+type StaticProps = {
```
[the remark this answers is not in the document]
kark: Yes, only `defaultProps` and `displayName` are recognized by default static props when using `forwardRef`. I will add a note about it.

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Ah! Ok, now I see.
I was confused by the name of the file.

packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -161,34 +208,76 @@ const Editor = (props: TEditorProps) => {
...
+                    disabled={props.isDisabled}
+                    onKeyDown={(event) => {
+                      for (const hotkey in HOTKEYS) {
+                        if (isHotkey(hotkey, event)) {
+                          event.preventDefault();
+                          const mark = HOTKEYS[hotkey as keyof typeof HOTKEYS];
+                          toggleMark(editor, mark);
+                        }
+                      }
+                    }}
```
CarlosCortizasCT, line 267: nit: Although the HOTKEYS object currently only have 4 attributes, we are iterating over all of them for every keystroke.

Maybe we could break after finding a hot key:
```suggestion
                  onKeyDown={(event) => {
                    for (const hotkey in HOTKEYS) {
                      if (isHotkey(hotkey, event)) {
                        event.preventDefault();
                        const mark = HOTKEYS[hotkey as keyof typeof HOTKEYS];
                        toggleMark(editor, mark);
                        break;
                      }
                    }
                  }}
```

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
```diff
@@ -181,160 +195,183 @@ const expandedTranslationsReducer = (
...
-      typeof props.onClickExpand === 'function',
-      'LocalizedRichTextInput: "onClickExpand" is required when showExpandIcon is true'
+    const [expandedTranslationsState, expandedTranslationsDispatch] =
+      useReducer(expandedTranslationsReducer, initialExpandedTranslationsState);
+
+    const defaultExpansionState =
+      props.hideLanguageExpansionControls ||
+      props.defaultExpandLanguages ||
+      // useToggleState's default is `true`, but we want `false`
+      false;
```
CarlosCortizasCT, line 241: nit: If we get to this point of the condition is because both `hideLanguageExpansionControls` and `defaultExpandLanguages` are falsy values, right?

Maybe this is because they might be `undefined` and if you pass that value to `useToggleState` it will use `true`, which we don't want.

If that's the case, perhaps we could use something like this:
```
const defaultExpansionState = Boolean(
      props.hideLanguageExpansionControls ||
      props.defaultExpandLanguages 
);
```

packages/components/inputs/rich-text-utils/src/is-empty/is-empty.ts
```diff
@@ -1,6 +1,44 @@
...
+} from 'slate';
 import html from '../html';
 
-const isEmpty = (rawValue: string): boolean =>
-  html.deserialize(rawValue).document.text === '';
+const isTextNodeNonEmpty = (node: TText) =>
+  Text.isText(node) && node.text !== '';
+
+const isElementNodeNonEmpty = (node: TElement) =>
+  node.children.some(isTextNodeNonEmpty);
```
CarlosCortizasCT, line 13: I'm a bit confused here since slate's Element.children prop seems to be defined as `Descendant[]` and that type is defined as `Element | Text` so we're not always passing `isTextNodeNonEmpty` function a `Text` typed object.

packages/components/inputs/localized-rich-text-input/src/editor.tsx
kark: nice improvement!

packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -115,26 +151,31 @@ const Editor = (props: TEditorProps) => {
...
-    prevIsFocused !== props.editor?.value.selection.isFocused &&
-    props.editor?.value.selection.isFocused &&
-    !props.isOpen
-  ) {
-    onToggle();
-  }
+  // resetting
+  const reset = useCallback(
+    (newValue: string) => {
+      resetEditor(editor, newValue);
```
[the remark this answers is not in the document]
CarlosCortizasCT: I was wondering if this issue been addressed.
Was it?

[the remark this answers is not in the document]
kark: Yes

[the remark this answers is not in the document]
kark: Yes, totally! in this case 130f0cf7348c18

CarlosCortizasCT (approved):
lgtm

👏

packages/components/inputs/rich-text-input/src/editor.tsx
```diff
@@ -90,22 +123,33 @@ const Editor = (props: TEditorProps) => {
 
   useEffect(() => {
     updateRenderToggleButton();
-  }, [props.editor.value.document, updateRenderToggleButton]);
+  }, [editor, updateRenderToggleButton]);
+
+  // resetting
+  const reset = useCallback(
```
emmenko, line 129: Nit: I would name this to `resetValue`.

packages/components/inputs/rich-text-input/src/rich-text-input.story.js
```diff
@@ -20,8 +20,10 @@ const Input = (props) => {
   const [value, setValue] = useState(initialValue);
   const onChange = useCallback(
     (event) => {
-      setValue(event.target.value);
-      action('onChange')(event);
+      {
+        setValue(event.target.value);
+        action('onChange')(event);
+      }
```
emmenko, line 26:
```suggestion
      setValue(event.target.value);
      action('onChange')(event);
```

emmenko (commented):
I tested out the component a bit in Storybook.
The reset functionality works as expected now, thanks!

I couldn't see the "issue" with the focus anymore. Is this still present or has it been fixed?

PS: I haven't finished the full review but wanted to leave my feedback so far. Thanks!

packages/components/inputs/rich-text-input/src/rich-text-input.story.js
```diff
@@ -14,7 +14,7 @@ import Readme from '../README.md';
 
 // Create our initial value...
 
-const initialValue = '<h1>H1 heading</h1>';
+const initialValue = '<h1>H1 <u>heading</u></h1>';
```
kark, line 17: I changed this as suggested by @emmenko to enable proper testing of mark button toggling state upon applying initial focus

.changeset/calm-rockets-grin.md
```diff
@@ -0,0 +1,30 @@
...
+return (
+  <>
+    <button onMouseDown={handleReset}>Reset</button>
+    <RichTextInput
+      // ...
+      ref={ref}
+    />
+  </>
+);
+```
```
emmenko, line 30: Nit: some line spacings are missing here.

```suggestion
This release contains some breaking changes to the following components, due to upgrading the internal `slate` libraries for rich-text editor functionality:
* `<RichTextInput>`
* `<LocalizedRichTextInput>`

> All other components do not have any breaking changes.

# Breaking changes

The change affects the way the editor is able to reset the `value`. This is usually relevant when using the components within a form and the user wants to reset the form to its initial state.

Previously the richt-text editor was working in a controlled way. This allowed the `value` passed to the input components to be used by the editor as-is. So if the user resets the form, the new value will be displayed.

Now the richt-text editor works in an uncontrolled way, meaning that it keeps its own state of the value and the `value` passed to the input components is used as the initial value. Therefore, if the user resets the form, the new value passed to the input won't do anything.

Instead, the reset functionality needs to be explicitly triggered using a dedicated function. To enable this, you need to use a `ref` object and pass it to the input component. The `ref` object will then contain the `resetValue` function that can be used to trigger the editor reset.

```jsx
const ref = useRef(null);
const handleReset = useCallback(() => {
  ref.current?.resetValue("<p><strong>Value after reset</strong></p>");
}, []);
return (
  <>
    <button onMouseDown={handleReset}>Reset</button>
    <RichTextInput
      // ...
      ref={ref}
    />
  </>
);
```
```

File: packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx [hunk 18]
```diff
@@ -181,160 +195,181 @@ const expandedTranslationsReducer = (
...
-      // https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
-      toggleLanguages();
+    const toggleLanguage = useCallback(
+      (language) => {
+        expandedTranslationsDispatch({ type: 'toggle', payload: language });
+      },
+      [expandedTranslationsDispatch]
+    );
+
+    const handleChange = useCallback(
```
emmenko, line 252: Nit: we're creating a new function here (because of the `language`), so to make it less confusing with normal event handlers we should name this something else.

```suggestion
    const createChangeHandler = useCallback(
```

packages/components/inputs/rich-text-utils/src/html/html.tsx
```diff
@@ -1,186 +1,251 @@
...
+      attrs = Object.entries(styleObj || {}).reduce(
+        (mappedAttrObj, [key, value]) => {
+          const values = value.split(' '); // to cover the case of space separated values e.g. `text-decoration-line: "underline line-through"`
+
+          values.forEach((splittedValue) => {
+            if (mapper[key] && mapper[key][splittedValue]) {
+              // checking if the parsed style attr value has representation in the mapper obj
+              mappedAttrObj = {
+                ...mapper[key][splittedValue],
+                ...mappedAttrObj,
```
emmenko, line 203: Shouldn't this be the other way around? First spread, then overwrite.

```suggestion
                ...mappedAttrObj,
                ...mapper[key][splittedValue],
```

packages/components/inputs/rich-text-utils/src/is-empty/is-empty.ts
emmenko: @kark Do you still want to adjust this?

packages/components/inputs/rich-text-utils/src/rich-text-body/dropdown.tsx
```diff
@@ -37,46 +41,37 @@ type THeadings = {
...
         getToggleButtonProps,
         getItemProps,
-      }: {
-        isOpen: boolean;
-        getToggleButtonProps: (
-          options?: GetToggleButtonPropsOptions
-        ) => unknown;
-        getItemProps: (options: GetItemPropsOptions<unknown>) => unknown;
-      }) => {
+      }: DownshiftRenderProps) => {
```
emmenko, line 74: I think Downshift has its one types. Can you check if we still need to provide our "own" types `DownshiftRenderProps`?

emmenko (commented):
Alright, did another review. To me everything works fine as expected now, great job again!

Just some minor nitpicks otherwise this is good to go now from my side.

.changeset/unlucky-stingrays-appear.md
```diff
@@ -0,0 +1,96 @@
+---
+'@commercetools-uikit/design-system': patch
```
kark, line 2: I used `patch` as suggested by @emmenko to verify what will be the resulting `changesets` PR

packages/components/inputs/rich-text-utils/src/is-empty/is-empty.ts
kark: I suppose that has already been covered but in 5e0bfd3b43d770c6f23921c12ef834915eb7868e I adjusted the use of `isElementNodeNonEmpty` within `isEmpty` so it is more straightforward

emmenko (approved):
I think this is good to go! Many thanks again for going through this 🚀 💪

PR #2160, opened 2022-04-01
fix(time-input): regression of handling change events

emmenko:
Small regression caused by #2147

### DIFF
--- a/packages/components/inputs/time-input/src/time-input.spec.js
+++ b/packages/components/inputs/time-input/src/time-input.spec.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/prop-types */
+import { useState } from 'react';
 import TimeInput from './time-input';
 import { screen, render, fireEvent } from '../../../../../test/test-utils';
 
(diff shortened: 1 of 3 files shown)


Rhotimee (approved):

packages/components/inputs/time-input/src/time-input.spec.js
```diff
@@ -97,6 +97,24 @@ describe('TimeInput', () => {
     );
   });
 
+  it('should call onChange when clearing the value', () => {
```
emmenko, line 100: There was no test for clearing the value, so the regression slipped through unfortunately.

kark (approved):
🙇‍♂️

tdeekens (approved):

emmenko:
The last commit https://github.com/commercetools/ui-kit/pull/2160/commits/6218281683607014722e531508c5fd47f3da7633 also fixes the remaining problems. Please take another look, thanks.

packages/components/inputs/time-input/src/time-input.tsx
```diff
@@ -129,12 +128,21 @@ const hasMilliseconds = (parsedTime: {
...
+  setValue?.call(node, value);
+  node.dispatchEvent(new Event('change', { bubbles: true }));
+};
+
 const TimeInput = (props: TTimeInputProps) => {
   const id = useFieldId(props.id, sequentialId);
   const intl = useIntl();
   const element = useRef<HTMLInputElement>(null);
-  // Keep internal state to allow clearing the value manually (`onClear`).
-  const [controlledValue, setControlledValue] = useState(props.value);
```
emmenko, line 137: This was the other source of the problems. Since we now set the value using our custom `dispatchReactChangeEvent`, we can simply forward the `props.value` as-is.

CarlosCortizasCT (approved):

ddouglasz (approved):

PR #2164, opened 2022-04-04
fix(time-input): pass `onChange` as-is

emmenko:
Follow up of #2160 
Caused by #2147

Another small regression. The only possible explanation why I refactored it that way is that I was drunk.

### DIFF
--- a/packages/components/inputs/time-input/src/time-input.spec.js
+++ b/packages/components/inputs/time-input/src/time-input.spec.js
@@ -20,6 +20,7 @@ const TestComponent = (props) => {
         {...props}
         value={time}
         onChange={(event) => setTime(event.target.value)}
+        onBlur={(event) => setTime(event.target.value)}
       />
       <div>{time ? `Time: ${time}` : 'No time'}</div>
     </div>
@@ -108,7 +109,7 @@ describe('TimeInput', () => {
     const event = { target: { value: '9' } };
     fireEvent.change(screen.getByLabelText('Input'), event);
 
-    await screen.findByText('Time: 9:00 AM');
+    await screen.findByText('Time: 9');
   });
 
   it('should call onChange when clearing the value', async () => {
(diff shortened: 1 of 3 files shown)


packages/components/inputs/time-input/src/time-input.spec.js
```diff
@@ -108,7 +109,7 @@ describe('TimeInput', () => {
     const event = { target: { value: '9' } };
     fireEvent.change(screen.getByLabelText('Input'), event);
 
-    await screen.findByText('Time: 9:00 AM');
+    await screen.findByText('Time: 9');
```
emmenko, line 112: TL;DR: change events should not format the time, otherwise the input becomes unusable. Only `onBlur` does the formatting.

tdeekens (approved):

packages/components/inputs/time-input/src/time-input.spec.js
CarlosCortizasCT: Perhaps I'm missing something but does this mean that users will see unformatted values while changing them until they move away the focus from the input field?

emmenko: Yes when users interact with the input they can potentially type whatever they want. This is important as otherwise the value keeps changing and the input becomes unusable.

You can see the issue in storybook: https://uikit.commercetools.com/?path=/story/components-inputs--timeinput

Rhotimee (approved):

CarlosCortizasCT (approved):

packages/components/inputs/time-input/src/time-input.spec.js
tdeekens: Yip, I think the (if I follow along) UX of an input formatting as you type is subjectively horrible. At least I go bananas when I encounter them. You type and it formats. Maintaining cursor position in an input which formats while you type at the right position is super shifty.

PR #2165, opened 2022-04-05
ci: check yarn constraints

emmenko:
Small leftover from the Yarn v3 migration

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Check for unmet constraints (fix w/ "yarn constraints --fix")
+        run: yarn constraints
+
       - name: Building packages
         run: yarn build
 


.github/workflows/main.yml
```diff
@@ -37,6 +37,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Check for unmet constraints (fix w/ "yarn constraints --fix")
+        run: yarn constraints
+
```
CarlosCortizasCT, line 42: Is it not necessary to install the _constraints_ plugin in advance?
Does the container already have it installed?

emmenko: Everything was already set up, we just forgot to run the check on CI.

https://github.com/commercetools/ui-kit/blob/16cf82b43db20da11be0612ea64c30ca4c23654a/.yarnrc.yml#L4-L5
https://github.com/commercetools/ui-kit/blob/main/constraints.pro

CarlosCortizasCT (approved):

PR #2168, opened 2022-04-08
fix unexpected warnings in some field components

### LINKED ISSUE #2162: `touched` is expected to be an array of boolean when isMulti is true
**Describe the bug**
When we are using `<SearchSelectField/>` and `<AsyncSelectFiled />` and isMulti is true, the below error is throwing when touched is undefined.

`Uncaught Error: Warning: SearchSelectField: touched is expected to be an array of boolean when isMulti is true`

CarlosCortizasCT:
#### Summary

Suppress invalid props warnings in some field components.

## Description

We were told that some field components are raising some warnings regarding the `touched` prop when they also have the `isMulti` prop with the value `true`.
This is because we are not correctly checking that `touched` property can be undefined while the input has not been gained focus yet.

Affected components:
* async-creatable-table-field
* async-select-field
* creatable-select-field
* search-select-field

closes #2162

### DIFF
--- a/.changeset/strange-bugs-vanish.md
+++ b/.changeset/strange-bugs-vanish.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-uikit/async-creatable-select-field': patch
+'@commercetools-uikit/async-select-field': patch
+'@commercetools-uikit/creatable-select-field': patch
+'@commercetools-uikit/search-select-field': patch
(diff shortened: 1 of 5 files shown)


emmenko (approved):
💯

ddouglasz (approved):

PR #2170, opened 2022-04-08
fix(radio-input): prop types

emmenko:
Fixes this issue:

image

### DIFF
--- a/packages/components/inputs/radio-input/src/radio-group.tsx
+++ b/packages/components/inputs/radio-input/src/radio-group.tsx
@@ -10,14 +10,11 @@ import {
 } from 'react';
 import { warning, filterDataAttributes } from '@commercetools-uikit/utils';
 import Constraints from '@commercetools-uikit/constraints';
-import Stack, { type TStackProps } from '@commercetools-uikit/spacings-stack';
-import Inline from '@commercetools-uikit/spacings-inline';
+import Stack from '@commercetools-uikit/spacings-stack';
+import Inline, {
+  type TInlineProps,
+} from '@commercetools-uikit/spacings-inline';
 import Option from './radio-option';
-
-const directionWrapper = {
-  stack: Stack,
-  inline: Inline,
-};
 
 export type TGroupProps = {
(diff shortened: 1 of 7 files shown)


packages/components/inputs/radio-input/src/radio-group.tsx
```diff
@@ -10,15 +10,12 @@ import {
...
+import Stack from '@commercetools-uikit/spacings-stack';
+import Inline, {
+  type TInlineProps,
+} from '@commercetools-uikit/spacings-inline';
 import Option from './radio-option';
 
-const directionWrapper = {
-  stack: Stack,
-  inline: Inline,
-};
```
emmenko, line 20: This apparently wasn't used, and it was causing the issue (see screenshot).

Removed now.

packages/components/inputs/radio-input/src/radio-group.tsx
```diff
@@ -43,8 +40,12 @@ export type TGroupProps = {
...
     | 'scale'
     | 'auto';
-  direction?: keyof typeof directionWrapper;
-  directionProps?: Partial<TStackProps>;
+  direction: 'stack' | 'inline';
+  directionProps: {
+    scale?: TInlineProps['scale'];
+    alignItems?: TInlineProps['alignItems'];
+    justifyContent?: TInlineProps['justifyContent'];
+  };
```
emmenko, line 48: Better to be explicit, as otherwise there are other props that do not make sense to allow (e.g. `children`)

packages/components/inputs/radio-input/src/radio-group.tsx
```diff
@@ -30,7 +27,7 @@ export type TGroupProps = {
   isReadOnly?: boolean;
   hasError?: boolean;
   hasWarning?: boolean;
-  horizontalConstraint?:
+  horizontalConstraint:
```
emmenko, line 30: No need to keep it optional as there are default props

CarlosCortizasCT (approved):
lgtm

PR #2177, opened 2022-04-19
fix: release workflow fix by stopping docs autogeneration for `rich-text-input` and `localized-rich-text-editor`

kark:
#### Summary

This PR fixes the problem with release CI workflow and running `yarn generate-readmes` command


emmenko (approved):
Thanks.

Can you create an issue to look into trying fixing the generation with components using `forwardRef`?

PR #2183, opened 2022-04-22
fix: typo in prop

emmenko:


kark (approved):

ddouglasz (approved):

packages/components/fields/search-select-field/src/search-select-field.tsx
```diff
@@ -74,7 +74,7 @@ type TSearchSelectFieldProps = {
    * <br>
    * [Props from React select was used](https://react-select.com/props)
    */
-  'aria-labelledby?': ReactSelectAsyncProps['aria-labelledby'];
+  'aria-labelledby': ReactSelectAsyncProps['aria-labelledby'];
```
CarlosCortizasCT, line 77: I just wanted to double check if this property should be required.
I was taking a look at [the commit](https://github.com/commercetools/ui-kit/commit/53e85593365fbe85938837122c1c4586c9147745) when we migrated this component to typescript and this prop was previously optional so maybe we should keep it like that.

emmenko: Good catch, it should be optional yes. Thanks! https://github.com/commercetools/ui-kit/pull/2183/commits/2023ec8c7f34119dd06080a44520a4dc3be30d4b

CarlosCortizasCT (approved):
👍

Issue #2074, opened 2022-01-25
DateTime input placeholders show hardcoded date/time format


emmenko:
**Describe the bug**
The default `placeholder` value for DateTime input uses a hardcoded date/time format message.

https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/date-time-input/src/messages.ts#L4-L13

**Expected behavior**

* No message is needed (there is nothing to be translated).
* No hardcoded date/time format.
* The default placeholder value should indicate the date/time format based on the current locale.

emmenko:
Yes good question. Locales usually consist of a **language** and optionally a **country code**.

```
en
en-US
en-GB

de
de-DE
de-AT
```

In your Merchant Center user profile you can select a local/language that also contains a more specific country code.

  <img width="556" alt="image" src="https://user-images.githubusercontent.com/1110551/151042010-9c5f1feb-0974-4139-9c16-378846978d47.png">

This locale information is then used throughout the application in different ways:
* For the Merchant Center application messages: here the translation language closely matching the user locale is used.
* For formatting dates: the format is usually locale + country specific.

CarlosCortizasCT:
Thanks for the explanation.

I was confused because the way I'm used to handle locale resolution fallbacks so if no translation is available for the complete locale (ex: `de-CH`) it would the use the language only one if available (ex: `de`).
This mean you need to have translations defined for all the allowed combinations, which I assume it is something we don't want in our scenario.

That said, I'm still wondering where should be define the placeholder values as I don't find how to get them in the runtime (I mean, if we want to show the users values such as `MM/DD/YYYY - HH:mm AM/PM` and `HH:mm AM/PM`).
Or, otherwise, we want to show a specific date (maybe today) as a placeholder? (ex: `26/01/2022 - 08:56 AM`)

emmenko:
Yes when it comes to translations it falls back to the language code.

As for getting the format date, I think we should use momentjs (for example with the `creationData` API https://momentjs.com/docs/#/parsing/creation-data/).

CarlosCortizasCT:
I still don't get how we can resolve the formats as they are translated. I mean, what we need to show to the user are abbreviations of words composing the format. In english `D` stands for day, but in german we should use `T` from tag.

Reading the moment documentations I see we can get the format, but it is always using the same letters (always `D` for day no matter the locale used) as it is not a translated format.

These are the texts we should be showing to our users for the long date-time format (based on the current translations we have):
[...]

PR #2190, opened 2022-05-04
Localized date and time pattern formats

CarlosCortizasCT:
#### Summary

The goal of this PR is to add support for rendering date and time format placeholders using the localized pattern and characters in input components where such a value is expected to be provided by the user.

Relates to this issue: #2074 

## Description

When we currently render a component such `DateInput` where we ask the user to type a date, we initially render it with a placeholder with the pattern of how we expect the user to provide the date:
image

Currently we are using the same placeholder for every language when both the pattern and the characters set used might be different.
For instance, if user is using deutsch, the day is expected to be before the month and also the characters should be different:
image

With this in mind, a new helper function has been created to calculate the correct localized pattern for dates and times.

Here is an example of the result:
Components _ Fields - DateField ⋅ Storybook

We can use [moment](https://momentjs.com/docs/#/i18n/locale-data/) to know about the localized patterns for dates and times, but we don't have any means to fetch the localized version of the pattern:
```
// Date pattern for german
DD/MM/YYY

// Localized date pattern for german
TT.MM.JJJJ
```
So I added a mapping configuration to implement this localization and we would need to extend it whenever we support a new language.


docs/.storybook/configs/intl-context.js
```diff
@@ -25,7 +27,9 @@ const getMessagesForLocale = (locale) => {
 const namifyLocale = (locale) => {
   switch (locale) {
     case 'en':
-      return 'English';
+      return 'English (US)';
+    case 'en-GB':
+      return 'English (British)';
```
CarlosCortizasCT, line 32: I added british english to Storybook so we can play also with that locale, although it loads the same translations that american english.

packages/calendar-time-utils/src/calendar-time.ts
```diff
@@ -162,3 +163,57 @@ export const parseInputText = (
...
+
+const localizedDateFormatPatternCache = new Map<string, string>();
+export const getLocalizedDateTimeFormatPattern = (
+  locale: string,
+  formatType: 'date' | 'time' | 'full' = 'date'
+): string => {
+  const key = `${locale}::${formatType}`;
+  if (localizedDateFormatPatternCache.has(key)) {
+    return localizedDateFormatPatternCache.get(key)!;
+  }
```
CarlosCortizasCT, line 175: Since this function will be used from the components every time they are rendered, I thought it would help to have a simple cache here to avoid calculating the pattern too many times.

packages/calendar-time-utils/src/formats.ts
```diff
@@ -0,0 +1,69 @@
...
+  YY = 'YY', // TWO_DIGITS_YEAR
+  YYYY = 'YYYY', // FOUR_DIGITS_YEAR
+  h = 'h', // HOUR_12_SINGLE
+  hh = 'hh', // HOUR_12_TWO_DIGITS
+  H = 'H', // HOUR_24_SINGLE
+  HH = 'HH', // HOUR_24_TWO_DIGITS
+  m = 'm', // MINUTE_SINGLE
+  mm = 'mm', // MINUTE_TWO_DIGITS
+  A = 'A', // PERIOD - https://unicode.org/reports/tr35/tr35-6.html#Date_Format_Patterns
+}
```
CarlosCortizasCT, line 16: Here we define an enumeration with all the possible characters moment locale pattern might include so we can then _translate_ them to the expected value based on the locale.

packages/calendar-time-utils/src/formats.ts
```diff
@@ -0,0 +1,69 @@
...
+export type TLocalizedDateFormatMapping = {
+  [locale: string]: TLocaleDateFormatMapping;
+};
+
+// Here is where we translate from the standard characters used in the pattern
+// (calculated in the moment.js library) to the characters used in the locale
+// We just need to provide the translation for the characters that are different
+// For instance, in english we don't need to translate date characters as
+// they are the same as the pattern characters
+export const DATE_FORMAT_LOCALIZED_MAPPINGS: TLocalizedDateFormatMapping = {
```
CarlosCortizasCT, line 31: This is where we would need to work if we start supporting new languages.
We just map the pattern characters to localized characters (if needed).

packages/calendar-time-utils/src/calendar-time.ts
emmenko: Sure, makes sense. Thanks

packages/calendar-time-utils/src/formats.ts
```diff
@@ -0,0 +1,69 @@
+// Reference: https://momentjs.com/docs/#/displaying/format/
+export enum FormatToken {
```
emmenko, line 2: We usually don't use enums, as there are issues with Babel (as far as I remember).

You can use instead the `as const` syntax.

```ts
const formatToken = {
  D: 'D',
  // ...
} as const
```

emmenko (commented):
Nice stuff, thanks for working on that. Besides a couple of small suggestions, looks good overall 👌

packages/calendar-time-utils/src/calendar-time.ts
```diff
@@ -162,3 +163,57 @@ export const parseInputText = (
...
+      break;
+    case 'time':
+      localizedFormat = localeData.longDateFormat('LT');
+      break;
+    case 'full':
+      localizedFormat = `${localeData.longDateFormat(
+        'L'
+      )} - ${localeData.longDateFormat('LT')}`;
+      break;
+  }
```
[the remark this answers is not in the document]
CarlosCortizasCT: I thought about it but it seemed to me it was more clear to have all the variants and no default since we know one of those would be used because we are typing it.
Thinking again about it, as users don't need to use Typescript, perhaps its better.

I can think in two options:
```
  let localizedFormat = '';
  switch (formatType) {
    case 'time':
      localizedFormat = localeData.longDateFormat('LT');
      break;
    case 'full':
      localizedFormat = `${localeData.longDateFormat(
        'L'
      )} - ${localeData.longDateFormat('LT')}`;
      break;
    case 'date': // <--- Not really needed, but it think having it here makes it more clear to the reader
    default:
      localizedFormat = localeData.longDateFormat('L');
      break;
  }
```

```
  // Default format is 'date'
  let localizedFormat = localeData.longDateFormat('L');
  switch (formatType) {
    case 'time':
      localizedFormat = localeData.longDateFormat('LT');
      break;
    case 'full':
      localizedFormat = `${localeData.longDateFormat(
        'L'
      )} - ${localeData.longDateFormat('LT')}`;
      break;
  }
```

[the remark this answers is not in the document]
emmenko: Ok so what we can do is throw an error in the `default` case, so that we make sure that we cover all known cases but throw an error otherwise.

packages/calendar-time-utils/src/formats.ts
CarlosCortizasCT: Updated in this commit: 57e106548f5d93026e3a5fb8f5d2b42b66f7653c

packages/calendar-time-utils/src/calendar-time.ts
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit: 6edef8f5850fc17e4134f815ad652e396aeafcb8

kark (approved):
Hats off! 👏

PR #2192, opened 2022-05-05
refactor(rich-text): return empty string instead of empty paragraph

kark:
#### Summary

In order to make rich text inputs work properly with Formik if there is no text entered the input should return empty string instead of empty paragraph (`<p></p>`). That implies some changes in the way serialization has worked before and what some unit tests expected.

Other than that, custom module declarations were moved to `html.tsx` file, since they were not visible previously during `tsc-files` type checks.

### DIFF
--- a/.changeset/old-jeans-yell.md
+++ b/.changeset/old-jeans-yell.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/rich-text-utils': patch
+---
+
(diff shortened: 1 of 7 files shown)


emmenko (approved):
Nice, thanks! 🙌

Rhotimee (approved):

ddouglasz (approved):

PR #2196, opened 2022-05-10
Improve LocalizedRichTextInput VRT

CarlosCortizasCT:
#### Summary

I've noticed the visual regression test for the `LocalizedRichTextInput` fails randomly in CI so I tried to updated it to make it more robust.

Pasted Graphic

## Description

When the test failed it was always trying to assert the state of the rich text input after some manipulation.
Since it fails randomly, I believe this is a timing issue where some changes in the underlying library (`slate`) are not reflected fast enough in the test environment.

Instead of trying to slow down the test a little more (we are already slowing it down a bit), I opted for changing some assertions to be async so we wait for the changes to be reflected in the UI.

### DIFF
--- a/packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
+++ b/packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
@@ -35,13 +35,15 @@ describe('LocalizedRichTextInput', () => {
     }, input);
   };
 
-  const getNumberOfTags = async (tagName) => {
-    // eslint-disable-next-line no-shadow
-    const numberOfTags = await page.evaluate((tagName) => {
-      return document.querySelectorAll(tagName).length;
-    }, tagName);
-
-    return numberOfTags;
+  const waitForNumberOfTags = (tagName, count) => {
+    return page.waitForFunction(
+      (_tagName, _count) => {
+        return document.querySelectorAll(_tagName).length === _count;
+      },
+      { timeout: 5000 },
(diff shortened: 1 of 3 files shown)


packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
```diff
@@ -35,13 +35,15 @@ describe('LocalizedRichTextInput', () => {
...
+  const waitForNumberOfTags = (tagName, count) => {
+    return page.waitForFunction(
+      (_tagName, _count) => {
+        return document.querySelectorAll(_tagName).length === _count;
+      },
+      { timeout: 5000 },
+      tagName,
+      count
+    );
   };
```
CarlosCortizasCT, line 47: This is the new helper function I use in the test for async assertions.

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
```diff
@@ -100,7 +100,7 @@ describe('LocalizedRichTextInput', () => {
     await input.type('Hello world');
 
     // check that there is now a strong tag in the document.
-    numOfTags = await getNumberOfTags('strong');
```
CarlosCortizasCT, line 103: Previously there was no actual check here.
It was needed the _focus_ mentioned before to actually test the expected behaviour.

emmenko (approved):
Very nice, thanks! 🙌

PR #2197, opened 2022-05-10
Add radio field to story book

Rhotimee:
#### Summary
The radio field was previously not added to story book. This PR fixes that by adding the radio field to storybook

### DIFF
--- a/.changeset/fast-turtles-compare.md
+++ b/.changeset/fast-turtles-compare.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/radio-field': patch
+---
+
+Add Radio Field to docs
--- a/.changeset/fast-turtles-compare.md
+++ b/.changeset/fast-turtles-compare.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/radio-field': patch
 ---
 
-Add Radio Field to docs
+Add Radio Field to storybook
--- a/packages/components/fields/radio-field/src/radio-field.story.js
+++ b/packages/components/fields/radio-field/src/radio-field.story.js
@@ -1,0 +1,152 @@
+import { createElement } from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import {
+  withKnobs,
+  boolean,
+  text,
+  select,
+  object,
+} from '@storybook/addon-knobs/react';
+import Constraints from '@commercetools-uikit/constraints';
+import { Value } from 'react-value';
+import Spacings from '@commercetools-uikit/spacings';
+import RadioInput from '@commercetools-uikit/radio-input';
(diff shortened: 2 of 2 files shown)


.changeset/fast-turtles-compare.md
```diff
@@ -0,0 +1,5 @@
+---
```
kark, line 1: All works great in storybook 🙌 I'm only wondering if changeset is needed in this case

Rhotimee: Good question. I'm not 100% sure as well. 
@commercetools/shield-team-fe, what do you guys think?

packages/components/fields/radio-field/src/radio-field.story.js
```diff
@@ -0,0 +1,152 @@
...
+                  false,
+                  radioOption1
+                )}
+                isHovered={boolean('isHovered #1', false, radioOption1)}
+                additionalContent={text('additionalContent', '', radioOption1)}
+              >
+                <Spacings.Inline scale="xs" alignItems="center">
+                  <div>{'🍎'}</div>
+                  {text('label #1', 'Apple', radioOption1)}
+                </Spacings.Inline>
```
CarlosCortizasCT, line 115: I think we need to wrap the text within an element for the spacing to actually works. Otherwise, the icon and the text don't have any space between them.
image

```suggestion
                <Spacings.Inline scale="xs" alignItems="center">
                  <div>{'🍎'}</div>
                  <div>{text('label #1', 'Apple', radioOption1)}</div>
                </Spacings.Inline>
```

CarlosCortizasCT: Maybe we could also take a look at the accessibility issues that are shown in the `Accessibility` panel in Storybook:
image

For instance, it seems `aria-readonly` is not a supported prop for [radio elements](https://www.w3.org/TR/wai-aria-1.1/#radio).

Perhaps this can be done as a follow-up task to not increasing this one's scope.

.changeset/fast-turtles-compare.md
CarlosCortizasCT: I think this change won't be included in the next release if we don't add the changeset so perhaps the [public Storybook site](https://uikit.commercetools.com/) won't be updated neither (not sure about this).

CarlosCortizasCT (approved):
Great work 💯 

I left some minor suggestions that maybe could be considered.

packages/components/fields/radio-field/src/radio-field.story.js
Rhotimee: Sure, we will work on this as a follow-up.

kark (approved):

tdeekens (approved):

Issue #2078, opened 2022-01-27
Add `<ViewSwitcher>` component


MicheleRezk:
**Is your feature request related to a problem? Please describe.**
Within our Operations App, we used a "core/shared component" to display a quick filter for our imported data. 
To now Improve MC consistency and usability, we should build a component out of the shared component we used.

**Describe the solution you'd like**
Introduce ViewSwitchers Component in UI Kit, which allow users to toggle between two or more different views of the same, similar or related content items within the same space on screen.

**When to use:**
ViewSwitchers are frequently used to let users toggle between different formattings, like with a grid view and a table view.
They are also often used to narrow content groups.
For example, use a content switcher if you have a list, such as “Import-results” and you want to divide it into views such as “All”, “Failed”, and “Success”.
![1](https://user-images.githubusercontent.com/30607195/151365212-b63ec82b-07db-40eb-86f4-dc4d7e869a93.png)

Shecki:
https://wiki.commercetools.com/display/shield/ViewSwitcher%3A+component+checklist

mmaltsev-ct:
Hey @MicheleRezk, could you clarify whether there was already any work done for this component in the ui-kit?

MicheleRezk:
Hey @mmaltsev-ct , I still didn't start working on this component yet

PR #2198, opened 2022-05-19
Implement ViewSwitcher component

Rhotimee:
#### Summary
Implement ViewSwitcher component. 

## Description
ViewSwitchers allow users to toggle between two or more different views of the same, similar or related content items within the same space on the screen.

## Screenshot
image

ref #2078 
https://wiki.commercetools.com/display/shield/ViewSwitcher%3A+component+checklist
https://commercetools.invisionapp.com/console/share/GH3699W04K/829507218


packages/components/view-switcher/src/view-switcher.spec.js
```diff
@@ -0,0 +1,148 @@
+import { warning } from '@commercetools-uikit/utils';
```
kark, line 1: Let me pin it here. I am wondering if visual regression tests should also be part of this PR

packages/components/view-switcher/src/view-switcher.spec.js
```diff
@@ -0,0 +1,148 @@
...
+        </Button>
+        <Button value="test-button">test button</Button>
+      </Group>
+    );
+
+    screen.getByLabelText('disabled button').click();
+    expect(handleChange).not.toHaveBeenCalled();
+  });
+
+  it('should mark button as selected', () => {
```
kark, line 75: In my opinion what this test does is something like:
```suggestion
  it('should call onChange by the selected button', () => {
```

packages/components/view-switcher/src/view-switcher.spec.js
```diff
@@ -0,0 +1,148 @@
...
+    expect(handleClick).not.toHaveBeenCalled();
+  });
+});
+
+describe('warnings', () => {
+  let props;
+  beforeEach(() => {
+    props = createGroupTestProps(0);
+  });
+  it('should warn when there are no view switcher button in the group', () => {
```
kark, line 141:
```suggestion
  it('should warn when there are no view switcher buttons in the group', () => {
```

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Indicates that the view switcher can be reduced to save space
+   */
+  isCondensed?: boolean;
+  /**
+   * Pass one or more `ViewSwitcher.Button` components
+   */
+  children: ReactNode;
+  /**
+   * Will be triggered whenever a `ViewSwitcher.Button` is clicked. Called with the ViewSwitcherButton value
```
kark, line 28:
```suggestion
   * Will be triggered whenever a `ViewSwitcher.Button` is selected. Called with the ViewSwitcherButton value
```

packages/components/view-switcher/src/view-switcher-button.tsx
```diff
@@ -0,0 +1,59 @@
+import {
```
kark, line 1: Let me pin it here - I noticed this error in the console, perhaps it's nothing serious, but I guess it's worth checking:
Screenshot 2022-05-20 at 10 43 56

kark (commented):
👏 I really like the API of the component and the fact that we don't have to wrap the buttons with a React.Fragment 🙌 I'm leaving only some minor comments.

packages/components/view-switcher/src/view-switcher-button.tsx
Rhotimee: Thanks for the feedback. I'm still trying to figure out how to solve the problem. 
If you have any clue, let me know.

kark: It seems to be caused by the conditional styling:
```js
if (isDisabled) {
    return [
      baseStyles,
      css`
        background-color: ${vars.colorAccent98};
        color: ${vars.colorNeutral60};
      `,
    ];
  }
  if (isActive) {
    return [
      baseStyles,
      css`
        background-color: ${vars.colorNeutral95};
        box-shadow: ${vars.shadow9};
      `,
    ];
  }
```

tdeekens:
Nice API!

packages/components/view-switcher/src/view-switcher.styles.ts
```diff
@@ -0,0 +1,74 @@
...
+    return [
+      baseStyles,
+      css`
+        background-color: ${vars.colorNeutral95};
+        box-shadow: ${vars.shadow9};
+      `,
+    ];
+  }
+
+  return baseStyles;
```
CarlosCortizasCT, line 71: Regarding the warning about hooks execution order, as @kark mentioned, it seems to be related to returning different array length between first and consecutive renders.

Maybe something like this (I've seen it in other components) might help:

```suggestion
return [
    css`
      align-items: center;
      color: ${vars.colorSolid};
      transition: background-color ${vars.transitionLinear80Ms};
      font-size: ${vars.fontSizeDefault};
      border-radius: ${borderRadius};
      box-shadow: 0 1px 1px 0 rgba(0, 0, 0, 0.24),
        0 -1px 1px 0 rgba(0, 0, 0, 0.12);
      &:focus {
        border: ${vars.borderRadius1} solid ${vars.colorInfo};
      }
      &:hover {
        background-color: ${vars.colorNeutral90};
      }
      &:active {
        background-color: ${vars.colorNeutral95};
      }
      ${getSizeStyles(isCondensed)}
    `,
    isActive && css`
      background-color: ${vars.colorNeutral95};
      box-shadow: ${vars.shadow9};
    `,
    isDisabled && css`
      background-color: ${vars.colorAccent98};
      color: ${vars.colorNeutral60};
    `,
  ];
```

Rhotimee: Thanks @CarlosCortizasCT

packages/components/view-switcher/src/view-switcher-button.tsx
Rhotimee: Thanks. It is fixed.

packages/components/view-switcher/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+ViewSwitchers allow users to toggle between two or more different views of the same, similar or related content items within the same space on screen.
+
+## When to use
+
+ViewSwitchers are frequently used to let users toggle between different formatting's, like with a grid view and a table view.
+
+They are also often used to narrow content groups.
+
+For example, use a content switcher if you have a list, such as “Import-results” and you want to divide it into views such as “All”, “Failed”, and “Success”.
```
CarlosCortizasCT, line 13: Personally I find difficult to differentiate between this example and when to use tabs.
Perhaps it's better to remove it.

packages/components/view-switcher/src/view-switcher.styles.ts
```diff
@@ -0,0 +1,63 @@
+import { css } from '@emotion/react';
+import { customProperties as vars } from '@commercetools-uikit/design-system';
```
CarlosCortizasCT, line 2: Personally I think `customProperties` is a more meaningful name than just `vars`.
Perhaps we don't need to create an alias here.

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -0,0 +1,93 @@
...
+const ViewSwitcher = (props: TViewSwitcherProps) => {
+  const [selectedButton, setSelectedButton] = useState<string>(
+    props.defaultSelected
+  );
+
+  useEffect(() => {
+    const childrenAsArray = Children.toArray(props.children) as TReactChild[];
+    const optionChildrenAsArray = childrenAsArray.filter(
+      (child) => child.type.displayName === ViewSwitcherButton.displayName
+    );
```
CarlosCortizasCT, line 46: I'm not sure if I'm missing something here because, since we don't expect the `ViewSwitcher` to have any other component than `ViewSwitcherButton`, perhaps we could avoid this filtering iteration. I mean, it seems it does not make sense to provide any other type of children, right?

Also, we use the hook because we expect the children to change dynamically? (perhaps some use case you have in mind)

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -0,0 +1,93 @@
...
+      const isButtonActive = selectedButton === child.props.value;
+      const clonedChild = cloneElement(child, {
+        onClick: () => {
+          setSelectedButton(child.props.value);
+          if (child.props.onClick && !isButtonActive) {
+            child.props.onClick(child.props.value);
+          }
+          if (props.onChange && !isButtonActive) {
+            props.onChange(child.props.value);
+          }
```
CarlosCortizasCT, line 69: nit: perhaps we could take advantage of optional chaining here.

```suggestion
          if (!isButtonActive) {
            child.props.onClick?.(child.props.value);
            props.onChange?.(child.props.value);
          }
```

CarlosCortizasCT (commented):
Awesome work here 👏 

I just left a couple of minor suggestions.

packages/components/view-switcher/src/view-switcher.tsx
Rhotimee: It's meant to be a check in case the component wasn't used properly. But, I think it makes sense to remove the filter and hook.

packages/components/view-switcher/README.md
Rhotimee: True. I agree with you.

CarlosCortizasCT: This has been updated in this commit: c7f430ef5530845eb75c43525902691340a711ef

packages/components/view-switcher/src/view-switcher.tsx
CarlosCortizasCT: It has been removed in this commit: c7f430ef5530845eb75c43525902691340a711ef

CarlosCortizasCT (approved):
💯

FilPob (approved):

PR #2205, opened 2022-05-31
fix typescript props declarations to make menuPortalZIndex optional

### LINKED ISSUE #2204: Alignment types and documentation (Create)SelectInput
**Describe the bug**
Documentation and types of CreateSelectInput and SelectInput seem not aligned:
The prop `isRequired` is not available according to the type of the props, but it is present in the documentation.
The prop `menuPortalZIndex` is not an optional prop according to its type, but is according to the documentation.

**To Reproduce**
Compare the props in this link:
https://uikit.commercetools.com/?path=/story/components-fields-selectfields--creatableselectfield
to this link
https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx

**Expected behavior**
I expected the prop 'descriptions' of types and documentation to be aligned.

**Additional context**
Basically I'd just like to know what is intended and what I should be going after 😀: 
- Can / can't I use the prop `isRequired`?
- Should / Shouldn't I set the prop `menuPortalZIndex`?

Thanks for helping!
cheers
Sjoerd

CarlosCortizasCT:
### Summary

We found we have a property called `menuPortalZIndex` in several components typed as mandatory for their props where it actually is optional.

### Description

This prop is present in several components and, in all of them, we have a default value for it but some of them have it defined as required in its props typescript declaration.

closes #2204

### DIFF
--- a/packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.tsx
+++ b/packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.tsx
@@ -204,7 +204,7 @@ type TAsyncCreatableSelectInputProps = {
   /**
    * z-index value for the menu portal
    */
-  menuPortalZIndex: number;
+  menuPortalZIndex?: number;
   /**
    * whether the menu should block scroll while open
    * <br>
--- a/packages/components/inputs/async-select-input/src/async-select-input.tsx
+++ b/packages/components/inputs/async-select-input/src/async-select-input.tsx
@@ -196,7 +196,7 @@ type TAsyncSelectInputProps = {
   /**
    * z-index value for the menu portal
    */
-  menuPortalZIndex: number;
+  menuPortalZIndex?: number;
   /**
    * whether the menu should block scroll while open
    * <br>
--- a/packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
+++ b/packages/components/inputs/creatable-select-input/src/creatable-select-input.tsx
@@ -203,7 +203,7 @@ type TCreatableSelectInputProps = {
   /**
    * z-index value for the menu portal
    */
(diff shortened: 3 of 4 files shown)


emmenko (commented):
That's not really a bug because these props have default values. TS understands that and the prop is technically optional.

This also helps in the implementation as the value can't be optional (because of the default value), which means that the prop is treated as required.

CarlosCortizasCT:
> That's not really a bug because these props have default values. TS understands that and the prop is technically optional.
> 
> This also helps in the implementation as the value can't be optional (because of the default value), which means that the prop is treated as required.

Ok, I get it.
What puzzles me is that the prop is defined as required but Typescript is clever enough to understand that it is actually optional because we assign a value in the default props. It is no clear to me (not the person who opened the issue) that the prop is optional just reading the type definitions, but I guess it's just a matter of learning how it works.

Personally I would prefer to have the _prop_ set as optional because, from the consumer point of view, it is and it would be more clear, but I understand this is just a personal preference.

I get your point but I would like to argue that props are the public API of components so, when typing them, we're shaping that public API.
From the consumer point of view, the `menuPortalZIndex` is not required since it is not mandatory to provide it when using the component and that's why I think the type should.

emmenko:
Yes you're right, I agree it's not ideal at the moment.

I think we need to make a decision how we want to handle this "pattern/setup", and refactor things consistently.

For example, there were rumors that `defaultProps` is kind of deprecated and the recommendation is to use props destructuring with default assigned values.

This would then allow the prop to be declared as optional and have the prop value initialized properly in the component implementation.

Maybe there are other approaches as well.

For me the important thing is that in the implementation code the props value is considered defined (not optional) when the prop is optional but with a default value.

PR #2206, opened 2022-06-02
Add aria error attributes to field components

CarlosCortizasCT:
### Summary

We found out that error aria attributes (`aria-invalid` and `aria-errormessage`) were missing in some field components.

### Description

The problem is that field components were not providing those attributes to their inner input components and so, the latter were not being marked as invalid at the aria level.

### DIFF
--- a/packages/components/fields/radio-field/src/radio-field.spec.js
+++ b/packages/components/fields/radio-field/src/radio-field.spec.js
@@ -178,21 +178,27 @@ describe('when showing an info button', () => {
 describe('when field is touched and has errors', () => {
   describe('when field empty', () => {
     it('should render a default error', () => {
-      const { getByText } = renderRadioField({
+      const { container, getByText } = renderRadioField({
         touched: true,
         errors: { missing: true },
       });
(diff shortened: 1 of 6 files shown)


packages/components/fields/radio-field/src/radio-field.spec.js
```diff
@@ -178,21 +178,27 @@ describe('when showing an info button', () => {
...
     it('should render a default error', () => {
-      const { getByText } = renderRadioField({
+      const { container, getByText } = renderRadioField({
         touched: true,
         errors: { missing: true },
       });
       expect(getByText(/field is required/i)).toBeInTheDocument();
+      expect(
+        container.querySelector('[aria-invalid="true"]')
+      ).toBeInTheDocument();
```
CarlosCortizasCT, line 188: `Radio.Group` component adds the aria error attributes to a wrapper `div` of the radio inputs. That's why I used a query selector.

Rhotimee (approved):

emmenko (approved):
Thanks!

PR #2218, opened 2022-06-22
Fix CollapsiblePanel default ID generation

CarlosCortizasCT:
### Summary

`CollapsiblePanel` component does not correctly autogenerate IDs. 

### Description

When the `CollapsiblePanel` component is used and no `id` property is provided, the autogenerated value is not correctly calculated so, if a consumer uses several instances of this component in the same view, they all will have the same ID in the rendered HTML.

I also found some room for improvements in the accessibility attributes of the component based of [this reference](https://w3c.github.io/aria-practices/examples/accordion/accordion.html).

### DIFF
--- a/packages/components/collapsible-panel/src/collapsible-panel.spec.js
+++ b/packages/components/collapsible-panel/src/collapsible-panel.spec.js
@@ -212,7 +212,7 @@ describe('aria attributes', () => {
       expect(panelContent).toHaveAttribute('hidden');
     });
 
-    it('should not have aria attribute when panel is open', () => {
+    it('should not have hidden attribute when panel is open', () => {
       const { panelContent } = renderPanel({ isClosed: false });
 
       expect(panelContent).not.toHaveAttribute('hidden');
--- a/packages/components/collapsible-panel/src/collapsible-panel.spec.js
+++ b/packages/components/collapsible-panel/src/collapsible-panel.spec.js
@@ -206,16 +206,16 @@ describe('aria attributes', () => {
   });
 
   describe('content', () => {
-    it('should have aria-hidden true when panel is closed', () => {
+    it('should have hidden attribute when panel is closed', () => {
       const { panelContent } = renderPanel({ isClosed: true });
 
-      expect(panelContent).not.toHaveAttribute('hidden');
+      expect(panelContent).toHaveAttribute('hidden');
     });
 
-    it('should have aria-hidden false when panel is open', () => {
+    it('should not have aria attribute when panel is open', () => {
       const { panelContent } = renderPanel({ isClosed: false });
 
-      expect(panelContent).toHaveAttribute('hidden');
+      expect(panelContent).not.toHaveAttribute('hidden');
     });
   });
 });
--- a/packages/components/collapsible-panel/src/collapsible-panel.spec.js
+++ b/packages/components/collapsible-panel/src/collapsible-panel.spec.js
@@ -160,85 +160,62 @@ it('should forward data-attributes', () => {
   expect(container.querySelector("[data-foo='bar']")).toBeInTheDocument();
 });
 
-describe('getPanelContentId', () => {
-  it('should return a string containing the given id', () => {
-    const panelContentId = CollapsiblePanel.getPanelContentId('example');
-
-    expect(panelContentId).toEqual(expect.stringContaining('example'));
-  });
-});
-
 describe('aria attributes', () => {
   const renderPanel = (props) => {
     const { container } = render(
-      <CollapsiblePanel
-        id="example"
-        header="Header"
-        onToggle={jest.fn()}
-        {...props}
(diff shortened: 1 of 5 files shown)


packages/components/collapsible-panel/src/collapsible-panel.spec.js
```diff
@@ -160,85 +160,62 @@ it('should forward data-attributes', () => {
...
+    const { panelHeader, panelContent } = renderPanel({
       id: 'test-id',
       isClosed: false,
     });
 
-    const panelContentId = CollapsiblePanel.getPanelContentId('test-id');
-
-    // assert that the header button has the aria attribute
-    const headerButton = getPanelHeader();
-    expect(headerButton).toHaveAttribute('aria-controls', panelContentId);
```
CarlosCortizasCT, line 206: This is now implicitly done in the `renderPanel` function to get access to the panel content element.

packages/components/collapsible-panel/src/collapsible-panel.tsx
```diff
@@ -245,7 +250,9 @@ const CollapsiblePanel = (props: TCollapsiblePanel) => {
                 <Spacings.Inset scale={scale}>
                   <SectionContent
                     id={panelContentId}
-                    aria-hidden={isOpen ? 'false' : 'true'}
+                    role="region"
+                    aria-labelledby={panelButtonId}
+                    hidden={isOpen}
```
CarlosCortizasCT, line 255: Updated attributes based on [this reference](https://w3c.github.io/aria-practices/examples/accordion/accordion.html).

packages/components/collapsible-panel/src/collapsible-panel.tsx
```diff
@@ -259,7 +266,10 @@ const CollapsiblePanel = (props: TCollapsiblePanel) => {
   );
 };
 
-CollapsiblePanel.getPanelContentId = getPanelContentId;
+/**
+ * @deprecated This function is no longer supported.
+ */
+CollapsiblePanel.getPanelContentId = () => '';
```
CarlosCortizasCT, line 272: I think this function [was introduced](https://github.com/commercetools/ui-kit/pull/1272/files#diff-8d9b9a02ab995f407cfe27c55cc92198345216e51dfbac565109a13f4b28fa55) when writing the first tests that took a11y into account but I updated them to not rely on it since it was also not working in every scenario.

Although this function is public, I don't think it's being actually used.
I search the Merchant Center frontend and App-kit repositories and didn't find any usage of this function.

I'm not sure if it would be ok to just remove it but it could happen that any `ui-kit` user could be using it. But again, It think that function was not meant to be publicly used not work in all scenarios.

Let me know what you think about it.

CarlosCortizasCT:
Percy check failing is because now the ID is not a default property but it's created once in the render of the component (using out `useFieldId`) and the properties are part of the screenshots that are compared.

So I think it is safe to accept the changes.

packages/components/collapsible-panel/src/collapsible-panel.spec.js
```diff
@@ -160,85 +160,62 @@ it('should forward data-attributes', () => {
...
 
   describe('content', () => {
     it('should have aria-hidden true when panel is closed', () => {
-      const { getPanelContent } = renderPanel({ isClosed: true });
-
-      const panelContent = getPanelContent();
+      const { panelContent } = renderPanel({ isClosed: true });
 
-      expect(panelContent).toHaveAttribute('aria-hidden', 'true');
+      expect(panelContent).not.toHaveAttribute('hidden');
```
emmenko, line 212: Can you explain the difference here?

Maybe I'm missing some context around how `aria-hidden` and `hidden` work 🤔

CarlosCortizasCT: The `hidden` attribute is more semantically correct in this case as we want the element not to be visible for the browser nor the screen readers.
`aria-hidden` attribute only "hides" the element to screen readers which it's ok for this case since we are also hiding the element with CSS.
(_`hidden` attribute also implies `aria-hidden`_).

I wanted to follow the reference I found in the [w3c example](https://w3c.github.io/aria-practices/examples/accordion/accordion.html).
I guess [this](https://stackoverflow.com/a/49807034) is a more detailed explanation about both attributes.

emmenko: Thanks, I get it.

However, I'm still confused why before we were checking that the `aria-hidden` was set to `true` and now checking that is NOT `hidden` 🧐

kark (approved):
Thanks! It looks good to me (and Lighthouse doesn't raise now any issues either). Many thanks for sharing the link to `ARIA Design Pattern Examples` 🙌

emmenko (approved):
Thanks for the improvements!

PR #2219, opened 2022-06-23
Allow generate-readmes script to work with components using forwardref 

ddouglasz:
Allow generate-readmes script to work with components using forwardref. Creating a patch for `typescript-react-function-component-props-handler` to allow components using forwardref.

This is an interim solution as I have submitted an [issue](https://github.com/Winner95/typescript-react-function-component-props-handler/issues/3) regarding the code snippet and also raised a [PR](https://github.com/Winner95/typescript-react-function-component-props-handler/pull/2) suggestion as follow up.


CarlosCortizasCT (approved):
💯

patches/typescript-react-function-component-props-handler+1.1.1.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/node_modules/typescript-react-function-component-props-handler/index.js b/node_modules/typescript-react-function-component-props-handler/index.js
+index 5c636b9..ca4c9a0 100644
+--- a/node_modules/typescript-react-function-component-props-handler/index.js
++++ b/node_modules/typescript-react-function-component-props-handler/index.js
+@@ -47,10 +47,6 @@ function checkForProptypes(path, paramTypeName) {
+ }
+ 
+ function setParamsTypeDefinitionFromFunctionType(documentation, path) {
+-    if (path.parentPath.node.init && path.parentPath.node.init.params.length === 0) {
```
kark, line 9: Great work overall 👏!
For our case I guess it works fine, but as for the external library PR this check might be crucial in some cases. Perhaps this would work better:
```suggestion
+ if (path.parentPath.node.init  && Array.isArray(path.parentPath.node.init.params) && path.parentPath.node.init.params.length === 0) {
        return;
    }
```

ddouglasz: This is true, I had this thought too as regards the check, there must be a reason for it and maybe not in our favor, but in favor of other users. I think this is a better check. 
Thanks! 
updated [here](https://github.com/commercetools/ui-kit/pull/2219/commits/1ea9a91c0663deddc79a051a926aa818db13d458)

kark (approved):

patches/typescript-react-function-component-props-handler+1.1.1.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/typescript-react-function-component-props-handler/index.js b/node_modules/typescript-react-function-component-props-handler/index.js
+index 5c636b9..2f3cb68 100644
+--- a/node_modules/typescript-react-function-component-props-handler/index.js
++++ b/node_modules/typescript-react-function-component-props-handler/index.js
+@@ -47,9 +47,9 @@ function checkForProptypes(path, paramTypeName) {
+ }
+ 
+ function setParamsTypeDefinitionFromFunctionType(documentation, path) {
+-    if (path.parentPath.node.init && path.parentPath.node.init.params.length === 0) {
++    if (path.parentPath.node.init  && Array.isArray(path.parentPath.node.init.params) && path.parentPath.node.init.params.length === 0) {
```
emmenko, line 10: I'm missing a bit of context here. Can you explain how this makes `forwardRef` work?
Thanks 🤗

emmenko (commented):
Thanks for picking this up! 🙏

patches/typescript-react-function-component-props-handler+1.1.1.patch
emmenko: I see, thanks.

Btw, this package does not seem maintaned since a couple of years.

We can either:
* keep using the patch.
* copy the implementation file and use it in our repo, without relying on the package.

emmenko:
@ddouglasz There are some packages with a `docs_disabled` folder, due to the usage of `forwardRef`.

Can you rename these folders back to `docs` and re-run the generator, to see if the issues are gone?

ddouglasz:
> @ddouglasz There are some packages with a `docs_disabled` folder, due to the usage of `forwardRef`.
> 
> Can you rename these folders back to `docs` and re-run the generator, to see if the issues are gone?

Oh, I can see one more package:`AccessibleButton`. This does not seem to fix it. I will investigate further to see why.

packages/components/buttons/accessible-button/src/accessible-button.tsx
```diff
@@ -91,7 +91,7 @@ const defaultProps: Pick<
 };
 
 const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
-  (props, ref) => {
+  (props: TAccessibleButtonProps, ref) => {
```
emmenko, line 94: So this makes it work? Interesting...

Any idea why the types in the `forwardRef` were not used? 🤔

emmenko: I mean this `forwardRef<HTMLButtonElement, TAccessibleButtonProps>`. This should make the `props` already typed.

ddouglasz: Yes, you are correct. This is how forwardRef is typed but I guess for some reason, where we are using [reactDocgen](https://github.com/commercetools/ui-kit/blob/84e45e0619e96b6c585c8f1f048a0025400881e6/generators/readme/src/index.ts#L332) , it does not parse the complete prop attributes using that syntax.

emmenko (approved):
Thanks again, I think we can proceed with this 👌

PR #2222, opened 2022-06-28
fix: `ClearIndicator` accessibility improvements

kark:
#### Summary

The aim of this PR is to add some accessibility improvements to the `ClearIndicator` component.
The warning from `Lighthouse`:
```
[aria-hidden="true"] elements contain focusable descendents. Focusable descendents within an `[aria-hidden="true"]` element prevent those interactive elements from being available to users of assistive technologies like screen readers.
```

### DIFF
--- a/packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
+++ b/packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
@@ -18,7 +18,7 @@ const ClearIndicator = (props: TClearIndicator) => {
   const intl = useIntl();
   const {
     getStyles,
-    innerProps: { ref, ...restInnerProps },
+    innerProps: { ref, onMouseDown, ...restInnerProps },
   } = props;
   return (
     <button
@@ -37,6 +37,10 @@ const ClearIndicator = (props: TClearIndicator) => {
       style={getStyles('clearIndicator', props) as CSSProperties}
       title={intl.formatMessage(messages.clearButtonLabel)}
       aria-label={intl.formatMessage(messages.clearButtonLabel)}
+      // overriding the default `aria-hidden` prop value to make the component accessible by keyboard - https://github.com/JedWatson/react-select/issues/4793
+      aria-hidden={false}
+      // only onMouseDown and onTouchEnd event handlers are passed by `react-select` to the component by default, which makes it not accessible by keyboard
+      onClick={onMouseDown}
     >
       <CloseIcon color="solid" size="medium" />
     </button>
--- a/.changeset/tall-cooks-sneeze.md
+++ b/.changeset/tall-cooks-sneeze.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/select-utils': patch
 ---
 
-ClearIndicator component accessibility improvements
+`<ClearIndicator>` component accessibility improvements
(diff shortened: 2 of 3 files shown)


packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
```diff
@@ -37,6 +37,10 @@ const ClearIndicator = (props: TClearIndicator) => {
       style={getStyles('clearIndicator', props) as CSSProperties}
       title={intl.formatMessage(messages.clearButtonLabel)}
       aria-label={intl.formatMessage(messages.clearButtonLabel)}
+      // overriding the default `aria-hidden` prop value to make the component accessible by keyboard - https://github.com/JedWatson/react-select/issues/4793
+      aria-hidden={false}
```
kark, line 41: `react-select` allows using customized components passed as `components` prop. `ClearIndicator` is one of the supported customizable components and our implementation of `SelectInput` uses a customized `ClearIndicator`.

The issue reported by Lighthouse is caused by the default props passed to the `ClearIndicator` component (namely `aria-hidden: true`). This is the [rationale](https://github.com/JedWatson/react-select/issues/4793) of the lib creators. Please mind though that they keep the `ClearIndicator` not accessible by keyboard by default. In the `react-select` [docs examples](https://react-select.com/components#adjusting-the-styling) in each of the cases the container element has attribute `aria-hidden=“true”` and the svg element has attributes: `aria-hidden:”true”` and `focusable="false”`.

 I am assuming this is not the way we wish that to work (as we use a button for the `ClearIndicator` component).

I have followed the advice from the GH issue to override the default bahavior.

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
```diff
@@ -37,6 +37,10 @@ const ClearIndicator = (props: TClearIndicator) => {
...
+      // only onMouseDown and onTouchEnd event handlers are passed by `react-select` to the component by default, which makes it not accessible by keyboard
+      onClick={onMouseDown}
```
kark, line 43: Another issue I noticed is that by default the `ClearIndicator` does not trigger any action on “Enter” key stroke. Therefore, the component is not accessible by keyboard when `isClearable` prop is passed to the component. The reason is that by default only `onMouseDown` and `onTouchEnd` event handlers are passed to the component. I am suggesting using the `onClick` instead, but perhaps for some reason this is not what we want

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
CarlosCortizasCT: So it seems we have two options:
* Do not hide the reset button to screen readers
* Do explicitly hide inner components of `ClearIndicator` (the button) to screen readers

It seems they defaulted to the second one in the library to mimic the behaviour of reset handle in HTML search inputs.

From my point of view, I prefer the option you've taken here 👍 .

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
CarlosCortizasCT: For me, it also makes sense to use the `onClick` handler.

CarlosCortizasCT (approved):
LGTM

PR #2224, opened 2022-07-01
fix: `Avatar` accessibility improvements

kark:
#### Summary

This PR attempts to fix the error reported by axe
```
Alternative text of images should not be repeated as text. 
Element contains <img> element with alt text that duplicates existing text
```

### DIFF
--- a/packages/components/avatar/src/avatar.tsx
+++ b/packages/components/avatar/src/avatar.tsx
@@ -28,9 +28,7 @@ export type TAvatarProps = {
 export type TGravatarImgProps = Pick<
   TAvatarProps,
   'gravatarHash' | 'isHighlighted' | 'size'
-> & {
-  altText?: string;
-};
+>;
 
 export type TInitialsProps = Pick<
   TAvatarProps,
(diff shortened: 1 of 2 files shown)


packages/components/avatar/src/avatar.tsx
```diff
@@ -94,7 +92,11 @@ const GravatarImg = (props: TGravatarImgProps) => (
       `${createGravatarImgUrl(props.gravatarHash, props.size)} 1x`,
       `${createGravatarImgUrl(props.gravatarHash, props.size, 2)} 2x`,
     ].join(',')}
-    alt={props.altText}
+    /**
+     * Based on https://www.w3.org/WAI/tutorials/images/decision-tree/:
+     * image contributes meaning to the current page or context and it shows content that is redundant to real text nearby
+     */
+    alt=""
```
kark, line 99: This might be a little controversial, so I wanted to give a little context.
The issue is reported in axe and this is the [writeup](https://dequeuniversity.com/rules/axe/4.4/image-redundant-alt?application=AxeChrome) referenced by axe.
 
My solution was based on the [alt decision tree](https://www.w3.org/WAI/tutorials/images/decision-tree/). 
Screenshot 2022-07-01 at 12 36 17

Additionally, please note that this is what e.g. Google does:
Screenshot 2022-07-01 at 12 39 28

On the other hand, I have also found many other solutions e.g.:
- Atlassian's [avatar component](https://atlassian.design/components/avatar/examples) contains full name as alt.
- GitHub, depending on the page, uses `alt=“”`, `alt=“Avatar”` or `alt=“@username”` 🤯

It would be cool to know your opinions.

emmenko: Thanks for looking into that!

I think using `alt=""` or `alt="Avatar"` would be my preference.

CarlosCortizasCT: I also agree having an empty `alt` is a good choice here 👍.

CarlosCortizasCT (approved):

emmenko (approved):

Issue #1661, opened 2020-11-09
Rethinking Theming support


emmenko:
We should also probably rethink a bit how we plan and want to continue support theming.
At the moment is more of a "hidden" feature but we do rely on it in other places.
It's also not super consistent as only a bunch of components support that.

Apparently, in some recent emotion updates there have been some "regressions" (#1660 and https://github.com/commercetools/commercetools-docs-kit/pull/745) which forces us to fix that here.

I think we need to make a decision how we want to continue the theming support.

PS: have a read at https://epicreact.dev/css-variables/ as an inspiration for an alternative approach (assuming it would be possible in our setup).

bradd123:
Has any discussion happened regarding this?

tdeekens:
Nope. It hasn't been a priority and we do not need it ourselves at the moment.

PR #2227, opened 2022-07-11
💅 Prepare theming support in design system

kark:
#### Summary

Implementation of https://wiki.commercetools.com/display/shield/%5BRFC%5D+Theming+feature
Part of #1661 initiative

#### ToDo
- [x] simplify alternative themes definitions
- [ ] invalid percy tests (after theme change) - Percy issue tracking https://github.com/kark/percy-snapshot/issues/1
- [x] enable theme persistence upon page refresh in the `useTheme` hook (?)


design-system/src/theme-provider.tsx
```diff
@@ -0,0 +1,68 @@
...
+      setTheme(newTheme as ThemeName);
+    } else {
+      setTheme('default');
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    const vars = toVars(themes[theme]);
+    Object.entries(vars).forEach(([key, value]) => {
+      root.current?.style.setProperty(key, value);
```
CarlosCortizasCT, line 45: With this implementation, we don't remove the old theme vars from the document.
This is not a problem if we want themes to cascade. I mean, if we want to have the `default` one always loaded and then the rest to use it as a blueprint where they only overwrite specific parts.

There is an edge case though that I'm not sure it's worth taking into account:
* Let's say we have the these CSS vars in the `default` theme:
```
primaryColor: '#123123';
secondaryColor: '#234234';
```
* Then we have another theme (`theme-A`) where want to overwrite the primary color:
```
primaryColor: '#567567';
```
* And another theme (`theme-B`) where we overwrite the secondaryColor:
```
secondaryColor: '#789789';
```

If the users loads `default` theme, then `theme-A` and then `theme-B`, the resulting state would be:
```
primaryColor: '#567567';
secondaryColor: '#789789';
```
when maybe, as we have the `theme-B` selected and we want all themes to overwrite `default`, the state should be:
```
primaryColor: '#123123';
secondaryColor: '#789789';
```

It's just a matter of how we want the themes transition to work.

CarlosCortizasCT (commented):
So far this is a great work towards supporting Theming in ui-kit 💪 

I just wanted to leave some early feedback with some ideas after a first review.

design-system/src/theme-provider.tsx
CarlosCortizasCT: I thought about this again and I think I was wrong.
Since we are inlining the styles, when we re-render the element with the new vars, the older ones would be gone.

design-system/src/theme-provider.tsx
```diff
@@ -0,0 +1,68 @@
...
+const ThemeProvider = (props: ThemeProviderProps) => {
+  const root = useRef<HTMLDivElement>();
+  const [theme, setTheme] = useState<ThemeName>('default');
+
+  const changeTheme = useCallback((newTheme: string) => {
+    if (allThemesNames.some((themeName) => themeName === newTheme)) {
+      setTheme(newTheme as ThemeName);
+    } else {
+      setTheme('default');
+    }
```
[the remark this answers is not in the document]
kark: Thanks, good points 👍 Updated in e72166682be41460347cde809efcd13c78e78397

design-system/src/theme-provider.tsx
```diff
@@ -0,0 +1,68 @@
...
+    });
+  }, [theme]);
+
+  const value = useMemo(() => {
+    return { theme, changeTheme };
+  }, [theme, changeTheme]);
+
+  return (
+    <ThemeContext.Provider value={value}>
+      <div ref={root as LegacyRef<HTMLDivElement>}>{props.children}</div>
```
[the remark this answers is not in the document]
kark: Yes, that is also a valid point, I guess @FilPob brought it up in one of our meetings. Updated in https://github.com/commercetools/ui-kit/commit/e72166682be41460347cde809efcd13c78e78397

design-system/src/theme-provider.tsx
```diff
@@ -28,31 +27,35 @@ type ThemeProviderProps = {
...
 
   const changeTheme = useCallback((newTheme: string) => {
-    if (allThemesNames.some((themeName) => themeName === newTheme)) {
-      setTheme(newTheme as ThemeName);
-    } else {
-      setTheme('default');
-    }
+    const isNewThemeValid = allThemesNames.some(
+      (themeName) => themeName === newTheme
+    );
```
CarlosCortizasCT, line 36: nit: since we are iterating over a string array, I guess it would be simpler (even faster) to use `.include()` instead:
```
const isNewThemeValid = allThemesNames.include(newTheme);
setTheme(isNewThemeValid ? (newTheme as ThemeName) : 'default');
warning(
  isNewThemeValid
  `ThemeProvider: the specified theme '${newTheme}' is not supported.`
);
```

design-system/src/theme-provider.tsx
```diff
@@ -28,31 +27,35 @@ type ThemeProviderProps = {
 };
 
 const ThemeProvider = (props: ThemeProviderProps) => {
-  const root = useRef<HTMLDivElement>();
+  const root = document.querySelector(':root') as HTMLElement;
```
CarlosCortizasCT, line 30: Although the we are not expecting the `ThemeProvider` to be re-rendered many times, perhaps we can keep using a ref here as the root element (`html`) will not change and so we don't need to query the document on every render.

kark: I'm sorry but I don't know if I follow it ([considering the previous comment](https://github.com/commercetools/ui-kit/pull/2227#discussion_r930015715)). Is your point here to use `forwardRef` for this component and expect the ref to `root` element be provided in the client code?

CarlosCortizasCT: Sorry, what I meant is to use `useRef` because the root element won't change and it does not need to be queried in every render:
```suggestion
  const root = useRef<HTMLElement>(document.querySelector(':root'));
```

design-system/src/theme-provider.tsx
```diff
@@ -0,0 +1,82 @@
...
+const validateTheme = (themeName?: string): ThemeName => {
+  if (!themeName) {
+    return 'default';
+  }
+  const isNewThemeValid = allThemesNames.includes(themeName);
+  if (isNewThemeValid) {
+    return themeName as ThemeName;
+  }
+  if (!isNewThemeValid) {
+    console.warn(
```
kark, line 41: I am not using `warning` function here because it would throw every time someone changes theme in storybook to `Dark theme (experimental)` (and therefore make using strorybook very cumbersome) until other themes are properly defined in `definition.yaml` file

design-system/src/theme-provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const ThemeProvider = (props: ThemeProviderProps) => {
+  const parentSelectorRef = useRef(props.parentSelector);
+  useLayoutEffect(() => {
+    applyTheme({
+      newTheme: props.theme,
+      parentSelector: parentSelectorRef.current,
+      themeOverrides: props.themeOverrides,
+    });
+  }, [props.theme, props.themeOverrides]);
```
emmenko, line 73: I was wondering a bit about this, as `props.themeOverrides` could be passed as inline object in the props, causing the reference to always change (thus executing the effect on each render).

Can you think of an approach we can use to avoid that?

CarlosCortizasCT: I've opened [this PR](https://github.com/commercetools/ui-kit/pull/2272) with a proposal.

presets/ui-kit/src/index.ts
```diff
@@ -51,6 +51,9 @@ export {
   useDataTableSortingState,
 } from '@commercetools-uikit/hooks';
 
-export { customProperties } from '@commercetools-uikit/design-system';
+export {
+  customProperties,
```
CarlosCortizasCT, line 55: Perhaps we can also add the deprecated comment here.
```suggestion
  /** @deprecated use `designTokens` instead */
  customProperties,
```

CarlosCortizasCT (commented):
Superb work, Kacper 👏 👏 👏 

I just left some final minor comments.

Also, I reviewed all the percy snapshot changes and there is [this one](https://percy.io/commercetools-GmbH/ui-kit/builds/21319643/changed/1188448271?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1024&widths=1024) I think we need to double check. It's probably ok, but it would be great to take another look at it.

.changeset/few-maps-smell.md
```diff
@@ -0,0 +1,78 @@
...
+'@commercetools-uikit/stamp': minor
+'@commercetools-uikit/tag': minor
+'@commercetools-uikit/text': minor
+'@commercetools-uikit/tooltip': minor
+'@commercetools-uikit/view-switcher': minor
+'@commercetools-frontend/ui-kit': minor
+'visual-testing-app': minor
+---
+
+Introduce theming support to ui-kit components based on the use of **CSS custom properties** (sometimes referred to as **CSS variables** or **cascading variables**).
```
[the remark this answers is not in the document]
kark: Sounds good. How about?

> Refactor UI kit components and `@commercetools-uikit/design-system` package for future theming support. `customProperties` object exporteded from the `@commercetools-uikit/design-system` is now deprecated (although still exported for backwards compatibility). It is advised to use the `designTokens` object instead.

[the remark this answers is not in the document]
CarlosCortizasCT: Personally, I wouldn't mention anything about theming yet but I'm ok with your suggestion. 👍

design-system/package.json
```diff
@@ -19,11 +19,16 @@
...
-    "build:tokens": "node ./scripts/generate-custom-properties.js",
+    "build:tokens": "node ./scripts/generate-design-tokens.js",
     "build:tokens:watch": "nodemon -e yaml --quiet --watch materials/internals --exec 'yarn build:tokens'"
   },
   "dependencies": {
     "@babel/runtime": "^7.17.2",
-    "@babel/runtime-corejs3": "^7.17.2"
+    "@babel/runtime-corejs3": "^7.17.2",
+    "@commercetools-uikit/utils": "15.2.1",
+    "@emotion/react": "^11.4.0",
```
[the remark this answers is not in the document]
CarlosCortizasCT: It seems we're keeping the `@commercetools-uikit/utils`.
I'm curious about what we need that dependency for in this package 🤔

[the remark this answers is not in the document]
kark: thanks, updated in d1ac5bfbe3cc460c86661dc356a6aa63afe52491

.changeset/few-maps-smell.md
```diff
@@ -0,0 +1,61 @@
...
+'@commercetools-uikit/text': minor
+'@commercetools-uikit/tooltip': minor
+'@commercetools-uikit/view-switcher': minor
+'@commercetools-frontend/ui-kit': minor
+'visual-testing-app': minor
+---
+
+Refactor UI kit components and `@commercetools-uikit/design-system` package for future theming support.
+
+`customProperties` object exporteded from the `@commercetools-uikit/design-system` is now deprecated (although still exported for backwards compatibility). It is advised to use the `designTokens` object instead.
```
emmenko, line 61:
```suggestion
Prepare theming support in our design system.

This is an internal change to restructure how we define and use design tokens, in particular by relying on CSS variables. Consumers are not affected by any of these changes.

Note that the `customProperties` object exported from the `@commercetools-uikit/design-system` package is now deprecated (although still exported for backwards compatibility) in favour of the `designTokens` object.
```

design-system/src/design-tokens.ts
```diff
@@ -1,38 +1,295 @@
...
+    sizeHeightTag: '26px',
+    standardInputHeight: '32px',
+  },
+  dark: {
+    colorPrimary: '#f78d26',
+    colorSolid: '#fff',
+    colorSurface: '#1a1a1a',
+    backgroundColorForTag: 'hsl(25.110132158590307, 89.0196078431%, 95%)',
+    backgroundColorForTagWarning: 'hsl(0, 0%, 95%)',
+  },
```
emmenko, line 137: What do we plan to do with this? Not sure we should merge this...or at least have a different name?

design-system/src/design-tokens.ts
```diff
@@ -1,38 +1,295 @@
...
-  fontColorForTag: 'colorSolid',
-  fontColorForTagWhenDisabled: 'colorNeutral60',
-  fontColorForTextWhenInverted: 'colorSurface',
-  fontFamily: 'fontFamilyBody',
-  placeholderFontColorForInput: 'colorNeutral60',
-  fontSizeForInput: 'fontSizeM',
-  shadowBoxTagWhenHovered: 'shadow1',
+export const themesNames = {
+  default: 'default',
+  dark: 'dark',
```
emmenko, line 142: Do we want to export these values for now? I want to avoid possible confusion if these values are available as exports 🤔

emmenko (commented):
Again, thank you! Big kudos here 🥳

Had a quick look and overall looks very good to me! 🙌

I probably won't check all changes in every file but I trust you and our test coverage to make sure everything works as expected. 😉

design-system/src/design-tokens.ts
kark: I'm not sure either. As this is only used [here](https://github.com/commercetools/ui-kit/blob/292cb29e0a2bc98570ba4f8c7eac137017362bad/design-system/src/theme-provider.spec.js#L108) and [here](https://github.com/commercetools/ui-kit/blob/292cb29e0a2bc98570ba4f8c7eac137017362bad/design-system/src/theme-provider.spec.js#L143), I renamed it to `test`. I think @CarlosCortizasCT [is in favour of leaving it be](https://github.com/commercetools/ui-kit/pull/2227/commits/0dca49ff90dfcfbb4d4c40903ac92e412a444e17).

CarlosCortizasCT (approved):
I took another look at the changes in the PR and also to the Percy diffs and I think this is good to go now.
I would prefer to have an isolated version for these changes only, but it's not mandatory at all.

Again, congrats @kark for such a great job 💪 💯 🚀

emmenko (approved):
Good to go from my side as well. Looking forward to see this in action.

Great job everyone! 🥳 🙌

design-system/src/design-tokens.ts
CarlosCortizasCT: Updated in this commit: 67e20d87b162991132e047c2cd58ca00c96ed93e

PR #2235, opened 2022-07-21
fix: enable deserialization of nested html markup

kark:
#### Summary

Fixes regression in `@commercetools-uikit/rich-text-utils` related to deserialization of nested html markup

PR #2236, opened 2022-07-21
fix: deserialization - wrap each resulting non-empty child text node with a paragraph

kark:
#### Summary

This is a follow up of #2235.
I'm very sorry this was done in turns, I should have added visual tests to the previous PR 😕
It turned out that each text node provided to `slate` editor has to be wrapped with a paragraph.

For instance:
`<p>this </p>will<p> fail</p>`
`<p>this </p><p>is</p><p> ok</>`

### DIFF
--- a/.changeset/slow-bags-cheer.md
+++ b/.changeset/slow-bags-cheer.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/rich-text-utils': patch
+---
+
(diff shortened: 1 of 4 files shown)


tdeekens (approved):
Nice. The visual spec is a good idea!

emmenko (approved):
❤️

PR #2240, opened 2022-07-26
Add usage as controlled component to ViewSwitcher

tdeekens:
> First apologies for not opening an issue first. I started on the code after inspecting it. Consider this an issue in terms of obligation and discussion starting point with code as additional context.

#### Summary

The ViewSwitcher is an uncontrolled component maintaining its state internally not accepting any control from the outside. This intends allow it to be controlled similar to e.g. a collapsible panel which also supports both modes.

## Description

Assume you render the `ViewSwitcher` and what to keep the selected "view" in the URL. The URL is external state to the ViewSwitcher. The `defaultSelected` will be passed as a default value to `useState`. When not navigating using `history.push` and _not_ clicking a `ViewSwithcer` button, the `ViewSwitcher` will not update its selected value. We have these use case of it being route driven internally. We navigate from "View 1" rendering a list to a details page. When saving on the details page we intend to navigate to "View 2" using the URL. The "View 2" should be selected on the `ViewSwitcher` but will not be as nobody clicked and the internal (uncontrolled state) of the `ViewSwitcher` no deviated from the URL (or controlled state).

The suggested implementation is similar to the collapsibles where we use `warning` to inform users about additionally required props in either mode.


packages/components/label/README.md
```diff
@@ -25,12 +25,12 @@ import Label from '@commercetools-uikit/label';
...
 | `children`                   | `node`         | ✅ (\*)  | -                         | -       | Value of the label                                                                                              |
 | `intlMessage`                | `intl message` | ✅ (\*)  | -                         | -       | An intl message object that will be rendered with `FormattedMessage`                                            |
 | `isBold`                     | `bool`         |    -     | -                         | `false` | Indicates if the label title should be in bold text                                                             |
-| `isRequiredIndicatorVisible` | `bool`         |    -     | -                         | `false` | Indicates if the labeled field is required in a form                                                            |  |
+| `isRequiredIndicatorVisible` | `bool`         |    -     | -                         | `false` | Indicates if the labeled field is required in a form                                                            |     |
```
tdeekens, line 33: Just ran prettier on markdown files.

packages/components/view-switcher/README.md
```diff
@@ -43,14 +43,14 @@ yarn add react
 npm --save install react
 ```
 
-## Usage
+## Uncontrolled usage
```
tdeekens, line 46: Just to give information about both options.

packages/components/view-switcher/src/view-switcher.spec.js
```diff
@@ -131,6 +132,34 @@ describe('rendering', () => {
     screen.getByLabelText('Test Button 1').click();
     expect(handleClick).not.toHaveBeenCalled();
   });
+
+  it('should be controlled when selectedValue is passed', () => {
```
tdeekens, line 136: In terms of naming I suggested `selectedValue` as a pairing to `defaultValue`. I tried implementing `value` and `initialValue` but see two downsides. One it requiring the component to be backwards compatible to avoid a breaking change plus the additional toil work. Two the naming on Group and Button becomes a bit confusing in terms of what `value` on a Group means vs. a Button.

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -16,28 +17,44 @@ type TReactChild = {
...
   children: ReactNode;
   /**
-   * Will be triggered whenever a `ViewSwitcher.Button` is selected. Called with the ViewSwitcherButton value
+   * Will be triggered whenever a `ViewSwitcher.Button` is selected. Called with the ViewSwitcherButton value.
+   * This function is only required when the component is controlled.
    */
   onChange?: (value: string) => void;
   /**
-   * Indicates the default selected button
+   * Passing this prop makes the component an uncontrolled component.
```
tdeekens, line 33: The descriptions here as similar to the collapsible.

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -16,28 +17,44 @@ type TReactChild = {
...
+   * Controlled components also require to pass a `onChange` callback function.
+   */
+  selectedValue?: string;
 };
 
 const ViewSwitcher = (props: TViewSwitcherProps) => {
-  const [selectedButton, setSelectedButton] = useState<string>(
+  const isControlledComponent = !isNil(props.selectedValue);
+  const hasOnChange = !isNil(props.onChange);
+  const [selectedButton, setSelectedButton] = useState<string | undefined>(
```
tdeekens, line 47: Could be a needless useState but if we want to avoid it we need two child components and increase complexity due to the cloning possibly needing to be shared. Let me know what you think.

packages/components/view-switcher/src/view-switcher.tsx
```diff
@@ -16,28 +17,44 @@ type TReactChild = {
...
+   * Passing this prop makes the component an uncontrolled component.
+   * Indicates the default selected button it is only used to as an initial state once, when the component mounts.
    */
-  defaultSelected: string;
+  defaultSelected?: string;
+  /**
+   * Passing this prop makes the component an controlled component.
+   * Controlled components also require to pass a `onChange` callback function.
+   */
+  selectedValue?: string;
```
emmenko, line 41: Should we also warn if both props are passed? Cause the `selectedValue` (controlled mode) will take precedence, right?

packages/components/view-switcher/src/view-switcher.tsx
emmenko: Yeah, I think it's fine yes. The state is only used anyway in uncontrolled mode.

Maybe let's add a comment to make it explicit.

packages/components/view-switcher/src/view-switcher.spec.js
emmenko: I think it makes sense yes, thanks for thinking about it.

packages/components/view-switcher/src/view-switcher.spec.js
```diff
@@ -131,6 +132,34 @@ describe('rendering', () => {
...
+          <Button value="test-button-2" onClick={handleClick}>
+            Test Button 2
+          </Button>
+        </Group>
+      );
+    }
+    render(<TestComponent />);
+
+    screen.getByLabelText('Test Button 1').click();
+    expect(handleClick).not.toHaveBeenCalled();
```
emmenko, line 155: To make this test more clear, maybe we should explicitly pass the initial value so that it's clear that the click handler won't be called.

```suggestion
    render(<TestComponent selectedValue="test-button-1" />);

    screen.getByLabelText('Test Button 1').click();
    expect(handleClick).not.toHaveBeenCalled();
```

emmenko (commented):
Thanks for improving the component functionalities! 🙌

Overall looks good and the changes make sense to me. Just left a few small suggestions.

packages/components/view-switcher/README.md
```diff
@@ -61,19 +61,47 @@ const Example = () => (
...
+export default UncontrolledExample;
+```
+
+## Controlled usage
+
+```jsx
+import { useState } from 'react';
+import ViewSwitcher from '@commercetools-uikit/view-switcher';
+
+const UncontrolledExample = () => {
```
Rhotimee, line 73:
```suggestion
const ControlledExample = () => {
```

packages/components/view-switcher/README.md
```diff
@@ -61,19 +61,47 @@ const Example = () => (
...
+      <ViewSwitcher.Button isDisabled value="button 1">
+        View 1
+      </ViewSwitcher.Button>
+      <ViewSwitcher.Button value="button 2">View 2</ViewSwitcher.Button>
+      <ViewSwitcher.Button value="button 3">View 3</ViewSwitcher.Button>
+    </ViewSwitcher.Group>
+  );
+};
+
+export default UncontrolledExample;
```
Rhotimee, line 91:
```suggestion
export default ControlledExample;
```

Rhotimee (commented):
Thanks for the update 🙌🏽

packages/components/view-switcher/src/view-switcher.spec.js
tdeekens: You say "explicitly pass the initial" then pass the "selectedValue". Not sure I follow. We now will have a warning if we pass both. You're saying we should pass different defaultSelected and selectedValue to then assert that the default isn't used, right?

tdeekens: I added a commit. Maybe you adjust it if you thought otherwise. Thanks.

emmenko:
Thanks @tdeekens I'll take over then and finish up the PR.

emmenko (approved):
🔥

packages/components/view-switcher/README.md
```diff
@@ -1,26 +1,28 @@
+<!-- THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY. -->
```
tdeekens, line 1: Thanks. I wondered actually why it wasn't generated when changing it.

packages/components/view-switcher/src/view-switcher.spec.tsx
```diff
@@ -1,22 +1,24 @@
 import { useState } from 'react';
```
tdeekens, line 1: Also wondered why it was js but didn't see it as a concern of my PR :) Thanks for changing it here.

packages/components/view-switcher/src/view-switcher.spec.tsx
```diff
@@ -145,4 +180,26 @@ describe('warnings', () => {
       'ViewSwitcher.Group must contain at least one ViewSwitcher.Button'
     );
   });
+  it('should warn when value is passed but no onChange', () => {
```
CarlosCortizasCT, line 183: nit:
```suggestion
  it('should warn when selectedValue is passed but no onChange', () => {
```

CarlosCortizasCT (approved):
💯

packages/components/view-switcher/README.md
emmenko: The package must have the `docs` folder, otherwise the generation is skipped.

emmenko:
@Rhotimee Anything else to add from your side?

Rhotimee (approved):
💯

Rhotimee:
>Anything else to add from your side?

All good 🙌🏽

PR #2243, opened 2022-07-28
[POC] Fix proposal for Percy related issues with CSS vars

CarlosCortizasCT:
### Summary

This PR includes a proposal fix for the problem we found when implementing Theming support with Percy snapshot testing.

## Description

It seems Percy does not support using CSS vars the way we want to use them (CSS in JS).

Since our components use the `custom-properties.ts` file to get access to design tokens values which we want to be CSS vars in the new version but that does not seem to be working with Percy snapshot testing, following their recommendation (not to use CSS vars), in this proposal we add a marker in the visual testing app document (the one we use for snapshot testing) and use it in the mentioned file to not return a CSS var but the fixed default value when rendering in that environment.

### DIFF
--- a/design-system/scripts/generate-custom-properties.js
+++ b/design-system/scripts/generate-custom-properties.js
@@ -157,7 +157,7 @@ let _canUseCssVars: Boolean | null = null;
 const canUseCssVars = (): Boolean => {
   if (_canUseCssVars === null) {
     _canUseCssVars =
-      Boolean(document.querySelector('meta[name="ui-kit-vrt-environment"]'));
+      !Boolean(document.querySelector('meta[name="ui-kit-vrt-environment"]'));
   }
   return _canUseCssVars;
 };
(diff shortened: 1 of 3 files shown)


kark (approved):
This is brilliant 🤩 

Since this is branched out from [poc-theme-support-in-uikit](https://github.com/commercetools/ui-kit/tree/poc-theme-support-in-uikit) perhaps we could revert one change to be 100% that old vrts setup works:
```js
// jest-puppeteer.config.js
server: {
    command: 'yarn visual-testing-app:preview', // 👈
    ...
  },
```

CarlosCortizasCT:
> This is brilliant 🤩
> 
> Since this is branched out from [poc-theme-support-in-uikit](https://github.com/commercetools/ui-kit/tree/poc-theme-support-in-uikit) perhaps we could revert one change to be 100% that old vrts setup works:
> 
> ```js
> // jest-puppeteer.config.js
> server: {
>     command: 'yarn visual-testing-app:preview', // 👈
>     ...
>   },
> ```

Since that change was done in the source branch, I'd prefer to change it in that branch.

I can merge this one and then push that change in the other one.

PR #2247, opened 2022-08-02
Enable readme auto-generation for DataTableManager

kark:
#### Summary

This PR enables auto-generation of the readme file for `<DataTableManager>` component


packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -6,14 +6,97 @@ import {
...
 export interface TRow {
   id: string;
 }
 
+export type TColumnData = {
+  key: string;
+  label: ReactNode;
+};
+
+export type TDisplaySettingsProps = {
```
CarlosCortizasCT, line 20: It seems now these types are duplicated here and it their _owning_ component (`DataTableSettings`). I guess this is a requirement for the readme generator to find the definitions but it feels somehow unwanted to duplicate our types because of a tool.

Perhaps as a follow up we could have a new task to investigate how we could avoid duplicating the types.

kark: 💯

emmenko: The readme generator can only parse the AST of this file, as `react-docgen` is not able to "look" into other files.

image

Not sure how much we can do about it...

emmenko: We could maybe include multiple React components (in this case also `DataTableSettings`) so that we document both.

See example:
https://github.com/commercetools/ui-kit/blob/9a09eeec1ff7ab4b0bd95a4f0317ee589743fada/packages/components/primary-action-dropdown/package.json#L44

kark: Thanks, I added that here 0d0c0f2e7a5728fe6d8b8ad071e3f66ee02a4974

emmenko: It seems like we would need to adjust the signature reference linking, in case we include multiple components and the component matches one of the signatures.

For example `displaySettings?: TDisplaySettingsProps` is rendered as
image

But I think we want it to be linked to `DataTableSettings`, right?

kark: Many thanks for noticing that 👍 I updated that in 95e18767375c114da8e17c608fb65386747f3f7b
This works for GitHub readmes, however, I guess links to headings don't work in storybook
https://ui-lublrlj8z-commercetools.vercel.app/?path=/story/components-datatable--datatablemanager
(for all of the components I checked, not only this one)

kark: Thanks 👍 I will change that. The script still needs some work related to `--all-workspace-packages` option

kark: Thanks again @CarlosCortizasCT for finding the solution 💯 I updated the script and all readme-s in 7b4b9e7eaccbd1acd86f88651eb0e0636d04e185

CarlosCortizasCT:
It seems `columns` signature renders a little bloated.
Is there anything we can do to render it more clear as the others?

image


image

kark:
> It seems `columns` signature renders a little bloated. Is there anything we can do to render it more clear as the others?
> 
> image
> 
> image

Thanks, this is definitely a valid point, since it's a change definitely for the worse.

I started fixing that, but it seems the `storybook-readme` strorybook add-on has real problems with parsing and properly displaying markdown code block with jsdoc comments. I guess it's doable if we add enough `str.replace()`s (also changing the way the `README`s are displayed in GitHub), but I honestly feel this is waste of time considering we are anyway [thinking about displaying those interfaces in a table for better readability](https://jira.commercetools.com/browse/SHIELD-557). 

So my conclusion here is to wait with this PR (perhaps close it) until work on the mentioned task can be included here.

CarlosCortizasCT:
> > It seems `columns` signature renders a little bloated. Is there anything we can do to render it more clear as the others?
> > image
> > image
> 
> Thanks, this is definitely a valid point, since it's a change definitely for the worse.
> 
> I started fixing that, but it seems the `storybook-readme` strorybook add-on has real problems with parsing and properly displaying markdown code block with jsdoc comments. I guess it's doable if we add enough `str.replace()`s (also changing the way the `README`s are displayed in GitHub), but I honestly feel this is waste of time considering we are anyway [thinking about displaying those interfaces in a table for better readability](https://jira.commercetools.com/browse/SHIELD-557).
> 
> So my conclusion here is to wait with this PR (perhaps close it) until work on the mentioned task can be included here.

I agree 👍

PR #2256, opened 2022-08-04
[PoC] Add Theming support in VRTs

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=NtPvMRz)
### Summary

This is a proposal for adding theming support to VRTs so we could add new visual tests using different themes.

This builds on top of the work done [here](https://github.com/commercetools/ui-kit/pull/2227).


kark (approved):
Thanks 🎉

PR #2257, opened 2022-08-09
refactor: remove @emotion/react theming from ui-kit components

kark:
For anyone struggling with `Load diff` clicking -> please run this:
```console
document.querySelectorAll('.load-diff-button').forEach(node => node.click())
```

TODOs:
- [ ] fix theme provider vrt snapshots


packages/components/buttons/flat-button/src/flat-button.tsx
```diff
@@ -7,19 +6,18 @@ import {
   ComponentPropsWithRef,
   cloneElement,
 } from 'react';
-import { css, useTheme } from '@emotion/react';
+import { css } from '@emotion/react';
 import omit from 'lodash/omit';
-import { customProperties as vars } from '@commercetools-uikit/design-system';
```
kark, line 12: I took chance to change how we import `customProperties` object to 
```js
```
If however that is not needed, please let me know. 
Also, please let me know if you wish me to refactor all the remaining components.

kark (commented):
Hey @commercetools/shield-team-fe, I'm wondering if you could give me some feedback here

packages/components/link/src/link.visualroute.jsx
```diff
@@ -1,17 +1,12 @@
...
 import { Suite, Spec } from '../../../../test/percy';
 
 export const routePath = '/link';
 
 const intlMessage = { id: 'link', defaultMessage: 'Link' };
 
-const purpleTheme = {
-  colorPrimary: 'purple',
-  colorPrimary25: 'deeppurple',
-};
```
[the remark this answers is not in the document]
CarlosCortizasCT: So my point of view is that it does not make sense anymore to have theming tests for specific components so I would remove them.
Theming is global and I guess what we would need to test is just that the `ThemeProvider` works when you provide and allowed theme name (and maybe also the `changeName` helper).

There's a chance we also would want to have VRTs for every component in every theme so we make sure they look like they're supposed to. I wouldn't implement that, at least for now. And, if we ever do, perhaps we should look into some way to automate it instead of writing different theming tests for every component.

Also, I would not implement (at least for now) the ability for the users to provide a custom theme, ui-kit is tailored for Merchant Center and we want to control how it looks like, so themes should be controlled by us.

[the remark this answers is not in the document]
emmenko: FYI: In the docs-kit we do use some of the UI Kit components and we've also been using the theme to override some of the UI Kit design tokens. For example:

https://github.com/commercetools/commercetools-docs-kit/blob/2d3bf6efc3c4768cc3056d47da9ecb3fe21a77ae/packages/ui-kit/src/design-system.ts#L303-L306

Something to keep in mind maybe. We don't have to keep the same functionality (and it's fine if we need to refactor/change some stuff in docs-kit) but the end result should be the same ideally.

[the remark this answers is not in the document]
CarlosCortizasCT: It seems then we would need to implement the functionality to allow users to overwrite a theme since what they want (in `docs-kit`) is to change the font used by `ui-kit` components. Maybe they are using `emotion` theming internally since they are initializing the provider with `ui-kit` properties but also another one which does not belong to `ui-kit` design tokens ([reference](https://github.com/commercetools/commercetools-docs-kit/blob/3610dc6f692d281f1a1ad37653aed2134b40175e/packages/gatsby-theme-docs/src/layouts/internals/layout-application.js#L46)).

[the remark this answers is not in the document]
kark: Thanks for the comments 🙏 and for bringing up the `docs-kit`. I will look for the best option to keep vrts passing and provide backwards compatibility in the `docs-kit`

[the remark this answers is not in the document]
emmenko: The docs-kit has its own "design system" tokens (partly referencing UI Kit design tokens) and its own version of "theming" (using Emotion theming yes).

However, I would be happy and ok to refactor things there to follow the same "theming setup" that we plan to use here.

Let's sync on that next week maybe, but I wouldn't worry much about it.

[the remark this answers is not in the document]
kark: What I did for now was adding 3 additional themes (apart from `vrtDark`): `vrtLink`, `vrtTag` and `vrtTextInput` without adding custom overrides. These themes are filtered out from the ["all tokens" story](https://ui-ivzahz34a-commercetools.vercel.app/?path=/story/basics-tokens--all-tokens).

It seems this approach doesn't cause any breaking changes in the `docs-kit`. This is the [deployment link](https://harmonious-stroopwafel-158412.netlify.app/) of the root folder of docs website with `@commercetools-uikit/design-system` package within `@commercetools-docs/ui-kit` built from this PRs codebase:
`"@commercetools-uikit/design-system": "https://uikit-theming.netlify.app/commercetools-uikit-design-system-15.2.0.tgz"`. There is no `<ThemeProvider>` added and all css vars fall back to fixed values.

packages/components/inputs/time-input/src/time-input-body.styles.ts
```diff
@@ -8,162 +8,131 @@ import { type TTimeInputProps } from './time-input';
...
+    return customProperties.fontColorForInputWhenReadonly;
   }
-  return vars.fontColorForInput;
+  return customProperties.fontColorForInput;
 };
-const getInputContainerStyles = (props: TTimeInputProps, theme: Theme) => {
-  const overwrittenVars = {
-    ...customProperties,
-    ...theme,
-  };
```
kark, line 131: Regarding the change in visual regression tests:
Screenshot 2022-08-17 at 09 27 04
it seems to me that the previous visual result was a bug.
`customProperties.backgroundColorForInput` used for `background-color` was not overwritten by the theme.

packages/components/inputs/time-input/src/time-input-body.styles.ts
```diff
@@ -8,162 +8,131 @@ import { type TTimeInputProps } from './time-input';
...
-
+const getInputContainerStyles = (props: TTimeInputProps) => {
   return css`
     appearance: none;
     background-color: ${props.isDisabled
-      ? overwrittenVars.backgroundColorForInputWhenDisabled
-      : overwrittenVars.backgroundColorForInput};
```
kark, line 137: Regarding the change in visual regression tests:
Screenshot 2022-08-17 at 09 27 04
it seems to me that the previous visual result was a bug.
`customProperties.backgroundColorForInput` used for `background-color` was not overwritten by the theme.

This should have rather been:
```
? overwrittenVars[designTokens.backgroundColorForInputWhenDisabled]
: overwrittenVars[designTokens.backgroundColorForInput]};
```
but now it works properly due to the use of css vars.

packages/components/link/src/link.visualroute.jsx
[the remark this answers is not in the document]
kark: I updated the PR with changes that we discussed:
- [custom theme overrides in `<ThemeProvider>`](https://github.com/commercetools/ui-kit/blob/bf1c8698cc73948f67d409038b7291a3855ac5af/design-system/src/theme-provider.tsx#L81:L84) + [inclued that in the vrt](https://github.com/commercetools/ui-kit/blob/bf1c8698cc73948f67d409038b7291a3855ac5af/design-system/src/theme-provider.visualroute.jsx#L64:L91)
- [passing a ref to `<ThemeProvider>` to make it scoped locally](https://github.com/commercetools/ui-kit/blob/bf1c8698cc73948f67d409038b7291a3855ac5af/design-system/src/theme-provider.tsx#L77:L79) - I used that approach instead of the `getDefaultParentSelector` prop but if you prefer that please let me know
- in vrts removed fragments using local custom themes (applied to Tag, Link, TextInput). I left the [dark theme](test/percy/local-theme-provider.jsx) in vrts though

design-system/src/theme-provider.tsx
```diff
@@ -35,7 +38,7 @@ const toVars = (obj: Record<string, string>) =>
 type ThemeProviderProps = {
   children: ReactNode;
   theme?: string;
-  scope?: 'global' | 'local';
+  customPropertiesOverrides?: Record<string, string>;
```
CarlosCortizasCT, line 41: nit: perhaps we could follow the "theme" naming and have something like `themeOverrides`.

design-system/src/theme-provider.tsx
```diff
@@ -54,34 +57,50 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+    const localRef = ref as unknown as RefObject<HTMLDivElement>;
+
+    useEffect(() => {
+      const targetElement = localRef?.current
+        ? localRef.current
+        : rootRef?.current;
+      const vars = toVars(
+        props.customPropertiesOverrides &&
+          isObject(props.customPropertiesOverrides)
+          ? { ...themes[theme], ...props.customPropertiesOverrides }
```
CarlosCortizasCT, line 83: I guess we would need to update this line because the way JS merge objects by default can potentially remove props from the source theme:
```
// Example
const foo = { a: { b: 10, c: 20, d: 30 } };

const bar = { a: { d: 400 } };

const result = { ...foo, ...bar }; // { a: { d: 400 } }
// we overwrote `a` property

Perhaps we can use some helper library to help us here (example: [lodash.merge](https://lodash.com/docs/4.17.15#merge)).

File: design-system/src/theme-provider.visualroute.jsx [hunk 7]
```diff
@@ -1,45 +1,93 @@
...
+          title={
+            <>
+              Title with overridden <i>colorSolid</i> design token
+            </>
+          }
+        />
+      </LocalThemeProvider>
+    </Spec>
+
+    <Spec label="custom property with double hyphen and kebab-case naming added to default theme">
```
CarlosCortizasCT, line 76: Sorry, but I don't understand what we are testing here (also in the next test).
Could you maybe explain a little further:

test/percy/local-theme-provider.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+const darkTheme = {
+  colorSurface: 'black',
+  colorSolid: 'white',
+  colorNeutral60: 'rgba(255,255,255,0.60)',
+  colorNeutral: 'rgba(255,255,255,0.60)',
+  colorAccent98: 'rgba(0,0,0,0.98)',
+};
+
+export const LocalDarkThemeProvider = (props) => (
```
CarlosCortizasCT, line 34: I'm not sure why we still need this dark theme theme provider. I thought we agreed on not having themed tests but in the `ThemeProvider`. But maybe I'm wrong.

design-system/src/theme-provider.tsx
```diff
@@ -54,34 +57,50 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+        targetElement?.style.setProperty(key, value);
+      });
+    }, [theme, props.customPropertiesOverrides, localRef]);
+
+    const value = useMemo(() => {
+      return { theme, changeTheme };
+    }, [theme, changeTheme]);
+
+    return (
+      <ThemeContext.Provider value={value}>
```
CarlosCortizasCT, line 97: I guess @emmenko wanted to get rid of the React provider. Will we give a try as a next step or will we skip it for now?

CarlosCortizasCT (commented):
Great improvements! 💪 

I just left some suggestions to maybe consider.

design-system/src/theme-provider.tsx
emmenko: I agree with Carlos, I think `themeOverrides` is more consistent

design-system/src/theme-provider.tsx
```diff
@@ -54,34 +57,50 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+    );
+    const [theme, setTheme] = useState<ThemeName>(validateTheme(props?.theme));
+
+    const changeTheme = useCallback((newTheme: string) => {
+      setTheme(validateTheme(newTheme));
+    }, []);
+
+    const localRef = ref as unknown as RefObject<HTMLDivElement>;
+
+    useEffect(() => {
```
emmenko, line 76: Nit: any particular reason for not using `useLayoutEffect`?

design-system/src/theme-provider.tsx
```diff
@@ -54,34 +57,50 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+      const targetElement = localRef?.current
+        ? localRef.current
+        : rootRef?.current;
```
emmenko, line 79: Nit: I'm not a big fan of using refs, as it's a bit tricky to manage and type. Also here you see that we create a `rootRef` even though we might not use it.

As mentioned in our last meeting, we can also approach this like `react-modal` does: https://github.com/reactjs/react-modal/blob/403093510eae51b5ee289ee0a08e79ee1be068dd/src/components/Modal.js#L101

Overriding the target element is simply a matter of implementing the selector function.

To think about it...the goal is to make the setup simpler.

design-system/src/theme-provider.visualroute.jsx
kark: Sure, the purpose was to override the theme with custom properties that are not part of standard `customProperties` exposed from `@commercetools-uikit/design-system` and test the theme provider and a component in that case.

Also, to make users aware that they can use `css variables` notation directly and camel cased naming (for anyone using tests as guidance how to use the feature).

I don't think this test is crucial, perhaps it belongs to unit tests of the `<ThemeProvider>`

test/percy/local-theme-provider.jsx
kark: Perhaps I got it wrong. I had in my notes to remove only the custom theme. Removed it 413e7c109f783d0dc0d5b1f4d75dbdd2bf7a0d3a

design-system/src/theme-provider.tsx
kark: Thanks, I got rid of refs

design-system/src/theme-provider.tsx
kark: I got back to using `useLayoutEffect`

design-system/src/theme-provider.tsx
```diff
@@ -54,41 +29,74 @@ const validateTheme = (themeName?: string): ThemeName => {
...
-    return { theme, changeTheme };
-  }, [theme, changeTheme]);
+const changeTheme = ({
+  newTheme,
+  parentSelector = defaultParentSelector,
+  themeOverrides,
+}: TChangeTheme): void => {
+  const target = isBrowser ? parentSelector() : null;
+  const validTheme = validateTheme(newTheme);
+  target?.setAttribute('data-theme', validTheme);
```
kark, line 51: The idea behind this solution is:
- adding a `data-theme` either to `:root` or the parent based on the provided selector function. `data-theme` stores the current theme
-  adding all the custom properties to `style`

On each change of `theme`, `customProperties` or `parentSelector` the theme change is triggered within `useLayoutEffect`.

design-system/src/theme-provider.tsx
```diff
@@ -54,41 +35,79 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+  const [theme, setTheme] = useState<string | null | undefined>(null);
+  const memoizedParentSelector = useCallback(
+    () => parentSelector(),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
 
-  warning(!isEmpty(context), `useTheme must be used within a ThemeProvider`);
+  useEffect(() => {
+    setTheme(isBrowser ? memoizedParentSelector()?.dataset.theme : null);
```
kark, line 105: For setting the new theme `data-theme` attribute from parent is accessed

design-system/src/theme-provider.tsx
```diff
@@ -54,41 +29,79 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+  themeOverrides,
+}: TChangeTheme): void => {
+  const target = isBrowser ? parentSelector() : null;
+  const validTheme = validateTheme(newTheme);
+  target?.setAttribute('data-theme', validTheme);
+  const vars = toVars(
+    themeOverrides && isObject(themeOverrides)
+      ? merge(themes[validTheme], themeOverrides)
+      : themes[validTheme]
+  );
```
emmenko, line 56: I was wondering one thing here. At the moment we have our `default` theme and a custom theme. The custom theme is [defined and generated](https://github.com/commercetools/ui-kit/blob/bbc9329bfb98066d33a6886ed62ac8b444d6c471/design-system/materials/internals/definition.yaml#L144-L258) as a full theme.

I would expect that custom themes only define whatever properties they need to override, there should be no need to specify and duplicate all other properties. It also makes it difficult to see what values are actually different from the default theme.

In general I would think that custom themes should simply override the default them rather than replace it.
So when computing and applying the properties, we should probably always use the default theme as a base and override properties from the custom theme.

What do you think?

emmenko (commented):
Thanks, this looks much simpler now! 🥳

I left some questions as I would like to clarify some points. See comments.

design-system/src/theme-provider.tsx
```diff
@@ -1,43 +1,18 @@
...
     Object.entries(obj).map(([key, value]) => [`--${kebabCase(key)}`, value])
   );
 
-type ThemeProviderProps = {
-  children: ReactNode;
-  theme?: string;
-  scope?: 'global' | 'local';
-};
-
 const validateTheme = (themeName?: string): ThemeName => {
```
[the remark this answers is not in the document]
kark: Thanks, 

I'm wondering if `getValidTheme` as function name could work here?

Regarding the throwing instead of logging, I actually explained this [here](https://github.com/commercetools/ui-kit/pull/2227#discussion_r932993805). Perhaps it's better to merge this branch to the feature branch for a better overview and all the previous changes in diffs?

The problem with warning or throwing is that once we delete the `dark` theme from `definitions.yaml` in this PoC it will start throwing every time someone changes theme in storybook. That will make it very inconvenient in use until the real `dark` theme is introduced.

Of course we could also change the validation logic to remove the `dark` keyword from throwing.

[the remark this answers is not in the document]
CarlosCortizasCT: The way I see it, as long as we have the option to select a dark theme from the Storybook, we would need to keep a dark theme in our `definitions.yaml`.

Another option would be to remove that option from the Storybook.
Currently it only exists as a "experimental" feature, but we don't support it, so I guess it would be better to remove that selector from Storybook until we have an actual public theme apart from the default one.

design-system/src/theme-provider.tsx
CarlosCortizasCT: I guess that's something we discussed but didn't agree on the answer.

Personally I agree with Nicola's suggestion and use the `default` theme as the blueprint for any other.

kark: Thanks, that is a valid point, I also added it [to the todo list](https://github.com/commercetools/ui-kit/pull/2227#issue-1300529104)

I discussed it with @CarlosCortizasCT and the conclusion was to perhaps leave it for later. But I can do that now as well.

design-system/src/theme-provider.tsx
```diff
@@ -54,41 +30,79 @@ const validateTheme = (themeName?: string): ThemeName => {
...
+  newTheme,
+  parentSelector = defaultParentSelector,
+  themeOverrides,
+}: TApplyTheme): void => {
+  const target = isBrowser ? parentSelector() : null;
+  const validTheme = validateTheme(newTheme);
+  target?.setAttribute('data-theme', validTheme);
+  const vars = toVars(
+    themeOverrides && isObject(themeOverrides)
+      ? merge(cloneDeep(themes[validTheme]), themeOverrides)
```
CarlosCortizasCT, line 55: I was wondering if we do actually need to use `cloneDeep` here. I guess the problem was that we were mutating the theme in `themes` but I guess we can also prevent from that without cloning using a different target as the merge destination:
```suggestion
      ? merge({}, themes[validTheme], themeOverrides)
```

What do you think?

design-system/scripts/generate-custom-properties.js
```diff
@@ -31,29 +31,35 @@ const allThemesNames = Object.keys(definitions.choiceGroupsPerTheme);
...
-        if (!key.startsWith(themeChoiceGroup.prefix))
-          endProgram(
-            `Expected token "${key}" to start with "${themeChoiceGroup.prefix}" as it is an "${themeChoiceGroup.label}" attribute.`
-          );
+const defaultTheme = definitions.choiceGroupsPerTheme.default;
 
-        tokens[themeName][key] = value;
-      });
-    });
+Object.keys(definitions.choiceGroupsPerTheme).forEach((themeName) => {
```
CarlosCortizasCT, line 36: It seems we're recreating the default theme in the `customProperties` file for alternative themes. 
Perhaps we could just have the overwritten properties and always merge from `default` theme in the `ThemeProvider`.
What do you think?

CarlosCortizasCT (commented):
I left a couple of suggestions to maybe add some improvements.

design-system/src/theme-provider.tsx
```diff
@@ -1,94 +1,108 @@
...
+        newTheme,
+        parentSelector: parentSelectorRef.current,
+        themeOverrides,
+      });
+      setTheme(newTheme || 'default');
+    }
+  ); */
+  return useMemo(() => {
+    return { theme, applyTheme };
+  }, [theme]);
```
kark, line 105: @CarlosCortizasCT I had to revert to using the `applyTheme` where user needs to provide the `parentSelector` again. Unfortunately, with this implementation `parentSelector` function is stale, perhaps you could have a look into it if you find time (as I am not sure how to fix that). That can be now tested with rtl tests and in visual testing app -> `/theme-provider/interactive`

emmenko: The parent selector should be a function, not a ref, right? 🤔

CarlosCortizasCT: In this case we were using a ref to have a static reference to a function which shouldn't change. It is not a DOM reference.

I still don't get the use case of providing a different `parentSelector` for an already rendered hook, but I will take a deeper look and update accordingly.

tsconfig.6jyxozowiar.json
```diff
@@ -0,0 +1,38 @@
+{
```
CarlosCortizasCT, line 1: What is this file needed for?

emmenko: It's an artifact of `tsc-files` when running through `lint-staged`. Can be removed.

.changeset/few-maps-smell.md
```diff
@@ -0,0 +1,74 @@
+---
+'@commercetools-uikit/design-system': minor
```
kark, line 2: I marked only changes in 'visual-testing-app' and '@commercetools-uikit/design-system' as minor, perhaps all changes should be marked as minor?

design-system/src/theme-provider.tsx
CarlosCortizasCT: So I guess commenting previous suggested implementation it's due to tests using the API in a different way of what that implementation was expecting.

With the current implementation I see a "confusing" API in the hook.
I'll try to describe the current API
```
declare type TApplyTheme = {
    newTheme?: string;
    parentSelector?: () => HTMLElement | null;
    themeOverrides?: Record<string, string>;
};

declare const useTheme: (parentSelector?: () => HTMLElement | null) => {
    theme: string;
    applyTheme: ({ newTheme, parentSelector, themeOverrides, }: TApplyTheme) => void;
};
```
As you can see, when we call the hook in our component we have the chance to provide the `parentSelector` but we also have that chance when calling the `applyTheme` function of its result. So, if we use the `parentSelector` parameter in the `applyTheme`, why do we need the param in the hook itself for?

Since the hook and the provider are not connected by a React context, we need some way of linking them per instance. The provider will set the theme styles in a DOM element the hook will need to know about when changing.

The initial idea is that a hook will be linked to a specific provider by providing the same `parentSelector` the user used when rendering the provider. If we initialize the hook with that, there's no need to provide that callback in every `applyTheme` function call:
```
const getThemeRootElement = () => document.getElementById('theme-root');

const { theme, applyTheme } = useTheme(getThemeRootElement);

<button onClick={() => applyTheme({ newTheme: 'dark' })} />
```
In this case, the API will look like this:
```
declare type TApplyTheme = {
    newTheme?: string;
    parentSelector?: () => HTMLElement | null;
    themeOverrides?: Record<string, string>;
};

declare const useTheme: (parentSelector?: () => HTMLElement | null) => {
    theme: string;
    applyTheme: ({ newTheme, themeOverrides, }: TApplyTheme) => void;
};
```


On the other hand, we might want to go for another alternative and expect the `applyTheme` function called from users to always require the `parentSelector`. In that case, the `parentSelector` at hook initialization won't make sense:
```
const getThemeRootElement = () => document.getElementById('theme-root');

const { theme, applyTheme } = useTheme();

<button onClick={() => applyTheme({ newTheme: 'dark', parentSelector: getThemeRootElement })} />
```

In this case, the API will look like this:
```
declare type TApplyTheme = {
    newTheme?: string;
    parentSelector?: () => HTMLElement | null;
    themeOverrides?: Record<string, string>;
};

declare const useTheme: () => {
    theme: string;
    applyTheme: ({ newTheme, parentSelector, themeOverrides, }: TApplyTheme) => void;
};
```

Perhaps we can have a quick sync to explain this further.

kark: Thanks for your very detailed reply.
Having read it through, it seems to be the latter API makes more sense to me (which is closer to what we have had so far).
This would allow to call the generic `useTheme` hook e.g. only once, and use the exposed `applyTheme` function many times with various `parentSelector`s.

For me personally that makes more sense than the below:
```ts
// ... one useTheme hook call per theme
const { applyTheme: applyGlobalTheme } = useTheme(globalParentSelector);
const { applyTheme: applyLocalTheme } = useTheme(localParentSelector);
```

CarlosCortizasCT: > This would allow to call the generic useTheme hook e.g. only once, and use the exposed applyTheme function many times with various parentSelectors.

Bear in mind it's just the opposite. Every call to `useTheme` would bind to a specific `parentSelector` which should be the same provided to the `ThemeProvider` component you want this hook instance to control.
It's just what happens in your example.

You want both properties returned by the hook to always be bound to the same `parentSelector`.
Let's consider the previous test implementation and reading also the current theme:
```
const TestComponentWithThemeProvider = () => {
  const { theme, applyTheme } = useTheme();
  return (
    <>
      <button
        onClick={() => {
          applyTheme({
          applyGlobalTheme({
            themeOverrides: {
              colorSolid: 'red',
              colorSurface: 'yellow',
            },
            newTheme: 'dark',
            parentSelector: globalParentSelector,
          });
        }}
      >
        change global theme
      </button>
      <button
        onClick={() => {
          applyTheme({
          applyLocalTheme({
            themeOverrides: {
              colorSolid: 'green',
              colorSurface: 'tomato',
              customColor: '#BADA55',
            },
            newTheme: 'dark',
            parentSelector: localParentSelector,
          });
        }}
      >
...
```
In this scenario, we have only one hook instance so the same `theme` state is shared between changing local and global scoped themes, but in reality, we have two themes in the document.
That's why I think we need to have one hook instance per theme you want to control.

Thanks for updating the tests to comply with this suggestion 🙇

PR #2258, opened 2022-08-10
[PoC] Spec theme switcher

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=IE4eaDN)
#### Summary

This is just a proof of concept to test the problems we are facing with theming support and Percy testing.


design-system/src/theme-provider.tsx
```diff
@@ -17,7 +17,15 @@ type ThemeName = keyof typeof themes;
...
 
-const ThemeContext = createContext({});
+type TThemeContext = {
+  theme: ThemeName;
+  changeTheme: (newTheme: string) => void;
+};
+
+const ThemeContext = createContext<TThemeContext>({
+  theme: themesNames.default,
+  changeTheme: () => {},
```
kark, line 27: Thanks!

design-system/src/theme-provider.visualroute.jsx
```diff
@@ -1,34 +1,42 @@
...
-  const { changeTheme } = useTheme();
-  useLayoutEffect(() => {
-    changeTheme('dark');
-  }, [changeTheme]);
-
-  return (<DummyComponent />);
-};
+    }}
+  >
+    Title with default theme <i>colorPrimary</i> design token
```
kark, line 12: The only issue I see is that `default theme` is hardcoded here and rendered in all cases. If you don't feel like changing that here I can take care of that later.

kark (approved):
Work great 🙌

design-system/src/theme-provider.visualroute.jsx
CarlosCortizasCT: Thanks for pointing that out 👍 

Updated in this commit: 9707ed017093bcaf474fe321006de4d3139725b7

kark (approved):

PR #2262, opened 2022-08-24
Acccessibility button: display role attribute when SecondaryButton component is disabled

ddouglasz:
SecondaryButton components rendered as` a` tags lose their role attribute when disabled. We should still see the role attribute even if the component is disabled.

### DIFF
--- a/.changeset/curvy-worms-invite.md
+++ b/.changeset/curvy-worms-invite.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/accessible-button': patch
 ---
 
-keep role attribute of the accessibility button even when disabled
+fix a bug with accessibility attributes when using the `as` property
--- a/.changeset/curvy-worms-invite.md
+++ b/.changeset/curvy-worms-invite.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/accessible-button': patch
+---
+
+keep role attribute of the accessibility button even when disabled
--- a/packages/components/buttons/accessible-button/src/accessible-button.tsx
+++ b/packages/components/buttons/accessible-button/src/accessible-button.tsx
@@ -143,14 +143,10 @@ const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
         type: props.type,
       };
     } else {
-      if (!props.isDisabled) {
-        buttonProps = {
-          onKeyPress: handleKeyPress,
-        };
-      }
       buttonProps = {
         role: 'button',
         tabIndex: '0',
+        onKeyPress: handleKeyPress,
       };
     }
 
--- a/packages/components/buttons/accessible-button/src/accessible-button.tsx
+++ b/packages/components/buttons/accessible-button/src/accessible-button.tsx
@@ -142,7 +142,7 @@ const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
       buttonProps = {
         type: props.type,
       };
-    } else if (!props.isDisabled) {
+    } else {
       buttonProps = {
         role: 'button',
         tabIndex: '0',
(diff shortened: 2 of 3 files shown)


CarlosCortizasCT (commented):
Perhaps we could extend the component's [tests](https://github.com/commercetools/ui-kit/blob/main/packages/components/buttons/accessible-button/src/accessible-button.spec.js#L94) to include checking for these accessible attributes are still present when disabled.

packages/components/buttons/accessible-button/src/accessible-button.tsx
```diff
@@ -142,7 +142,7 @@ const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
       buttonProps = {
         type: props.type,
       };
-    } else if (!props.isDisabled) {
```
emmenko, line 145: I think the `onKeyPress` should still not be included when the button is disabled.

emmenko (commented):
Thanks for looking into this. Please also try to adjust the tests, as Carlos suggested.

packages/components/buttons/accessible-button/src/accessible-button.tsx
CarlosCortizasCT: Perhaps we can keep it simple since [the callback](https://github.com/commercetools/ui-kit/blob/main/packages/components/buttons/accessible-button/src/accessible-button.tsx#L128) already take into account the `disabled` state.

ddouglasz: Yes, I was going to state that but also, it 's a tiny change. so I can just avoid getting to the function as well. 
Updated here: https://github.com/commercetools/ui-kit/pull/2262/commits/79d797993da297e414bdce4149da03670f318093
Thank you

CarlosCortizasCT: Sure. I prefer the other way since it seems cleaner and more maintainable to me, but no big deal.

packages/components/buttons/accessible-button/src/accessible-button.tsx
```diff
@@ -142,11 +142,15 @@ const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
...
+      if (!props.isDisabled) {
+        buttonProps = {
+          onKeyPress: handleKeyPress,
+        };
+      }
       buttonProps = {
         role: 'button',
         tabIndex: '0',
-        onKeyPress: handleKeyPress,
       };
```
CarlosCortizasCT, line 154: Now it seems the element will never have the `onKeyPress` prop since we're overwriting the `buttonProps` variable after the `if` statement, wiping out that property.

ddouglasz: I let the `onkeyPress` callback check for the disabled property instead.
Updated here: https://github.com/commercetools/ui-kit/pull/2262/commits/96d4627e36fa2cd0363e666a8e6abfbe6be9473b

.changeset/curvy-worms-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/accessible-button': patch
+---
+
+keep role attribute of the accessibility button even when disabled
```
CarlosCortizasCT, line 5: Perhaps we want to be more generic here.
I'm suggesting this since I got a similar comment from @emmenko in a past PR.

```suggestion
fix a bug with accessibility attributes when using the `as` property
```

CarlosCortizasCT (approved):
💯

packages/components/buttons/accessible-button/src/accessible-button.tsx
emmenko: Alright, fine by me. Thanks for double checking.

.changeset/curvy-worms-invite.md
ddouglasz: Thank you.
 updated here: https://github.com/commercetools/ui-kit/pull/2262/commits/8efe0f98487925bae033392e5de1d8476b01c8c0

emmenko (approved):
💯

PR #2264, opened 2022-08-25
Poc remove emotion theming - alternative proposal

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=5zy9iP_)

#### Summary

Just a temporary PR to discuss about an alternative proposal.

### DIFF
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -1,33 +1,17 @@
-import { useLayoutEffect, useMemo, useState, useCallback } from 'react';
+import { useLayoutEffect, useMemo, useState, useRef } from 'react';
 import kebabCase from 'lodash/kebabCase';
 import isObject from 'lodash/isObject';
 import merge from 'lodash/merge';
 import { themes, themesNames } from './custom-properties';
 
-type ThemeName = keyof typeof themes;
-
 const allThemesNames = Object.keys(themesNames);
+
+type ThemeName = keyof typeof themes;
 
 const toVars = (obj: Record<string, string>) =>
   Object.fromEntries(
     Object.entries(obj).map(([key, value]) => [`--${kebabCase(key)}`, value])
   );
-
-const validateTheme = (themeName?: string): ThemeName => {
-  if (!themeName) {
-    return 'default';
-  }
-  const isNewThemeValid = allThemesNames.includes(themeName);
-  if (isNewThemeValid) {
-    return themeName as ThemeName;
-  }
-  if (!isNewThemeValid) {
-    console.warn(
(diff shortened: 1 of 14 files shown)


design-system/src/theme-provider.tsx
```diff
@@ -47,38 +31,46 @@ const changeTheme = ({
...
+  const parentSelectorRef = useRef(props.parentSelector);
   useLayoutEffect(() => {
     changeTheme({
       newTheme: props.theme,
-      parentSelector,
+      parentSelector: parentSelectorRef.current,
       themeOverrides: props.themeOverrides,
     });
-  }, [props.theme, props.themeOverrides, parentSelector]);
+  }, [props.theme, props.themeOverrides]);
```
kark, line 73: Thanks, this is a great improvement 🙌 

As discussed, perhaps we would need to also use [the latest ref pattern](https://epicreact.dev/the-latest-ref-pattern-in-react/) for certain scenarios (one I can think of is having a select for choosing a theme and a separate button for confirming the choice). I'm just leaving this for reference if we need to get back to this in the future.

I guess when we update to React v18 and [`useEvent`](https://github.com/facebook/react/pull/24505) hook is released, that would be a candidate for refactor.

kark (approved):
Thank you!

design-system/src/theme-provider.tsx
```diff
@@ -88,19 +80,26 @@ ThemeProvider.defaultProps = {
 };
 
 const useTheme = (parentSelector = defaultParentSelector) => {
-  const [theme, setTheme] = useState<string | null | undefined>(null);
-  const memoizedParentSelector = useCallback(
-    () => parentSelector(),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
+  const [theme, setTheme] = useState<string>(
+    parentSelector()?.dataset.theme || 'default'
```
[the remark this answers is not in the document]
CarlosCortizasCT: Absolutely right. Thanks!

Updated in this commit: d2a061d15ef2ebe832a17fac80e25c397cdf241d

design-system/src/theme-provider.tsx
CarlosCortizasCT: I agree this is something we might check later when we support theming change.

On the other hand, I'm sorry I don't think I understand the scenario you mention.
I think the problem with no having the pattern you mention is expecting the users to change the value of `parentSelector` provided to the hook in different renders of their components.

Maybe we can discuss about it in a quick meeting if you want to.

kark: Yes, you're right about my scenario

kark (approved):

PR #2268, opened 2022-08-31
Simplify alternative themes configuration

CarlosCortizasCT:
#### Summary

Make it easier to configure a new theme.

## Description

Currently, if anyone wants to add a new theme, all the tokens would need to be configured in the `definition.yaml` file.
With this PR, we will always use the `default` theme as a blueprint so a new theme would only require to configure the tokens that need to be changed.


design-system/materials/internals/definition.yaml
```diff
@@ -142,6 +142,14 @@ choiceGroupsPerTheme:
         break-point-jumbodesktop: 1920px
 
   dark:
+    colors:
```
CarlosCortizasCT, line 145: Temporary definition until we have a proper new theme configuration values set.

design-system/src/custom-properties.ts
```diff
@@ -129,128 +129,9 @@ export const themes = {
     standardInputHeight: '32px',
   },
   dark: {
-    colorPrimary: '#00b39e',
```
CarlosCortizasCT, line 132: Now alternate themes will only contain the properties they need to overwrite from the `default` theme.

docs/.storybook/configs/theme-context.js
```diff
@@ -4,33 +4,22 @@ import {
...
-
-  return null;
-};
-
 const ThemeWrapper = (props) => {
   return (
     <>
-      <ThemeProvider theme={defaultThemeName} />
-      <ThemeToggler {...props} />
+      <ThemeProvider
```
CarlosCortizasCT, line 19: Unrelated, but I found we don't need the toggler here so it's better for our future selves maintaining this.

design-system/src/theme-provider.tsx
```diff
@@ -46,7 +46,7 @@ const applyTheme = ({
 
   const vars = toVars(
     themeOverrides && isObject(themeOverrides)
-      ? merge({}, themes[validTheme], themeOverrides)
+      ? merge({}, themes.default, themes[validTheme], themeOverrides)
```
emmenko, line 49: Nice 💯

emmenko (approved):
Very nice, thanks for simplifying it! 🙌

PR #2269, opened 2022-08-31
Fix theming support Percy failing test

CarlosCortizasCT:
## Summary

`ThemeProvider` vrt interactive tests don't work as expected.

### Description

We expect the snapshot taken after changing theme will show the new theme applied but it's not working like that. No visual tests are visible after changing the theme.


CarlosCortizasCT:
It seems Percy does not support what we want to achieve.

PR #2270, opened 2022-09-01
Remove unnecessary helper from theming React hook

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=QALXUDg)
### Summary

Remove `applyTheme` from `useTheme` hook since we don't see the need for it.

## Description

We expect users to handle the current theme on their side and provide it to the `ThemeProvider` as a prop.
When they'll need to change it, they would just provide a new value in the `theme` prop of the provider.

`useTheme` hook will only return a string with the current provider theme.

### DIFF
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -88,21 +88,7 @@ const useTheme = (parentSelector = defaultParentSelector) => {
     setTheme(parentSelectorRef.current()?.dataset.theme || 'default');
   }, []);
 
-  // So consumers don't have to provide 'parentSelector' again as
-  // they already provided it in the hook call
-  const updateTheme = useRef(
-    ({ newTheme, themeOverrides }: Omit<TApplyTheme, 'parentSelector'>) => {
-      applyTheme({
-        newTheme,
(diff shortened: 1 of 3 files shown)


design-system/src/theme-provider.tsx
```diff
@@ -88,21 +88,7 @@ const useTheme = (parentSelector = defaultParentSelector) => {
...
-        parentSelector: parentSelectorRef.current,
-        themeOverrides,
-      });
-      setTheme(newTheme || 'default');
-    }
-  );
-  return useMemo(() => {
-    return { theme, applyTheme: updateTheme.current };
-  }, [theme]);
+  return useMemo(() => theme, [theme]);
```
emmenko, line 91: It seems that `useMemo` is not useful anymore now.

```suggestion
  return theme;
```

emmenko (approved):

design-system/src/theme-provider.tsx
CarlosCortizasCT: Thanks 👍 

It's been updated in this commit: e57b8f14cba6f13161def8d9b9801c42c122530e

PR #2271, opened 2022-09-01
Support themed decision groups desing tokens

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=hfsqhHi)
### Summary

Allow to configure decisions per theme.

## Description

Currently we support to configure choice tokens per theme so, for example, `color-primary` token can have different values based on the theme.

With this PR we also allow configuring decision tokens as, for example, `background-color-for-tag`.

After implementing the first iteration I found an unexpected behaviour in some components which were not using the tokens properly. Instead of using the decision token, they were configured to use the choice token _behind_ the decision token.
All of those components where using values from the `design-tokens.ts` file to access tokens from the `customProperties.ts` one, when the latter already contain all the tokens with the right references ([example](https://github.com/commercetools/ui-kit/blob/main/packages/components/tag/src/tag.tsx#L98)).
I refactored all those components to use only the tokens from the `customProperties.ts` file so they reference the right values that can be dynamically updated in runtime by changing the theme.

I took a look both at `app-kit` and frontend to check wether this file (`design-tokens.ts`) is being used over there and didn't find any occurrence.
Since the file contents are available through the `@commercetools-uikit/design-system` public package, I suggest to deprecate it and no using it anymore.
(If we'd want to keep it, we would need also to refactor it to contain themed tokens, similar to what we did with `custom-properties.ts` file)


CarlosCortizasCT:
After a sync with @emmenko we decided that it's not worthy to have both `custom-properties.ts` and `design-tokens.ts` files containing both tokens (at the beginning they were intended to group different tokens by semantic level).
We decided to just keep `design-tokens.ts` as the source of truth of all tokens and remove the `custom-properties.ts`.
Since the latter is exported from the `@commercetools-uikit/design-system` we decided to keep that export but it is an alias of the `desing-tokens.ts` file (which has all the tokens) and mark it as deprecated.

As a follow-up, we will update the references to that export in all components: `ui-kit`, `app-kit` and `mc-frontend`.

design-system/src/index.ts
```diff
@@ -1,9 +1,10 @@
 export {
+  default as designTokens,
+  /** @deprecated use `designTokens` instead */
   default as customProperties,
```
emmenko, line 4: Nice trick! 👌

emmenko (approved):
Cool, thanks! 💯

PR #2272, opened 2022-09-02
Improve `ThemeProvider` renders

CarlosCortizasCT:
#### Summary

Improve `ThemeProvider` to avoid unnecessary renders.

## Description

With our current implementation, a user providing an inline object for the `themeOverrides` prop of the `ThemeProvider` component, would make the provider to apply a theme change on every of its renders:
```
const MyApp = () => 
  (
    <ThemeProvider themeOverrides={{ colorSolid: 'red' }}>
      ...
    </ThemeProvider>
  );
```

In this PR we update the provider to handle such situations and only apply the theme change if the provided `themeOverrides` object is deeply different from the previous one.


design-system/src/theme-provider.tsx
```diff
@@ -64,12 +65,26 @@ type ThemeProviderProps = {
...
+      themeRef.current = props.theme;
+      themeOverridesRef.current = props.themeOverrides;
+
+      applyTheme({
+        newTheme: props.theme,
+        parentSelector: parentSelectorRef.current,
+        themeOverrides: props.themeOverrides,
+      });
+    }
   }, [props.theme, props.themeOverrides]);
```
emmenko, line 88: Thanks for looking into that!

I was wondering if there is also a way of avoiding passing these values as "dependencies" in the first place 🤔

CarlosCortizasCT: Do you mean removing the effect?

I guess the way it is right now helps us not triggering the effect when users don't provide a `themeOverrides` prop or they don't provide an inline object. If they do, then we have the new check in place.

If we remove the effect, the actual result will be the same, but the check would be run on every render.
Also, we need to apply the theme in the first render.

Maybe if you have some idea in mind you can share it and I can explore it.

emmenko: Sure, I was just thinking about computing (merging) the theme properties outside of the effect and then pass the result into the effect. So the effect is only responsible to apply the value to the DOM but the computation can be done outside of it.

What do you think?

CarlosCortizasCT: I'm sorry but I think I'm not following you. If we compute the theme properties (do you mean creating the `var` that will be applied to the parent selector?) outside the effect, we will be computing those props on every render, even when the original props did not change and we will not need to apply the change.

Perhaps we can arrange a quick call to go over this topic.

emmenko: No problem, I'm just "thinking out loud" here. Sorry for not being too clear.

I was referring more to the fact that the effect should only apply the theme but the computed theme can be calculated before (for example using `useMemo` or something). Still, to ensure referencial equality of the `themeOverrides` (being an object) we need to keep the ref around, as you have already implemented.

Anyway, this is perfectly fine (and better than before) so let's go with this 🤗

CarlosCortizasCT: So, if this is fine, could I have an approval, please 🙏 ?

emmenko (approved):
Sorry, forgot to approve it last time 🙃

PR #2273, opened 2022-09-02
Update css tokens imports

CarlosCortizasCT:
#### Summary

Update design tokens imports to use the new exported object from the `@commercetools-uikit/design-token` package.

## Description

This is a follow-up PR for #2271.

We change imports to use the new exported property from mentioned package.

```js
// Before

// After
```


CarlosCortizasCT:
Invalid destination target branch.

PR #2274, opened 2022-09-02
Update css tokens imports

CarlosCortizasCT:
#### Summary

Update design tokens imports to use the new exported object from the `@commercetools-uikit/design-token` package.

## Description

This is a follow-up PR for #2271.

We change imports to use the new exported property from mentioned package.

```js
// Before

// After
```

### DIFF
--- a/packages/components/buttons/flat-button/src/flat-button.tsx
+++ b/packages/components/buttons/flat-button/src/flat-button.tsx
@@ -8,7 +8,7 @@ import {
 } from 'react';
 import { css } from '@emotion/react';
 import omit from 'lodash/omit';
-import { customProperties } from '@commercetools-uikit/design-system';
+import { designTokens } from '@commercetools-uikit/design-system';
 import { filterInvalidAttributes } from '@commercetools-uikit/utils';
 import Text from '@commercetools-uikit/text';
 import AccessibleButton from '@commercetools-uikit/accessible-button';
@@ -16,7 +16,7 @@ import { getTextColor, getButtonIconColor } from './flat-button.styles';
 
 const propsToOmit = ['type'];
 
-export type CustomProperties = typeof customProperties;
+export type TDesignTokens = typeof designTokens;
 
 export type TFlatButtonProps<
   TStringOrComponent extends ElementType = 'button'
@@ -132,19 +132,19 @@ const FlatButton = <TStringOrComponent extends ElementType = 'button'>(
 
         span {
           color: ${props.isDisabled
-            ? customProperties.colorNeutral
-            : getTextColor(props.tone, false, customProperties)};
+            ? designTokens.colorNeutral
+            : getTextColor(props.tone, false, designTokens)};
         }
 
         svg * {
           fill: ${props.isDisabled
-            ? customProperties.colorNeutral
-            : getTextColor(props.tone, false, customProperties)};
+            ? designTokens.colorNeutral
+            : getTextColor(props.tone, false, designTokens)};
         }
 
         * + span,
         * + svg {
-          margin-left: ${customProperties.spacingXs};
+          margin-left: ${designTokens.spacingXs};
         }
 
         ${!props.isDisabled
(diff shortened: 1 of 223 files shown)


packages/components/tooltip/src/tooltip.styles.ts
```diff
@@ -44,9 +46,7 @@ export const getBodyStyles = ({
   return {
     fontFamily: 'inherit',
     margin: `${getOffsetMargin({ placement })} !important`,
-    maxWidth: (customProperties as Record<string, string>)[
-      `constraint${constraint}`
-    ],
+    maxWidth: designTokens[`constraint${constraint}` as TDesignTokenName],
```
CarlosCortizasCT, line 49: I needed to cast since it's a dynamic value.

packages/components/constraints/src/helpers.ts
```diff
@@ -1,10 +1,11 @@
...
+import { designTokens } from '@commercetools-uikit/design-system';
+
+type TDesignTokenName = keyof typeof designTokens;
 
 // `null` is derived from `getMaxPropEquivalent`
 const getMaxPropTokenValue = (max: number | string | null) => {
   if (!max) return null;
-  // @ts-expect-error
-  return customProperties[`constraint${max}`];
+  return designTokens[`constraint${max}` as TDesignTokenName];
```
CarlosCortizasCT, line 8: I needed to cast since it's a dynamic value

packages/components/buttons/flat-button/src/flat-button.tsx
```diff
@@ -8,15 +8,15 @@ import {
...
+import { designTokens } from '@commercetools-uikit/design-system';
 import { filterInvalidAttributes } from '@commercetools-uikit/utils';
 import Text from '@commercetools-uikit/text';
 import AccessibleButton from '@commercetools-uikit/accessible-button';
 import { getTextColor, getButtonIconColor } from './flat-button.styles';
 
 const propsToOmit = ['type'];
 
-export type CustomProperties = typeof customProperties;
+export type TDesignTokens = typeof designTokens;
```
CarlosCortizasCT, line 19: I renamed this type for consistency.
Although it is exported, it is only used in the `flat-button.styles.ts` file and no exported from the package, so it's safe.

packages/components/constraints/src/helpers.ts
emmenko: Nit: if we want to be more "type safe" in these cases (accessing a property dynamically), we can extract this logic into a function with a switch statement.

Similar examples:
* https://github.com/commercetools/ui-kit/blob/e3b726cc7684e6a4da0fdd251b6b0ee8ffda7815/packages/components/spacings/spacings-stack/src/stack.styles.ts#L17-L32
* https://github.com/commercetools/merchant-center-application-kit/blob/a4e7a3452d8a4bd0cf3f7cef65098c8de2ba0740/packages/application-components/src/components/dialogs/internals/dialog.styles.ts#L9-L26

This way we ensure we pick the correct value and we can decide what to do for unknown values.

Not a must, maybe something to think about.

emmenko (approved):
Love it, thanks! 🤗

kark (approved):
💯

CarlosCortizasCT:
@emmenko regarding the type safety improvement, what do you think about having something like this? (following with the constraints helper example):
```js
const getMaxPropTokenValue = (max: number | string | null) => {
  if (!max) return null;
  const tokenName = `constraint${max}`;
  if (tokenName in designTokens) {
    return designTokens[tokenName as TDesignTokenName];  
  }
  return null;
};
```

We would be validating the dynamic token name exists in the `designTokens` object, we won't need to explicitly write all the different _switch_ cases.

This is just a proposal. If you prefer to be explicit and list all the switch cases, I will use that pattern to update this PR.

emmenko:
Sounds good, thanks for looking into it! 💯

CarlosCortizasCT:
> @emmenko regarding the type safety improvement, what do you think about having something like this? (following with the constraints helper example):
> 
> ```js
> const getMaxPropTokenValue = (max: number | string | null) => {
>   if (!max) return null;
>   const tokenName = `constraint${max}`;
>   if (tokenName in designTokens) {
>     return designTokens[tokenName as TDesignTokenName];  
>   }
>   return null;
> };
> ```
> 
> We would be validating the dynamic token name exists in the `designTokens` object, we won't need to explicitly write all the different _switch_ cases.
> 
> This is just a proposal. If you prefer to be explicit and list all the switch cases, I will use that pattern to update this PR.

@emmenko This is the commit (cf14e91e874b38f19b9c62a12081bcfd00126595) where I applied suggested change.

emmenko (approved):
👌

PR #2278, opened 2022-09-07
fix: add missing `label` property to `options` prop type of select inputs and select fields

### LINKED ISSUE #2277: Typings: `TOption` is missing label property
**Describe the bug**
TS typings for `SelectField` and `SelectInput` are missing `label` property in `TOption`

Current: 
```
declare type TOption = {
    value: string;
};
```

I believe it should also include `label`.

image

image

kark:
Fixes #2277


emmenko (approved):
Thanks!

CarlosCortizasCT (approved):
👌

PR #2279, opened 2022-09-08
Fix `label` property type in `options prop` type of select inputs and select fields

CarlosCortizasCT:
#### Summary

Fixes the type of options `label` property to allow more types than a string.

## Description

When providing the value for the `label` property of the options, users might use not only strings but other types as well (ex: an element for a rich label).
We widen the type of that property to support different use cases.


emmenko (approved):
Good catch, thanks! 💯

kark (approved):
Thanks!

Rhotimee (approved):
🙌🏽

PR #2284, opened 2022-09-13
Fix info button arrangement in the data table component

### LINKED ISSUE #2283: Adjust info button in the data table component
The optional info button which can be set in the table header cell is currently 24px big and is placed left to the label. This causes a bad looking alignment with the rest of the table cells and due to the size, the padding of the header ell seems to be ignored

- Set the size of the info button to **16px**
- Place the info button on the **right side** of the label
- The margin between the label and the info button should be **8px.**
image

ddouglasz:
This PR fixes the issue opened regarding data-table info button arrangement. 
image

Fixes #2283

### DIFF
--- a/.changeset/slimy-walls-worry.md
+++ b/.changeset/slimy-walls-worry.md
@@ -1,0 +1,7 @@
+---
+'@commercetools-uikit/data-table': minor
+---
+
+Fix arrangement of data-table info button.
+
+The optional info button which can be set in the table header cell is placed left to the label. This causes a bad looking alignment with the rest of the table cells and due to the size, the padding of the header cell seems to be ignored.
--- a/.changeset/slimy-walls-worry.md
+++ b/.changeset/slimy-walls-worry.md
@@ -1,7 +1,5 @@
 ---
-'@commercetools-uikit/data-table': minor
+'@commercetools-uikit/data-table': patch
 ---
 
-Fix arrangement of data-table info button.
-
-The optional info button which can be set in the table header cell is placed left to the label. This causes a bad looking alignment with the rest of the table cells and due to the size, the padding of the header cell seems to be ignored.
+Fix position of header cell icon to the right of the text.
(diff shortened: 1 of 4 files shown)


.changeset/slimy-walls-worry.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/data-table': minor
```
emmenko, line 2: Please be careful when to use which version types. Minor is mostly used for new features or some substantial changes. For bug fixes, minor changes, normal dep updates, etc. we should use `patch`.

```suggestion
'@commercetools-uikit/data-table': patch
```

.changeset/slimy-walls-worry.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Fix arrangement of data-table info button.
+
+The optional info button which can be set in the table header cell is placed left to the label. This causes a bad looking alignment with the rest of the table cells and due to the size, the padding of the header cell seems to be ignored.
```
emmenko, line 7:
```suggestion
Fix position of header cell icon to the right of the text.
```

emmenko (commented):
Thanks for the fix! I left a couple of comments about the changeset otherwise 👌

.changeset/slimy-walls-worry.md
ddouglasz: > FYI: if a PR closes or fixes an issue, we should use the syntax `Closes` or `Fixes`, which then instructs GitHub to automatically close the issue.
> 
> image

Thank you. I get it now. Updated here:

https://github.com/commercetools/ui-kit/pull/2284/commits/13f01a6db493060ec99c82a2e33e8f846aa07727

emmenko (approved):
Thanks! 👌

Rhotimee (approved):

kark (approved):

PR #2288, opened 2022-09-16
Fix icon displaced to far right when data table cell is expanded

ddouglasz:
Fix icon displaced to far right when data table cell is expanded
Before: 
image

After:
image (1)

### DIFF
--- a/packages/components/data-table/src/header-cell.styles.tsx
+++ b/packages/components/data-table/src/header-cell.styles.tsx
@@ -1,7 +1,7 @@
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
-import { designTokens } from '@commercetools-uikit/design-system';
 import { getCellInnerStyles } from './cell.styles';
+import { designTokens } from '@commercetools-uikit/design-system';
 import type { THeaderCell } from './header-cell';
 
 const getButtonStyle = () => css`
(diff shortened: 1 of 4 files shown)


emmenko (approved):
Thanks! 👌

emmenko (changes requested):
I checked the VRTs. The column alignments (left, right) are broken now.

image

ddouglasz:
Oh, will take a look now. Thank you.

packages/components/data-table/src/header-cell.styles.tsx
```diff
@@ -123,8 +123,8 @@ const HeaderIconWrapper = styled.div`
   display: inline-flex;
   align-items: center;
   justify-content: center;
-  margin-left: ${vars.spacingS};
-  padding: 0 ${vars.spacingXs};
+  margin: 0 ${vars.spacingS};
```
emmenko, line 126: I think the margin should be left only. When the cell content is right-aligned it's visible that the space on the right is bigger than normal.

image

emmenko (commented):
Thanks, just one minor thing about the `margin`.

packages/components/data-table/src/header-cell.styles.tsx
ddouglasz: Thank you. 
Updated here: https://github.com/commercetools/ui-kit/pull/2288/commits/5d18b00a597a722dbaf8b1b4eea7dd20f1ae9ba8

Rhotimee (approved):

emmenko (approved):
Looks good now, thanks!

PR #2293, opened 2022-09-19
fix(view-switcher): Use white as button background color

Rhotimee:
#### Summary
Change ViewSwitcher background colour from transparent to white.


emmenko (approved):
💯

tdeekens (approved):

Issue #2261, opened 2022-08-23
Unpredictable behavior DateInput when input by keyboard


planetcrypton:
**Describe the bug**
When entering dates by keyboard (in stead of clicking the calendar UI), I expect to be able to enter any valid formatted date (and have that selected on the calendar UI) and hitting enter will change the date accordingly.
The example at https://uikit.commercetools.com/?path=/story/components-inputs--dateinput works perfectly, but my version of the component (v15.2.1) doesn't.

**To Reproduce & Expected behavior & Screenshots****
Compare the example in the documentation:

https://user-images.githubusercontent.com/1587059/186252866-d0ff9f59-bfe8-482e-b567-f836ae879b35.mov

with my example:

https://user-images.githubusercontent.com/1587059/186253000-b40bba15-fdff-4ce0-9bae-3c51c2184c43.mov

As you can see it only works as expected for months earlier than October 🤔; For month 10+11+12 the date never gets selected in the UI and thus hitting enter won't change the date. Whereas months < 10 work fine

**Additional context**
The app does nothing but displaying this `DateTest` component on this particular route. Code for that is in the screencast video

ddouglasz:
Hello @planetcrypton ,
Thank you for reporting this. We are currently investigating the cause.

PR #2304, opened 2022-10-06
Date inputs with months greater or equal to the 10th month do not highlight and submit.

ddouglasz:
fix issue where date inputs with months greater or equal to the 10th month do not highlight and submit.

**How can I test this?**
Reproducing the issue is a bit tricky. The fastest way to reproduce the issue is to comment out the `minValue` and `maxValue` props of the` <DateInput>` component in `packages/components/inputs/date-input/src/date-input.story.js`.
Go to date-input on your storybook, check to see that values with, for example, 12/12/2022 or any date with month >=10 (Oct, Nov and Dec) do not submit. Checkout to this current work branch to check again that this PR fixes it. 

Fixes: #2261


kark (approved):
Thanks, it seems it solves the issue. I still don't get why the bug only appears for month value >= 10.

emmenko:
Wow, thanks for finding the issue. I have no idea how this relates to the bug, but if it works then let's go for it.

A couple of things maybe to avoid this in the future:
* leave a comment of why it needs to be `if-else`
* (if possible) write a test that ensures this behavior is fixed

Thanks

Rhotimee (approved):
Good catch 💪🏽

ddouglasz:
> Wow, thanks for finding the issue. I have no idea how this relates to the bug, but if it works then let's go for it.
> 
> A couple of things maybe to avoid this in the future:
> 
> * leave a comment of why it needs to be `if-else`
> * (if possible) write a test that ensures this behavior is fixed
> 
> Thanks

Updated here: https://github.com/commercetools/ui-kit/pull/2304/commits/e70016cf7d00dca93d35f7107b9b15084d51dbb7

Thank you.

emmenko (approved):
Thanks!

PR #2317, opened 2022-10-26
Typography tokens

chloe0592:
#### Summary

Adding typography tokens and introducing new specific token `Variant` (which maybe needs a different name?)

## Description

Tokens are defined in [Miro board file](https://miro.com/app/board/uXjVPSlrfYY=/)


design-system/materials/internals/definition.yaml
```diff
@@ -74,6 +74,29 @@ choiceGroupsByTheme:
       prefix: font-size
       choices:
         font-size-m: '1rem'
+        font-size-10: '0.75rem'
```
CarlosCortizasCT, line 77: I would also update the comment in the header of the file to give another example using variants.
Now:
```
# The names of decisions in this file need to follow this format:
#   <attribute>-for-<component-group>-when-<state>-
```

Proposal:
```
# The names of decisions in this file need to follow this format:
#   <attribute>-for-<component-group>-when-<state>
#   <attribute>-for-<component-group>-as-<state>
```

Also, I think it might be possible to use both combined (nothing we need right now):
```
<attribute>-for-<component-group>-as-<state>-when-<state>
font-size-for-text-as-h2-when-hovered
```

design-system/materials/internals/TOKENS.md
```diff
@@ -17,8 +17,7 @@ The tokens give semantic meaning to our plain palette of choices..
   - [Aspect: Component Group](#aspect-component-group)
   - [Aspect: State](#aspect-state)
     - [`active` vs `focused`](#active-vs-focused)
-  - [Aspect: Modifier](#aspect-modifier)
-    - [on-dark](#on-dark)
+  - [Aspect: Variant](#aspect-variant)
```
CarlosCortizasCT, line 20: It seems there are a couple of images referenced in this document which contain the `on` modifier.
Perhaps we would need to create new ones.

CarlosCortizasCT (commented):
Awesome! 👏 

I've just left a couple of minor suggestions.

kark (commented):
This looks great 🙌

design-system/materials/internals/definition.yaml
chloe0592: Thank you, totally makes sense! Updated  it 😊 
And I agree we might down the line include also the combination.

design-system/materials/internals/TOKENS.md
chloe0592: I haven’t even noticed that, thanks for pointing! Do you have any idea how this image was created? Maybe we could manually add some examples?

design-system/materials/internals/definition.yaml
```diff
@@ -74,6 +74,29 @@ choiceGroupsByTheme:
...
+        font-size-20: '0.875rem'
+        font-size-30: '1rem'
+        font-size-40: '1.125rem'
+        font-size-50: '1.25rem'
+        font-size-60: '1.5rem'
+    fontWeight:
+      label: Font weight
```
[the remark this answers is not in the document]
chloe0592: Eagle eyes! Thanks for noticing 🙌

design-system/materials/internals/definition.yaml
```diff
@@ -169,6 +192,23 @@ states:
...
+  h3:
+    description: 'H3 text element'
+  h4:
+    description: 'H4 text element'
+  h5:
+    description: 'H5 text element'
+  body:
+    description: 'Body text element'
+  detail:
+    description: 'Detail text element'
```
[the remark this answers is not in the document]
chloe0592: That's a great point, I think you are right. I will add new variants for the body - it'll be default, small and extra small. But what should we do with the detail component? We cannot remove it now, right? It'll be then a breaking change..

design-system/materials/internals/definition.yaml
```diff
@@ -74,6 +74,29 @@ choiceGroupsByTheme:
       prefix: font-size
       choices:
         font-size-m: '1rem'
```
emmenko, line 76: Maybe we can mark this as deprecated?

design-system/scripts/generate-design-tokens.js
```diff
@@ -22,10 +22,11 @@ const endProgram = (message) => {
 };
 
 const TOKEN_REGEX =
-  /^(\w+(?:-\w+)(?:-\w+)?)(?:-for-(\w+(?:-\w+)?))?(?:-when-([\w-]+?))?(?:-on-([\w-]+?))?$/i;
+  /^(\w+(?:-\w+)(?:-\w+)?)(?:-for-(\w+(?:-\w+)?))?(?:-when-([\w-]+?))?(?:-as-([\w-]+?))?$/i;
```
emmenko, line 25: I think `when` and `as` should be mutually exclusive, right? 🤔

design-system/scripts/generate-design-tokens.js
```diff
@@ -79,6 +80,7 @@ Object.entries(definitions.decisionGroupsByTheme).forEach(
         if (match) {
           const componentGroup = match[2];
           const state = match[3];
+          const variant = match[4];
```
emmenko, line 83: In theory this should be one.

```suggestion
          const stateOrVariant = match[3];
```

packages/components/text/src/text.styles.ts
```diff
@@ -56,15 +56,49 @@ const getTone = (tone: string) => {
 const getElementFontSize = (elementType?: string) => {
   switch (elementType) {
     case 'h1':
-      return '2.4615rem';
+      return `${designTokens.fontSizeForTextAsH1};`;
```
emmenko, line 59: Nit: you don't need string templates here.

```suggestion
      return designTokens.fontSizeForTextAsH1;
```

design-system/src/design-tokens.ts
```diff
@@ -273,6 +324,27 @@ const designTokens = {
   placeholderFontColorForInput:
     'var(--placeholder-font-color-for-input, hsl(0, 0%, 60%))',
   fontSizeForInput: 'var(--font-size-for-input, 1rem)',
+  fontSizeForTextAsH1: 'var(--font-size-for-text-as-h-1, 1.5rem)',
```
emmenko, line 327: I would expect the name to be `-h1` and not `-h-1`. Might want to look into the script.

design-system/materials/internals/definition.yaml
```diff
@@ -301,6 +341,79 @@ decisionGroupsByTheme:
         font-size-for-input:
           description: ''
           choice: font-size-m
+        font-size-for-text-as-h1:
+          description: ''
+          choice: font-size-60
```
emmenko, line 346: I suppose this should be the same as h2 by default?

emmenko (commented):
Exciting, thanks for working on this! 🙌

design-system/materials/internals/TOKENS.md
CarlosCortizasCT: Sorry but I don't know.
Maybe @emmenko remembers?

emmenko: No idea. We can also simply use a code block or something, I don't think we need an image necessarily.

design-system/materials/internals/definition.yaml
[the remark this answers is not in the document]
CarlosCortizasCT: I guess we could update the choice we have for the font-size of that component to use the new `body-xs` token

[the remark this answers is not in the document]
CarlosCortizasCT: This is an interesting topic. 🤔 

From my point of view, I think `body` is more a component group than a variant (the same as text, input or tag).

design-system/materials/internals/TOKENS.md
chloe0592: Okay, I'll create a block then.

design-system/materials/internals/definition.yaml
chloe0592: ✅

design-system/materials/internals/definition.yaml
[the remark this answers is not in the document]
chloe0592: @CarlosCortizasCT so the it wouldn’t be a part of the `text` component anymore? It would be the separate and individual component?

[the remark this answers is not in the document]
CarlosCortizasCT: After giving it a second thought, I believe it makes more sense to have `body` as a variant as well as sizes and allow to combine them in the name: `font-size-for-text-as-body-as-small`.

As we already discussed, support for that is in #2319 PR.
If we can add that as a follow-up, great.
Another option is to extract the changes in that PR related to the tokens generation script to a new one so we can all work upon it.

design-system/src/theme-provider.tsx
```diff
@@ -46,9 +46,12 @@ const applyTheme = ({
...
 
   const vars = toVars(
-    themeOverrides && isObject(themeOverrides)
-      ? merge({}, themes.default, themes[validTheme], themeOverrides)
-      : themes[validTheme]
+    merge(
+      {},
+      themes.default,
+      themes[validTheme],
+      isObject(themeOverrides) ? themeOverrides : {}
```
emmenko, line 53: PS: Let's merge the other PR first #2327

design-system/materials/internals/definition.yaml
```diff
@@ -22,7 +22,7 @@
 # files in different format (css, json, js, ...).
 #
 # The names of decisions in this file need to follow this format:
-#   <attribute>-for-<component-group>-when-<state>-
+#   <attribute>-for-<component-group>-as-<variant>-when-<state>
```
emmenko, line 25: Should we include the same examples we have in the generate script?

https://github.com/commercetools/ui-kit/blob/0a42178c2c733cf31120fa208db182399887677e/design-system/scripts/generate-design-tokens.js#L39-L51

packages/components/text/src/text.styles.ts
```diff
@@ -73,7 +107,9 @@ const getElementFontSize = (elementType?: string) => {
 export const bodyStyles = (props: TBodyProps) => css`
   ${getBaseStyles()}
   margin: 0;
-  font-size: 1rem;
+  font-size: ${designTokens.fontSizeForTextAsBody};
+  line-height: ${designTokens.lineHeightForTextAsBody};
+  font-weight: ${designTokens.fontWeightForTextAsBody};
   ${props.isBold && bold}
```
emmenko, line 113: Should we use a token also for `bold`?

Also, at moment bold is equivalent to `700`. Is it going to be the same for the new design? If so, it's missing in the token choices.

packages/components/link/src/link.tsx
```diff
@@ -38,6 +43,13 @@ type TLinkProps = {
    * Color of the link
    */
   tone?: 'primary' | 'inverted';
+
+  /**
+   * Handler when the link is clicked.
+   */
+  onClick?: (
```
emmenko, line 50: Wasn't this part of another PR? #2323

design-system/materials/internals/definition.yaml
```diff
@@ -323,6 +445,87 @@ decisionGroupsByTheme:
...
+        font-weight-for-text-as-h5:
+          description: ''
+          choice: font-weight-500
+        font-weight-for-text-as-body:
+          description: ''
+          choice: font-weight-400
+        font-weight-for-text-as-detail:
+          description: ''
+          choice: font-weight-400
+
```
emmenko, line 528: I think the case for "bold" text is missing, isn't it?

emmenko (commented):
Super nice, thanks for working on this!

I left a couple of questions but otherwise looks good to me!

design-system/materials/internals/definition.yaml
CarlosCortizasCT: In the board we're using as reference we didn't see any reference to `bold` weight so we assumed we will still using `700` in the new theme so no token would be needed for that.

Perhaps @FilPob might confirm this point.

emmenko: I think the board is missing the bold styles.

design-system/materials/internals/definition.yaml
chloe0592: Yes, was thinking about it as well. I'll add it :)

chloe0592: Added [3a4141a](https://github.com/commercetools/ui-kit/pull/2317/commits/3a4141a69b03e085e3adffd61f7456a8fe0832ee) ✅

packages/components/link/src/link.tsx
emmenko: Yes but the changes then shouldn't be visible in your PR.

I would suggest to rebase the branch instead of merging it.

chloe0592: Ah, yes you're right. 😅

packages/components/text/src/text.styles.ts
```diff
@@ -56,15 +56,49 @@ const getTone = (tone: string) => {
...
-      return '1.5385rem';
+      return designTokens.fontSizeForTextAsH3;
     case 'h4':
-      return '1.2308rem';
+      return designTokens.fontSizeForTextAsH4;
     case 'h5':
-      return '1.0769rem';
+      return designTokens.fontSizeForTextAsH5;
+    default:
+      return '';
```
emmenko, line 69: Shouldn't we return a value here? 🤔 (same in the other functions)

emmenko (commented):
Looking good! 🥳

design-system/src/theme-provider.tsx
```diff
@@ -67,6 +67,7 @@ type ThemeProviderProps = {
 };
 
 const ThemeProvider = (props: ThemeProviderProps) => {
+  const themeName = props.theme || 'default'; // Use 'default' theme if none is provided
```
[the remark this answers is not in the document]
chloe0592: Could we `return elementType;` since we don't have default font size for all of them?
And for the tone `return designTokens.colorPrimary;`?

packages/components/text/src/text.styles.ts
chloe0592: Could we `return elementType`; since we don't have default font size for all of them?
And for the tone return `designTokens.colorPrimary;`?

design-system/materials/internals/definition.yaml
```diff
@@ -73,7 +78,41 @@ choiceGroupsByTheme:
...
+        font-size-70: '2rem'
+
+    fontWeights:
+      label: Font weights
+      prefix: font-weight
+      choices:
+        font-weight-300: '300'
+        font-weight-400: '400'
+        font-weight-500: '500'
+        font-weight-600: '600'
```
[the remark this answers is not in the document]
chloe0592: I don't know but to me that made most sense because we already use a number scale and this way it's most straightforward which value are we aiming for

packages/components/text/src/text.styles.ts
emmenko: > Could we `return elementType;` since we don't have default font size for all of them?

Maybe we can return `1rem`?

> And for the tone return `designTokens.colorPrimary;`?

For the tone is fine to return an empty string, because the other values return the full style `color: ???`.

For the `font-size`, `line-height`, `font-weight` the function returns the actual value (not the full style), so I think it would be better to return a value rather than having potentially an empty style (e.g. `font-size: ;`).

What do you think?

chloe0592: Ah, yes I get your point and I agree. Thank you!

design-system/materials/internals/definition.yaml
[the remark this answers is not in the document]
CarlosCortizasCT: Bear in mind we currently are using three different weights (300, 400 and 700) but in the new theme we will be using four (400, 500, 600 and 700).
If we want to use a more semantic name for the tokens, we need to mind all the options.

In the past I used light, normal and bold, but in this context we will have two values between normal and bold.
We can tailor the naming to match our current scale, but I don't see it as a positive change for us in the future.

Personally I would rather keep using the numbered naming.

chloe0592 (commented):
Accidentally removed link in the previous commit

packages/components/text/src/text.styles.ts
```diff
@@ -56,24 +56,60 @@ const getTone = (tone: string) => {
...
+    case 'h2':
+      return designTokens.lineHeightForTextAsH2;
+    case 'h3':
+      return designTokens.lineHeightForTextAsH3;
+    case 'h4':
+      return designTokens.lineHeightForTextAsH4;
+    case 'h5':
+      return designTokens.lineHeightForTextAsH5;
+    default:
+      return '1.625rem';
```
emmenko, line 86: Shouldn't we use a design token here?

packages/components/text/src/text.styles.ts
```diff
@@ -56,24 +56,60 @@ const getTone = (tone: string) => {
...
+    case 'h2':
+      return designTokens.fontWeightForTextAsH2;
+    case 'h3':
+      return designTokens.fontWeightForTextAsH3;
+    case 'h4':
+      return designTokens.fontWeightForTextAsH4;
+    case 'h5':
+      return designTokens.fontWeightForTextAsH5;
+    default:
+      return '400';
```
emmenko, line 103: Same here? We can also use `normal` maybe

emmenko (commented):
Overall looks good to me, great work!

CarlosCortizasCT:
We need to take a look at the tokens Storybook story as it's failing:

image

design-system/src/theme-provider.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: We still need to undo this changes.

chloe0592:
> We need to take a look at the tokens Storybook story as it's failing:
> 
> image

@CarlosCortizasCT is this related to the buttons? `shadow-for-button-when-disabled` It seems to me it's just missing a value

design-system/src/theme-provider.tsx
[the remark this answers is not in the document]
chloe0592: Yes, true, thanks for pointing that out. So I removed them from the `theme-provider.tsx `file - [here](https://github.com/commercetools/ui-kit/pull/2317/commits/b59c1a3bd960fa9389650ce076a0e2578433406d).
It seems that the other two `contexts.js` and `theme-context.js` files are alright, because they are the same as what is on the main branch. Let me know if there's anything else that I'm missing out.

CarlosCortizasCT:
> > We need to take a look at the tokens Storybook story as it's failing:
> > image
> 
> @CarlosCortizasCT is this related to the buttons? `shadow-for-button-when-disabled` It seems to me it's just missing a value

So it seems the problem is related to a check we have in that story to make sure we only use choices that are actually available.
This is no longer true because we are also using css values for some specific choices so I remove that constraint (we alraedy have our constraints at tokens build time).

Updated in this commit: 182653996560a76ed5bc3f551b39844c8303c3d2

emmenko (approved):
🥳

CarlosCortizasCT:
> PS: the VRTs look good to me (just font rendering adjustments). From my side they can be approved ✅

I was working in adjusting the default theme legacy tokens but I had to stop for a while and now the branch is already merged 😂 

Since I also have to adjust some tokens related with buttons, I'll add these pending adjustments in a next PR.

PR #2319, opened 2022-10-28
Update Buttons to support new theme

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=0VLbozp)

#### Summary

Update buttons components to support the new theme while keeping its visual styles in the default one.

## Description

We need to update the different buttons components to use new design tokens so we can tailor the stylings to keep the current visualization for the default theme but allow for a different one in the new theme.

TODO:

- [x] border color for secondary button and icon button
- [x] border radius for buttons in all their sizes
- [x] font weight for text buttons (semi-bold means 500 or 600?)
- [x] do `FlatButton` use the same font weight as the primary and secondary buttons?
- [x] height for buttons in all their sizes
- [x] use disabled hover styles that were hidden?


design-system/materials/internals/definition.yaml
```diff
@@ -414,11 +477,23 @@ decisionGroupsByTheme:
...
+        shadow-for-button:
+          choice: shadow-7
+        shadow-for-button-when-focused:
+          choice: shadow-8
+        shadow-for-button-when-hovered:
+          choice: shadow-8
+        shadow-for-button-when-active:
+          choice: shadow-9
+        shadow-for-button-when-disabled:
+          choice: '0 0 0 1px var(--color-neutral) inset'
```
CarlosCortizasCT, line 496: Currently we are using this value, but for the new theme we don't want any shadow so I need to use the whole value as a choice to be able to have then `none` value in the new theme choice.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +504,102 @@ decisionGroupsByTheme:
...
         background-color-for-tag-warning:
           description: ''
           choice: color-neutral-95
 
+    borders:
+      label: Borders
+      prefix: border
+      decisions:
+        border-for-button-as-secondary:
+          choice: '1px solid hsla(0, 0%, 81%, .60)'
```
CarlosCortizasCT, line 521: Not sure yet about this color. Need to double check with UX folks.

CarlosCortizasCT, line 521: Need to use the whole value as there are no borders in the default theme.

design-system/materials/internals/definition.yaml
```diff
@@ -101,6 +104,7 @@ choiceGroupsByTheme:
       label: Shadows
       prefix: shadow
       choices:
+        shadow-0: 'none'
```
CarlosCortizasCT, line 107: In the new theme we have buttons having no shadow at all so having a token for it makes sense to me.

design-system/materials/internals/definition.yaml
```diff
@@ -170,9 +174,7 @@ choiceGroupsByTheme:
       prefix: color
       description: Dark theme colors overwrite
       choices:
-        color-primary: '#f78d26'
-        color-solid: '#fff'
-        color-surface: '#1a1a1a'
+        color-primary: '#00b39e'
```
CarlosCortizasCT, line 177: Currently the generate tokens script needs the test section to exist so I'm just copying one of the values here.
Maybe we could look into it as a follow-up.

design-system/materials/internals/definition.yaml
```diff
@@ -208,8 +210,22 @@ variants:
     description: 'Body text element'
   detail:
     description: 'Detail text element'
+  icon:
```
CarlosCortizasCT, line 213: All these are new variants I need for the different buttons combinations.

design-system/materials/internals/definition.yaml
```diff
@@ -244,6 +262,17 @@ decisionGroupsByTheme:
           description: ''
           choice: color-warning-95
 
+    borders:
+      label: Borders
+      prefix: border
+      decisions:
+        border-for-button-as-secondary:
+          choice: 'none'
```
CarlosCortizasCT, line 270: Default theme buttons don't use borders but shadows for border-like visualization.

design-system/materials/internals/definition.yaml
```diff
@@ -326,6 +369,21 @@ decisionGroupsByTheme:
           description: ''
           choice: font-family-body
 
+    heights:
+      label: Height
+      prefix: height
+      decisions:
+        height-for-button-as-big:
+          choice: '32px'
```
CarlosCortizasCT, line 377: I'm using actual CSS values here as I don't see it makes sense to have a height scale since different components might have different needs so I guess we would end up with tokens like `height-32: 32px` and we would be adding new ones for every new use case.
However, if you think it would be better to have such a scale, I can refactor this part.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+          choice: '1px solid hsla(0, 0%, 81%, .80)'
+        border-for-button-as-secondary-when-active:
+          choice: '1px solid hsla(0, 0%, 81%)'
+
+    borderColors:
+      label: Border Colors
+      prefix: border-color
+      decisions:
+        border-color-for-button-as-icon:
+          choice: 'hsla(0, 0%, 81%, .60)'
```
CarlosCortizasCT, line 530: This color is not final yet.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+        border-color-for-button-as-icon-when-disabled:
+          choice: color-surface
+
+    borderRadiuses:
+      label: Border radiuses
+      prefix: border-radius
+      decisions:
+        border-radius-for-button-as-big:
+          choice: border-radius-4
+        # TODO: We don't have the right value from UX yet. This one is a placeholder
```
CarlosCortizasCT, line 540: Missing values. We're still need UX folks to provide them.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+          # TODO: In Miro they say "semi-bold" which could be 500 or 600
+          choice: font-weight-600
+
+    heights:
+      label: Heights
+      prefix: height
+      decisions:
+        height-for-button-as-big:
+          choice: '40px'
+        # TODO: We're missing this value from UX team (this is a placeholder)
```
CarlosCortizasCT, line 574: Missing values. We're still need UX folks to provide them.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+        font-size-for-button:
+          description: ''
+          choice: font-size-40
+
+    fontWeight:
+      label: Font Weights
+      prefix: font-weight
+      decisions:
+        font-weight-for-button:
+          # TODO: In Miro they say "semi-bold" which could be 500 or 600
```
CarlosCortizasCT, line 565: Missing value. We're still need UX folks to provide it.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+      prefix: height
+      decisions:
+        height-for-button-as-big:
+          choice: '40px'
+        # TODO: We're missing this value from UX team (this is a placeholder)
+        height-for-button-as-small:
+          choice: '32px'
+        height-for-button-as-icon-as-big:
+          choice: '40px'
+        # TODO: We're missing this value from UX team (this is a placeholder)
```
CarlosCortizasCT, line 579: Missing values. We're still need UX folks to provide them.

design-system/materials/internals/definition.yaml
```diff
@@ -429,13 +502,102 @@ decisionGroupsByTheme:
...
+          choice: '32px'
+        # TODO: We're missing this value from UX team (this is a placeholder)
+        height-for-button-as-icon-as-small:
+          choice: '24px'
+
+    shadows:
+      label: Shadows
+      prefix: shadow
+      decisions:
+        shadow-for-button:
```
CarlosCortizasCT, line 590: New theme buttons don't use shadows but borders for border-like visualization.

design-system/package.json
```diff
@@ -28,5 +28,8 @@
     "lodash": "4.17.21",
     "prop-types": "15.8.1",
     "react": "17.0.2"
+  },
+  "devDependencies": {
+    "nodemon": "^2.0.20"
```
CarlosCortizasCT, line 33: The `build:tokens:watch` command is using this dependency but it was missing.

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,15 +21,99 @@ const endProgram = (message) => {
   process.exit(1);
 };
 
-const TOKEN_REGEX =
```
CarlosCortizasCT, line 24: We were using this regex to parse the tokens names and to check its syntax.

Since I think we need to support more combinations I replaced this regex with a couple of functions which parse and check syntax independently.

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,15 +21,99 @@ const endProgram = (message) => {
...
+    - <attribute>-for-<component-group>-as-<state>-
+      + border-radius-for-button-as-big
+    - <attribute>-for-<component-group>-as-<state>-when-<state>
+      + border-for-button-as-secondary-when-hovered
+    - <attribute>-for-<component-group>-as-<state>-as-<state>-
+      + border-radius-for-button-as-icon-as-small
+    - <attribute>-for-<component-group>-as-<state>-as-<state>-when-<state>-
+      + border-radius-for-button-as-icon-as-small-when-disabled
+*/
+function parseToken(token) {
```
CarlosCortizasCT, line 50: A new function to parse token names supporting for new combinations:
* a token can have several variants
* a token can have variant/s and states

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,15 +21,99 @@ const endProgram = (message) => {
...
+
+/*
+  We make sure the order of the token parts is
+    1. component group
+    2. variants (might have several of this)
+    3. state
+
+  Eg: <attribute>-for-<component-group>-as-<state>-when-<state>
+*/
+function isValidTokenName(tokenName, tokenParts) {
```
CarlosCortizasCT, line 99: A function to check the syntax of tokens names. Basically we want to enforce the order of the different parts.

design-system/scripts/generate-design-tokens.js
```diff
@@ -68,45 +152,53 @@ Object.entries(definitions.decisionGroupsByTheme).forEach(
         }
         if (
           !designTokens[themeName][decision.choice] &&
-          !designTokens.default[decision.choice]
+          !designTokens.default[decision.choice] &&
+          !isAllowedCssChoice(decision.choice)
```
CarlosCortizasCT, line 156: This is to support some CSS values in the design tokens choices.
We covering cases like this: secondary buttons currently don't use borders values but they do in the new themes. I added a new token to handle the difference and I'm using the value `none` for the default theme but `1px solid hsla(0, 0%, 81%, .60)` in the new theme.

Also, height values don't use a scale because they are very coupled to the component needing them so I'm using pixel values: `height-for-button-as-big: 32px`

docs/.storybook/configs/theme-context.js
```diff
@@ -1,20 +1,10 @@
 import PropTypes from 'prop-types';
 import { designTokens, ThemeProvider } from '../../../design-system';
 
-const defaultTheme = designTokens;
```
CarlosCortizasCT, line 4: Now that we're working with the new theme, I updated Storybook to actually use it instead of a custom one defined here.

packages/components/buttons/accessible-button/src/accessible-button.tsx
```diff
@@ -160,7 +160,9 @@ const AccessibleButton = forwardRef<HTMLButtonElement, TAccessibleButtonProps>(
         css={css`
           ${normalizedButtonStyles}
           display: inline-flex;
-          font-size: ${designTokens.fontSizeDefault};
+          font-size: ${designTokens.fontSizeForButton};
+          /* TODO: We need to know if we want the same weight for primary, secondary and flat buttons */
```
CarlosCortizasCT, line 164: We need to confirm this with UX folks.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -3,12 +3,6 @@ import { designTokens } from '@commercetools-uikit/design-system';
 import { css } from '@emotion/react';
 import type { TIconButtonProps } from './icon-button';
 
-const buttonSizes = {
-  small: '16px',
-  medium: '24px',
-  big: '32px',
-};
```
CarlosCortizasCT, line 10: We moved this values to the design-systems definitions so we can have different ones per theme.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -84,15 +78,6 @@ const getStateStyles = (
       case 'info':
         return [
           activeStyle,
-          isDisabled &&
```
CarlosCortizasCT, line 87: This styles were never used. If the `isDisabled` prop is true, we return other styles previously in this function.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -106,16 +91,6 @@ const getStateStyles = (
       case 'primary':
         return [
           activeStyle,
-          // When the button is active and somehow is disabled it should have
-          // a different color to indicate that it's active but can't receive any actions
-          isDisabled &&
```
CarlosCortizasCT, line 111: This styles were never used. If the `isDisabled` prop is true, we return other styles previously in this function.

packages/components/buttons/icon-button/src/icon-button.tsx
```diff
@@ -119,17 +119,17 @@ const IconButton = <TStringOrComponent extends ElementType = 'button'>(
...
+          box-shadow: ${designTokens.shadowForButton};
           color: ${designTokens.colorSolid};
           transition: background-color ${designTokens.transitionLinear80Ms},
             box-shadow 150ms ease-in-out;
         `,
+        getThemeStyles(props.theme),
         getStateStyles(props.isDisabled, isActive, props.theme),
         getShapeStyles(props.shape, props.size),
         getSizeStyles(props.size),
-        getThemeStyles(props.theme),
```
CarlosCortizasCT, line 132: This need to be confirmed with UX folks but there were some styles when button is disabled that weren't being applied. It seemed a bug to me.

packages/components/buttons/primary-button/src/primary-button.styles.ts
```diff
@@ -34,7 +34,6 @@ const getButtonStyles = (
     align-items: center;
     color: ${designTokens.colorSurface};
     transition: background-color ${designTokens.transitionLinear80Ms};
-    font-size: ${designTokens.fontSizeDefault};
```
CarlosCortizasCT, line 37: This font size was the same as the one in `AccessibleButton` (which this component relies on).
UX folks told me all buttons must have the same font size so I think it's better to only have the value set over there.

packages/components/buttons/secondary-button/src/secondary-button.styles.ts
```diff
@@ -31,19 +32,14 @@ const getStateStyles = (
...
-        box-shadow: ${designTokens.shadow9};
+        border: ${designTokens.borderForButtonAsSecondaryWhenActive};
+        box-shadow: ${designTokens.shadowForButtonWhenActive};
         background-color: ${designTokens.colorSurface};
         &:focus,
         &:hover {
           background-color: ${designTokens.colorNeutral95};
         }
       `,
-      isDisabled &&
```
CarlosCortizasCT, line 41: This styles were never used. If the `isDisabled` prop is true, we return other styles previously in this function.

design-system/src/theme-provider.tsx
```diff
@@ -64,6 +64,7 @@ type ThemeProviderProps = {
 };
 
 const ThemeProvider = (props: ThemeProviderProps) => {
+  const themeName = props.theme || 'default'; // Use 'default' theme is none is provided
```
CarlosCortizasCT, line 67: We were technically requiring the theme prop for the provider to actually work.

design-system/materials/internals/definition.yaml
```diff
@@ -36,23 +36,33 @@ choiceGroupsByTheme:
         color-primary-40: 'hsl(172.9608938547486, 100%, 40%)'
         color-primary-85: 'hsl(172.9608938547486, 100%, 85%)'
         color-primary-95: 'hsl(172.9608938547486, 100%, 95%)'
+        color-primary-op-90: 'hsl(173deg 100% 35% / 90%)'
+        color-primary-op-95: 'hsl(173deg 100% 35% / 95%)'
```
CarlosCortizasCT, line 40: There are some temporary color tokens (the ones with `op`) since I'm not sure how to handle the new requirement about using opacity for buttons background color (5% for hover and 10% for active).

For primary color we already have a `color-primary-95` token which, by name, would fit to be used when a primary button is hovered, however, its value does not seem to be what we expect if a compare it with the base primary color with 95% opacity.

Here is an visual comparision:
image

CarlosCortizasCT:
We have a failing Percy test related to icon buttons which is intentional because I wanted to double check something that seems to be currently broken.

When an icon button is toggled and we have a theme selected, the border color is not consistent:
* for `primary` theme, the color is white
* for `info` theme, the color is blur

I guess we should use the same pattern in both situations:
* use white for both cases
* use specific color (green or blue) for both cases

@FilPob could you please share your feedback about this?

FilPob:
It is actually quite weird to use border-color in general as these buttons have only a fill color and no border. The toggled state is then visualised with an inset box shadow. But anyways you are right. The border color should be green for the toggled primary button and blue for the toggled info button

CarlosCortizasCT:
To check the visualization of the components with the new theme tokens you can use the theme switcher:

Screenshot 2022-11-07 at 12 48 53

docs/.storybook/configs/contexts.js
```diff
@@ -1,8 +1,14 @@
 import intlContext from './intl-context';
 import themeContext from './theme-context';
 
+// We only want to show the theme switcher in local environment
+// or in preview deployments
+const shouldIncludeThemeContext = () =>
+  process.env.NODE_ENV !== 'production' ||
+  window.location.hostname.includes('commercetools.vercel.app');
```
CarlosCortizasCT, line 8: I tried using the `VERCEL_ENV` env variable but it didn't work so I used this alternative.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -175,43 +60,91 @@ const getSizeStyles = (size: TIconButtonProps['size']) => {
...
+        &:active {
+          background-color: ${designTokens.backgroundColorForButtonAsIconAsInfoWhenActive};
+          box-shadow: ${designTokens.shadowForButtonWhenActive};
+        }
+        ${isActive ? '&,' : ''}
+        &:hover,
+        &:active {
           border-color: ${designTokens.colorInfo};
-          color: ${designTokens.colorSurface};
+          border-color: ${designTokens.borderColorForButtonAsIconAsInfo};
```
kark, line 147: duplicated `border-color` rule

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -175,43 +60,91 @@ const getSizeStyles = (size: TIconButtonProps['size']) => {
...
+        &:active {
+          background-color: ${designTokens.backgroundColorForButtonAsIconAsPrimaryWhenActive};
+          box-shadow: ${designTokens.shadowForButtonWhenActive};
+        }
+        ${isActive ? '&,' : ''}
+        &:hover,
+        &:active {
           border-color: ${designTokens.colorPrimary};
-          color: ${designTokens.colorSurface};
+          border-color: ${designTokens.borderColorForButtonAsIconAsPrimary};
```
kark, line 129: duplicated `border-color` rule

kark (commented):
Kudos for making successful way through all the design changes 🙌 it was very tricky

design-system/materials/internals/definition.yaml
CarlosCortizasCT: We had a meeting where we discussed this topic and we decided to not create new low level tokens for the new colors needed but instead use the new values in the decision tokens for the buttons where they are needed.

Updated in this commit: f70abe822d56e2ee755bdefb06605319a94db8ca

packages/components/buttons/icon-button/src/icon-button.styles.ts
CarlosCortizasCT: Thanks for noticing 👍 

Updated in this commit: ef1c6f88e890539eb7dba2918d557d38ccbad782

packages/components/buttons/icon-button/src/icon-button.styles.ts
CarlosCortizasCT: Thanks for noticing 👍 

Updated in this commit: ef1c6f88e890539eb7dba2918d557d38ccbad782

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -175,43 +60,91 @@ const getSizeStyles = (size: TIconButtonProps['size']) => {
...
+    `;
+  }
+
+  if (!theme || theme === 'default') {
+    return css`
+      &:hover {
+        background-color: ${designTokens.backgroundColorForButtonWhenHovered};
+        box-shadow: ${designTokens.shadowForButtonWhenHovered};
+      }
+      ${isActive ? '&,' : ''}
```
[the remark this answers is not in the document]
CarlosCortizasCT: Maybe I'm missing something but I do seem to see the active styles applied.
Could you maybe provide extra hints of what you're missing? 🙏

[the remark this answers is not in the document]
kark: Sure - it's about `IconButton` in default theme.
With this line:

Without this line:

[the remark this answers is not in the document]
CarlosCortizasCT: I think that's the point.
I guess what's happening here is that you have both `isToggleButton` and `isToggled` checks checked. That means the icon must be rendered as it's being clicked.

CarlosCortizasCT:
@FilPob it seems all visual issues have been addressed. Could you please take another look and leave your feedback?
Thanks.

FilPob:
@CarlosCortizasCT thanks a lot! All styles look correct to me. One thing remaining is that the icon-button should be squared on default. Or better say: In the new vision we only have square buttons. But we can address this in a follow-up as you already mentioned there is something tricky.

FilPob (approved):

design-system/materials/internals/definition.yaml
```diff
@@ -74,10 +79,25 @@ choiceGroupsByTheme:
...
+        font-size-50: '1.25rem'
+        font-size-60: '1.5rem'
+    fontWeight:
+      label: Font weight
+      prefix: font-weight
+      choices:
+        font-weight-400: '400'
+        font-weight-500: '500'
+        font-weight-600: '600'
+        font-weight-700: '700'
```
emmenko, line 95: Are these changes relevant to this PR? I thought they are implemented in the typography PR #2317 🤔

design-system/materials/internals/definition.yaml
emmenko: I think it's fine this way, thanks 😊

design-system/src/theme-provider.tsx
```diff
@@ -67,6 +67,7 @@ type ThemeProviderProps = {
 };
 
 const ThemeProvider = (props: ThemeProviderProps) => {
+  const themeName = props.theme || 'default'; // Use 'default' theme if none is provided
```
emmenko, line 70: Not needed, as the value is defined in  React's `defaultProp` object.

emmenko (commented):
Overall looks good, thanks a lot! 💯 

I played a bit around the button variations in storybook and I didn't notice anything strange.

Just a couple of questions as there seem to be some leftovers or duplicate changes (with other PRs).

design-system/materials/internals/definition.yaml
CarlosCortizasCT: Buttons have different font weights based on the theme.
In this PR I only need `400` and `500` but I opted for including the other ones as I thought we would need them anyway. I can remove `600` and `700` if you think it's better.

design-system/src/theme-provider.tsx
CarlosCortizasCT: Updated in this commit: 3909ff9415b69ef3d077becfb386651d4400d852

Thanks!

emmenko (approved):
💯

PR #2320, opened 2022-11-02
Update Cards to support new theme (Shield-625)

Rhotimee:
#### Summary
Update cards component to support new design tokens.


design-system/materials/internals/definition.yaml
```diff
@@ -492,6 +496,8 @@ decisionGroupsByTheme:
           choice: shadow-9
         shadow-for-button-when-disabled:
           choice: '0 0 0 1px var(--color-neutral) inset'
+        shadow-for-card-when-raised:
```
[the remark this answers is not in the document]
Rhotimee: Updated [0fd3638](https://github.com/commercetools/ui-kit/pull/2320/commits/0fd3638807f294372914fad1e7bddbdf90db5729). Thanks you!

packages/components/card/src/card.tsx
```diff
@@ -34,8 +34,13 @@ const Card = (props: TCardProps) => (
...
       font-size: 1rem;
-      box-shadow: ${props.type === 'raised' ? designTokens.shadow1 : 'none'};
-      border-radius: ${designTokens.borderRadius6};
+      box-shadow: ${props.type === 'raised'
+        ? designTokens.shadowForCardWhenRaised
+        : 'none'};
+      border-radius: ${designTokens.borderRadius4};
+      border-color: ${props.type === 'raised'
+        ? designTokens.colorNeutral90
+        : 'none'};
```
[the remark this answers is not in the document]
Rhotimee: Good catch. Updated [dbfc38f](https://github.com/commercetools/ui-kit/pull/2320/commits/dbfc38f4b81078bc2507c75450d3376a3db153ba), [c7d5b50](https://github.com/commercetools/ui-kit/pull/2320/commits/c7d5b5046444861b3ae42b8a5e23a6acd9daece1)

design-system/materials/internals/definition.yaml
```diff
@@ -308,6 +312,8 @@ decisionGroupsByTheme:
           choice: color-surface
         border-color-for-button-as-icon-when-disabled:
           choice: color-neutral
+        border-color-for-card-when-raised:
+          choice: color-neutral-90
```
CarlosCortizasCT, line 316: This token is inside the default `theme` group so the value here should be `color-surface` since we don't want the border to be visible in this theme.
Also, you would need to add this token in the `test` group with this value (`color-neutral-90`).

design-system/materials/internals/definition.yaml
```diff
@@ -329,6 +335,8 @@ decisionGroupsByTheme:
         border-radius-for-tag:
           description: ''
           choice: border-radius-2
+        border-radius-for-card:
```
CarlosCortizasCT, line 338: This token is inside the `default` theme group so the value here should be `border-radius-6` (the current value).
Also, you would need to add this token in the test group with this value (`border-radius-4`).

design-system/materials/internals/definition.yaml
```diff
@@ -492,6 +500,8 @@ decisionGroupsByTheme:
           choice: shadow-9
         shadow-for-button-when-disabled:
           choice: '0 0 0 1px var(--color-neutral) inset'
+        shadow-for-card-when-raised:
+          choice: '0 1px 2px rgb(25 40 81 / 5%)'
```
CarlosCortizasCT, line 504: This token is inside the `default` theme group so the value here should be `shadow-1` (the current value).
Also, you would need to add this token in the test group with this value (`0 1px 2px rgb(25 40 81 / 5%)`).

design-system/materials/internals/definition.yaml
```diff
@@ -597,6 +607,8 @@ decisionGroupsByTheme:
           choice: shadow-0
         shadow-for-button-when-disabled:
           choice: shadow-0
+        shadow-for-card-when-raised:
+          choice: 'none'
```
CarlosCortizasCT, line 611: This token is in the new theme group so we should use the new value provided by UX: `0 1px 2px rgb(25 40 81 / 5%)`.

CarlosCortizasCT (commented):
This is looking better but we have some VRTs failing.
It is due to the new tokens only existing in the `default` theme but with the values we want to use for the new one.

I've left some comments with suggestions about how to adjust the tokens.

Rhotimee:
Thanks @CarlosCortizasCT, I have updated all here [dedb912](https://github.com/commercetools/ui-kit/pull/2320/commits/dedb912e7ca7b77b4e4f61dd86a8947d230aa311)

CarlosCortizasCT:
To check the visualization of the components with the new theme tokens you can use the theme switcher:

Screenshot 2022-11-07 at 12 48 53

kark (approved):
Thanks @Rhotimee and @CarlosCortizasCT! Looking good!

FilPob:
I am not sure if I see the correct changes in the deploy link.

For the raised card:

- the **border color** should be: color-neutral-90 (in the inspect mode, the value seems to be there but it's somehow not applied)
- the **box shadow** should be:  0px 1px 5px 0px rgba(0, 0, 0, 0.05) ...The provided value in the Miro board is wrong. Sorry for that
- The **border radius** should be 4px. I am not sure if this is the case. I also don't see this in inspect elements as i am used to.

CarlosCortizasCT:
> I am not sure if I see the correct changes in the deploy link.
> 
> For the raised card:
> 
> * the **border color** should be: color-neutral-90 (in the inspect mode, the value seems to be there but it's somehow not applied)
> * the **box shadow** should be:  0px 1px 5px 0px rgba(0, 0, 0, 0.05) ...The provided value in the Miro board is wrong. Sorry for that
> * The **border radius** should be 4px. I am not sure if this is the case. I also don't see this in inspect elements as i am used to.

@FilPob,

I've updated the value for the **box-shadow** with the one you've provided in your comment.

Apart from that, it seems I do see the expected values son perhaps it's better to have a quick sync and go through this changes screen sharing.

CarlosCortizasCT (approved):
Super! 🚀

CarlosCortizasCT:
@FilPob the issue with the border has been addressed so please [take another look](https://ui-e850ibyqt-commercetools.vercel.app/?path=/story/components-cards--raised) whenever you can 🙇

FilPob:
@CarlosCortizasCT  thanks! Looks all good now

FilPob (approved):

PR #2321, opened 2022-11-03
Extend design tokens naming rules

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=9-mOGm8)

#### Summary

Extend design tokens naming so we can support more use cases.

## Description

While working with the designs for the new theme and making our components compatible with both current and new UI we-re finding finding some restrictions in tokens naming that it seems we need to extend.
So now the naming pattern we allow is like this:
```
<attribute>-for-<component-group>(-[when|as]-[<variant|state>])?
```

This pattern allows for this token names:
```
background-color-for-tag
font-color-for-text-when-inverted
font-size-for-text-as-h1
```

What we are missing is:
* ability to combine variants with states: `<attribute>-for-<component-group>-as-<variant>-when-<state>` (eg: `border-for-button-as-secondary-when-hovered`)
* ability to combine variants with variants: `<attribute>-for-<component-group>-as-<variant>-as-<variant>` (eg: `border-radius-for-button-as-icon-as-small`)
  * In this case both icon and small are treated as variants for buttons

### DIFF
--- a/design-system/scripts/generate-design-tokens.js
+++ b/design-system/scripts/generate-design-tokens.js
@@ -38,13 +38,13 @@ const designTokens = {};
       + background-color-for-tag
     - <attribute>-for-<component-group>-when-<state>-
       + background-color-for-button-when-disabled
-    - <attribute>-for-<component-group>-as-<state>-
+    - <attribute>-for-<component-group>-as-<variant>-
       + border-radius-for-button-as-big
-    - <attribute>-for-<component-group>-as-<state>-when-<state>
-      + border-for-button-as-secondary-when-hovered
-    - <attribute>-for-<component-group>-as-<state>-as-<state>-
-      + border-radius-for-button-as-icon-as-small
-    - <attribute>-for-<component-group>-as-<state>-as-<state>-when-<state>-
+    - <attribute>-for-<component-group>-as-<variant>-when-<state>
+    + border-for-button-as-secondary-when-hovered
+    - <attribute>-for-<component-group>-as-<variant>-as-<variant>-
+    + border-radius-for-button-as-icon-as-small
+    - <attribute>-for-<component-group>-as-<variant>-as-<variant>-when-<state>-
       + border-radius-for-button-as-icon-as-small-when-disabled
 */
 function parseToken(token) {
@@ -94,7 +94,7 @@ function parseToken(token) {
     2. variants (might have several of this)
     3. state
 
-  Eg: <attribute>-for-<component-group>-as-<state>-when-<state>
+  Eg: <attribute>-for-<component-group>-as-<variant>-when-<state>
 */
 function isValidTokenName(tokenName, tokenParts) {
   const componentGroupIndex = tokenName.indexOf(tokenParts.componentGroup);
--- a/design-system/scripts/generate-design-tokens.js
+++ b/design-system/scripts/generate-design-tokens.js
@@ -21,13 +21,98 @@ const endProgram = (message) => {
   process.exit(1);
 };
 
-const TOKEN_REGEX =
-  /^(\w+(?:-\w+)(?:-\w+)?)(?:-for-(\w+(?:-\w+)?))?(?:-when-([\w-]+?))?(?:-on-([\w-]+?))?$/i;
+const ALLOWED_CSS_VALUES_IN_CHOICES = /px|none|hsla/;
+
+const isAllowedCssChoice = (choice) =>
+  choice.match(ALLOWED_CSS_VALUES_IN_CHOICES) !== null;
 
 const supportedStates = Object.keys(definitions.states);
 const supportedComponentGroups = Object.keys(definitions.componentGroups);
+const supportedVariants = Object.keys(definitions.variants || {});
 
 const designTokens = {};
+
+/*
+  Allowed patterns with examples:
+    - <attribute>-for-<component-group>
+      + background-color-for-tag
+    - <attribute>-for-<component-group>-when-<state>-
+      + background-color-for-button-when-disabled
+    - <attribute>-for-<component-group>-as-<state>-
+      + border-radius-for-button-as-big
+    - <attribute>-for-<component-group>-as-<state>-when-<state>
+      + border-for-button-as-secondary-when-hovered
+    - <attribute>-for-<component-group>-as-<state>-as-<state>-
+      + border-radius-for-button-as-icon-as-small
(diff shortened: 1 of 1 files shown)


design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
   process.exit(1);
 };
 
-const TOKEN_REGEX =
```
CarlosCortizasCT, line 24: We were using this regex to parse the tokens names and to check its syntax.

Since I think we need to support more combinations I replaced this regex with a couple of functions which parse and check syntax independently.

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
...
+    - <attribute>-for-<component-group>-as-<state>-
+      + border-radius-for-button-as-big
+    - <attribute>-for-<component-group>-as-<state>-when-<state>
+      + border-for-button-as-secondary-when-hovered
+    - <attribute>-for-<component-group>-as-<state>-as-<state>-
+      + border-radius-for-button-as-icon-as-small
+    - <attribute>-for-<component-group>-as-<state>-as-<state>-when-<state>-
+      + border-radius-for-button-as-icon-as-small-when-disabled
+*/
+function parseToken(token) {
```
CarlosCortizasCT, line 50: A new function to parse token names supporting for new combinations:

* a token can have several variants
* a token can have variant/s and states

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
...
+
+/*
+  We make sure the order of the token parts is
+    1. component group
+    2. variants (might have several of this)
+    3. state
+
+  Eg: <attribute>-for-<component-group>-as-<state>-when-<state>
+*/
+function isValidTokenName(tokenName, tokenParts) {
```
CarlosCortizasCT, line 99: A function to check the syntax of tokens names. Basically we want to enforce the order of the different parts.

design-system/scripts/generate-design-tokens.js
```diff
@@ -67,41 +152,53 @@ Object.entries(definitions.decisionGroupsByTheme).forEach(
         }
         if (
           !designTokens[themeName][decision.choice] &&
-          !designTokens.default[decision.choice]
+          !designTokens.default[decision.choice] &&
+          !isAllowedCssChoice(decision.choice)
```
CarlosCortizasCT, line 156: This is to support some CSS values in the design tokens choices.
We covering cases like this: secondary buttons currently don't use borders values but they do in the new themes. I added a new token to handle the difference and I'm using the value `none` for the default theme but `1px solid hsla(0, 0%, 81%, .60)` in the new theme.

Also, height values don't use a scale because they are very coupled to the component needing them so I'm using pixel values: `height-for-button-as-big: 32px`

chloe0592 (approved):

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
...
+  );
+}
+
+/*
+  We make sure the order of the token parts is
+    1. component group
+    2. variants (might have several of this)
+    3. state
+
+  Eg: <attribute>-for-<component-group>-as-<state>-when-<state>
```
chloe0592, line 97: Small typo here: `Eg: <attribute>-for-<component-group>-as-<variant>-when-<state>`

CarlosCortizasCT: Thanks for noticing 🙇 

It's been fixed in this commit: 9596b45c20e85d6609ae5aa156fc16e654aa1ed4

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
...
+      if (['for', 'as', 'when'].includes(part)) {
+        partType = part;
+        if (part === 'as') newVariant = true;
+        return agg;
+      }
+
+      if (partType === 'for') {
+        agg.componentGroup = `${agg.componentGroup}${
+          !!agg.componentGroup ? '-' : ''
+        }${part}`;
```
kark, line 65: Since this pattern is used in a few places we could have an external function:
```js
const joinParts = (preceding, following) =>
  `${preceding}${preceding.length > 0 ? '-' : ''}${following}`;
```
and use it like
```js
agg.componentGroup = joinParts(agg.componentGroup, part);
```

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
...
+          agg.variants[lastIndex] = `${agg.variants[lastIndex]}-${part}`;
+        }
+      } else {
+        agg.cssRule = `${agg.cssRule}${!!agg.cssRule ? '-' : ''}${part}`;
+      }
+
+      return agg;
+    },
+    {
+      cssRule: '',
```
kark, line 83:
```suggestion
      cssProperty: '',
```

kark (commented):
This looks good and works great 🙌 It was a great idea to replace the regex.
I'm just leaving a few minor suggestions

design-system/scripts/generate-design-tokens.js
```diff
@@ -21,14 +21,99 @@ const endProgram = (message) => {
   process.exit(1);
 };
 
-const TOKEN_REGEX =
-  /^(\w+(?:-\w+)(?:-\w+)?)(?:-for-(\w+(?:-\w+)?))?(?:-when-([\w-]+?))?(?:-on-([\w-]+?))?$/i;
+const ALLOWED_CSS_VALUES_IN_CHOICES = /px|none|hsla/;
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in b008efe0cce4e6cffd60e61ff554e5480a400525

design-system/scripts/generate-design-tokens.js
CarlosCortizasCT: cleaner 👌 

Updated in b008efe0cce4e6cffd60e61ff554e5480a400525

design-system/scripts/generate-design-tokens.js
CarlosCortizasCT: Updated in b008efe0cce4e6cffd60e61ff554e5480a400525

kark (approved):

PR #2323, opened 2022-11-08
Export missing prop in `Link` typescript type

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=mcxXD_f)

#### Summary

`Link` component allows for the `onClick` prop but it is not defined in its props typescript type.

## Description

We just add the `onClick` to the `Link` component props typescript definition and also include a couple of tests to verify this prop is called when provided after clicking the component, both when we render a `react-router-dom` link or a native HTML one.

### DIFF
--- a/.changeset/three-poets-rule.md
+++ b/.changeset/three-poets-rule.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/link': patch
+---
+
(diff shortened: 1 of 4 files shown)


emmenko (approved):
Thanks!

kark (approved):

PR #2324, opened 2022-11-08
refactor(design-system): expose reset styles and loading fonts

emmenko:
This PR experiments loading fonts via NPM packages.

We need to do this for a couple of reasons:
1. The new [Inter font](https://rsms.me/inter/) distributed by Google Fonts does not support _italic_, so we need to load it using the NPM library [inter-ui](https://www.npmjs.com/package/inter-ui).
2. For China, we probably can't load fonts directly from Google Fonts API, so we need to pack them with the application anyway.

In addition to that, I wanted to move some of the global styles "closer" to the design system, as some of the styles there are actually related to the basic design. This is now extracted into the `materials/resets.css` file (name can change).

As I was testing things out, I noticed a couple of issues in the theme provider. Happy to discuss this for clarification.


design-system/src/theme-provider.tsx
```diff
@@ -48,7 +48,7 @@ const applyTheme = ({
   const vars = toVars(
     themeOverrides && isObject(themeOverrides)
       ? merge({}, themes.default, themes[validTheme], themeOverrides)
-      : themes[validTheme]
+      : merge({}, themes.default, themes[validTheme])
```
emmenko, line 51: I believe here we need to merge the new theme with the default one, otherwise only the new theme tokens are used.

docs/.storybook/configs/theme-context.js
```diff
@@ -27,11 +30,11 @@ ThemeWrapper.propTypes = {
 const themeParams = [
   {
     name: 'Default Theme',
-    props: { themeName: 'default', themeOverrides: defaultTheme },
```
emmenko, line 30: The `themeOverrides` here should not be using the design tokens variables

emmenko:
Closing as the approach has some drawbacks.

PR #2326, opened 2022-11-09
refactor: font-family tokens and use Google Fonts V2 API

emmenko:
~~Trying to debug VRT issues from #2324~~

This PR adds design tokens for the font family, including the new `Inter` font.

The actual fonts will still be loaded via Google Fonts for now, using the v2 API. This will also need to be changed in the App Kit.


docs/.storybook/preview-head.html
```diff
@@ -1,11 +1,20 @@
-<link href='https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,600,600i,700,700i' rel='stylesheet' type='text/css'></link>
+<link rel="preconnect" href="https://fonts.googleapis.com"></link>
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin></link>
+<link href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,500;0,600;0,700;1,400&display=swap" rel="stylesheet"></link>
```
emmenko, line 3: Using the new font v2 API.

Notice that we now load the following styles: `300, 400, 500, 600, 700` to cover both new and old requirements.

docs/.storybook/preview-head.html
```diff
@@ -1,11 +1,20 @@
...
 <style>
   html,
   body {
+    color: var(--font-color-for-text);
+    font-family: var(--font-family);
```
emmenko, line 8: Using CSS variables here as well.

docs/.storybook/preview-head.html
```diff
@@ -1,11 +1,20 @@
...
     margin: 0;
     padding: 0;
-    font-size: 13px;
-    font-family: 'Open Sans', sans-serif;
+  }
+
+  @supports (font-variation-settings: normal) {
+    html, body {
+      font-family: var(--font-family-variable);
+    }
```
emmenko, line 17: This enables the [new variable fonts](https://fonts.google.com/knowledge/introducing_type/introducing_variable_fonts).

emmenko:
The VRTs seem to be somehow flaky in regards to rendering fonts. We have the same issue in other PRs, which hints to this being an unrelated issue with the changes but rather with Percy.

Have a look at the VRTs anyway but I think we can approve them and proceed with this.

design-system/materials/internals/definition.yaml
```diff
@@ -278,14 +288,6 @@ decisionGroupsByTheme:
           description: ''
           choice: color-surface
 
-    fontFamily:
-      label: Font family
-      prefix: font-family
-      decisions:
-        font-family:
-          description: ''
-          choice: font-family-body
```
emmenko, line 287: This decision group does not make sense to keep, as we have and use the choices (also for the theme overrides).

design-system/materials/internals/definition.yaml
```diff
@@ -253,6 +260,9 @@ decisionGroupsByTheme:
       label: Font Colors
       prefix: font-color
       decisions:
+        font-color-for-text:
+          description: ''
+          choice: color-solid
```
emmenko, line 265: Just to use a design token also for the default text color.

CarlosCortizasCT:
> The VRTs seem to be somehow flaky in regards to rendering fonts. We have the same issue in other PRs, which hints to this being an unrelated issue with the changes but rather with Percy.
> 
> Have a look at the VRTs anyway but I think we can approve them and proceed with this.

So I compared some visual tests locally with the visual testing app comparing the main branch results with the branch of this PR and I noticed something different with the fonts.
In this branch it seems we are using the variable fonts and, apparently, Helvetica font is loaded instead of Open sans. 
Helvetica seems to be taller but have lower letter spacing.

Screenshot 2022-11-11 at 09 53 16

emmenko:
🤔 That's surprising...

Can you check the same in the other PR #2328 that does not change the font family, as we have the same Percy issues?

CarlosCortizasCT:
> Can you check the same in the other PR https://github.com/commercetools/ui-kit/pull/2328 that does not change the font family, as we have the same Percy issues?

It seems in that PR the problem is that the values for the font-size tokens in the default theme are not correct yet. It is due to the fact we need to use different root font-size by theme. We updated the root font-size by theme but we need to adjust font-size tokens accordingly.
cc/ @chloe0592

emmenko:
Ah, `Open SansVariable` was the way to do it when using the `@fontsource` package. Let me fix that, I've been mixing things up.

emmenko:
Ah much better now 🥳

cc @CarlosCortizasCT

CarlosCortizasCT:
Could it be that new font does not support italic bold?
In the Tests that's the difference I can see.

Screenshot 2022-11-11 at 14 36 05

emmenko:
Let me check the query 🧐

emmenko:
Yes it was indeed missing. Good catch!

emmenko:
🙌

emmenko:
Seems all good now!

CarlosCortizasCT (approved):
🎉

kark (approved):
💯

chloe0592:
@emmenko you can merge this one first then I'll merge mine after :)

PR #2327, opened 2022-11-09
fix(design-system): themes merging and default theme

emmenko:
Extracting some fixes across different PRs around design tokens work.

### DIFF
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -46,9 +46,12 @@ const applyTheme = ({
   }
 
   const vars = toVars(
-    themeOverrides && isObject(themeOverrides)
-      ? merge({}, themes.default, themes[validTheme], themeOverrides)
-      : merge({}, themes.default, themes[validTheme])
+    merge(
+      {},
+      themes.default,
+      themes[validTheme],
+      isObject(themeOverrides) ? themeOverrides : {}
+    )
   );
 
   Object.entries(vars).forEach(([key, value]) => {
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -48,7 +48,7 @@ const applyTheme = ({
   const vars = toVars(
     themeOverrides && isObject(themeOverrides)
       ? merge({}, themes.default, themes[validTheme], themeOverrides)
-      : themes[validTheme]
+      : merge({}, themes.default, themes[validTheme])
   );
 
   Object.entries(vars).forEach(([key, value]) => {
@@ -59,7 +59,7 @@ const applyTheme = ({
 
 type ThemeProviderProps = {
   parentSelector: typeof defaultParentSelector;
-  theme?: string;
+  theme: string;
   themeOverrides?: Record<string, string>;
 };
 
@@ -91,6 +91,7 @@ const ThemeProvider = (props: ThemeProviderProps) => {
 };
 ThemeProvider.defaultProps = {
   parentSelector: defaultParentSelector,
+  theme: 'default',
 };
 
 const useTheme = (parentSelector = defaultParentSelector) => {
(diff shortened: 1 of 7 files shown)


docs/.storybook/configs/theme-context.js
```diff
@@ -27,11 +20,11 @@ ThemeWrapper.propTypes = {
 const themeParams = [
   {
     name: 'Default Theme',
-    props: { themeName: 'default', themeOverrides: defaultTheme },
+    props: { themeName: 'default' },
```
CarlosCortizasCT, line 23: I would also remove the `themeOverrides` prop in the `ThemeProvider` above.

design-system/src/theme-provider.tsx
```diff
@@ -48,7 +48,7 @@ const applyTheme = ({
   const vars = toVars(
     themeOverrides && isObject(themeOverrides)
       ? merge({}, themes.default, themes[validTheme], themeOverrides)
-      : themes[validTheme]
+      : merge({}, themes.default, themes[validTheme])
   );
```
CarlosCortizasCT, line 52: **nit**: I would maybe unify branches:
```
  const vars = toVars(
    merge(
      {},
      themes.default,
      themes[validTheme],
      isObject(themeOverrides) ? themeOverrides : {}
    )
  );
```

docs/.storybook/webpack.config.js
```diff
@@ -11,6 +12,12 @@ const sourceFolders = [
 ];
 
 module.exports = ({ config }) => {
+  config.plugins.push(
+    new DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+      'process.env.VERCEL_ENV': JSON.stringify(process.env.VERCEL_ENV),
```
CarlosCortizasCT, line 18: ❤️

CarlosCortizasCT (approved):

docs/package.json
```diff
@@ -37,7 +37,7 @@
   "devDependencies": {
     "@commercetools-frontend/babel-preset-mc-app": "21.14.3",
     "prismjs": "1.29.0",
-    "webpack": "5.74.0"
+    "webpack": "4.46.0"
```
kark, line 40: Just wondering, is it because `DefinePlugin` doesn't work in webpack 5?

kark (approved):
🎉

docs/package.json
emmenko: No, actually Storybook v5 (the one we're currently using) doesn't use Webpack v5, so we should use also v4.

image

Rhotimee (approved):
Looks great!

emmenko:
@kark @CarlosCortizasCT Could you check the failing VRT and confirm if it's expected or not? 🤔

ddouglasz (approved):
💯

kark:
> @kark @CarlosCortizasCT Could you check the failing VRT and confirm if it's expected or not? 🤔

I don't think that's expected but I can't see any reason for this behavior in this PR. I will investigate further.
The same behavior as here can be observed by commenting out [this line](https://github.com/commercetools/ui-kit/blob/0a42178c2c733cf31120fa208db182399887677e/packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx#L175) so I'm assuming it could be in some way related to [react-textarea-autosize](https://github.com/Andarist/react-textarea-autosize) library

emmenko:
If we are fine with the changes and are aware of it, we can approve them.

I checked the VRT locally and it behaves correctly. Another Percy hiccup? 🤷‍♂️

CarlosCortizasCT:
> @kark @CarlosCortizasCT Could you check the failing VRT and confirm if it's expected or not? 🤔

I also took a look at it and it doesn't seem related to the changes in this PR at all.
I remember seeing this error randomly in other PRs and it seems to me it's a timing issue where the component resize does not happen quick enough to Percy to catch it.

I would merge this PR.

PR #2328, opened 2022-11-09
link-design-tokens

chloe0592:
## Link design token

This is[ the relevant commit](https://github.com/commercetools/ui-kit/commit/3bad720774c8df140f179bd8e9522e8de85b0051) for this PR


emmenko:
@chloe0592 There seems to be some "unexpected" visual changes in the VRTs. Maybe before approving should we double check that these are expected?

chloe0592:
> @chloe0592 There seems to be some "unexpected" visual changes in the VRTs. Maybe before approving should we double check that these are expected?

@emmenko did you mean all the text? Or you've noticed something else, that I missed?
If it's the text - I won't merge anyway this link changes until we have all the typography set. I was planning to merge one into the other and all together to main.

CarlosCortizasCT:
It seems we are including changes from other PRs in this one.
I would suggest to close this one and open a new one from `main` as the change here should be very simple.

chloe0592:
> It seems we are including changes from other PRs in this one. I would suggest to close this one and open a new one from `main` as the change here should be very simple.

@CarlosCortizasCT opened up a new [PR here](https://github.com/commercetools/ui-kit/pull/2330).

PR #2330, opened 2022-11-11
Update `Link` component to support new theme

chloe0592:
## Summary

Refactor `Link` component to use a new design token to control its font size.


emmenko (approved):
Thanks! Don't forget to add a changeset 🤗

chloe0592:
> Thanks! Don't forget to add a changeset 🤗

Thanks for the reminder! 🍫 It's here - [chore: adding changeset](https://github.com/commercetools/ui-kit/pull/2330/commits/c3a8c5efe51c3cda47c4cb0b95eeb655158e230d)

.changeset/famous-ads-count.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/design-system': minor
+'@commercetools-uikit/link': minor
+---
+
+Adding inherit property for the links in the new theme
```
CarlosCortizasCT, line 6: **nit**: perhaps we don't need to be so specific and we can use something like:
```
Add new design token for `Link` component.
```

CarlosCortizasCT (approved):
Thanks 👍

.changeset/famous-ads-count.md
chloe0592: Yes, that sounds better. Thank you. 

- [chore: adding changeset 2](https://github.com/commercetools/ui-kit/pull/2330/commits/2a2fcc35ddd6ace763aa9b854d0e27baec534688)

CarlosCortizasCT: We just need one changeset file so perhaps you can remove the first one (`famous-ads-count`).

emmenko: Also, I think `patch` would also suffice here instead of `minor` as we're simply refactoring some internal stuff (no new features or something similar).

PR #2331, opened 2022-11-11
feat(design-system): export `resets.css` styles

emmenko:
Extracting some of the changes from #2324 

Most of the resets styles are currently [defined in App Kit](https://github.com/commercetools/merchant-center-application-kit/blob/87853f778b03fc86363333e7c1bc1953a7728ef4/packages/application-shell/src/components/application-shell/global-styles.tsx#L53-L98). However, some of them are tightly related to basic styles which are more a concern of the design system.

This PR proposes to expose a `resets.css` file from the `design-system` package that can be used by the consumers of the UI Kit. This should make it easier to manage basic changes in one place.

Happy to hear your thoughts and if this makes sense or not.

### DIFF
--- a/design-system/materials/resets.css
+++ b/design-system/materials/resets.css
@@ -1,0 +1,40 @@
+*,
+*::before,
+*::after {
+  box-sizing: inherit;
+}
+
+html,
+body {
+  margin: 0;
+  padding: 0;
+}
+
+html {
+  box-sizing: border-box;
+}
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  margin: 0;
+}
+
+button {
+  cursor: pointer;
+}
+
+a {
+  color: var(--color-primary);
+  cursor: pointer;
+  text-decoration: none;
+  transition: color 0.2s ease-in;
+}
+
+p {
+  color: var(--font-color-for-text);
+  margin: 0;
+}
--- a/design-system/materials/resets.css
+++ b/design-system/materials/resets.css
@@ -6,6 +6,8 @@
 
 html,
 body {
+  color: var(--font-color-for-text);
+  font-family: var(--font-family);
   margin: 0;
   padding: 0;
 }
@@ -35,6 +37,5 @@ a {
 }
 
 p {
-  color: var(--font-color-for-text);
   margin: 0;
 }
--- a/.changeset/purple-flowers-provide.md
+++ b/.changeset/purple-flowers-provide.md
@@ -1,0 +1,9 @@
+---
+'@commercetools-uikit/design-system': minor
+---
+
+Expose global reset styles from `/materials` entry point.
+
+```js
+import '@commercetools-uikit/design-system/materials/resets.css';
+```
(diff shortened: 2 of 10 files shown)


CarlosCortizasCT:
I agree with you.
Having this new `resets` file in `ui-kit` makes more sense since it's tailoring defaults for component defined here.

chloe0592:
Could I just ask what is the purpose of having these _resets_ styles? Is it just naming of global styling or is there something else that the name 'reset' suggests?

emmenko:
Yeah mostly some basic defaults. For now it's just being backwards compatible with what we already had in App Kit.
In the future, I suppose we can change it or clean it up if we need to.

design-system/materials/resets.css
```diff
@@ -0,0 +1,41 @@
...
+a {
+  color: var(--color-primary);
+  cursor: pointer;
+  text-decoration: none;
+  transition: color 0.2s ease-in;
+}
+
+p {
+  margin: 0;
+}
```
emmenko, line 41: Small change! The paragraph should not have the `color` style assigned as it conflicts with disabled styles.

design-system/materials/resets.css
```diff
@@ -0,0 +1,41 @@
+*,
+*::before,
+*::after {
+  box-sizing: inherit;
+}
+
+html,
+body {
+  color: var(--font-color-for-text);
+  font-family: var(--font-family);
```
emmenko, line 10: I'm going a bit back and forth with this, but I think it kind of makes sense to have this defined here instead of relying on the application/consumer to do so.

Thoughts? @commercetools/shield-team-fe

design-system/materials/resets.css
emmenko: Could be although it seemed an edge case according to the VRTs: the `LocalizedRichTextInput` in read-only mode.

See https://percy.io/commercetools-GmbH/ui-kit/builds/22876206/changed/1276880837?browser=chrome&browser_ids=31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1024&widths=1024

Do you have other suggestions?

CarlosCortizasCT: Ok, so then this is an actual bug in `app-kit` (I checked it locally) and this change would fix it.

Thanks.

design-system/materials/resets.css
```diff
@@ -0,0 +1,41 @@
+*,
+*::before,
+*::after {
+  box-sizing: inherit;
+}
```
CarlosCortizasCT, line 5: I tracked the currently failing Percy test related to `CollapsiblePanel` component down to this new rules. I makes the icon a little bigger and it also moves the title to the right.

Perhaps this is intended since we were actually using it like this in the applications (through `app-kit`) so maybe we can approve those changes in Percy.

emmenko: Yeah I think we can approve that.

CarlosCortizasCT (approved):
💯

PR #2332, opened 2022-11-15
Export utility helpers to transform from design tokens to CSS vars

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=ZwmdL-5)

### Summary

Export utility helpers to transform from design tokens to CSS vars

## Description

We realized there are a couple of consumer repositories that are implementing transformer utilities to convert from design tokens map to CSS vars.

Having an input design tokens map like this:
```
{
  colorPrimary: '#00b39e',
  borderRadius4: '4px',
}
```

We might want to transform that object to be a map where keys are CSS var names and values are the CSS values:
```
{
  '--color-primary': '#00b39e',
  '--border-radius-4': '4px',
}
```

Another option would be the transform the values to be CSS var references to be used:
```
{
  colorPrimary: 'var(--color-primary, #00b39e)',
  borderRadius4: 'var(--border-radius-4, 4px)',
}
```

### DIFF
--- a/.changeset/few-toes-roll.md
+++ b/.changeset/few-toes-roll.md
@@ -1,0 +1,32 @@
+---
+'@commercetools-uikit/design-system': patch
+---
+
+New helper functions to transform from design tokens to CSS vars.
+
+Let's say we have an object with design tokens data like this:
+```
+{
+  colorPrimary: '#00b39e',
+}
+```
+
+We can transform the object to use CSS var names keys:
+```
+transformTokensToCssVarsValues(tokens);
+
+// Output
+{
+  '--color-primary': '#00b39e',
+}
+```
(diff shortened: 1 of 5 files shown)


emmenko (approved):
Nice! 🥳

emmenko:
PS: changeset 😉

CarlosCortizasCT:
> PS: changeset 😉

I can't understand why I always forget about the changeset 🤦 

Added here: 64b0435aeb48040cc7ad5d34819b5e2859bd46f7

emmenko:
> I can't understand why I always forget about the changeset 

You will get used to it sooner or later 😜

PR #2333, opened 2022-11-16
Buttons font size design tokens adjustments

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=aWAvxJj)

#### Summary

Buttons font size design tokens adjustments

## Description

Buttons font size in new the theme were calculated with the previous root font size which changed in the Typography PR so an adjustment was needed.
Also, `SecondayButton` was using an incorrect token.

On the other side, the font-size design tokens used in Typography for the current theme are not exactly the same we are using right now in our latest published `ui-kit` version so I updated it to match.


emmenko (approved):
Thanks

PR #2336, opened 2022-11-17
Design tokens adjustments

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=5RM0TQg)

### Summary

A couple of design tokens leftovers from previous work.

#### Typography headers

We implemented the `H1` header to have the same font size of `H2` (`Text.Headline` component).
It was supposed to be a temporary measure to adjust pages titles: currently we are not using `H1` but `H2` for that, but we want to change it. We need titles to use `H1` as they are semantically and from accessibility point of view more correct.
In the meantime in the new theme, we would like to make `H2` to use the same size `H1` (`24px`). Once we update all `H2`s to be `H1` as expected, we will come back and update `H2` in the new theme with its final font size (`20px`).

#### FlatButton
The label of this button was using a `Text.Body` component but that does not adhere to the new font size we want for all the buttons in the new theme. Actually, all buttons use `AccessibleButton` under the hood and it sets up font CSS rules common to all buttons so we don't need to use `Text.Body`.


emmenko (approved):
Thanks!

PR #2340, opened 2022-11-21
Fix typography headings tokens

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=KfNUZ-5)

#### Summary

Typography headings design tokens had incorrect values.

## Description

This is a leftover from #2336

We updated `font-size` but we forgot about `line-height` and `font-weight` in the new theme.


emmenko (approved):
Thanks!

CarlosCortizasCT:
There's a weird error in Vercel deployment job and it seems it has nothing to do with changes in this PR. Actually, building locally both library and Storybook runs with no errors.

It seems Vercel is using node v18 by default [since last Friday](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version) and maybe the problem is related to that.
I'll try to reproduce the error locally using the same node version.

emmenko:
> It seems Vercel is using node v18 by default [since last Friday](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version) and maybe the problem is related to that.

This is only for new projects. We still use v16 in ui-kit.
image

However, it does say that it will look at the `engine` version

```
Warning: Detected "engines": { "node": ">=16" } in your `package.json` that will automatically upgrade when a new major Node.js Version is released. Learn More: http://vercel.link/node-version
```

image

We should adjust the `engine` version then and if the error relates to v18 we should exclude that.

CarlosCortizasCT:
> > It seems Vercel is using node v18 by default [since last Friday](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version) and maybe the problem is related to that.
> 
> This is only for new projects. We still use v16 in ui-kit. image
> 
> However, it does say that it will look at the `engine` version
> 
> ```
> Warning: Detected "engines": { "node": ">=16" } in your `package.json` that will automatically upgrade when a new major Node.js Version is released. Learn More: http://vercel.link/node-version
> ```
> 
> image
> 
> We should adjust the `engine` version then and if the error relates to v18 we should exclude that.

Restricting the version to `16` makes the deployment to work again.

emmenko (approved):

PR #2342, opened 2022-11-22
Fix `Tag` component design tokens values in new theme

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=SI3CGuH)

## Summary

Fix `Tag` component design tokens values in new theme

## Description

The values for the background color used in the new theme were invalid. Actually, the background color should not change by theme.

Also, I found a visual bug with the right border of the component when using type warning, onClick and the mouse over the component:
### Before
image

### After
image

### DIFF
--- a/packages/components/tag/src/tag-body.tsx
+++ b/packages/components/tag/src/tag-body.tsx
@@ -99,6 +99,9 @@ const TagBody = (props: TTagBodyProps) => {
                 right: -1px;
                 content: '';
                 background-color: ${designTokens.borderColorForTagWhenFocused};
+                background-color: ${props.type === 'warning'
+                  ? designTokens.borderColorForTagWarning
+                  : designTokens.borderColorForTagWhenFocused};
                 width: 1px;
                 height: 100%;
               }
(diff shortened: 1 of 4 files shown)


packages/components/tag/src/tag-body.tsx
```diff
@@ -99,6 +99,9 @@ const TagBody = (props: TTagBodyProps) => {
                 right: -1px;
                 content: '';
                 background-color: ${designTokens.borderColorForTagWhenFocused};
+                background-color: ${props.type === 'warning'
+                  ? designTokens.borderColorForTagWarning
+                  : designTokens.borderColorForTagWhenFocused};
```
kark, line 104: Thanks for the fix 🎉 A this fixes the bug, I just wanted to point out that the `border-color` is duplicated

CarlosCortizasCT: Ouch! I forgot to remove the previous one 🤦 
Thanks for noticing it 👍 

Removed in this commit: 57c0076f15e130aa4916bdf209c525da94f9669a

CarlosCortizasCT:
@FilPob just mentioning you so you are aware of this.

emmenko (approved):
Thanks! Was this a regression?

kark (approved):

CarlosCortizasCT:
> Thanks! Was this a regression?

If you're asking about the right border color, no, it was already there before we started working on theming support.

PR #2343, opened 2022-11-22
Update spacings

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=Y7FYizT)

#### Summary

Update design tokens and some components spacing related values.

## Description

We're adding two new values for the `scale` property in `Stack` and `Inline` spacing components: `xxl` (48px) and `xxxl` (64px).

Also, we are updating the paddings in `Tag` and `Spacing` components in the new theme.


design-system/materials/internals/definition.yaml
```diff
@@ -165,11 +165,20 @@ choiceGroupsByTheme:
...
         spacing-l: 24px
         spacing-xl: 32px
+        # New tokens
+        spacing-10: 4px
+        spacing-20: 8px
+        spacing-30: 16px
+        spacing-40: 24px
+        spacing-50: 32px
+        spacing-60: 48px
+        spacing-70: 64px
```
CarlosCortizasCT, line 181: Current t-shirt spacing scale is marked as deprecated as we are changing it for a numbered scale which allows for more flexibility.
The new scale is used in the components updated in this PR.
A follow-up PR will be created to migrate from the t-shirt scale to the numbered one in all other components (didn't want to introduce more noise in this PR).

design-system/materials/internals/definition.yaml
```diff
@@ -543,6 +554,15 @@ decisionGroupsByTheme:
...
           choice: font-weight-400
 
+    paddings:
+      label: Paddings
+      prefix: padding
+      decisions:
+        padding-for-stamp:
+          choice: 'var(--spacing-xs) var(--spacing-s)'
+        padding-for-tag:
+          choice: '5px var(--spacing-s)'
```
CarlosCortizasCT, line 564: We're waiting for @FilPob to provide final values for these components.

design-system/src/index.ts
```diff
@@ -5,6 +5,7 @@ export {
   themes,
 } from './design-tokens';
 export { ThemeProvider, useTheme } from './theme-provider';
+export type { ThemeName } from './theme-provider';
```
CarlosCortizasCT, line 8: Since we will need to modify `ui-kit` consumer components code based on the selected theme, I think it would be a good idea to export this type for better typing.

design-system/src/theme-provider.tsx
```diff
@@ -92,17 +92,20 @@ ThemeProvider.defaultProps = {
   theme: 'default',
 };
 
-const useTheme = (parentSelector = defaultParentSelector) => {
-  const [theme, setTheme] = useState<string>('default');
+const useTheme = (parentSelector = defaultParentSelector): ThemeName => {
```
CarlosCortizasCT, line 95: I'm starting to think that current return type of this hook (a `string`) is too restrictive. If we need to return something else from the hook in the future, consumers would be affected.
Perhaps it would be better to return and object with the selected theme as a property so it would be easier to add something in the future (I'm thinking in an imperative function to update the theme, for instance).

What do you think?

packages/components/tag/src/tag-body.tsx
```diff
@@ -72,7 +72,10 @@ const getContentWrapperStyles = (props: TTagBodyProps) => {
 };
 
 const TagBody = (props: TTagBodyProps) => {
+  const currentTheme = useTheme();
   const textTone = props.isDisabled ? 'secondary' : undefined;
+  const TextComponent = currentTheme === 'default' ? Text.Detail : Text.Body;
```
CarlosCortizasCT, line 77: We need to use a different font-size per theme so we read current theme to calculate which `Text` component we must use.

FilPob (approved):

design-system/src/theme-provider.tsx
emmenko: I guess it depends how we envision the usage of it. Also, I'm wondering about the `useEffect` implementation, as it seems it only runs once but what about when the theme changes due to other factors?

Maybe it's too soon to think about it (for the redesign work) and it's more interesting when we might have multiple themes to choose from.

packages/components/tag/src/tag-body.tsx
emmenko: Maybe leave a comment that this would be removed at some point when we make the new theme the default one?

emmenko (approved):
Thanks, looks good to me 🥳

Rhotimee (approved):
Looks nice

kark (approved):

design-system/src/theme-provider.tsx
CarlosCortizasCT: I think having a string as a return type is too restricting. I don't know how this will evolve in the future, but having an object as return type would make potential changes easier to digest for consumers.

Regarding the effect, yes it was only running once as we didn't have requirements to have it in sync. But now I already found an issue with this (when we need to re-render a component using `useTheme`, for example the `Tag` component now).
I already have an idea about how to tackle that issue but I prefer to open a separate PR for that.

PR #2345, opened 2022-11-23
Update `Card` inset scale property allowed values

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=s0Wazr2)

#### Summary

Update `Card` inset scale property allowed values

## Description

`Card` component uses internally the `Inset` component to handle its paddings. The `insetScale` property is used to configure how much padding space we want in the component.
Although the `Inset` component allows for `xs`, `s`, `m`, `l` and `xl` values, the `Card` component only allowed for `s` and `m` at typescript definition level.
Since we will be needing `Cards` with bigger paddings, we're adding `l` and `xl` values as also allowed.


kark (approved):
👍

emmenko (approved):
Thanks!

tdeekens (approved):
👯

PR #2347, opened 2022-11-29
Add new tokens for Input component

Rhotimee:
#### Summary
Add design tokens for Input component


design-system/materials/internals/definition.yaml
```diff
@@ -409,6 +409,19 @@ decisionGroupsByTheme:
     # We don't have font families in choices yet, so we can't have decisions for
     # them
 
+    borderWidth:
+      label: Border width
+      prefix: border-width
+      decisions:
+        border-width-for-input:
+          choice: '1px'
```
CarlosCortizasCT, line 417: Perhaps we can have new choice tokens with `border-widths` similar to what we already have with `border-radius`.

design-system/materials/internals/definition.yaml
```diff
@@ -427,7 +440,7 @@ decisionGroupsByTheme:
           choice: color-error
         font-color-for-input-when-readonly:
           description: ''
-          choice: color-neutral-60
+          choice: color-neutral-40
```
CarlosCortizasCT, line 443: Here we are changing the value for the current theme and we're not supposed to change anything in that theme.
Is this a new UX requirement?

design-system/materials/internals/definition.yaml
```diff
@@ -757,6 +789,13 @@ decisionGroupsByTheme:
         font-weight-for-button:
           choice: font-weight-500
 
+    fontColors:
+      label: Font Colors
+      prefix: font-color
+      decisions:
+        font-color-for-input-when-readonly:
+          choice: color-neutral-60
```
CarlosCortizasCT, line 797: Shouldn't this value be `color-neutral-40`?

packages/components/inputs/input-utils/src/styles.ts
```diff
@@ -42,29 +42,39 @@ const getInputFontColor = (props: TInputProps) => {
...
-    min-height: ${designTokens.sizeHeightInput};
+    height: ${designTokens.heightForInput};
+    min-height: ${designTokens.heightForInput};
     opacity: ${props.isDisabled || props.disabled
       ? '1'
       : 'unset'}; /* fix for mobile safari */
     outline: none;
     overflow: hidden;
-    padding: 0 ${designTokens.spacing20};
+    padding: 0 ${designTokens.spacingM};
```
CarlosCortizasCT, line 77: I guess we need a new decision token here.
In the default theme we should keep using the `spacing20` (8px) choice token but in the new theme we need to use `spacing30` instead.

We could have a new decision token like this (in the `definitions.yaml` file):
```
// Default theme
paddingForInput: spacing-20

// New theme
paddingForInput: spacing-30

**Review by emmenko — COMMENTED** [review]
Same questions as Carlos, otherwise is looking good!

File: packages/components/inputs/input-utils/src/styles.ts (hunk 4 above)
Comment (Rhotimee) [reply packages/components/inputs/input-utils/src/styles.ts]: Updated here [9bbdd7f](https://github.com/commercetools/ui-kit/pull/2347/commits/9bbdd7ff1f26ae4896b6164aeee65a48447b15c7). Thanks

File: design-system/materials/internals/definition.yaml (hunk 1 above)
Comment (Rhotimee) [reply design-system/materials/internals/definition.yaml]: Updated here [a6465b9](https://github.com/commercetools/ui-kit/pull/2347/commits/a6465b999def586cb373a8148e735873420465ed)

File: design-system/materials/internals/definition.yaml (hunk 2 above)
Comment (Rhotimee) [reply design-system/materials/internals/definition.yaml]: Updated [38d3d54](https://github.com/commercetools/ui-kit/pull/2347/commits/38d3d543be3051df492edcd0d5722413e61a5499)

File: design-system/materials/internals/definition.yaml (hunk 3 above)
Comment (Rhotimee) [reply design-system/materials/internals/definition.yaml]: Updated [38d3d54](https://github.com/commercetools/ui-kit/pull/2347/commits/38d3d543be3051df492edcd0d5722413e61a5499)

**Comment by CarlosCortizasCT** [comment]
Checking the preview deployment I think I found some things that are not as in the designs:
* Border color on hover: it's green when it should be grey
* Border size when focused: the border is too big (I think it's due to box-shadow, which I guess we should not use in the new theme)
* Background color on focus: it's grey when it should be white
* Background color when disabled: it's blueish when it should be grey

**Comment by Rhotimee** [comment]
@CarlosCortizasCT, thanks for catching that. I only updated the token based on the table. 
I've now implemented the rest of the token based on the design. 
Please look at the [2356364](https://github.com/commercetools/ui-kit/pull/2347/commits/23563642d34dc2498fd3539ce6a11f9cbde51c96). Thanks.

File: design-system/materials/internals/definition.yaml [hunk 5]
```diff
@@ -799,6 +868,8 @@ decisionGroupsByTheme:
           choice: shadow-0
         shadow-for-card-when-raised:
           choice: '0 1px 5px 0 rgba(0, 0, 0, 0.05)'
+        shadow-for-input-when-focused:
+          choice: none
```
CarlosCortizasCT, line 872: You might want to use the `shadow-0` choice token instead.

CarlosCortizasCT:
I think there's a different behaviour in the new theme when handling `hover` state as it seems it overrides `focus` state and that's something that do not happen in the default theme.
I think `focus` state should have more priority than `hover`

This is how it behaves in the new theme:
[Components _ Inputs-TextInput-New_Theme.webm](https://user-images.githubusercontent.com/97907068/205252804-91657433-e746-4f28-b19f-43a59d098306.webm)

And this is how it works in the default theme:
[Components _ Inputs-TextInput-Default_theme.webm](https://user-images.githubusercontent.com/97907068/205252864-6ae46dee-b351-45f9-8a56-0120420ccbdb.webm)


We also need to resolve the conflicts with the target branch.

chloe0592:
> Checking the preview deployment I think I found some things that are not as in the designs:
> 
> * Border color on hover: it's green when it should be grey
> * Border size when focused: the border is too big (I think it's due to box-shadow, which I guess we should not use in the new theme)
> * Background color on focus: it's grey when it should be white
> * Background color when disabled: it's blueish when it should be grey


I'd just add also the border color when it's error state. (When it's unhovered it turns green even though it's still erred).
Probably should be gray? (Like in the default theme) 
And 2px instead of 1px?

https://user-images.githubusercontent.com/52276952/205263042-a8fae99d-b2ae-4851-b40c-7faede0c2ee9.mov

chloe0592:
> I think there's a different behaviour in the new theme when handling `hover` state as it seems it overrides `focus` state and that's something that do not happen in the default theme. I think `focus` state should have more priority than `hover`
> 
> This is how it behaves in the new theme: [Components _ Inputs-TextInput-New_Theme.webm](https://user-images.githubusercontent.com/97907068/205252804-91657433-e746-4f28-b19f-43a59d098306.webm)
> 
> And this is how it works in the default theme: [Components _ Inputs-TextInput-Default_theme.webm](https://user-images.githubusercontent.com/97907068/205252864-6ae46dee-b351-45f9-8a56-0120420ccbdb.webm)
> 
> We also need to resolve the conflicts with the target branch.

I would honestly expect, that after you click on the input that the hover state is then disabled. And focus state stays, nbut not both at the same time. But maybe that's just weird to me...

design-system/materials/internals/definition.yaml
Rhotimee: Thanks @CarlosCortizasCT, Updated [4fd6a93](https://github.com/commercetools/ui-kit/pull/2347/commits/4fd6a934c68d96a095d68c7f4d60766a6590df94)

CarlosCortizasCT:
This is a question for @FilPob regarding the _hover_ state when the input has an error on warning state: currently we change the border and the background colors but not the font color. I find it a little weird since I would expect at least the border to stay in the state color.
Just wanted to double check if the current UI for this combinations are the expected ones.

[Components _ Inputs - TextInput ⋅ Storybook.webm](https://user-images.githubusercontent.com/97907068/205607529-e4ce2415-30c3-4ee7-9664-cc6e40bed77e.webm)

FilPob:
> > I suppose is due to the input border size change between the default state and any other state. The default state should probably have the border with the same size.
> 
> Yes, this is due to the border-width change between 1px and 2px. @FilPob, should the states have the same border-width?

yes that's correct. Border-width for focus, error and warning should be 2px. Other states have 1px. But that little "resize/ content jump" is not intended. Wondering why that is because right now we also have 1px border per default and 3px when focused. And the content stays at the same spot even though the border-width changes when focused

FilPob:
> This is a question for @FilPob regarding the _hover_ state when the input has an error on warning state: currently we change the border and the background colors but not the font color. I find it a little weird since I would expect at least the border to stay in the state color. Just wanted to double check if the current UI for this combinations are the expected ones.
> 
>  Components._.Inputs.-.TextInput.Storybook.webm

Oh yeah that's right. The border-color and width should be given by the state. Hovering the Input should only affect the background color

Rhotimee:
> > > I suppose is due to the input border size change between the default state and any other state. The default state should probably have the border with the same size.
> > 
> > 
> > Yes, this is due to the border-width change between 1px and 2px. @FilPob, should the states have the same border-width?
> 
> yes that's correct. Border-width for focus, error and warning should be 2px. Other states have 1px. But that little "resize/ content jump" is not intended. Wondering why that is because right now we also have 1px border per default and 3px when focused. And the content stays at the same spot even though the border-width changes when focused

The difference is that, in the default theme, we are not changing the width of the border instead, added a `box-shadow` for the focus state.

Rhotimee:
The [resizing issue](https://github.com/commercetools/ui-kit/pull/2347#issuecomment-1337003793) and [border-color issue](https://github.com/commercetools/ui-kit/pull/2347#issuecomment-1337051968) has been fixed in this [cbf6693](https://github.com/commercetools/ui-kit/pull/2347/commits/cbf6693150769649e7f40974ac37a7dcc1c25c98)

FilPob (approved):

emmenko:
I noticed a small regression in other input components, for example when hovering the borders "disappear"

image

emmenko:
Please check all other input components for the default theme that there are no unwanted regressions.

FilPob:
@emmenko very good catch. The issue you mentioned above occurs in: **Date, Time, Datetime, Daterange inputs**

Another issue I found is:
- There is no hover effect when going trough the items in a select menu... Present in all **select inputs and moneyinput**

Rhotimee:
Thanks, @emmenko, @FilPob I'm looking into them.

design-system/materials/internals/definition.yaml
```diff
@@ -65,6 +65,7 @@ choiceGroupsByTheme:
         color-solid-05: 'hsl(0deg 0% 10% / 5%)'
         color-solid-10: 'hsl(0deg 0% 10% / 10%)'
         color-surface: '#fff'
+        color-transparent: 'transparent'
```
CarlosCortizasCT, line 68: Not sure that we need this color. I think we should use `color-surface` instead. That would help if we implement a dark theme in the future.

CarlosCortizasCT (commented):
Visually I see everything ok expect for the issue @FilPob [mentioned](https://github.com/commercetools/ui-kit/pull/2347#issuecomment-1340889131) about the dropdowns

CarlosCortizasCT (changes requested):
I used "Request changes" so we don't merge the PR before reviewing the issue Filip pointed out.

design-system/materials/internals/definition.yaml
Rhotimee: My initial idea was to use `color-surface` but it introduced a regression. See the recording below. 
https://user-images.githubusercontent.com/6726467/207265047-53512bf1-7dd2-4151-befe-7f79f343a363.mov

emmenko: Yeah cause the border is defined in the parent container, not on the input.

I wouldn't add a new choice for that though, we can allow the values like `unset` or `transparent` to be used in the decision groups

Rhotimee: I updated to use `unset` instead - [302da5f](https://github.com/commercetools/ui-kit/pull/2347/commits/302da5ff5cfa26b531a68cd76fbfa6af3769387e)

Rhotimee:
@FilPob @CarlosCortizasCT, the select input hover issue was fixed here- [0518c8c](https://github.com/commercetools/ui-kit/pull/2347/commits/0518c8c089a480f74e05271f98774163d12f5679)

Please take a look again. Thanks.

FilPob:
Looks good now from my side

PR #2349, opened 2022-11-29
Fix `useTheme` hook

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=a27jD5v)

#### Summary

`useTheme` hook returns the selected theme but the value gets out of sync if it's changed.

## Description

Initially we thought the theme would not needed to be changed for the purpose of the migration but as we started working on it, we realized it's quite useful and also that's something we would like to have in the future (if we support official themes).

The current selected theme is not stored in a context state but in the DOM element where we set the design tokens up.
The `useHook` was just reading that value after first render so consumers can access to that value, but if the theme is changed, the hook keeps the old value.

We didn't realize about this problem as we weren't using the hook in our components yet, but as we start needing it to refactor the code of some components to adjust its UI for the new theme, we now see the problem.
Actually, the current problem is that the theme returned from the hook never changes, so the components using the hook does not get re-rendered when the theme changes in the page.

Since the value is stored in an HTML DOM node, I opted for using a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to get a notification whenever the attribute where we store the theme value is modified.

Also, I decided to change the return type of the hook: instead of returning a string with the current theme name, we now return an object containing a property with that value. This would make potential hook updates easier to digest by its consumers (currently we are using it almost nowhere, so this is a good time to do it).

### DIFF
--- a/.changeset/heavy-hornets-pretend.md
+++ b/.changeset/heavy-hornets-pretend.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/design-system': patch
 ---
 
-Fix `useHook` to keep the returned current selected theme name in sync with the provider.
+Fix `useTheme` hook to keep the returned current selected theme name in sync with the provider.
--- a/.changeset/heavy-hornets-pretend.md
+++ b/.changeset/heavy-hornets-pretend.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/design-system': patch
+---
+
+Fix `useHook` to keep the returned current selected theme name in sync with the provider.
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
   theme: 'default',
 };
 
-const useTheme = (parentSelector = defaultParentSelector): ThemeName => {
+type TUseThemeResult = {
+  currentTheme: ThemeName;
+};
+const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
   const [theme, setTheme] = useState<ThemeName>('default');
   const parentSelectorRef = useRef(parentSelector);
+  const observerRef = useRef(
+    new MutationObserver((mutationList, _observer) => {
+      // Since we are only observing the theme DOM node for changes in its
+      // `data-theme` attribute (configured below when calling `observe` function),
+      // we will receive a single element in the list
+      setTheme(
+        (mutationList[0].target as HTMLElement).dataset.theme as ThemeName
+      );
+    })
+  );
 
   // If we use 'useLayoutEffect' here, we would be trying to read the
   // data attribute before it gets set from the effect in the ThemeProvider
   useEffect(() => {
-    setTheme(
-      (parentSelectorRef.current()?.dataset.theme as ThemeName) || 'default'
-    );
+    const themeDomNode = parentSelectorRef.current();
+    const observer = observerRef.current;
+
+    if (themeDomNode) {
+      // We need to read the current theme after the provider is rendered
+      // to have the actual selected theme (calulated client-side) in the
+      // hook local state
+      setTheme((themeDomNode.dataset.theme as ThemeName) || 'default');
+
+      // We observe the theme DOM node for changes in its `data-theme`
+      // attribute, which is the one we update in the `applyTheme` function
+      observer.observe(themeDomNode, {
+        attributes: true,
+        attributeFilter: ['data-theme'],
+      });
+    }
+
+    return () => observer.disconnect();
   }, []);
 
-  return theme;
+  return { currentTheme: theme };
 };
 
 export { ThemeProvider, useTheme };
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -93,7 +93,7 @@ ThemeProvider.defaultProps = {
 };
 
 type TUseThemeResult = {
-  currentTheme: ThemeName;
+  theme: ThemeName;
 };
 const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
   const [theme, setTheme] = useState<ThemeName>('default');
@@ -135,7 +135,7 @@ const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
     return () => observer.disconnect();
   }, []);
 
-  return { currentTheme: theme };
+  return { theme };
 };
 
 export { ThemeProvider, useTheme };
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -117,7 +117,7 @@ const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
 
     if (themeDomNode) {
       // We need to read the current theme after the provider is rendered
-      // to have the actual selected theme (calulated client-side) in the
+      // to have the actual selected theme (calculated client-side) in the
       // hook local state
       setTheme((themeDomNode.dataset.theme as ThemeName) || 'default');
 
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -119,7 +119,10 @@ const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
       // We need to read the current theme after the provider is rendered
       // to have the actual selected theme (calculated client-side) in the
       // hook local state
-      setTheme((themeDomNode.dataset.theme as ThemeName) || 'default');
+      const nextTheme = themeDomNode.dataset.theme as ThemeName;
+      if (nextTheme) {
+        setTheme(nextTheme);
+      }
 
       // We observe the theme DOM node for changes in its `data-theme`
       // attribute, which is the one we update in the `applyTheme` function
--- a/design-system/src/theme-provider.visualroute.jsx
+++ b/design-system/src/theme-provider.visualroute.jsx
@@ -16,7 +16,7 @@ export const routePath = '/theme-provider';
 const parentSelector = (id) => () => document.getElementById(id);
 
 const DummyComponent = (props) => {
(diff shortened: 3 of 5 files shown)


.changeset/heavy-hornets-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/design-system': patch
+---
+
+Fix `useHook` to keep the returned current selected theme name in sync with the provider.
```
emmenko, line 5: What's `useHook`? 🤔

CarlosCortizasCT: Fixed in this commit: d0f30d5cfefdd6fdb2ceb0de59b65b47e0a78f62

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
 
-const useTheme = (parentSelector = defaultParentSelector): ThemeName => {
+type TUseThemeResult = {
+  currentTheme: ThemeName;
+};
+const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
   const [theme, setTheme] = useState<ThemeName>('default');
   const parentSelectorRef = useRef(parentSelector);
+  const observerRef = useRef(
+    new MutationObserver((mutationList, _observer) => {
```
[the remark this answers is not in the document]
CarlosCortizasCT: I think this is a great suggestion but I would like to do it in a follow-up PR since I'll be introducing something new and I think the code from app-kit should be updated. I expect some discussion around it so I think it's better to do it as a follow-up.

kark (commented):
Great idea with using the `MutationObserver` 🙌 
I had some problems checking if the hook works properly 🤔

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
   useEffect(() => {
-    setTheme(
-      (parentSelectorRef.current()?.dataset.theme as ThemeName) || 'default'
-    );
+    const themeDomNode = parentSelectorRef.current();
+    const observer = observerRef.current;
+
+    if (themeDomNode) {
+      // We need to read the current theme after the provider is rendered
+      // to have the actual selected theme (calulated client-side) in the
```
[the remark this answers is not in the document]
CarlosCortizasCT: Fixed in this commit: 0defdc09af5f1ddcbbeda312f4ebc11d40bffc9d

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
+      // hook local state
+      setTheme((themeDomNode.dataset.theme as ThemeName) || 'default');
+
+      // We observe the theme DOM node for changes in its `data-theme`
+      // attribute, which is the one we update in the `applyTheme` function
```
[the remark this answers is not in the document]
kark: Thank Carlos for your reply, and sorry. My bad. What got me confused was the `DummyComponent` implementation in our `ThemeProvider` visual tests (that I think I authored 😅), where I suppose we have  a bug:

https://github.com/commercetools/ui-kit/blob/0defdc09af5f1ddcbbeda312f4ebc11d40bffc9d/design-system/src/theme-provider.visualroute.jsx#L19

As I see it that should be:
```jsx
const { currentTheme } = useTheme(
    props.parentId
      ? parentSelector(props.parentId)
      : () => document.querySelector(':root') // or imported `defaultParentSelector`
  );
```
As we're not passing any `parentId` to the `<DummyComponent>`:

https://github.com/commercetools/ui-kit/blob/0defdc09af5f1ddcbbeda312f4ebc11d40bffc9d/design-system/src/theme-provider.visualroute.jsx#L47-L49 

a wrong `parentSelector` is passed to `useTheme` inside the `<DummyComponent>`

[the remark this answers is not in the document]
CarlosCortizasCT: Thanks for the explanation, Kacper 👍 

I applied your suggested update in this commit: 7febc6fd9f2195f84dfb456d0683389eae4fa302

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
   const [theme, setTheme] = useState<ThemeName>('default');
   const parentSelectorRef = useRef(parentSelector);
+  const observerRef = useRef(
+    new MutationObserver((mutationList, _observer) => {
+      // Since we are only observing the theme DOM node for changes in its
+      // `data-theme` attribute (configured below when calling `observe` function),
+      // we will receive a single element in the list
+      setTheme(
+        (mutationList[0].target as HTMLElement).dataset.theme as ThemeName
+      );
```
emmenko, line 108: Wondering if we should access these properties in a more safer way.

For example, looking at the [MutationObserver docs](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver#the_callback_function) they iterate on the `mutationList` and check the mutation `type`.

What do you think?

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
-      (parentSelectorRef.current()?.dataset.theme as ThemeName) || 'default'
-    );
+    const themeDomNode = parentSelectorRef.current();
+    const observer = observerRef.current;
+
+    if (themeDomNode) {
+      // We need to read the current theme after the provider is rendered
+      // to have the actual selected theme (calculated client-side) in the
+      // hook local state
+      setTheme((themeDomNode.dataset.theme as ThemeName) || 'default');
```
emmenko, line 122: The `default` value is already set in the theme state. Maybe we can simply check if the theme is defined, otherwise we skip updating the state.

```suggestion
      const nextTheme = themeDomNode.dataset.theme as ThemeName;
      if (nextTheme) {
        setTheme(nextTheme);
      }
```

design-system/src/theme-provider.visualroute.jsx
```diff
@@ -16,7 +16,11 @@ export const routePath = '/theme-provider';
 const parentSelector = (id) => () => document.getElementById(id);
 
 const DummyComponent = (props) => {
-  const theme = useTheme(parentSelector(props.parentId));
+  const { currentTheme } = useTheme(
+    props.parentId
+      ? parentSelector(props.parentId)
+      : () => document.documentElement
```
emmenko, line 22: Do we need this given that `useTheme` has a `defaultParentSelector` function?

```suggestion
      : undefined
```

design-system/src/theme-provider.tsx
```diff
@@ -92,19 +92,47 @@ ThemeProvider.defaultProps = {
...
+        attributes: true,
+        attributeFilter: ['data-theme'],
+      });
+    }
+
+    return () => observer.disconnect();
   }, []);
 
-  return theme;
+  return { currentTheme: theme };
```
emmenko, line 135: Nit: any particular reason for naming this `currentTheme` instead of `theme`?

emmenko (commented):
Looking good, I left some suggestions 🤗

design-system/src/theme-provider.tsx
CarlosCortizasCT: We can do that but I thought is not necessary since we are observing only a change on the `data-theme` attribute, we are not supposed to receive another mutations.

design-system/src/theme-provider.tsx
CarlosCortizasCT: My reasoning is that theme can potentially be changed on runtime so I thought having something like "current" or "selected" in the property name makes it clear the value is for the active theme.

design-system/src/theme-provider.tsx
CarlosCortizasCT: Yes, that's a great idea.

Updated in this commit: bc1aca29d24e65d565b04c30f86c69a4872c8ef8

design-system/src/theme-provider.visualroute.jsx
CarlosCortizasCT: Updated in this commit: bc1aca29d24e65d565b04c30f86c69a4872c8ef8

design-system/src/theme-provider.tsx
emmenko: I guess it's fine, just wanted to double check and be more on the safe side.

design-system/src/theme-provider.tsx
emmenko: Trying to think of similar cases that use this kind of naming convention 🤔

As far as it concerns the hook, given that it's named `useTheme` I would expect the value to be `theme`. If we use `currentTheme` I would expect the hook to be named `useCurrentTheme`.

Anyway, not super important. Up to you to decide.

emmenko (approved):
Thanks!

design-system/src/theme-provider.tsx
CarlosCortizasCT: Updated in this commit: 13e39429a2f6cc42a06257d405e6cd5130b12b32

design-system/src/theme-provider.tsx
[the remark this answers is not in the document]
gabriele-ct: @emmenko @nkuehn and @CarlosCortizasCT 
Upgrading `docs` repo here, and I'm seeing gatsby SSR failing because of the use of `MutationObserver`.
Any chance we can get @emmenko suggestion implemented soon?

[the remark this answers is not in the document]
CarlosCortizasCT: Hi @gabriele-ct 

We have [some work in progress](https://github.com/commercetools/ui-kit/pull/2351) but its priority has not been big enough since other MC redesign tasks need more attention.
We will try to find some time to finish it.

Just for us to know, is [this workaround](https://github.com/commercetools/commercetools-docs-kit/pull/1418) not working anymore?

PR #2350, opened 2022-11-30
chore: replacing the icons

chloe0592:
Replacing wrongly sized icons with the right ones.


packages/components/icons/src/generated/AngleDownReact.tsx
```diff
@@ -139,11 +139,15 @@ const SvgAngleDown = (props: SVGProps) => (
     width={24}
     height={24}
     viewBox="0 0 24 24"
+    fill="none"
```
CarlosCortizasCT, line 142: This `fill="none"` property is a problem and I already had to remove it from the previous new icons.
I think we need to remove it again from all of them and let UX know so they can also remove it from the original icons we're importing.

chloe0592: Ah okay thanks for the heads-up, will remove it and let Natalia know also.

chloe0592: Here's the fix - [d557b34](https://github.com/commercetools/ui-kit/pull/2350/commits/d557b34039e25c05d65d2485911aa2506a80ec32).

chloe0592:
@CarlosCortizasCT do you maybe now why the test 'Checking bundle size of visual testing app' is failing? And how can I fix it? 😅

tdeekens:
> @CarlosCortizasCT do you maybe now why the test 'Checking bundle size of visual testing app' is failing? And how can I fix it? 😅

I think it's in this file https://github.com/commercetools/ui-kit/blob/main/bundlesize.config.json

emmenko:
Hard to spot but you have to look for the `FAIL` log

image

The config is defined here: https://github.com/commercetools/ui-kit/blob/main/bundlesize.config.json

CarlosCortizasCT:
> Hard to spot but you have to look for the `FAIL` log
> 
> image
> 
> The config is defined here: https://github.com/commercetools/ui-kit/blob/main/bundlesize.config.json

Perhaps we can extend the configuration to allow for a bigger size in the icons package.

chloe0592:
> > Hard to spot but you have to look for the `FAIL` log
> > image
> > The config is defined here: https://github.com/commercetools/ui-kit/blob/main/bundlesize.config.json
> 
> Perhaps we can extend the configuration to allow for a bigger size in the icons package.

In which cases are we allow to extend it? And if this is one of them is it okay to set the `maxSize` to `60 kB`?

emmenko:
I guess we can add a new entry for the icons with a higher threshold, as it's usually the biggest package in size.

chloe0592:
> I guess we can add a new entry for the icons with a higher threshold, as it's usually the biggest package in size.

How do I do that? Is it just adjusting the `maxSize` in `bundlesize.config.json` file?

emmenko:
You can try adding a new entry for the icons bundle with a higher value yes

chloe0592:
> chore: adjusting maxSize for UI Kit icons

I can't seem to resolve it. There must be something wrong here that I did? 
Commit - [chore: adjusting maxSize for UI Kit icons](https://github.com/commercetools/ui-kit/pull/2350/commits/0414450e5c27d9c6733d552b61a963509e06f315)

chloe0592:
> You can try adding a new entry for the icons bundle with a higher value yes

I can't seem to resolve it. There must be something wrong here that I did? 
Commit - [chore: adjusting maxSize for UI Kit icons](https://github.com/commercetools/ui-kit/pull/2350/commits/0414450e5c27d9c6733d552b61a963509e06f315)

emmenko:
I suppose the first glob should ignore the icons. Let's just bump the max size then for now.

Thanks for looking into that!

chloe0592:
> should ignore the icons



> I suppose the first glob should ignore the icons. Let's just bump the max size then for now.
> 
> Thanks for looking into that!

So you mean it's okay if I remove the icon bundle and just max it out for all the bundles? 

```
"files": [
    {
      "path": "packages/**/dist/*.cjs.prod.js",
      "maxSize": "60 kB"
    }
  ]
```

emmenko:
Yes

CarlosCortizasCT:
I guess the problem when configuring several entries is that they are not exclusive. Although you add a configuration specific for the icons package, the other configuration rule is also used to check that package.

It seems something like this should work:
```
{
  "files": [
    {
      "path": "packages/**/icons/dist/*.cjs.prod.js",
      "maxSize": "60 kB"
    },
    {
      "path": "packages/**/!(icons)/dist/*.cjs.prod.js",
      "maxSize": "50 kB"
    }
  ]
}
```

CarlosCortizasCT:
I would ask @NataliaSclipet to take a look.

Personally, I find the "X" icon bigger and it looks quite big in the [Tag component](https://percy.io/commercetools-GmbH/ui-kit/builds/23364054/changed/1305222898?browser=chrome&browser_ids=31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1024&widths=1024%2C1600), but maybe that's intended.

CarlosCortizasCT (approved):
Looks good to me 💯

chloe0592:
> I guess the problem when configuring several entries is that they are not exclusive. Although you add a configuration specific for the icons package, the other configuration rule is also used to check that package.
> 
> It seems something like this should work:
> 
> ```
> {
>   "files": [
>     {
>       "path": "packages/**/icons/dist/*.cjs.prod.js",
>       "maxSize": "60 kB"
>     },
>     {
>       "path": "packages/**/!(icons)/dist/*.cjs.prod.js",
>       "maxSize": "50 kB"
>     }
>   ]
> }
> ```

Thanks for the help! 🤗  It worked :)

chloe0592:
> I would ask @NataliaSclipet to take a look.
> 
> Personally, I find the "X" icon bigger and it looks quite big in the [Tag component](https://percy.io/commercetools-GmbH/ui-kit/builds/23364054/changed/1305222898?browser=chrome&browser_ids=31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1024&widths=1024%2C1600), but maybe that's intended. It will look ok with the new theme, so maybe it's acceptable in the meantime.

Yes, @NataliaSclipet if could take a look at once more in Percy if everything looks like you'd expect it to? 🙏

kark (approved):
It looks good to me too 👍

@NataliaSclipet I also wanted to double check if:
- dot icon is ok - looks much smaller (we use it in a bunch of places in the MC)
- minimize icon changed "alignment"
- plus bold icon is exactly same as plus thin (just checking if there was no export error)

chloe0592:
**Note**: Natalia will double check some icons with Filip then will get back to us.

emmenko:
Is there anything left to do here or can we merge this?

NataliaSclipet (approved):
All good! Thank you for making these changes

PR #2351, opened 2022-12-01
Add new hook to use `MutationObserver` API

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=9Tthkrf)

#### Summary

Include a new hook to easily use [MutationObserver API](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) in components.

## Description

The new implementation has been copied from the `app-kit` [repository](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-components/src/hooks/use-mutation-observer/use-mutation-observer.ts).

There are two changes over the original implementation:
1. `observer` configuration is not passed to the mutation observer factory but to the `observe` function instead. The way it was, since we only have one `observer` instance, if we configure several observers, only the configuration from the first one is used. (_I think we should change this also in `app-kit`_)
2. Instead of calling the consumer callback with one mutation record at a time, we call it with a list of all mutations received by the observer to follow the [MutationObserver behaviour](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver#parameters).

### DIFF
--- a/packages/hooks/src/use-mutation-observer/use-mutation-observer.ts
+++ b/packages/hooks/src/use-mutation-observer/use-mutation-observer.ts
@@ -1,0 +1,107 @@
+// NOTE: This React hook is essentially the same as the [@react-hook/intersection-observer](https://www.npmjs.com/package/@react-hook/intersection-observer)
+// except for the usage of `MutationObserver` instead of `IntersectionObserver`.
+import { useEffect, useLayoutEffect, useRef } from 'react';
+import rafSchd from 'raf-schd';
+
+type TUseMutationObserverCallback = (
+  mutationsList: MutationRecord[],
+  observer: MutationObserver
+) => void;
+
+let _mutationObserver: ReturnType<typeof createMutationObserver>;
+
+const useLatest = <T>(current: T) => {
+  const storedValue = useRef(current);
+  useEffect(() => {
+    storedValue.current = current;
+  });
+  return storedValue;
+};
+
+function createMutationObserver() {
+  const callbacks = new Map<Node, TUseMutationObserverCallback[]>();
+  const observer = new MutationObserver(
+    rafSchd((mutationsList, observer) => {
+      const mutationsByTarget = mutationsList.reduce<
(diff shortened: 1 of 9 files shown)


packages/hooks/src/use-mutation-observer/use-mutation-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+      });
+    })
+  );
+
+  return {
+    observer,
+    subscribe(
+      target: HTMLElement,
+      callback: TUseMutationObserverCallback,
+      options?: MutationObserverInit
```
CarlosCortizasCT, line 40: We need to receive the options per `subscribe` call so we can have observe different elements with different configurations in the same observer.

packages/hooks/src/use-mutation-observer/use-mutation-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+  observer: MutationObserver
+) => void;
+
+let _mutationObserver: ReturnType<typeof createMutationObserver>;
+
+function createMutationObserver() {
+  const callbacks = new Map<Node, TUseMutationObserverCallback[]>();
+  const observer = new MutationObserver(
+    rafSchd((mutationsList, observer) => {
+      const mutationsByTarget = mutationsList.reduce<
```
CarlosCortizasCT, line 18: We group the mutations by target and deliver the complete list to the consumer callback instead of calling it once per mutation.
This is closer to how the [MutationObserver API works](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver#parameters).

emmenko (approved):
This is good to go now 🚀

emmenko:
@CarlosCortizasCT The issue was mainly with Jest having to load an ESM module (the `@react-hooks/latest` package) which is still tricky to configure in Jest.

Since the module was implementing a simple `useEffect` function, I just copied the implementation 😄

PR #2354, opened 2022-12-04
Update Collapsible panel component to support theming

ddouglasz:
Summary
Update Collapsible panel component to support new design tokens.


design-system/materials/internals/definition.yaml
```diff
@@ -782,6 +804,14 @@ decisionGroupsByTheme:
           choice: '0 var(--spacing-20)'
         padding-for-tag:
           choice: '2px 12px'
+        padding-for-collapsible-panel-header:
+          choice: 'var(--spacing-30) 0'
```
ddouglasz, line 808: I am not so sure about this values of padding left and right = 0. It makes the side paddings for the header out of alignment.

image

ddouglasz: Asking from Filip, this is actually the expected behaviour.

design-system/materials/internals/definition.yaml
```diff
@@ -861,6 +885,14 @@ decisionGroupsByTheme:
...
         padding-for-table-cell-as-condensed:
           choice: 'var(--spacing-20) var(--spacing-40)'
+        padding-for-collapsible-panel-header:
+          choice: 'var(--spacing-30) 0'
+        padding-for-collapsible-panel-header-as-condensed:
+          choice: 'var(--spacing-20) 0'
+        padding-for-collapsible-panel-content:
+          choice: 'var(--spacing-l) 0 var(--spacing-70) 28px'
+        padding-for-collapsible-panel-content-as-condensed:
+          choice: 'var(--spacing-l) 0 var(--spacing-70) 28px'
```
kark, line 895: `var(--spacing-l)` is deprecated

packages/components/collapsible-panel/src/header-icon.tsx
```diff
@@ -1,19 +1,34 @@
...
+    } else {
+      return 'warning';
+    }
+  }
+
   return 'solid';
 };
 
+const checkIfTestOrDefault = (
+  theme?: 'default' | 'test',
```
kark, line 25: I don't think the name `test` will stay for long-term, so I wouldn't couple our names with it. (same in other places)

packages/components/collapsible-panel/src/header-icon.tsx
```diff
@@ -39,11 +55,12 @@ const HeaderIcon = (props: THeaderIcon) => {
...
-          flex-shrink: 0;
-          box-shadow: ${designTokens.shadow7};
-          background-color: ${backgroundColor};
-          border: 1px solid ${backgroundColor};
+
+          border-radius: ${checkIfTestOrDefault(theme, '50%')};
+          flex-shrink: ${checkIfTestOrDefault(theme, '0')};
+          box-shadow: ${checkIfTestOrDefault(theme, designTokens.shadow7)};
+          background-color: ${checkIfTestOrDefault(theme, backgroundColor)};
+          border: ${checkIfTestOrDefault(theme, backgroundColor)};
```
kark, line 63: Perhaps if it's not possible to do with design tokes we could add those styles as a separate array item:
```js
[
...,
theme === 'default' &&
  css`
    border-radius: 50%;
    flex-shrink: 0;
    box-shadow: ${designTokens.shadow7};
    background-color: ${backgroundColor};
    border: 1px solid ${backgroundColor};
  `,
...
]
```

kark (commented):
So far so good 🙌 I am leaving some comments to consider

packages/components/collapsible-panel/src/header-icon.tsx
ddouglasz: Thank you very much @kark , I am going to update soon.

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2354/commits/a1d3831fa56c3c1c8ec1f3ba0f4a5e3cc666c4bf

Thank you.

ddouglasz:
> Other issues I noticed:
> 
> * please mind that we should update markup based on theme
> 
> Screenshot 2022-12-07 at 17 36 51
> 
> - I’m not sure if the `box-shadow` around the collapsible panel should be visible (and if border radius is correct)

I have a call with Filip today. I will get more insight on this and make the necessary changes after the call.

packages/components/collapsible-panel/src/header-icon.tsx
CarlosCortizasCT: There's a `ThemeName` type exported from the `design-system` package you could use here.

I wouldn't worry too much about the naming of the test as we will need to just remove all this transitioning code after the new theme is finally released.

FilPob:
the description is a good one. I never saw that being used somewhere. My suggestion would be to use a body.detail for that with text tone: secondary. The same should be used for the SecondaryHeader.

CarlosCortizasCT:
> the description is a good one. I never saw that being used somewhere. My suggestion would be to use a body.detail for that with text tone: secondary. The same should be used for the SecondaryHeader.

@FilPob what would be the vertical spacing to use between the description and the actual content?

CarlosCortizasCT:
I think I've addressed all the ui issues I found in my previous review, but I found a couple of things that might worth UX feedback:
* We didn't have a design for the `disabled` state of the component where the icon should look different: I updated the styles to just use a lighter color of the arrow but I would like UX to review
* The component has the option to hide the icon for opening/closing the panel so I'm wondering if, when the icon is not rendered, we should change the left padding of its content to be vertically aligned with the title

image

Please @FilPob take a look when you have some time 🙏

packages/components/collapsible-panel/src/collapsible-panel.styles.tsx
```diff
@@ -83,22 +98,34 @@ const HeaderControlsWrapper = styled.div`
...
+const SectionContent = styled.div<Pick<TCollapsiblePanel, 'condensed'>>`
   width: 100%;
   display: flex;
   flex-wrap: wrap;
   align-items: flex-start;
 `;
 
-const SectionWrapper = styled.div`
-  border-top: 1px solid ${designTokens.colorNeutral60};
+const SectionDescriptionWrapper = styled('div')<{ condensed?: boolean }>`
```
kark, line 108: nit:
```suggestion
const SectionDescriptionWrapper = styled.div<{ condensed?: boolean }>`
```

packages/components/collapsible-panel/src/collapsible-panel.styles.tsx
```diff
@@ -83,22 +98,34 @@ const HeaderControlsWrapper = styled.div`
   cursor: auto;
 `;
 
-const SectionContent = styled.div`
+const SectionContent = styled.div<Pick<TCollapsiblePanel, 'condensed'>>`
```
kark, line 101:
```suggestion
const SectionContent = styled.div`
```

packages/components/collapsible-panel/src/collapsible-panel.tsx
```diff
@@ -20,9 +21,11 @@ import {
   HeaderControlsWrapper,
   SectionWrapper,
   SectionContent,
+  getBaseContainerStyles,
+  SectionDescriptionWrapper,
 } from './collapsible-panel.styles';
 import CollapsiblePanelHeader from './collapsible-panel-header';
-
+import { ThemeName, useTheme } from '@commercetools-uikit/design-system';
```
kark, line 28:
```suggestion
import { type ThemeName, useTheme } from '@commercetools-uikit/design-system';
```

kark (commented):
That's just great work 👏 Looks good from my side.
Just wanted to double check if this will be handled later on as FIlip suggested:
Screenshot 2022-12-14 at 12 52 31
Screenshot 2022-12-14 at 12 55 44

CarlosCortizasCT:
> That's just great work 👏 Looks good from my side. Just wanted to double check if this will be handled later on as FIlip suggested: Screenshot 2022-12-14 at 12 52 31 Screenshot 2022-12-14 at 12 55 44

Good point! I think I missed that one.

Let me fix it.

packages/components/collapsible-panel/src/collapsible-panel.styles.tsx
CarlosCortizasCT: Updated in this commit: b2c37088f4d207992c3a579236e88828daa233b6

packages/components/collapsible-panel/src/collapsible-panel.styles.tsx
CarlosCortizasCT: Updated in this commit: b2c37088f4d207992c3a579236e88828daa233b6

packages/components/collapsible-panel/src/collapsible-panel.tsx
CarlosCortizasCT: Updated in this commit: b2c37088f4d207992c3a579236e88828daa233b6

CarlosCortizasCT:
> > That's just great work 👏 Looks good from my side. Just wanted to double check if this will be handled later on as FIlip suggested: Screenshot 2022-12-14 at 12 52 31 Screenshot 2022-12-14 at 12 55 44
> 
> Good point! I think I missed that one.
> 
> Let me fix it.

Fixed in this commit: b2c37088f4d207992c3a579236e88828daa233b6

I also noticed in the designs that the icon size does not change when the panel is in condensed mode for the new theme so I implemented that as well.

packages/components/collapsible-panel/src/collapsible-panel.tsx
```diff
@@ -208,23 +244,20 @@ const CollapsiblePanel = (props: TCollapsiblePanel) => {
...
                 )}
                 <Spacings.Inline alignItems="center" scale={scale}>
-                  {props.condensed ? (
-                    <Text.Detail as="span" isBold truncate>
-                      {props.header}
-                    </Text.Detail>
-                  ) : (
-                    props.header
-                  )}
+                  {HeadLineText(theme, props)}
```
kark, line 260: I think this function should be refactored a little and rendered in JSX as a React component (the name suggests that) or the function name should be changed

kark (approved):
🙌

packages/components/collapsible-panel/src/collapsible-panel.tsx
```diff
@@ -241,13 +275,21 @@ const CollapsiblePanel = (props: TCollapsiblePanel) => {
               )}
             </HeaderContainer>
             <div style={containerStyles}>
-              <SectionWrapper ref={registerContentNode}>
+              <SectionWrapper
+                // @ts-ignore
```
CarlosCortizasCT, line 279: I tried several things but I don't find how to make the typing work for this reference.

Any help would be much appreciated 🙏

kark: I think I used type assertion in `<RichTextBody>` for forwarded refs so I'd suggest:
```suggestion
import type { ReactNode, Ref } from 'react';

<SectionWrapper
  ref={registerContentNode as Ref<HTMLDivElement>}
```

FilPob (approved):

CarlosCortizasCT (approved):

PR #2355, opened 2022-12-06
style: fix `<DataTable>` column separator color

kark:
#### Summary

After introducing design tokens changes `<DataTable>` column separator lines are a bit too bright and therefore not easily discoverable. The purpose of this PR is to make them a bit darker (use `color-neutral` instead of `color-neutral-90` in the new theme)


FilPob (approved):

ddouglasz (approved):
Tested and it looks good 🚀

PR #2356, opened 2022-12-08
Label and Field label

chloe0592:
This PR has a purpose of updating `Label`, `FieldLabel ` and  `RequiredIndicator` components to the newest design requirements.

It also has a [follow up PR](https://github.com/commercetools/merchant-center-frontend/pull/13403) in `RequiredIndicator` component in the MC repository as we wanted to have the consistent style in both.


packages/components/label/src/label.tsx
```diff
@@ -59,6 +61,17 @@ type TLabelProps = {
 };
 
 const Label = (props: TLabelProps) => {
+  const { theme } = useTheme();
+  // TODO: This is a temporary solution due to theme migration. After the new
+  // theme is published, we must remove this and just use the `Text.Detail` component
```
kark, line 66: ❤️

packages/components/label/src/label.tsx
```diff
@@ -59,6 +61,17 @@ type TLabelProps = {
...
+  const TextComponent = theme === 'default' ? Text.Body : Text.Detail;
+
+  const getLabelStyles = () => css`
+    > div {
+      font-weight: 500;
+    }
+  `;
```
kark, line 73: I think we shouldn't hardcode values and use a design token instead.
I suppose we don't need to define as a function, but if that is preferred then not define it inside the component.

packages/components/field-label/src/field-label.tsx
```diff
@@ -81,6 +82,17 @@ export type TFieldLabelProps = {
...
+  const { theme } = useTheme();
+
+  const getInfoTextTone = (
+    { tone }: Pick<TFieldLabelProps, 'tone'>,
+    theme: 'default' | 'test'
+  ) => {
+    if (theme === 'default') {
+      return tone;
+    } else return 'secondary';
+  };
```
kark, line 94:
```suggestion
import { useTheme, type ThemeName } from '@commercetools-uikit/design-system';
const getInfoTextTone = (theme: ThemeName, tone?: TFieldLabelProps['tone']) => {
  if (theme === 'default') {
    return tone;
  }
  return 'secondary';
};
```
Other than that I don't think we need to define it inside the component.

FilPob (dismissed):

kark (commented):
Looking good! Great work 🙌

packages/components/label/src/label.tsx
chloe0592: I wasn't really sure how to implement this change and if my implementation in correct. Any other suggestion is more than welcome! 😄

packages/components/field-label/src/field-label.tsx
chloe0592: Thank you @kark! 🍫 So you mean I should create a `field-label.styles.tsx` file and move the function there?

kark: Cool 🙌 I would only opt for defining the function outside the scope of `const FieldLabel = (props: TFieldLabelProps) => {  }` since it's a pure function, to not define the function on each render of the component. A separate file is not necessary.

chloe0592: Sounds good! Thanks!

chloe0592:
@FilPob could you check again if the alignment is alright now? Thank you!

[chore: for the test theme center items](https://github.com/commercetools/ui-kit/pull/2356/commits/32502c671daba02369b9e360742db798a94edde1)

packages/components/field-label/src/field-label.tsx
```diff
@@ -96,10 +106,13 @@ const FieldLabel = (props: TFieldLabelProps) => {
...
       <Stack scale="xs">
-        <Inline alignItems="flexStart" scale="xs">
+        <Inline
+          alignItems={theme === 'default' ? 'flexStart' : 'center'}
+          scale="xs"
+        >
           <Text.Wrap>
             <Label
-              isBold={true}
+              isBold={theme === 'default' ? true : false}
```
CarlosCortizasCT, line 115: **nit**
```suggestion
              isBold={theme === 'default'}
```

packages/components/label/src/label.tsx
```diff
@@ -75,15 +88,15 @@ const Label = (props: TLabelProps) => {
   );
 
   return (
-    <label id={props.id} htmlFor={props.htmlFor}>
-      <Text.Body tone={props.tone} isBold={props.isBold}>
+    <label css={getLabelStyles()} id={props.id} htmlFor={props.htmlFor}>
+      <TextComponent tone={props.tone} isBold={props.isBold}>
```
CarlosCortizasCT, line 92: I'm not sure if this is intended but now the `isBold` prop does not visually affect the component. It has `font-weight: 500` in both cases.

packages/components/label/src/label.tsx
```diff
@@ -75,15 +88,15 @@ const Label = (props: TLabelProps) => {
   );
 
   return (
-    <label id={props.id} htmlFor={props.htmlFor}>
-      <Text.Body tone={props.tone} isBold={props.isBold}>
+    <label css={getLabelStyles()} id={props.id} htmlFor={props.htmlFor}>
```
CarlosCortizasCT, line 91: As @kark [suggested](https://github.com/commercetools/ui-kit/pull/2356/files#r1043192538), perhaps it would be simpler to use the css value here:
```suggestion
    <label
      css={css`
         > div {
           font-weight: ${designTokens.fontWeight500};
         }
      `}
      id={props.id}
      htmlFor={props.htmlFor}
    >
```

CarlosCortizasCT:
There's a problem with the required indicator since not it's rendered in _italic_ font style instead of normal:
image

That symbol is rendered in the `RequiredIndicator` component using an `em` HTML tag which, by default, uses _italic_ font style.
I think this was not an issue before because the font used in the default theme does not contain that character in italic but the new font does.

I can think of two options:
1. Restyle `RequiredIndicator` component to overwrite the default styling by the browser
2. Add an entry in the [global reset](https://github.com/commercetools/ui-kit/blob/main/design-system/materials/resets.css) for this HTML element to change its global default font style

Personally I would vote for the second option as the default behaviour up until now in MC is for `em`s to not use italic font style.

packages/components/field-label/src/field-label.tsx
```diff
@@ -119,7 +132,10 @@ const FieldLabel = (props: TFieldLabelProps) => {
...
 
         {props.hint && (
-          <Inline alignItems="flexStart" scale="xs">
+          <Inline
+            alignItems={theme === 'default' ? 'flexStart' : 'center'}
+            scale="xs"
+          >
             {props.hintIcon && (
               <Inline>
                 {cloneElement(props.hintIcon, {
```
CarlosCortizasCT, line 141: I think we're missing here using the square shape for the icon in the new theme.

FilPob: Good point. I assumed we will continue with the rounded buttons first, as it was not that easy to change to squared ones everywhere. I would rather prefer to keep the rounded ones, than having a mix of squared and rounded

FilPob:
@chloe0592  Looks all good from my side. Just the one issue with the italic required indicator which Carlos mentioned above

chloe0592:
> There's a problem with the required indicator since not it's rendered in _italic_ font style instead of normal: image
> 
> That symbol is rendered in the `RequiredIndicator` component using an `em` HTML tag which, by default, uses _italic_ font style. I think this was not an issue before because the font used in the default theme does not contain that character in italic but the new font does.
> 
> I can think of two options:
> 
> 1. Restyle `RequiredIndicator` component to overwrite the default styling by the browser
> 2. Add an entry in the [global reset](https://github.com/commercetools/ui-kit/blob/main/design-system/materials/resets.css) for this HTML element to change its global default font style
> 
> Personally I would vote for the second option as the default behaviour up until now in MC is for `em`s to not use italic font style.

Thanks @CarlosCortizasCT for all the feedback! 🙏 I'll go now, one by one and try to improve it.

-> Great catch for the *, I have't even noticed it! And I agree it might be better to change it the global styling if we don't need the `em` tag to have italic style

packages/components/label/src/label.tsx
chloe0592: Fixed in this commit - [9fab790](https://github.com/commercetools/ui-kit/pull/2356/commits/9fab79008b6f56b6e1f368aa32ab473ab39f55f0)

packages/components/label/src/label.tsx
chloe0592: Fixed in this commit - [9fab790](https://github.com/commercetools/ui-kit/pull/2356/commits/9fab79008b6f56b6e1f368aa32ab473ab39f55f0)

packages/components/field-label/src/field-label.tsx
chloe0592: Improved with this commit - [e3aaf6b](https://github.com/commercetools/ui-kit/pull/2356/commits/e3aaf6b87520c5b8d055aafebf7bfdc1a69c3bcb)

packages/components/field-label/package.json
```diff
@@ -22,6 +22,7 @@
     "@babel/runtime": "^7.19.0",
     "@babel/runtime-corejs3": "^7.19.1",
     "@commercetools-uikit/constraints": "15.5.1",
+    "@commercetools-uikit/design-system": "15.5.1",
     "@commercetools-uikit/icon-button": "15.5.1",
     "@commercetools-uikit/icons": "15.5.1",
     "@commercetools-uikit/label": "15.5.1",
```
emmenko, line 28: I think you need to rebase your branch, as the latest published version is `15.6.0`

packages/components/field-label/src/field-label.tsx
```diff
@@ -130,13 +146,17 @@ const FieldLabel = (props: TFieldLabelProps) => {
               </Inline>
             )}
             {props.hint && (
-              <Text.Detail tone={props.tone}>{props.hint}</Text.Detail>
+              <Text.Detail tone={getInfoTextTone(theme, props.tone)}>
```
emmenko, line 149: Nit: I would keep using a ternary here, to keep it consistent with other similar cases in this file.

emmenko (commented):
Looking good, just some minor remarks. Thanks!

emmenko:
>  it might be better to change it in the global styling if we don't need the em tag to have italic style anywhere.

@chloe0592 You can do that in the [resets.css file](https://github.com/commercetools/ui-kit/blob/main/design-system/materials/resets.css).

packages/components/field-label/src/field-label.tsx
chloe0592: That's a good point! Thanks! [commit](https://github.com/commercetools/ui-kit/pull/2356/commits/3ae00e8a13eeead2db040e1408dcebf337455bcc)

packages/components/field-label/src/field-label.tsx
```diff
@@ -14,6 +14,7 @@ import Constraints from '@commercetools-uikit/constraints';
 import Stack from '@commercetools-uikit/spacings-stack';
 import Inline from '@commercetools-uikit/spacings-inline';
 import Label from '@commercetools-uikit/label';
+import { useTheme } from '@commercetools-uikit/design-system';
```
CarlosCortizasCT, line 17: Please remember to add the `// TODO: @redesign cleanup` comment at the top of the files where we're using `useTheme` hook.

chloe0592:
@CarlosCortizasCT, @emmenko if we apply the `em` font style in `resets.css` file then it'll override all the possible `em` tags in other places as well (such as `RichTextUtils`), which we might not want. 

So I applied design tokens only in the `RequiredIndicator` component - [see here](https://github.com/commercetools/ui-kit/pull/2356/commits/d1dbe33a743de900ee3a89b3d820d218bcfe4616).

My question is do we also need to apply some changes (and if yes, what changes) to the [RequiredIndicator](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/experimental-components/src/components/required-indicator/required-indicator.jsx) from MC?

packages/components/label/src/required-indicator.tsx
```diff
@@ -5,6 +5,7 @@ const RequiredIndicator = () => (
   <em
     css={css`
       color: ${designTokens.colorWarning};
+      font-style: ${designTokens.fontStyleForRequiredIndicator};
```
CarlosCortizasCT, line 8: I would expect us just to use `normal` here since we don't want to use _italic_ for this indicator in any of the themes.

CarlosCortizasCT:
> My question is do we also need to apply some changes (and if yes, what changes) to the [RequiredIndicator](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/experimental-components/src/components/required-indicator/required-indicator.jsx) from MC?

I think we should for consistency as a new follow-up PR and I would just add the `font-style` rule in the CSS module of that component.

emmenko:
@chloe0592 thanks!

Maybe we can have the `em` style always as `normal` (old and new theme), not sure if this needs to be using italic. @FilPob what do you think?

CarlosCortizasCT:
> > @chloe0592 thanks!
> > Maybe we can have the `em` style always as `normal` (old and new theme), not sure if this needs to be using italic. @FilPob what do you think?
> 
> Yes, maybe @FilPob would be the best person to ask. 😄
> 
> Right now for the 'default' theme the indicator's `em` is italic. That's why I created a separate design token for the 'test' theme (`font-style: normal`) since I assumed that we don't want any changes in the default theme. But if we can use `normal` style for both themes, that's even better.
> 
> cc @CarlosCortizasCT

The way I see it, the current `issue` is that we are using the `em` tag in the `RequiredIndicator` component and the behaviour for the asterisk have changed because of the new font.
In the current theme, even though the `font-style` for the `em`s is `italic` (as per browser defaults), the asterisk it not rendered as _italic_ because (I think) the Open Sans font we're using does not include that character in _italic_.
In the new theme we are using the `Inter` font which do include the asterisk character in _italic_, so that's why we see it like this only in the new theme.

At first I thought we should just add a resetting global style to make `em`s `font-style` to be `normal` but, if we do this, any other usage of `em`s (outside the `RequiredIndicator` component) would have a regression.
After we searched for `em` usages across MC frontend, app-kit and ui-kit, we found two places ([here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/html/html.tsx#L59) and [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/slate-helpers.tsx#L107)) related Rich text where we are using `em`s to make text _italic_, so that's why now I don't think resetting the style globally is a good idea.

We just want to make sure the `RequiredIndicator` does not use _italic_ for the asterisk and that can be done changing the HTML tag we use to wrap the asterisk (which I don't recommend because that tag makes semantically sense in this context), or by just adjusting the `font-style` of the `em` to be `normal` **only in this component**.
Since the change would be local and we want to just be `normal` in both themes, I think the change it's just something like this:
```
const RequiredIndicator = () => (
  <em
    css={css`
      color: ${designTokens.colorWarning};
      font-style: normal; // <---- HERE IS THE ACTUAL CHANGE
    `}
  >
    {'*'}
  </em>
);
```

chloe0592:
> > > @chloe0592 thanks!
> > > Maybe we can have the `em` style always as `normal` (old and new theme), not sure if this needs to be using italic. @FilPob what do you think?
> > 
> > 
> > Yes, maybe @FilPob would be the best person to ask. 😄
> > Right now for the 'default' theme the indicator's `em` is italic. That's why I created a separate design token for the 'test' theme (`font-style: normal`) since I assumed that we don't want any changes in the default theme. But if we can use `normal` style for both themes, that's even better.
> > cc @CarlosCortizasCT
> 
> The way I see it, the current `issue` is that we are using the `em` tag in the `RequiredIndicator` component and the behaviour for the asterisk have changed because of the new font. In the current theme, even though the `font-style` for the `em`s is `italic` (as per browser defaults), the asterisk it not rendered as _italic_ because (I think) the Open Sans font we're using does not include that character in _italic_. In the new theme we are using the `Inter` font which do include the asterisk character in _italic_, so that's why we see it like this only in the new theme.
> 
> At first I thought we should just add a resetting global style to make `em`s `font-style` to be `normal` but, if we do this, any other usage of `em`s (outside the `RequiredIndicator` component) would have a regression. After we searched for `em` usages across MC frontend, app-kit and ui-kit, we found two places ([here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/html/html.tsx#L59) and [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/rich-text-utils/src/slate-helpers.tsx#L107)) related Rich text where we are using `em`s to make text _italic_, so that's why now I don't think resetting the style globally is a good idea.
> 
> We just want to make sure the `RequiredIndicator` does not use _italic_ for the asterisk and that can be done changing the HTML tag we use to wrap the asterisk (which I don't recommend because that tag makes semantically sense in this context), or by just adjusting the `font-style` of the `em` to be `normal` **only in this component**. Since the change would be local and we want to just be `normal` in both themes, I think the change it's just something like this:
> 
> ```
> const RequiredIndicator = () => (
>   <em
>     css={css`
>       color: ${designTokens.colorWarning};
>       font-style: normal; // <---- HERE IS THE ACTUAL CHANGE
>     `}
>   >
>     {'*'}
>   </em>
> );
> ```

Ah, okay now it makes sense to me. It changed because we changed the font family! Okay, yeah, was missing that part. I also then think we should just need to add the `font-size: normal` to the `RequiredIndicator` component. 
Thank you Carlos! ❤️

emmenko:
I think we can use `normal` style and add some space (margin) between the text and the asterisk.

FilPob:
The font style doesn't need to be italic (sorry if that was not clear.) Normal is fine. And putting a spacing of 2px between the label and the asterix would be sweet. 4px works as well though

chloe0592:
@CarlosCortizasCT, @emmenko, @FilPob - I've added margin to the asterisk, so let me know if it all looks good now. Then I can also add that styling to the MC component. 😊

CarlosCortizasCT (approved):
Looks good to me now.

Thanks @chloe0592 👏

FilPob:
Good from my side as well :)

PR #2359, opened 2022-12-13
`DataTableManager` support for the new theme

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=u3XyU0e)

#### Summary

Adapt `DataTableManager` component to support the new theme

## Description

Creates new design tokens to use in the `DataTableManager` subcomponents and refactored them to allow for new theme ui.

These are some things to bear in mind:
* The settings dropdown does not look as expected as `SelectInput` component will be updated in another specific PR
* The header for the settings panels is using an H2 element but now it looks bigger than expected because H2 (temporary) uses the same `font-size` than H1
* layout settings panel radio inputs don't have the expected spacing (between icon and text) as that change will be addressed when we work in `RadioInput` component


FilPob (approved):

kark (approved):
all good as far as I'm concerned 🎉

PR #2361, opened 2022-12-14
New helper in `useTheme` hook

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=eCpoPab)

#### Summary

Added a new temporary helper to the resulting object from `useTheme` hook to help in the new theme migration process.

### DIFF
--- a/.changeset/few-dancers-trade.md
+++ b/.changeset/few-dancers-trade.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tag': patch
+---
+
+Refactor tag text component calculation
--- a/.changeset/old-bobcats-know.md
+++ b/.changeset/old-bobcats-know.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/design-system': minor
+---
+
+Added internal temporary helper to `useTheme` hook which will help with new theme migration
--- a/design-system/src/theme-provider.tsx
+++ b/design-system/src/theme-provider.tsx
@@ -1,4 +1,10 @@
-import { useLayoutEffect, useState, useRef, useEffect } from 'react';
+import {
+  useLayoutEffect,
+  useState,
+  useRef,
+  useEffect,
+  type ReactNode,
+} from 'react';
 import isObject from 'lodash/isObject';
 import merge from 'lodash/merge';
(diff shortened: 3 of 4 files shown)


Rhotimee (approved):
Thanks for this. Very useful.

tdeekens (commented):
This is nice. What I additionally wondered if we can (for other cases) encapsulate the `=== 'test` bit also. Test is sometimes a confusing name. Maybe we can export a `isNewTheme` or similar?

emmenko (approved):
Nice! 🥳

kark (approved):

CarlosCortizasCT:
> This is nice. What I additionally wondered if we can (for other cases) encapsulate the `=== 'test` bit also. Test is sometimes a confusing name. Maybe we can export a `isNewTheme` or similar?

@tdeekens 
Yes, I thought about that but I wasn't comfortable with being so specific. I think keeping the conditionals using the `default` keyword would be enough (`theme === 'default'` and `theme !== 'default'`) since all this will be removed anyway after migration is done.

However, if you think that would be more helpful for developers working in the migration, we can also add that one.

tdeekens:
> I thought using the names would be ok since they are typed and you would only be allowed to use one the predefined strings, but in the MC frontend we're not using typescript yet so perhaps it makes sense the addition you suggested.

Yeah it's just that we don't use TS everywhere and switching on default is fine but test is a confusing name. So maybe a helper would be nice.

CarlosCortizasCT:
> Yeah it's just that we don't use TS everywhere and switching on default is fine but test is a confusing name. So maybe a helper would be nice.

Ok. So I added a new prop to the hook result in this commit: eca1eda1a52fd71ba5f964e82869e2e8c3d7a749

CarlosCortizasCT:
> Yeah it's just that we don't use TS everywhere and switching on default is fine but test is a confusing name. So maybe a helper would be nice.

@tdeekens 
Ok. So I added a new prop to the hook result in this commit: eca1eda1a52fd71ba5f964e82869e2e8c3d7a749

tdeekens (approved):

PR #2365, opened 2022-12-16
refactor(tag): add prop `isDraggable`

emmenko:
As suggested by @FilPob 

image


FilPob:
What about the few styling changes specified in the Miro board? will they be part of an another PR?

emmenko:
@FilPob Yes a follow up would be better.

PR #2366, opened 2022-12-19
Avatars design tokens

chloe0592:
## Description

Updating `Avatar` component with new design tokens


design-system/materials/internals/definition.yaml
```diff
@@ -98,6 +98,8 @@ choiceGroupsByTheme:
         font-size-40: '1.125rem'
         font-size-50: '1.25rem'
         font-size-60: '1.5rem'
+        font-size-80: '2.5rem'
+        font-size-90: '3rem'
```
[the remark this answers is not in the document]
chloe0592: It was bugging my eyes, too. Like it much more better now! Thanks!

chloe0592:
@FilPob could you also, please, check if everything looks good from your side?

chloe0592:
@FilPob I've adjusted changes we talked about - there's no `hover` anymore and the color for `highlighted` state is `accent-30`. Let me know if it looks all good now? 🙏

CarlosCortizasCT (approved):
Loos good to me 👍

FilPob:
perfect! thanks

FilPob (approved):

design-system/materials/internals/definition.yaml
```diff
@@ -946,6 +984,13 @@ decisionGroupsByTheme:
         height-for-input:
           choice: '40px'
 
+    widths:
+      label: Width
+      prefix: width
+      decisions:
+        width-for-avatar-as-medium:
+          choice: '40px'
```
emmenko, line 992: Should we include the other sizes as well?

emmenko (approved):
Thanks! 🤗

design-system/materials/internals/definition.yaml
CarlosCortizasCT: I also thought about this and my take here is that it depends on what we want to do with design tokens after we finish the migration: do we want to populate the `definition.yaml` file with all the css values to have a centralized source of truth so all our components should be tokenized and they wouldn't have any css value but just tokens references? If that's so, I guess it would make sense to bring the other values. Otherwise, I think we would be removing this token after the migration and just updating the values in the `avatar.tsx` file.

This is one of the questions I think we need to discuss about after the theme migration to decide how we want to manage styles in `ui-kit` for the future.

chloe0592: Should I merge this PR or should I wait until we figure this out?

emmenko: Let's merge this and continue the discussion separately.

PR #2371, opened 2022-12-22
Update H2 heading sizes

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=-zcNSXL)

#### Summary

Update `font-size`, `line-height` and `font-weight` for `Text.Headline` component when used as an H2 element.

## Description

As part of the MC redesign, we updated the [headings](https://uikit.commercetools.com/?path=/story/basics-typography-text--headline) sizes for the new theme but we had to do a little trick with the H2 variant of the `Text.Headline` component.
As most of the pages were using H2s for their titles and we want them to be H1s (for semantic and accessibility reasons), we set the same values for H1 and H2 at the design tokens level. This way, even though the pages were still using H2 tags, the look like H1s.

Now that we refactored the page layout components we expose from `app-kit` repository and are used to build pages in the MC frontend, we think it's a good time to actually set the final values for H2s.


FilPob (approved):

emmenko (approved):
🥳

PR #2372, opened 2022-12-22
refactor(design-system): add new design tokens for color-accent

emmenko:
~~In the new designs we only have 3 shadows. I tried to give them some meaningful names (got some [inspiration from atlaskit](https://atlassian.design/foundations/elevation)).~~

In addition to that, there are some additional colors that we need to add (see https://github.com/commercetools/merchant-center-application-kit/blob/caee9e27e76ca6bef05388d8cc11c754bfe773b9/packages/application-components/src/theming.ts#L7-L10)


design-system/materials/internals/definition.yaml
```diff
@@ -151,6 +155,10 @@ choiceGroupsByTheme:
         shadow-14: '0 0 0.5px rgba(0, 0, 0, 0.1)'
         shadow-15: '0 2px 2px rgba(0, 0, 0, 0.24), 0 1px 4.75px rgba(0, 0, 0, 0.12)'
         shadow-16: '0.5px 1.5px 4px 1px rgba(0, 0, 0, 0.25), -0.5px -0.5px 4px 1px rgba(0, 0, 0, 0.25)'
+        # New shadows
+        shadow-overlay-5: '0 1px 5px 0 rgb(0 0 0 / 5%)'
+        shadow-overlay-15: '0 2px 5px 0 rgb(0 0 0 / 15%)'
+        shadow-raised-25: '0 0 2px 0 rgb(0 0 0 / 25%)'
```
emmenko, line 161: @FilPob What do you think about this naming structure? Does it make sense to you?

PS: inspiration from [here](https://atlassian.design/foundations/elevation).

emmenko: My intention is to provide somehow some meaningful naming structure. The current `shadow-*` choices are pretty much inconsistent and the scale does not make much sense.

These new names are more of a proposal and we can change them but I think it would make sense to have a structure along these lines. In a sense, somehow similar to e.g. `color-primary`, `color-primary-25`, `color-neutral-90`, etc.

The `overlay` and `raised` I tried to map them from the atlaskit elevation document but we can change them around if for us the meaning is the opposite way.

CarlosCortizasCT: Ok, the example about colors is a very good analogy so I think it makes sense 👍

CarlosCortizasCT (approved):
💯

emmenko (commented):
Thanks!

PR #2368, opened 2022-12-20
Review and adapt date & time components to the new theme

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=QGee2hj)

#### Summary

Review and adapt date & time components to the new theme

## Description

After the changes we made to the TextInput component for the new theme, we need to review the other input components to check whether an unexpected side effect has been made.

Since we are doing this review in chunks, in this particular one I reviewed these components:
- [x] DateInput
- [x] DateRangeInput
- [x] DateTimeInput
- [x] TimeInput

Also, I took the opportunity to adjust some colors regarding the icons. These changes where requested by the UX team.

PR #2373, opened 2022-12-29
Update Search Select input component to support new design tokens.

ddouglasz:
Summary
Update Search Select input component to support new design tokens.


CarlosCortizasCT (commented):
This is starting to look good, but I found some things that need to be adjusted when testing the component in Storybook with the new theme:
* The height of the input is smaller than it should be (you might use the `height-for-input` design token
* _onHover_ state is not like the expected result as per the Miro requirements (border and background)
* _onFocus_ state uses a border that looks bigger than expected (maybe it's just 1px)
* Cross and Dropdown icons should have an 8px horizontal separation
* Error and Warning states are using a border thinner than expected
* When using `isMulti`, selected elements tag in the input should not have a vertical separator between the value and the cross icon

Perhaps you can take a look at [this PR](https://github.com/commercetools/ui-kit/pull/2368) since some of these challenges had also been addressed over there and might provide some ideas.

ddouglasz:
Thank you for the feedback: I updated here together with the new input changes:
https://github.com/commercetools/ui-kit/pull/2373/commits/f1a1207f15e2cfb7dd94dbe8fa38fc42f7b0b14f

ddouglasz:
Setting this to WIP for now since until select menu design is concluded on.

Rhotimee (commented):
Good work @ddouglasz 

A few things I noticed. 
- Background color for Select Option when on hover state doesn't match Miro
- `Clear` Icon should be color-neutral-40
- Border color for Select Input  in error state doesn't match Miro
- Border color for Select Input when in a warning state doesn't match Miro

ddouglasz:
> * Background color when Select Input is disabled doesn't match Miro

- Thank you for the feedback. There are some updates that have been made to the design. However, I will update feedback as expected. 🙏

FilPob (approved):

CarlosCortizasCT:
There are a couple of things I noticed when testing the component in Storybook:
* The background color in read only mode is the same that when in disabled mode. Is this intentional? This is different in the rest of the input components.
* Both disabled and read only mode styles are being overwritten when hovering over the input and that's something not supposed to happen.

FilPob:
@CarlosCortizasCT Yes, for the rad only state there were some [adjustments ](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764542492751225&cot=14) requested by the design team. So the read only in this PR looks correct and should be applied to other Inputs as well. 

The 2nd point is right. The disabled and read only states should not have any hover state. Good catch!

CarlosCortizasCT:
> @CarlosCortizasCT Yes, for the rad only state there were some [adjustments ](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764542492751225&cot=14) requested by the design team. So the read only in this PR looks correct and should be applied to other Inputs as well.d only states should not have any hover effect. Good catch!

Ah! Ok. good to know 😄 

@ddouglasz, perhaps you can do it as a follow-up PR maybe?

ddouglasz:
> > @CarlosCortizasCT Yes, for the rad only state there were some [adjustments ](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764542492751225&cot=14) requested by the design team. So the read only in this PR looks correct and should be applied to other Inputs as well.d only states should not have any hover effect. Good catch!
> 
> Ah! Ok. good to know 😄
> 
> @ddouglasz, perhaps you can do it as a follow-up PR maybe?

Thanks Carlos, I did not notice that initially. 
updated here:  https://github.com/commercetools/ui-kit/pull/2373/commits/0481ec2842f1326881566266915b632c93e3565c

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -43,25 +44,35 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
...
       if (props.hasError) return designTokens.borderColorForInputWhenError;
       if (props.hasWarning) return designTokens.borderColorForInputWhenWarning;
-      if (props.isReadOnly) return designTokens.borderColorForInputWhenReadonly;
+      if (props.isReadOnly)
+        return designTokens.borderColorForSelectInputWhenReadonly;
       return designTokens.borderColorForInput;
     })(),
+    borderWidth: (() => {
+      if (props.hasError) return designTokens.borderWidthForSelectInput;
+      if (props.hasWarning) return designTokens.borderWidthForSelectInput;
```
CarlosCortizasCT, line 65: It seems both conditions use the same value:
```suggestion
      if (props.hasError || props.hasWarning) return designTokens.borderWidthForSelectInput;
```

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -158,8 +182,10 @@ const optionStyles = () => (base: TBase, state: TState) => {
     transition: `border-color ${designTokens.transitionStandard},
       background-color ${designTokens.transitionStandard},
       color ${designTokens.transitionStandard}`,
-    paddingLeft: designTokens.spacing20,
-    paddingRight: designTokens.spacing20,
+    paddingLeft: designTokens.paddingLeftForSelectInputOptions,
+    paddingRight: designTokens.paddingRightForSelectInputOptions,
+    paddingTop: `${designTokens.paddingTopForSelectInputOptions}`,
+    paddingBottom: `${designTokens.paddingBottomForSelectInputOptions}`,
```
CarlosCortizasCT, line 188: We don't need the template literal here:
```suggestion
    paddingTop: designTokens.paddingTopForSelectInputOptions,
    paddingBottom: designTokens.paddingBottomForSelectInputOptions,
```

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -285,45 +315,50 @@ const menuPortalStyles = (props: TProps) => (base: TBase) => ({
...
-    backgroundColor: designTokens.backgroundColorForTag,
-    padding: '0',
+const multiValueStyles =
+  (props: TProps & { isNewTheme: boolean }) => (base: TBase) => {
+    return {
+      ...base,
+      height: designTokens.sizeHeightTag,
+      backgroundColor: designTokens.backgroundColorForTag,
+      padding: '0',
+      border: props.isNewTheme && `1px solid ${designTokens.colorNeutral85}`,
```
CarlosCortizasCT, line 325: In case `props.isNewTheme` is _false_, the `border` value would be _'false'_. I guess we would want something like this:
```suggestion
      border: props.isNewTheme ?? `1px solid ${designTokens.colorNeutral85}` : 'none',
```

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -285,45 +315,50 @@ const menuPortalStyles = (props: TProps) => (base: TBase) => ({
...
+  (props: TProps & { isNewTheme: boolean }) => (base: TBase, state: TState) => {
     return {
       ...base,
       borderColor: designTokens.borderColorForTag,
       padding: `0 ${designTokens.spacing10}`,
       borderRadius: `0 ${designTokens.borderRadiusForTag} ${designTokens.borderRadiusForTag} 0`,
-      borderStyle: 'solid',
-      borderWidth: '1px',
+      borderStyle: !props.isNewTheme && 'solid',
+      borderWidth: !props.isNewTheme && designTokens.borderWidth1,
```
CarlosCortizasCT, line 361: We need alternative values for the use case when `props.isNewTheme` is _false_.

```suggestion
      borderStyle: !props.isNewTheme ? 'solid' : 'none',
      borderWidth: !props.isNewTheme ? designTokens.borderWidth1 : '0',
```

CarlosCortizasCT (commented):
Great progress 👍 

I've left some comment with some minors things to maybe adjust.

CarlosCortizasCT:
> @CarlosCortizasCT Yes, for the rad only state there were some [adjustments ](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764542492751225&cot=14) requested by the design team. So the read only in this PR looks correct and should be applied to other Inputs as well.

@ddouglasz please also don't forget to apply the adjustments to the other inputs as a follow-up 👍

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: Thank you.
All updated here: https://github.com/commercetools/ui-kit/pull/2373/commits/9123d3cfb0ba7a0b5371e8cd341a0e679b265e4d

CarlosCortizasCT (approved):
Awesome! 👏 👏 👏

.changeset/poor-hats-compete.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/design-system': patch
+'@commercetools-uikit/search-select-input': patch
+'@commercetools-uikit/select-utils': patch
+---
+
+Add new design tokens for search select input component.
```
Rhotimee, line 7:
```suggestion
Add new design tokens for select input components.
```

The changeset should highlight the select-input components you worked on in this PR. Most of them are missing here.

Rhotimee (commented):
Generally looks good. 
I noticed that the fontColor when in error state (also applies to warning) is initially color-error. However, when a value is selected, it changes to color-solid. 
I don't think the color should change after selecting a value. 

image

.changeset/poor-hats-compete.md
ddouglasz: Thank you, 
Updated here: https://github.com/commercetools/ui-kit/pull/2373/commits/49476539925ae803f7cbad5b5d5ae9da7c71da16

Rhotimee (approved):
Nice work

tdeekens (approved):

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -285,45 +300,50 @@ const menuPortalStyles = (props: TProps) => (base: TBase) => ({
...
+        borderRadius: designTokens.borderRadiusForTag,
+        borderWidth: '1px',
+      },
+    };
   };
-};
 
 const multiValueRemoveStyles =
-  (props: TProps) => (base: TBase, state: TState) => {
+  (props: TProps & { isNewTheme: boolean }) => (base: TBase, state: TState) => {
```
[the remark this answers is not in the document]
ddouglasz: Thanks for the catch! 
Updated here: https://github.com/commercetools/ui-kit/pull/2373/commits/1437fa46d6faa20159c1ea01e4d6ac8de64accaf

PR #2379, opened 2023-01-10
Increase `Tooltip` component `z-index` value

### LINKED ISSUE #2378: Increase `Tooltip` component `z-index` value
The problem we might find is for a `Tooltip` component not being rendered on top of other elements in the page.

Currently this component has a `z-index` CSS property with a value of *1* which is a lower value so, if any other component nearby the location where the tooltip is supposed to be rendered has a greater `z-index`, the tooltip will be displayed behind it.

Since tooltips use to contain very important context information, I think they should always be on top of everything else.
I would suggest to increase the `z-index` to a much higher value (for instance, *1000*). It's not 100% safe, but it will highly improve the chances of tooltip to not be rendered behind another elements in the page.

None.

**Additional context**
I've based this request on [this documentation](https://atlassian.design/foundations/elevation#z-index) from Atlassian design system where they have a `z-index` scale having the tooltips with the highest value.

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=xoF6yx3)

#### Summary

Increase `Tooltip` component `z-index` value

## Description

Raise the value to 1000 to play safe with other elements in the page since we want to tooltip to always be in the top.

Closes #2378


.changeset/late-donkeys-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Increate `z-index` CSS property to 1000.
+_(previously was 1)_
```
emmenko, line 6:
```suggestion
Increase `z-index` CSS property to `1000` _(previously was `1`)_
```

emmenko (approved):
Thanks!

kark (approved):

PR #2382, opened 2023-01-12
Fix `TextInput` background color style when hovered

CarlosCortizasCT:
#### Summary

Fix `TextInput` background color style when hovered

## Description

It seems we introduced a regression bug when working in `TextInput` adjustments to support the new theme.
When hovering the component, its background color is changed to transparent so, if the component is in a container with any background color, the input itself changes its background color as well.

This is an example of this behaviour:
Components _ Inputs - TextInput ⋅ Storybook (1)

This is the behaviour after applying the fix:
Components _ Inputs - TextInput ⋅ Storybook (2)


design-system/materials/internals/definition.yaml
```diff
@@ -387,10 +387,10 @@ decisionGroupsByTheme:
           choice: color-accent-98
         background-color-for-input-when-hovered:
           description: ''
-          choice: 'unset'
+          choice: color-surface
         background-color-for-input-when-focused:
           description: ''
-          choice: 'unset'
+          choice: color-surface
```
CarlosCortizasCT, line 393: This is the actual fix

docs/.storybook/decorators/section/section.js
```diff
@@ -15,6 +15,9 @@ const Section = (props) => {
   );
 };
 
-Section.propTypes = { children: PropTypes.node.isRequired };
+Section.propTypes = {
```
CarlosCortizasCT, line 18: This is to allow changing the background color for stories containers.

packages/components/inputs/text-input/src/text-input.story.js
```diff
@@ -17,7 +17,16 @@ storiesOf('Components|Inputs', module)
   })
   .add('TextInput', () => {
     return (
-      <Section>
+      <Section
```
CarlosCortizasCT, line 20: This allows to change the background color of the story container to check the correct behaviour.

tdeekens (approved):

Rhotimee (approved):
💯

PR #2383, opened 2023-01-12
Refactor Stamp component to accept only text and message inline props.

ddouglasz:
#### Summary

We want the Stamp component to be more strict regarding what it can render as children.
Instead of accepting any children, the component should have a new mandatory prop with the text to render and another optional prop with the icon to use.

- [x] `overrideTextColor` code should only be used when props.children is used
- [x] pass inherit property to child component
- [x] log warning if users use old implementation.
- [x] implementation should be backwards compatible 
- [x] prevent using both new and old implementation at the same time


packages/components/stamp/src/stamp.tsx
```diff
@@ -21,10 +24,24 @@ type Props = {
    */
   isCondensed: boolean;
   children: ReactNode;
+  icon?: ReactElement;
+  message: MessageDescriptor & {
```
CarlosCortizasCT, line 28: In think `message` property should be optional for backwards compatibility.

packages/components/stamp/src/stamp.tsx
```diff
@@ -21,10 +24,24 @@ type Props = {
...
+    values?: Record<string, ReactNode>;
+  };
+  iconSize?: string;
```
CarlosCortizasCT, line 31: It seems we don't need this prop.

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,16 +150,42 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
         getToneStyles({ ...props, overrideTextColor }),
         getPaddingStyle(props),
       ]}
     >
-      {props.children}
+      <SpacingsInline alignItems="center">
+        {Icon}
+        <Text.Detail tone={themedValue(undefined, 'inherit')}>
+          <FormattedMessage {...props.message} />
+        </Text.Detail>
```
CarlosCortizasCT, line 187: You can use the `intlMessage` prop from the `Text.Detail` component to pass it the message so you don't have to use the `FormattedMessage` component in this one.

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,16 +150,42 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
   const overrideTextColor = themedValue(false, true);
+  const Icon =
+    props.icon &&
+    cloneElement(props.icon, {
+      size: 'medium',
+      color: props.tone ? iconColorsMap[props.tone] : null,
+    });
 
-  return (
+  return props.children ? (
```
CarlosCortizasCT, line 160: I would give priority to the new use case. Also, perhaps it would more clear to not use a ternary here:
```jsx
if (props.message) {
  return (....);
}

// use children
return (....)
```

CarlosCortizasCT (commented):
This is going in the right direction 👏 

I've left some suggestions to maybe improve little things.

packages/components/stamp/src/stamp.tsx
ddouglasz: Thank you. Updated here: https://github.com/commercetools/ui-kit/pull/2383/commits/f9e56c683f1769bbd086e070a943603014706791

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,17 +150,74 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
+    return (
+      <div
+        css={css`
+          fill: ${designTokens.colorError};
+          color: ${designTokens.colorError};
+        `}
+      >
+        <SpacingsInline alignItems="center">
+          <ErrorIcon />
+          <Text.Detail tone={'critical'}>
```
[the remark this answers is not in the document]
CarlosCortizasCT: The idea here is not to render anything different to the end user but to let the developer know that we now prefer the `Stamp` component to be used with the inline props.

For some period of time, we want to allow the component used both ways where inline props would have more priority.

I think we don't need this code branch but just replace the `console.log` we're using in the `children` branch by a `warning` as we do in other components ([example](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/password-field/src/password-field.tsx#L181))

CarlosCortizasCT:
Bear in mind we would like to remove the `important` keywords from the CSS rules as we won't be needing them anymore.

Perhaps you can take the opportunity to also refactor the way we calculate the tone styles using [this example](https://github.com/commercetools/ui-kit/pull/2380/files#diff-dee34db27718f903dbdc29a7a1d220ef5f7aa14410e90a325eedc5e1d9aa0063) as a guideline.

packages/components/stamp/src/stamp.tsx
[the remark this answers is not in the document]
ddouglasz: Oh yes, that is correct. But this one is in the case where the component is not use as expected. where the user passes the inline prop and at the same time, uses prop.children too. I believe we should give the user some notice that these two cannot be used at the same time. 
something like this: 
```
<Stamp tone={tone} isCondensed={boolean('isCondensed', false)} message={{ id: 'stamp-message', defaultMessage: tone }} icon={<Icon />} >
    <Text.Detail>{`tone="${tone}"`}</Text.Detail>
</Stamp>
```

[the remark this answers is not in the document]
CarlosCortizasCT: Yes, I understand, but we need to differentiate between a warning for the developer and a warning for the end user.

In case a user provides both the inline props and the children, we would simply ignore the children and use the inline props.
Personally I think this is not an expected use case for a developer so perhaps we don't need to add code for it, but if you feel it's better to catch this scenario, we can use the `warning` utility as we do in other components.
Example:
```js
warning(
  this.message && this.children,
  'Stamp: `children` prop is ignored as `message` was provided and it has more priority'
);
```

Does it make sense to you?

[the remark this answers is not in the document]
ddouglasz: Yes this makes sense now. Thank you.

packages/components/stamp/src/stamp.tsx
```diff
@@ -1,7 +1,11 @@
 // TODO: @redesign cleanup
-import type { ReactNode } from 'react';
+import { cloneElement, ReactElement, ReactNode } from 'react';
```
emmenko, line 2:
```suggestion
import { cloneElement, type ReactElement, type ReactNode } from 'react';
```

packages/components/stamp/src/stamp.tsx
```diff
@@ -1,7 +1,11 @@
...
 import { css } from '@emotion/react';
 import { designTokens, useTheme } from '@commercetools-uikit/design-system';
+import { MessageDescriptor } from 'react-intl';
```
emmenko, line 5:
```suggestion
import type { MessageDescriptor } from 'react-intl';
```

packages/components/stamp/src/stamp.tsx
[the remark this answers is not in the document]
emmenko: Yes I agree with @CarlosCortizasCT , we should only log a warning.

> In case a user provides both the inline props and the children, we would simply ignore the children and use the inline props.

⬆️ That would also be my expectation about the behavior.

packages/components/stamp/src/stamp.tsx
```diff
@@ -21,10 +25,53 @@ type Props = {
...
   children: ReactNode;
+  icon?: ReactElement;
+  message?: MessageDescriptor & {
+    values?: Record<string, ReactNode>;
+  };
 };
 
 type StylesFunctionParams = Props & { overrideTextColor?: boolean };
 
+const tonesStylesMap = {
```
emmenko, line 36: Let's rely on TS to make sure that the keys here are valid.

```suggestion
const tonesStylesMap: Record<Tone, CSSProperties> = {
```

I would also decouple the `iconTone` into a separate function (see other comment).

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,17 +117,55 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
   const { themedValue } = useTheme();
   const overrideTextColor = themedValue(false, true);
+  const iconTone = props.tone ? tonesStylesMap[props.tone].iconTone : '';
+  console.log({ iconTone });
+  const Icon =
+    props.icon &&
+    cloneElement(props.icon, {
+      size: 'medium',
+      color: overrideTextColor ? iconTone : 'inherit',
+    });
```
emmenko, line 127: I would suggest to extract the logic around the icon color into a function as it relies on the `props.tone` and `overrideTextColor`. Remove it then from the `tonesStylesMap`.

```ts
const getIconColor = (props: Props, overrideTextColor: boolean) => {
  if (!overrideTextColor) {
    return 'inherit';
  }
  switch (props.tone): {
    case 'critical':
      return 'error';
    // ...
    default:
      return '';
  }
}
```

We should also think if we want to have a default `tone`, or validate that one is provided.

packages/components/stamp/package.json
```diff
@@ -22,6 +22,10 @@
     "@babel/runtime": "^7.19.0",
     "@babel/runtime-corejs3": "^7.19.1",
     "@commercetools-uikit/design-system": "15.9.0",
+    "@commercetools-uikit/icons": "15.9.0",
```
emmenko, line 25: I think this can be removed.

emmenko (commented):
Good work, I left some minor suggestions 🤗

CarlosCortizasCT:
There's something I was warned about about this component that maybe we can include in this PR.

When using the `Stamp` component in a _flexed_ container, it can span container's whole width by default when that would not be the expected behaviour. Here is an example:

Default behaviour
image

Expected behaviour
image

I think the `Stamp` root `div` element should have `display: inline-block` and it should only use its contents width.

packages/components/stamp/src/stamp.tsx
CarlosCortizasCT: I'm hesitant about this. I believe having the map of styles tied to the `tone` provided to the `Stamp` helps in grouping concerns since that prop is the one that controls the other styles.

I created [this PR](https://github.com/commercetools/ui-kit/pull/2393/) with the way I see this could work where I think it would be easy to follow how the styles of the component work.

CarlosCortizasCT (approved):
Awesome! 👏

packages/components/stamp/src/stamp.tsx
```diff
@@ -20,11 +24,75 @@ type Props = {
    * If `true`, renders a condensed version of the stamp.
    */
   isCondensed: boolean;
-  children: ReactNode;
+  children?: ReactNode;
+  icon?: ReactElement;
+  message?: MessageDescriptor & {
```
emmenko, line 29: Usually we name this prop `intlMessage`. Any reason for naming it `message`?

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,6 +151,17 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
+  const Icon =
+    props.icon &&
+    cloneElement(props.icon, {
+      size: 'medium',
+      color: getIconColor(props, overrideTextColor),
+    });
+
+  warning(
+    !props.children,
+    'Stamp: Please pass messages or icons as inline props, this method will be deprecated soon. For more information, see documentation: https://uikit.commercetools.com/?path=/story/components-stamps--stamp'
```
emmenko, line 163:
```suggestion
    'Stamp: The `children` prop has been deprecated. Please use the `intlMessage` and `icon` prop to render the content.'
```

packages/components/stamp/src/stamp.tsx
emmenko: Also, I think we need to add a new prop here. The `intlMessage` prop is usually an alternative to another prop that can be used if you don't use messages.

If we are deprecating `children`, we should add something like `label` (which is a `string`) to accept the text value. Then the user should chose between `label` or `intlMessage`.

emmenko (changes requested):
I think we missed one prop as we are deprecating the `children` prop (see comment).

Otherwise the rest looks good 👌

packages/components/stamp/src/stamp.tsx
CarlosCortizasCT: Super useful hint about the purpose of `intlMessage` 🙇 

I was wondering though about what you suggest of having both `label` and `intlMessage`. It seems the stamp it's about rendering text (and an optional icon) so I would expect (as a consumer) to only have a (maybe) `label` property which could be a `string` (in case you don't use translations) or a `MessageDescriptor`.
Aren't we supposed to have such typing for a single property?

packages/components/stamp/src/stamp.tsx
ddouglasz: Thank you.
updated here: https://github.com/commercetools/ui-kit/pull/2383/commits/18626634a4f71e9f068758455e03af90a99a8d9f

packages/components/stamp/src/stamp.tsx
CarlosCortizasCT: +1 to have only the `label` prop as a _string_.

After all, we want to control the typography from the component so it's perhaps better to now allow `Text.*` components.

Thanks!

ddouglasz: Good points. So what we are suggesting is the trade-off is worth it, that is no translations. Just use the label property to pass in string text only.

CarlosCortizasCT: Actually, what we suggest is to only have the `label` property with the _string_ type, but it's up to consumers to provide a translated message. The could do it using the `useIntl` hook.
For instance:
```jsx
const const intl = useIntl();

return (
  <Stamp label={intl.formatMessage(messages.ssoConfigStatusLabel)} />
);
```

ddouglasz: Ah, I get it now. that makes sense. I will update it now

ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2383/commits/05a1ecd9e5cc1d174b6efa700f439ad63c10a264

Thank you

packages/components/stamp/README.md
```diff
@@ -44,9 +44,11 @@ export default Example;
...
+| `isCondensed` | `boolean`                                                                                                     |          | `false`         | If `true`, renders a condensed version of the stamp. |
+| `children`    | `ReactNode`                                                                                                   |          |                 |                                                      |
+| `icon`        | `ReactElement`                                                                                                |          |                 |                                                      |
+| `label`       | `string`                                                                                                      |          |                 | used as an alternative to `intlMessage` prop         |
+| `intlMessage` | `intersection`                                                                                                |          |                 |                                                      |
```
CarlosCortizasCT, line 54: we need to re-generate the readme ♻️

ddouglasz: Thanks! I missed that part. 
Updated here: https://github.com/commercetools/ui-kit/pull/2383/commits/0a0ead4b3f8d3b5ff707c972d064b93470513995

CarlosCortizasCT (approved):
Awesome! 🎉

packages/components/stamp/src/stamp.tsx
```diff
@@ -133,6 +148,17 @@ const getStampStyles = (props: StylesFunctionParams) => {
...
+    props.icon &&
+    cloneElement(props.icon, {
+      size: 'medium',
+      color: getIconColor(props, overrideTextColor),
+    });
+
+  warning(
+    !props.children,
+    'Stamp: The `children` prop has been deprecated. Please use the `label` and `icon` prop to render the content.'
+  );
```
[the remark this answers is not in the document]
ddouglasz: I think we can use both since this one just warns the user that old implementation is deprecated.
```js
  warning(
    !props.children,
    'Stamp: The `children` prop has been deprecated. Please use the `label` and `icon` prop to render the content.'
  );
```
while this one seems to check if the component is used correctly (provide either one of the props expected) 
```js
   warning(
    hasContent,
    [
      'ui-kit/Stamp:',
      '"label" is required when "children" is `undefined` and vice versa.',
      'Provide "label" or "children"',
    ].join(' ')
  );
```
What do you think ?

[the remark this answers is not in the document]
emmenko: Ah nevermind, we only want to deprecate the `children` prop yes. All good then, thanks for double checking!

emmenko (approved):
🤗

PR #2386, opened 2023-01-17
Update svgr config to include role=img

Rhotimee:
#### Summary
`role="img"` is no longer present by default. Instead, it can be introduced as an optional config. 
This PR is to update the `svgr.config.js` to include `role="img"`.

https://github.com/gregberge/svgr/compare/v6.5.0...v6.5.1


CarlosCortizasCT:
When generating the icons in this PR all of them have small changes regarding removing break lines.
If this is caused from the new version, I think we should include those changes in this PR.

Rhotimee:
> When generating the icons in this PR all of them have small changes regarding removing break lines. If this is caused from the new version, I think we should include those changes in this PR.

Yes, you're right. I didn't include it here because that was already in the Tag PR and I didn't want to deal with the conflict.
Since that has been reverted, I have included it [3b83d5c](https://github.com/commercetools/ui-kit/pull/2386/commits/3b83d5c78ad199e8d142492bf9a69c01550778f7)

svgr.config.js
```diff
@@ -92,6 +92,8 @@ module.exports = {
       'prefixIds',
     ],
   },
+  svgProps: { role: 'img' },
+  typescript: true,
```
emmenko, line 96: 💯

emmenko (approved):
Thanks!

CarlosCortizasCT (approved):
👍

PR #2391, opened 2023-01-20
Add HOC to access theme context utilities in React class components

CarlosCortizasCT:
#### Summary

Add HOC to access theme context utilities in React class components

## Description

Currently we provide the `useTheme` hook to access some theme utilities that can be used to tailor components based on the selected theme.

Since hooks can only be used in functional components, in this PR we're introducing a new HOC which can be used in class components to access the same utilities.

Example:
```jsx
class DummyComponent extends Component {
    render() {
      return (
        <div>
          <p>Current theme is: {this.props.theme}</p>
          <p>Themed value: {this.props.themedValue('foo', 'bar')}</p>
        </div>
      );
    }
  }

  const WrappedComponent = withThemeContext(DummyComponent);
```


tdeekens (approved):

CarlosCortizasCT:
@commercetools/shield-team-fe I would appreciate some help with the [build error](https://github.com/commercetools/ui-kit/actions/runs/3968574439/jobs/6801913144).

It complains about a dependency not listed in the `package.json` but I don't see where am I using that dependency.

image

tdeekens:
@CarlosCortizasCT Is it cause you're using JSX now? Could it be that adding react, react-dom and @types/react solves it? It mainly preconsturct's analysis being not happy I think.

design-system/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@babel/runtime": "^7.19.0",
     "@babel/runtime-corejs3": "^7.19.1",
+    "@emotion/react": "^11.4.0",
```
emmenko, line 28: Is this needed? 🤔

emmenko (commented):
Thanks! 💯

design-system/package.json
emmenko: Ah I saw the [comment](https://github.com/commercetools/ui-kit/pull/2391#issuecomment-1400043487) now

PR #2394, opened 2023-01-24
Allow using icon only buttons in `ViewSwitcher`

CarlosCortizasCT:
#### Summary

Allow using icon only buttons in `ViewSwitcher`

## Description

We found out that there's one use case for the `ViewSwitcher` component that is not 100% supported currently: when someone wants to use buttons with no text but only icons.

Example:
image

Allowing this use case also implied to include a new optional property called `label` in the `ViewSwitcher.Button` which can be used to label the button in such case (when text it's used, that's what we use as the label).


packages/components/view-switcher/src/view-switcher-button.tsx
```diff
@@ -7,52 +7,66 @@ import {
...
+  );
+
+  warning(
+    Boolean(props.label || props.children),
+    'uikit/ViewSwitcherButton: You need to provide a `label` when rendering only an `icon` within the button.'
+  );
+
+  return (
+    <AccessibleButton
+      label={props.label ?? props.children ?? ''}
```
tdeekens, line 41: Note to self: `children` can only only be a string.

.changeset/slimy-donuts-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+'@commercetools-uikit/view-switcher': patch
+---
+
+Update `ViewSwitcher.Button` component to allow using it using no text and just an icon.
+
+It's `children` prop is now options.
```
tdeekens, line 7:
```suggestion
It's `children` prop is now optional.
```

packages/components/view-switcher/src/view-switcher-button.tsx
```diff
@@ -7,52 +7,66 @@ import {
...
-  </AccessibleButton>
-);
+const ViewSwitcherButton = (props: TViewSwitcherButtonProps) => {
+  warning(
+    Boolean(props.children || props.icon),
+    'uikit/ViewSwitcherButton: You need to provide at least the `children` to render inside the button or an `icon`.'
+  );
+
+  warning(
+    Boolean(props.label || props.children),
```
tdeekens, line 35: Ok, I looked at the conditions for 3m but it's late-ish: aren't both conditions the same?

CarlosCortizasCT: Previously the component had the `children` prop as required as we expected to always us it with some text. The `icon` property was optional.
Now we can use just an `icon` so `children` both props are optional, but we expect consumers to use at least one of them.
This is what the first condition is for.

Then, if consumer only wants to use an icon, we need him/her to provide a label for the icon (when using children, we use it as the label).
The second condition makes sure we have something for the button label.

Does it makes sense?

.changeset/slimy-donuts-sort.md
CarlosCortizasCT: Thanks!

Updated here: 7358cb5c06e336e6b5a42d69be74b48bcdddd197

packages/components/view-switcher/src/view-switcher-button.tsx
tdeekens: Thanks. Got it.

tdeekens (approved):

.changeset/slimy-donuts-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+'@commercetools-uikit/view-switcher': patch
+---
+
+Update `ViewSwitcher.Button` component to allow using it using no text and just an icon.
+
+It's `children` prop is now optional.
```
tdeekens, line 7: Nit: It is "Its" here.

Its (without an apostrophe) is a possessive pronoun, like his or her, for nouns that don't have a defined gender. In contrast, it's (with an apostrophe) is the shortened form, or contraction, of it is or it has.

tdeekens (approved):

.changeset/slimy-donuts-sort.md
CarlosCortizasCT: Sorry, I missed that one 🤦 

Updated here: 11fc8a45c24b60d79b8191e700c5e87999e5782f

CarlosCortizasCT:
> Looks good. Only one detail: When only icons are visible, the icon keeps the margin-right of 8px which makes the icon not centered properly. The margin is only needed when there is icon+label image

Thanks for spotting that one, @FilPob 👍 

It has been addressed in this commit: 11fc8a45c24b60d79b8191e700c5e87999e5782f
_(you can already check again)_

FilPob (approved):

CarlosCortizasCT:
FYI @YahiaElTai

YahiaElTai:
> FYI @YahiaElTai

Thanks so much for the quick update! Once this is ready in the MC, I will create a PR to remove all the unneeded code!

packages/components/view-switcher/src/view-switcher-button.tsx
emmenko: What do we need to do about that?

.changeset/slimy-donuts-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+'@commercetools-uikit/view-switcher': patch
+---
+
+Update `ViewSwitcher.Button` component to allow using it using no text and just an icon.
```
emmenko, line 5:
```suggestion
Update `ViewSwitcher.Button` component to allow using it without text and just an icon.
```

emmenko (approved):
Thanks, looks good 👌

Rhotimee (approved):

kark (approved):

CarlosCortizasCT:
> What do we need to do about that?
_([reference](https://github.com/commercetools/ui-kit/pull/2394#discussion_r1087716407))_

This is the case where a consumer would be providing only an icon and no label for the button. We have a warning for this case as we don't have that value.
In that case, I don't think we have a label to use. The only thing I can think of is trying to use the icon name.

What would you expect for us to use in the scenario?

ddouglasz (approved):

emmenko:
> What would you expect for us to use in the scenario?

It's fine the way it is. Thanks for checking.

PR #2412, opened 2023-02-02
Update background colors for inputs when readonly and background colors when hover

ddouglasz:
Summary:

Minor Input changes for input fields - background color when read-only and on hover:
Input fields affected include:

1. localizedmultilinetextinput
2. localizedrichtextinput
3. localizedtextinput
4. richtextinput
5. timeinput


CarlosCortizasCT:
Some issues I found (asides the ones in Percy) in the new theme:
* Checkbox: check input is not visible when component is read only
* MoneyInput: hover state in currency selector is not correct (background color) when component is read only or disabled
* SearchTextInput: background color is not correct when component is read only
* TimeInput: unexpected vertical border in icons when component is read only or disabled

FilPob:
+1 to the points from Carlos

Basically none of the inputs should have a hover effect when they are disabled or read-only. This is currently not the case in:

- Date, Time, Daterange Input
- Currency dropdown in Moneyinput

The point about the checkbox not being visible in read only state. I think that will be fixed after applying the new tokens to the checkbox component

PR #2415, opened 2023-02-08
Fix `FieldLabel` vertical spacing issue

### LINKED ISSUE #2413: Unexpected bottom margin in `FieldLabel` component
**Describe the bug**
[`FieldLabel`](https://uikit.commercetools.com/?path=/story/components-fieldlabel--fieldlabel) component has two **optional** properties (`hint` and `label`) that, when they receive another component as a value and this component render conditionally, when it does not render anything, we can see an unexpected bottom margin in the `FieldLabel`.

**To Reproduce**
This is an example of a code leading to this issue:

```jsx
const OrderDescription = () => {
  if (some_condition) {
    return 'You need to include your door number.';
  } else {
    return null;
  }
}

const App () => (
  <FieldLabel
    title="Address"
    description={<OrderDescription>}
  />
);
```

When `some_condition` is *falsy*, we will see the unexpected bottom margin.

[Here's a codesandbox](https://codesandbox.io/s/charming-silence-xy6its?file=/src/App.js) to demonstrate the issue. You can see the second field label has an extra vertical space since we're using a component which renders null for its description.

**Expected behavior**
No extra bottom margin should exists if there is no `description` or `hint` rendered.

**Screenshots**
Here is a screenshot where the first example is using a `FieldLabel` but we're not passing anything to its `description` value. For the second example, we're passing a component as `description` value which renders nothing:
image

**Additional context**
The `FieldLabel` [uses](https://github.com/commercetools/ui-kit/blob/main/packages/components/field-label/src/field-label.tsx#LL103C6-L103C25) a `Stack` component to vertically separate its children, but it also uses some wrappers for the `hint` and `description`. It checks whether these values are *truth* before rendering ([example](https://github.com/commercetools/ui-kit/blob/main/packages/components/field-label/src/field-label.tsx#L152)) but, when providing a component which might render *null*, the component itself it *truthy* so it gets rendered (empty) within its wrappers.
Since we're using the `Stack` components, even though there is no description, the wrappers get rendered and the CSS rules to separate `Stack` children kicks in and include the unexpected bottom margin.

CarlosCortizasCT:
#### Summary

This PR closes #2413

## Description

The idea is to not use the `Stack` component and instead use a custom styled component to deal with vertical spacing between `FieldLabel` *rows* so we don't actually use the vertical margin if those *rows* don't render anything.

We're using the [`:empty`](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) pseudo-class to achieve the expected behavior.


packages/components/field-label/src/field-label.tsx
```diff
@@ -83,6 +83,28 @@ export type TFieldLabelProps = {
...
+  when provided `hint` or `description` is a React component which
+  might render nothing.
+  Previously we were using the `Stack` component but, as we are wrapping
+  those props values with some elements, `Stack` was including some vertical
+  spacing even when the received values did not render anything.
+
+  The implementation is tightly coupled to how we currently wrap
+  those props, so if we change that, we will need to adjust this as well.
+*/
+const LabelRowWrapper = styled.div`
```
CarlosCortizasCT, line 97: This is the new component we use to conditionally add the vertical spacing between `FieldLabel` *rows*.

First selector is for the `hint` wrapper and the second one is for the `description` one.

packages/components/field-label/src/field-label.tsx
```diff
@@ -100,33 +122,33 @@ const FieldLabel = (props: TFieldLabelProps) => {
 
   return (
     <Constraints.Horizontal max={props.horizontalConstraint}>
-      <Stack scale="xs">
```
CarlosCortizasCT, line 103: We get rid of the `Stack` component.

packages/components/field-label/src/field-label.visualroute.jsx
```diff
@@ -3,10 +3,16 @@ import {
   WarningIcon,
   FlatButton,
 } from '@commercetools-frontend/ui-kit';
+import styled from '@emotion/styled';
 import { Suite, Spec } from '../../../../test/percy';
 
 export const routePath = '/field-label';
 
+const BorderedBox = styled.div`
```
CarlosCortizasCT, line 11: These are components to help with the new VRT where we render the `FieldLabel` with a description which is a component rendering nothing to make sure the unexpected vertical spacing does not exist anymore.

packages/components/field-label/src/field-label.tsx
```diff
@@ -83,6 +83,28 @@ export type TFieldLabelProps = {
...
+*/
+const LabelRowWrapper = styled.div`
+  & > span,
+  & > div > div {
+    margin-top: ${designTokens.spacing10};
+  }
+  & > span:empty,
+  & > div > div:empty {
+    margin-top: 0;
+  }
```
emmenko, line 105: I think we should we able to pass data attributes to the `Text.Detail` components and use them as selectors. Have you tried that?

emmenko (commented):
Thanks for looking into it. I think this is more or less the only way to fix that, I also can't think of another approach.

The VRT will definitely help with accidental regressions, so thanks for that!

packages/components/field-label/src/field-label.tsx
CarlosCortizasCT: Nope, I didn't think of that. Many thanks for the suggestion 🙇 

Updated here: 1b68ed7b748cbd3ff70b03226a251a950bb1e609

emmenko (approved):
💯

emmenko:
VRTs look good ✅

kark (approved):
Great idea with using `:empty` 👏

Rhotimee (approved):

packages/components/field-label/src/field-label.tsx
```diff
@@ -83,6 +83,28 @@ export type TFieldLabelProps = {
...
+  those props values with some elements, `Stack` was including some vertical
+  spacing even when the received values did not render anything.
+
+  The implementation is tightly coupled to how we currently wrap
+  those props, so if we change that, we will need to adjust this as well.
+*/
+const LabelRowWrapper = styled.div`
+  & [data-key='field-label-hint-wrapper'],
+  & [data-key='field-label-description-wrapper'] {
+    margin-top: ${designTokens.spacing10} !important;
```
CarlosCortizasCT, line 100: I noticed a little misalignment between hint icon and hint text:
// how it was
Screenshot 2023-02-09 at 09 26 28

// how it should be
Screenshot 2023-02-09 at 09 26 14

This is due to the `Spancing.Inline` component we're using for the horizontal separation between both elements:
Screenshot 2023-02-09 at 09 19 14

As you can see, the `Inline` component is overwriting not only the horizontal margins but also the vertical ones so the rule here does not apply.

I had three ideas:
* Using `!important` here
 * the easiest one
* Remove the `Spacing.Inline` element and using a custom one for this component to separate the hint icon and text
 * not makes a lot of sense to me to have a custom inline component here
* Update the `Spacing.Inline` css to not overwrite vertical margins as its concern is only about horizontal spacing
 * not sure about possible implications

@kark @emmenko please let me know what you think.

emmenko: Hmm let's use `!important` here, as this is mostly an issue within the field label.

CarlosCortizasCT: Thanks 👍 

I merged the PR with the `!important` solution but I would like to maybe have a second thought about the `Spacing.Inline` styles. I mean, is it on purpose that we want to have 0 vertical margin in that component? It feels that component is all about horizontal stuff to me so that's why I'm wondering.

Perhaps also @FilPob knows better 🤔

PR #2416, opened 2023-02-08
[`TimeInput`] Adjust clear icon visibility

CarlosCortizasCT:
#### Summary

We don't want to display the clear icon when there is no value in the input.

## Description

Here is an example of the new behavior:

[Components _ Inputs - TimeInput ⋅ Storybook.webm](https://user-images.githubusercontent.com/97907068/217501168-85176e67-a306-4be1-ab41-374466c9a9a4.webm)

### DIFF
--- a/.changeset/gorgeous-snails-sip.md
+++ b/.changeset/gorgeous-snails-sip.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/time-input': patch
+---
+
+Hide the clear icon when there's not value in the input.
--- a/packages/components/inputs/time-input/src/time-input-body.tsx
+++ b/packages/components/inputs/time-input/src/time-input-body.tsx
@@ -78,7 +78,7 @@ const TimeInputBody = forwardRef<HTMLInputElement, TTimeInputBodyProps>(
               : {})}
           />
 
-          {!props.isDisabled && !props.isReadOnly && (
+          {!props.isDisabled && !props.isReadOnly && Boolean(props.value) && (
             <ClearSection
               isDisabled={props.isDisabled}
               hasError={props.hasError}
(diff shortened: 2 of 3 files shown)


FilPob (approved):

kark (commented):
Thanks, it looks good to me in general. 

I only wanted to share some strange behavior - when we input some gibberish data and we clear the input, the clear icon is still visible.

https://user-images.githubusercontent.com/49066275/217508453-36547f54-1b22-4fd0-b555-65c25864737b.mp4

emmenko (approved):

kark (approved):

CarlosCortizasCT:
> Thanks, it looks good to me in general.
> 
> I only wanted to share some strange behavior - when we input some gibberish data and we clear the input, the clear icon is still visible.
> 

Ok, so this is a tricky one which relates to how we treat the internal state of the component and [some changes](https://github.com/commercetools/ui-kit/pull/2160/files#diff-4101c38f195dd12511293b301721434ca927780ddeffa5ec9686b55dc29d702c) @emmenko  made last year to better handle that part.

I was checking the value received in the `TimeInputBody` component to know whether to render the clear icon or not, but when handling the clear action, we were not resetting internal state nor the external one but clearing the value at the DOM level.

In this commit (3ed8beeef0638624962d4ed7bc2cdca1ae870e61) I updated the clear handling to also let consumer know about the new value (empty string) to it can re-render the component with it.

I also tried using the `ref` in the `TimeInputBody` using the DOM value, but it gets out of sync.

Let's see what you think about this change.

kark (approved):
> Let's see what you think about this change.

This is great 💯 thank you for looking into that 🙏

emmenko (approved):
Thanks @CarlosCortizasCT for looking into it! ❤️

PR #2418, opened 2023-02-09
Fix select input tags styles in the new theme

CarlosCortizasCT:
#### Summary

Fix select input tags styles in the new theme

## Description

The tags we use inside the select input components when using multiselection mode were too small in the new theme.

**BEFORE**
image

**AFTER**
image


CarlosCortizasCT:
FYI @jmcreasman

kark (approved):
Thanks! Looks good to me 💯

emmenko (approved):

PR #2420, opened 2023-02-15
Input components common styles adjustments

CarlosCortizasCT:
#### Summary

Input components common styles adjustments

## Description

This PR tries to address the new common UI requirements for input components around the `read-only` status and `hover` state.

All the input components have been addressed.

I've noticed the `RichTextInput` and  `LocalizedRichTextInput` components are still missing redesign work but not sure if this is something we will get from UX or if we should address them following other inputs patterns (@FilPob)


packages/components/inputs/localized-text-input/src/localized-text-input.styles.ts
```diff
@@ -1,5 +1,6 @@
 import { css } from '@emotion/react';
 import { designTokens } from '@commercetools-uikit/design-system';
+import { TLocalizedInputProps } from './localized-text-input';
```
emmenko, line 3:
```suggestion
import type { TLocalizedInputProps } from './localized-text-input';
```

emmenko (approved):
💯

emmenko:
Don't forget the changeset 😉

emmenko (approved):
Thanks!

PS: the changeset is still missing 😇

CarlosCortizasCT:
> Thanks!
> 
> PS: the changeset is still missing 😇

Thanks for the reminder 👍 

Added here: 4e974b3d05b817ef6fcbe50a32879961f5b5bfa8

.changeset/real-stingrays-battle.md
```diff
@@ -0,0 +1,97 @@
...
+'@commercetools-local/generator-readme': patch
+'@commercetools-uikit/spacings': patch
+'@commercetools-uikit/buttons': patch
+'@commercetools-uikit/hooks': patch
+'@commercetools-uikit/utils': patch
+'@commercetools-uikit/fields': patch
+'@commercetools-uikit/inputs': patch
+'@commercetools-frontend/ui-kit': patch
+'@commercetools-uikit/design-system': patch
+'@commercetools-uikit/i18n': patch
```
kark, line 94: I'm getting different results when running changestets 🤔 perhaps it's worth double-checking, these are the packages in my output:
```
'@commercetools-uikit/localized-multiline-text-input': patch
'@commercetools-uikit/localized-rich-text-input': patch
'@commercetools-uikit/localized-text-input': patch
'@commercetools-uikit/search-text-input': patch
'@commercetools-uikit/rich-text-utils': patch
'@commercetools-uikit/select-utils': patch
'@commercetools-uikit/input-utils': patch
'@commercetools-uikit/money-input': patch
'@commercetools-uikit/time-input': patch
'@commercetools-uikit/calendar-utils': patch
'@commercetools-uikit/design-system': patch
```

kark (approved):
🚀

FilPob:
I found there is no hover effect in the SearchTextInput and TimeInput. Other Inputs look good

Rhotimee:
For the new theme, there are currently different behaviour for inputs background color when isReadonly and focused. Also, for isReadonly, focused and hover. 
I think the background color should not change when input is readonly and then focussed or hovered. 
is that right @FilPob?

ddouglasz (approved):
🚀

Rhotimee:
Nice work @CarlosCortizasCT.
I noticed an inconsistency between the text input and select input when the components hasError and isReadOnly. 
For SelectInput, it doesn't show the error border but, text input shows it. I think we need to update SelectInput to act like text Input. 
This is also affecting MoneyInput since it uses both component styles. 

image

FilPob (approved):

CarlosCortizasCT:
> Nice work @CarlosCortizasCT. I noticed an inconsistency between the text input and select input when the components hasError and isReadOnly. For SelectInput, it doesn't show the error border but, text input shows it. I think we need to update SelectInput to act like text Input. This is also affecting MoneyInput since it uses both component styles.
> 
> image

Yes, thanks for pointing that out.

I hadn't finished with reviewing all components yet but, if I'm not mistaken, we agreed on leaving those cases out of the scope of this PR since they are considered edge cases.

I'll take a look though in case it's not too much work.

Rhotimee:
> Yes, thanks for pointing that out.
> 
> I hadn't finished with reviewing all components yet but, if I'm not mistaken, we agreed on leaving those cases out of the scope of this PR since they are considered edge cases.
> 
> I'll take a look though in case it's not too much work.


True, we can leave it out then. Everything else looks good.

Rhotimee (approved):
Nicely done 🙌🏽

.changeset/real-stingrays-battle.md
CarlosCortizasCT: Hey @kark 

I re-run the changesets command and I get the same previous results.

Do you mean you only see those 11 packages in your changeset file?

kark: Hmm, that's right - this is the list (checking on 981296c76):
```
'@commercetools-uikit/localized-multiline-text-input': patch
'@commercetools-uikit/localized-rich-text-input': patch
'@commercetools-uikit/selectable-search-input': patch
'@commercetools-uikit/localized-text-input': patch
'@commercetools-uikit/search-text-input': patch
'@commercetools-uikit/rich-text-utils': patch
'@commercetools-uikit/select-utils': patch
'@commercetools-uikit/input-utils': patch
'@commercetools-uikit/money-input': patch
'@commercetools-uikit/radio-input': patch
'@commercetools-uikit/time-input': patch
'@commercetools-uikit/calendar-utils': patch
'@commercetools-uikit/design-system': patch
```

CarlosCortizasCT: So there might be something I'm missing.

For instance, `@commercetools-uikit/notifications` does appear in my list but not in yours. This package depends on `@commercetools-uikit/design-system` so I guess it should be in the list. But then I also have `@commercetools-uikit/i18n` in my list which does not depend on any local package, so it should not be in the list.

I've double checked my `changeset` installed version is `2.26.0`.

@emmenko Did anything like this happened to you before?

emmenko: Make sure to have the latest `main` pulled and that the branch is rebased.

image

PR #2421, opened 2023-02-15
Refactor warning utility to not throw errors but log warnings

CarlosCortizasCT:
#### Summary

Refactor warning utility to not throw errors but log warnings

## Description

We currently have a `warning` utility we use when we want to warn about (for example) deprecated props.

Its current implementation throws an error (not in production) when the deprecated condition is met which is a little too much for our consumers because we might force them to update their codebase right away.

To have a less aggressive strategy and to follow the same one we use in `app-kit`, we're changing it to just log warning errors into the console, using the [`tiny-warning`](https://github.com/alexreardon/tiny-warning#readme) dependency.

Besides refactoring our existing helpers, I thought it would help to also export some custom hooks because I've seen this pattern were we just wrap the helpers into `useEffect` hooks so we just render the warning upon first render of the component and not every time it re-renders ([reference](https://github.com/commercetools/ui-kit/blob/main/packages/components/buttons/link-button/src/link-button.tsx#L102) in `ui-kit` and [reference](https://github.com/commercetools/merchant-center-application-kit/blob/4377b3642c08fd9480016a287a18ada780191ad6/packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.tsx#L37) in `app-kit`).

The *issue* here comes from the factor that to do such thing, I need to include `React` as a dependency of the `@commercetools-uikit/utils` package. Following the same pattern we use in the other packages, I'm declaring it as a peer dependency.
This should not be a problem since most of `ui-kit` packages already rely on that, but it would technically be a braking change for people using the package and not having `React` installed in their projects.
I found this a super edge case so I think we should not treat this change as a breaking one but I would like to read your opinions.


packages/components/stamp/src/stamp.tsx
```diff
@@ -164,6 +165,13 @@ const Stamp = (props: Props) => {
       color: getIconColor(props, overrideTextColor),
     });
 
+  useWarnDeprecatedProp(
```
CarlosCortizasCT, line 168: I'm taking back the warning about the `children` prop in the `Stamp` component which now will be easier to silence in the MC frontend.

packages/utils/package.json
```diff
@@ -22,7 +22,12 @@
     "@babel/runtime": "^7.20.13",
     "@babel/runtime-corejs3": "^7.20.13",
     "@emotion/is-prop-valid": "1.2.0",
-    "@types/warning": "^3.0.0",
-    "warning": "4.0.3"
+    "tiny-warning": "1.0.3"
```
CarlosCortizasCT, line 25: `warning` logs console errors instead of warnings so I decided to use `tiny-warning` as it's what we are already using in `app-kit`.

packages/utils/package.json
```diff
@@ -22,7 +22,12 @@
...
+  },
+  "devDependencies": {
+    "react": "17.0.2"
+  },
+  "peerDependencies": {
```
CarlosCortizasCT, line 30: New peer dependency for this package.

emmenko (approved):
🤩 👏

kark (approved):
\o/

CarlosCortizasCT:
Given the [simplicity of tiny-warning](https://github.com/alexreardon/tiny-warning/blob/master/src/index.js) package, I decided to drop it since I don't think it's worthy to maintain another dependency when we can just have a couple of code lines more.

emmenko (approved):
Awesome, thanks!

PR #2422, opened 2023-02-16
refactor: radio input new theme styles

kark:
#### Summary

Old theme:
Screenshot 2023-02-21 at 15 31 41

New theme:
Screenshot 2023-02-21 at 15 32 04


kark:
Hey @FilPob,
Could I ask you for an early review? Here are my concerns:
-  Would you say that the sizing of the radio input is right. I mean, it is not, but if I change it to the correct values, it kind of looks off compared to the designs
Screenshot 2023-02-16 at 11 04 30
- Would you say that the current font color of the label is ok for the warning/error state? Or should it be adjusted for the old theme?
Screenshot 2023-02-16 at 11 10 35
- spacing between the label and the button should be 4px - but we consider the hover area part of the button?
Screenshot 2023-02-16 at 11 13 53
We could sync about those points 👆 anytime it works for you

FilPob:
Hey @kark 

- The hover area should be part of the radio button. Meaning the visible radio button is 18x18 but the effective size incl. the hover area is 26x26
- read only and disabled state should not have a hover effect. But their size should still be 26x26

The preview link looks great! I just noticed that the radio button size changes, when toggeling it on and off. It seems to be 18x18 when toggled, and 16x16 when not toggled

tdeekens (approved):
💯 Looks good. Can't wait to see it inside the MC 😄

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -133,25 +120,96 @@ const getLabelCursor = (props: TStylesProps) => {
...
+  const hoverStyles = css`
+    [class*='getContainerStyles'] {
+      border-color: ${designTokens.colorNeutral90};
+    }
+  `;
+
+  return [
+    !props.isDisabled &&
+      !props.isReadOnly &&
+      /* prettier-ignore */
```
kark, line 158: This line is necessary, otherwise formatting of `&:hover` is quite off (the code still works though)

packages/components/inputs/radio-input/src/radio-option.tsx
```diff
@@ -111,22 +127,20 @@ const Option = (props: TOptionProps) => {
...
-        <div css={getContainerStyles(props)}>
-          {props.isChecked ? (
-            <RadioOptionCheckedIcon size="medium" />
-          ) : (
-            <RadioOptionUncheckedIcon size="medium" />
-          )}
+        <div css={getContainerStyles(stylesProps)}>
+          <RadioOptionBorder {...stylesProps}>
+            {props.isChecked ? <RadioOptionKnob {...stylesProps} /> : null}
+          </RadioOptionBorder>
```
kark, line 133: This is the major change that is applied to both the old and the new theme. We ditch svgs in favour of a bunch of styled divs.

My additional question is - should I delete the `icons` folder? Or leave it for backwards compatibility

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -133,25 +120,96 @@ const getLabelCursor = (props: TStylesProps) => {
...
+
+const Input = styled.input`
+  &:focus + div > ${RadioOptionBorder} {
+    border-color: ${designTokens.borderColorForInputWhenFocused};
+  }
+`;
+
+const getNewThemeHoverStyles = (props: TStylesProps) => {
+  const hoverStyles = css`
+    [class*='getContainerStyles'] {
```
kark, line 150: Any suggestions for an alternative solution so would be most welcome 🙏

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -133,25 +120,96 @@ const getLabelCursor = (props: TStylesProps) => {
...
+};
+
+const getDefaultThemeHoverAndFocusStyles = (props: TStylesProps) => css`
+  &:hover ${RadioOptionBorder} {
+    border-color: ${getLabelBorderColor(props)};
   }
   :focus-within ${LabelTextWrapper} {
     outline: auto 2px ${designTokens.borderColorForInputWhenFocused};
     outline-offset: 3px;
   }
```
kark, line 173: Please mind this outline in the `focus` state is only added in the current theme (we don't have it in the new theme designs). Not sure if this is expected 
cc @FilPob 

Screenshot 2023-02-21 at 15 21 27

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -133,25 +120,96 @@ const getLabelCursor = (props: TStylesProps) => {
...
+        &:hover ${hoverStyles};
+      `,
+    props.isHovered && hoverStyles,
+  ];
+};
+
+const getDefaultThemeHoverAndFocusStyles = (props: TStylesProps) => css`
+  &:hover ${RadioOptionBorder} {
+    border-color: ${getLabelBorderColor(props)};
   }
```
kark, line 169: Also, this is not directly mentioned in the designs, but for the there is a `focus` state indication (when the input is additionally in the `readonly`/`error`/`warning` state):
Screenshot 2023-02-21 at 15 26 22

Otherwise, there is no visual information for the user that the input has focus.
Perhaps this is not expected.
cc @FilPob

kark (commented):
My two cents

packages/components/inputs/radio-input/src/radio-option.styles.ts
emmenko: In the CheckboxInput we have it in both. We should do the same here.

image

image

packages/components/inputs/radio-input/src/radio-option.styles.ts
emmenko: What are you trying to do here? 🤔

emmenko (commented):
Thanks! Looking good overall, just the hover effect does not seem to work (in storybook) 😔

packages/components/inputs/radio-input/src/radio-option.styles.ts
kark: All in all it turned out that using this class attribute selector makes no sense. Emotion production class names are hashed differently than in development mode. And that is why the hover effect hasn't been working in the preview. Thanks for catching the bug.

packages/components/inputs/radio-input/src/radio-option.styles.ts
kark: Added here b074705fbfb7916e895d334e776cb1c2d1ac0ce3
I had to decrease the outline for the new theme since it was overlapping with the radio option hover area.
I might add a design token instead if it works better for you.

packages/components/inputs/radio-input/src/radio-option.styles.ts
emmenko: In the CheckboxInput I didn't need to do this...maybe worth using the same/similar implementation for the elements structure and styles? 🤔

FilPob:
Looks great! Can be merged

FilPob (approved):

packages/components/inputs/radio-input/src/radio-option.styles.ts
kark: I noticed that checkbox and radio inputs work a bit differently. Checkbox doesn't have hover effect when hovering over the label (and radio input does). Not sure if needs to be unified and if yes then what way 🤔 And if for both the old and the new theme.

emmenko: Ah, good point I didn't think about it.

@FilPob Should the hover effect be triggered also when hovering over the label?

FilPob: Oh missed this one. I think it actually looks great when triggering the hover also when hovering just the label. Would be nice to add that also in the Checkbox Input @emmenko

packages/components/inputs/radio-input/src/radio-option.styles.ts
FilPob: @kark The indication of a focus state is given by the green border around the label now right? I think given that, we don't need a `focused` indication in the radio input itself

packages/components/inputs/radio-input/src/radio-option.styles.ts
emmenko: Sure, here you go: https://github.com/commercetools/ui-kit/pull/2427

emmenko: @kark Up to you to see if we can use a similar implementation structure.

kark: Thanks, I used the same targeting. Updated here 7cff85436acf8dd54d2869823cae78ccb1c64663

packages/components/inputs/radio-input/src/radio-option.styles.ts
kark: Right 👍 done here 48efc33eb644a73ce2f3474ded55977d4c51c90c

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -1,36 +1,33 @@
...
-  margin-left: ${designTokens.spacing20};
+  margin-left: ${designTokens.marginLeftForRadioInputLabel};
   font-size: 1rem;
   font-family: inherit;
-  color: ${(props) =>
-    props.isDisabled
-      ? designTokens.fontColorForInputWhenDisabled
-      : designTokens.fontColorForInput};
+  display: flex;
+  ${(props) => !props.isNewTheme && getDefaultThemeLabelColor(props)}
```
emmenko, line 19: Nit: you can handle the if-else logic within the `getDefaultThemeLabelColor` function.

This then becomes simply:

```suggestion
  ${getDefaultThemeLabelColor}
```

packages/components/inputs/radio-input/src/radio-option.styles.ts
```diff
@@ -133,25 +118,103 @@ const getLabelCursor = (props: TStylesProps) => {
...
+  background-color: ${(props) => getKnobColor(props)};
+`;
+
+const RadioOptionBorder = styled.div<TStylesProps>`
+  width: ${designTokens.heightForRadioInputOption};
+  height: ${designTokens.heightForRadioInputOption};
+  border-radius: 50%;
+  background-color: ${designTokens.backgroundColorForInput};
+  border: ${designTokens.borderForRadioInputOption} solid
+    ${(props) => getBorderColor(props)};
```
emmenko, line 134: Nit: maybe split it up to make it a bit easier to read?

```
border-width
border-style
border-color
```

emmenko (approved):
🥳

PR #2426, opened 2023-02-17
Implement Search Selectable Input

Rhotimee:
#### Summary
Implementation of new Search Selectable Input Component


packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,434 @@
...
+   * <br>
+   * [Props from React select was used](https://react-select.com/props)
+   */
+  noMenuOptionsMessage?: ReactSelectProps['noOptionsMessage'];
+  /**
+   * Whether to enable search functionality.
+   * <br>
+   * [Props from React select was used](https://react-select.com/props)
+   */
+  IsMenuSearchable?: ReactSelectProps['isSearchable'];
```
tdeekens, line 199: Should this be lowercase? 🤔

tdeekens (commented):
I am super excited for this! Thanks for starting to work on it :)

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
Rhotimee: fixed here [4bae45c](https://github.com/commercetools/ui-kit/pull/2426/commits/4bae45c97abfcdf9ff54bbfff2bcde28afa5e400)

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+  const handleTextInputFocus = useCallback(() => {
+    if (props.onFocus)
+      props.onFocus({
+        target: {
+          id: SelectableSearchInput.getTextInputId(selectablSearchInputId),
+          name: getTextInputName(props.name),
+        },
+      });
+    toggleTextInputHasFocus(true);
```
CarlosCortizasCT, line 260: I would consider adding brackets to the `if` statements for better readability (not only here)

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,435 @@
...
+   */
+  onBlur?: (event: TCustomEvent) => void;
+  /**
+   * Called when input is focused
+   */
+  onFocus?: (event: TCustomEvent) => void;
+  /**
+   * Handler when the search button is clicked.
+   */
+  onSubmit: (searchValue: string) => void;
```
CarlosCortizasCT, line 105: I think the `onSubmit` handler should be called with both the searchValue and the selected option from the dropdown so the consumers can know what the user selected to trigger the right query.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,435 @@
...
+        target: {
+          id: dropdownId,
+          name: dropdownName,
+        },
+      });
+    toggleDropdownHasFocus(true);
+  }, [props.onFocus, toggleDropdownHasFocus, dropdownName, dropdownId]);
+
+  const handleDropdownBlur = useCallback(() => {
+    console.log('eyy???>>>>>>>');
```
CarlosCortizasCT, line 319: Don't forget to remove this 😄

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,435 @@
...
+};
+
+SelectableSearchInput.displayName = 'SelectableSearchInput';
+SelectableSearchInput.defaultProps = defaultProps;
+SelectableSearchInput.isEmpty = (
+  formValue: TSelectableSearchInputProps['value']
+) =>
+  !formValue ||
+  formValue.textValue.trim() === '' ||
+  formValue.dropdownValue === '';
```
CarlosCortizasCT, line 431: Not sure about this: perhaps a consumer would like to offer some options to filter its query but selecting one is not mandatory.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,435 @@
...
+                onClick={handleClear}
+                css={getClearIconButtonStyles(props)}
+              />
+            )}
+          <SecondaryIconButton
+            icon={<SearchIcon />}
+            label={'search-button'}
+            onClick={handleSubmit}
+            css={getSearchIconButtonStyles(props)}
+          />
```
CarlosCortizasCT, line 417: Perhaps we want this to be configurable. I'm suggesting because maybe we would like to use this component as a base for the `MoneyInput`.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
Rhotimee: I was thinking about this also. I decided the handler should be called with on the seachTerm because the component on MC does the same but, It makes more sense for the consumer to have both values. Thanks.

Rhotimee:
Thanks @CarlosCortizasCT. 
Here is the commit for the feedback -> [6242c48](https://github.com/commercetools/ui-kit/pull/2426/commits/6242c48578d640492457cca11e1f8ccb3e4d393a).

design-system/materials/internals/definition.yaml
```diff
@@ -1013,6 +1015,8 @@ decisionGroupsByTheme:
           choice: 'inherit'
         padding-for-money-input-currency-dropdown:
           choice: '0 var(--spacing-20)'
+        padding-for-selectable-search-input-dropdown:
+          choice: '0 var(--spacing-20)'
```
emmenko, line 1019: Do we need a design token for this? 🤔

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,446 @@
...
+          css={[
+            getSelectableSearchInputContainerStyles(props),
+            dropdownHasFocus &&
+              isNewTheme &&
+              css`
+                border-left-color: ${designTokens.borderColorForInputWhenFocused};
+              `,
+          ]}
+        >
+          <input
```
emmenko, line 389: Might be good to also support submitting on <kbd>Enter</kbd>

emmenko (commented):
Overall looks good, thanks!

Did some manual testing in storybook and couldn't spot any obvious issues ✅

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
tdeekens: Don't we get that it we wrap the input into a form nicely?

design-system/materials/internals/definition.yaml
Rhotimee: It may not be needed. What do you think @CarlosCortizasCT

packages/components/inputs/selectable-search-input/src/selectable-search-input.spec.js
```diff
@@ -0,0 +1,237 @@
...
+        {...props}
+      />
+    </div>
+  );
+};
+
+TestComponent.defaultProps = {
+  id: 'test-id',
+  name: 'test-name',
+  value: {
```
tdeekens, line 32: 🤔 Just an idea but an alternative would be

```js
values: {
    text: 'foo',
    option: 'bar'
}
```

then the structure of options and values feels a bit more symmetrical.

Rhotimee: Thats a good idea. Here is the [commit](https://github.com/commercetools/ui-kit/pull/2426/commits/340b32043822221ad7ad8efe89933a97f6fe5ce2). Thanks.

design-system/materials/internals/definition.yaml
tdeekens: Seems like the token above. Maybe follow up and remove both if that's decided.

packages/components/inputs/selectable-search-input/README.md
```diff
@@ -0,0 +1,150 @@
...
+The component further forwards all `data-` attributes to the underlying `input` component.
+
+## Static methods
+
+### `SelectableSearchInput.isEmpty`
+
+Returns `true` when the value is considered empty, which is when the value is empty or consists of spaces only.
+
+```js
+SelectableSearchInput.isEmpty(''); // -> true
```
tdeekens, line 143: I think this needs to be updated taking [this](https://github.com/commercetools/ui-kit/pull/2426/files#diff-c0601f2ee89d0052b13493f19c29394110f25ef8689fa42ba3f7addb227e7ff6R73) API into consideration.

tdeekens (commented):
Looks good. Only one thing I could notice :)

tdeekens (approved):

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,453 @@
...
+   */
+  name?: string;
+  /**
+   * Value of the input. Consists of text input and selected option.
+   */
+  value: TValue;
+  /**
+   * Called with the event of the input or dropdown when either the selectable dropdown or the text input have changed.
+   */
+  onChange?: (event: TCustomEvent) => void;
```
emmenko, line 95: We might want to specify a bit more what the `name` format of the event is. For instance, change events from the text input have the suffix `.textInput` and the ones from the dropdown have the suffix `.dropdown`.

emmenko (approved):
Just one minor suggestion, otherwise seems good to go from my side.

Thanks again for working on this! 🙌

design-system/materials/internals/definition.yaml
CarlosCortizasCT: If we're having the same value for both themes, I would consider not adding a token for it.

As for the token above, it does not have the same value for both themes.

packages/components/inputs/selectable-search-input/src/selectable-select.tsx
```diff
@@ -0,0 +1,195 @@
...
+    hasError,
+    menuPortalZIndex,
+    isNewTheme,
+  });
+
+  return {
+    ...selectStyles,
+    control: (base: TBase, state: ReactSelectProps) => ({
+      ...selectStyles.control(base, state),
+      padding: designTokens.paddingForSelectableSearchInputDropdown,
```
CarlosCortizasCT, line 67: **nit**: Since we already have a styles file for the component, perhaps we can move this to it.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -0,0 +1,454 @@
...
+    if (props.onFocus) {
+      props.onFocus({
+        target: {
+          id: SelectableSearchInput.getTextInputId(selectablSearchInputId),
+          name: getTextInputName(props.name),
+        },
+      });
+    }
+    toggleTextInputHasFocus(true);
+  }, [textInputHasFocus, props.onFocus, selectablSearchInputId, props.name]);
```
CarlosCortizasCT, line 266: There are some eslint warnings about `props` and `toggleTextInputHasFocus` not being included in the dependencies array (not only for this `useCallback`).
We might need to destructure properties such `onFocus`, `onBlur`,...) to not include the whole `props` object

Rhotimee: Thanks. I didn't have the eslint extension on the new laptop, so I couldn't see the warnings. 
Updated [cd72889](https://github.com/commercetools/ui-kit/pull/2426/commits/cd72889b1c645bc9d296d204c3808bf5f9450647)

packages/components/inputs/selectable-search-input/src/selectable-select.tsx
Rhotimee: Updated [cd72889](https://github.com/commercetools/ui-kit/pull/2426/commits/cd72889b1c645bc9d296d204c3808bf5f9450647)

packages/components/inputs/selectable-search-input/src/selectable-search-input.spec.js
```diff
@@ -0,0 +1,235 @@
...
+      <TestComponent
+        id="some-id"
+        name="some-name"
+        onChange={onChange}
+        onSubmit={onSubmit}
+      />
+    );
+    const event = { target: { value: 'avengers' } };
+
+    fireEvent.change(screen.getByTestId('selectable-input'), event);
```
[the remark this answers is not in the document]
Rhotimee: Updated [832feb7](https://github.com/commercetools/ui-kit/pull/2426/commits/832feb794b0d1ce0261829cf65a2117817f80c35)

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
Rhotimee: Good catch. Updated here [0915616](https://github.com/commercetools/ui-kit/pull/2426/commits/091561624cbfef92d021eea944e1d8051f085908)

design-system/materials/internals/definition.yaml
Rhotimee: Thanks, Updated here [d1e1958](https://github.com/commercetools/ui-kit/pull/2426/commits/d1e1958c7c0c1bb0ddf1fd144c80cae646684d0f)

packages/components/inputs/selectable-search-input/README.md
Rhotimee: Thanks, [e811206](https://github.com/commercetools/ui-kit/pull/2426/commits/e811206c4f7c216e22f85871c4a16990c6ea9188)

CarlosCortizasCT:
Some thing I noticed when testing manually:
*default theme*
* dropdown changes its background color when hovering even when the component is disabled
 * also, it's color is different from the search input, but this might be on purpose
image
image
* dropdown text color does not change when the component is read-only. Is this on purpose?
image
* dropdown text color does not change when the component has errors or warnings. In this on purpose?
image

*new theme*
* dropdown text color does not change when the component is disabled
image

* dropdown background color changes on hover when the component is disabled (also the cursor does not change)
image

* dropdown gets the focus after clicking in it when the component is read-only and it seems it never loses it. Even if you click outside the component, it keeps its border green. Also, the border is narrower than the text input.
image


* dropdown right border changes on hover when component is in readonly and text is hovered when dropdown was previously clicked
image

FilPob:
I listed my findings **[here](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764541184225370&cot=14)**

I got one question here, as we dont havesomething like this currently. My idea was to have a property that controls the width of the dropdown. I proposed it also [here](https://5jpcut.axshare.com/?id=3028ea91-d7cb-4d60-bb6a-2b6a9c7311d8&g=5) The possible values for the prop would be hor.constraint 3, 4, or 5. 

**Why having this?** 
To offer better flexibility. The dropdown can contain labels that can be really short, or really long. In a case with short labels, it would be a waste of space having a dropdown with the width of  `hor. constraint 5`... However, when having long labels they would get truncated too often if the the dropdown would be e.g `hor. constraint 3` (like it currently is) Not sure if I explained it in a good way :D lets have a follow up if you want cc @CarlosCortizasCT  @Rhotimee

CarlosCortizasCT:
> I listed my findings **[here](https://miro.com/app/board/uXjVPSlrfYY=/?moveToWidget=3458764541184225370&cot=14)**
> 
> I got one question here, as we dont havesomething like this currently. My idea was to have a property that controls the width of the dropdown. I proposed it also [here](https://5jpcut.axshare.com/?id=3028ea91-d7cb-4d60-bb6a-2b6a9c7311d8&g=5) The possible values for the prop would be hor.constraint 3, 4, or 5.
> 
> **Why having this?** To offer better flexibility. The dropdown can contain labels that can be really short, or really long. In a case with short labels, it would be a waste of space having a dropdown with the width of `hor. constraint 5`... However, when having long labels they would get truncated too often if the the dropdown would be e.g `hor. constraint 3` (like it currently is) Not sure if I explained it in a good way :D lets have a follow up if you want cc @CarlosCortizasCT @Rhotimee

I agree that having the way to control the width of the dropdown would be better.
Just bear in mind we don't control the values that will be rendered in it, so it would be just a guess from the consumers point of view.

FilPob:
@CarlosCortizasCT yes, that is then up to the user to choose the best fitting size.

Rhotimee:
@FilPob, the horizontal control for the dropdown was implemented but not exposed on storybook. Take a look at it now.

FilPob:
@Rhotimee  Thanks! Looks great now. I don't see any more issues :)

FilPob (approved):

PR #2429, opened 2023-02-23
 Export the Typescript types for the props of uikit public components.

ddouglasz:
**Background:**
After finishing the migration of all `ui-kit` components to Typescript, consumers are starting to use this new version and some of them complain about our library not exporting the Typescript definitions for the component properties.

**Goals:**
We want to make it easier for `ui-kit` components consumers to use them and to integrate them in their applications/libraries.
This mean for them to be able to access the Typescript definitions for the components props.

**Technical context:**
We need to cover all our component packages and make sure we export the Typescript types for the props of our public components.


kark (commented):
Thanks for working on this 🙌
Overall it looks good to me.
I’m not sure why packages like `@commercetools-uikit/accessible-hidden` and `@commercetools-uikit/constraints` were omitted 🤔 (perhaps there are more).
I guess the changeset is also another missing part.
I was also wondering about covering types from packages other than `components` as part of this PR (mainly `hooks`)

ddouglasz:
Thanks Kacper, updated here: https://github.com/commercetools/ui-kit/pull/2429/commits/ef70457b2685b6b9b43c859983586320a96f9905

packages/components/accessible-hidden/src/export-types.ts
```diff
@@ -0,0 +1,3 @@
+import type { TAccessibleHiddenProps as AccessibleHiddenProps } from './accessible-hidden';
+
+export type TAccessibleHiddenProps = AccessibleHiddenProps;
```
kark, line 3: I have a general remark that even if we add `export-types.ts` file, the types won’t get through to the bundle unless this file is imported in `index.ts` of the package. For instance, with types exported like this there is no possibility of importing types like:
```ts
import AccessibleHidden, {
  type TAccessibleHiddenProps,
} from '@commercetools-uikit/accessible-hidden';
```

You might check e.g. `money-input` for reference.

kark (commented):
Thanks for including changes 🙏
Couple more issues I noticed:

1. `localized-input-toggle` & `multiline-input`:
I think the types are already exported from the `index.ts`

2 Still omitted packages:
`select-utils`
`selectable-search-input`

3. Icons:
I think it’s up for discussion if we need to do something with icons. Just for the record, now users have to import types like:
```import type { Props } from '@commercetools-uikit/icons/dist/declarations/src/generated/InformationReact'```

File: packages/components/accessible-hidden/src/export-types.ts (hunk 1 above)
Comment (ddouglasz) [reply packages/components/accessible-hidden/src/export-types.ts:3]: Good point, I will take a look and see how this works now.

**Comment by ddouglasz** [comment]
> Thanks for including the changes 🙏 Couple more issues I noticed:
> 
> 1. `localized-input-toggle` & `multiline-input`:
>    I think the types are already exported from the `index.ts`
> 2. Packages still omitted:
>    `select-utils`
>    `selectable-search-input`
> 3. Icons:
>    I think it’s up for discussion if we need to do something with icons. Just for the record, now users have to import types like:
>    `import type { Props } from '@commercetools-uikit/icons/dist/declarations/src/generated/InformationReact'`

We can have a further conversation on how to handle the icon cases. 
These have been handled here: https://github.com/commercetools/ui-kit/pull/2429/commits/83fbf1dd6fd64d8a2eb185427aae317dc7ad7462

File: packages/components/accessible-hidden/src/export-types.ts (hunk 1 above)
Comment (kark) [reply packages/components/accessible-hidden/src/export-types.ts:3]: Circling back to my previous remark, I still think this:
https://github.com/commercetools/ui-kit/blob/5e37083198755a7b94a77d0c4c843e5609d5f2d0/packages/components/inputs/money-input/src/index.ts#L3
needs to be added to each `src/index.ts` to export the types.

Otherwise your `export-types.ts` file gets tree-shaken and `dist/declarations/src/index.d.ts` looks like this:
```ts
// dist/declarations/src/index.d.ts
export { default } from './accessible-hidden';
```
While, I think, we expect something like this:
```ts
// dist/declarations/src/index.d.ts
export { default } from './accessible-hidden';
export * from './export-types';
```
in the bundle

kark (commented):
Nice 💪 Thanks, these are the issues I noticed.

Regarding the icons, the way I would personally approach it would be by exporting the `Props` and `SVGProps` types from `packages/components/icons/src/templates/icon.styles.tsx` since they are the same for all the icon components.

packages/components/accessible-hidden/src/export-types.ts
ddouglasz: Oh No. Totally cool! 😁 Thanks for the extra eyes 🙌🏾

packages/components/constraints/src/horizontal/export-types.ts
```diff
@@ -0,0 +1,3 @@
+import type { THorizontalProps as HorizontalProps } from './horizontal';
+
+export type THorizontalProps = HorizontalProps;
```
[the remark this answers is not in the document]
ddouglasz: Thank you 🙏🏾
Updated here: https://github.com/commercetools/ui-kit/pull/2429/commits/0481305a4ca10d6378c58a40b993ca6738002b1c

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -46,7 +46,7 @@ type TCustomEvent = {
   persist: () => void;
 };
 
-type TSelectInputProps = {
+export type TSelectInputProps = {
```
kark, line 49: nit: In case of this components (and in several other cases) we're only exporting the `Props` type. This should be enough, but this limits users to referencing `TSelectInputProps` when typing the options object:

```tsx
import SelectInput, {
  type TSelectInputProps,
} from '@commercetools-uikit/select-input';

const options: TSelectInputProps['options'] = [
  { value: 'one', label: 'One' },
  { value: 'two', label: 'Two' },
];
const Example = (props: TSelectInputProps) => (
  <SelectInput
    options={options}
    // ...
  />
);
```

In case someone wished to be more explicit like:
```tsx
import /* ... */, {
  type TOption,
} from '@commercetools-uikit/select-input';

const options: TOption[] = [
  { value: 'one', label: 'One' },
  { value: 'two', label: 'Two' },
];
```
they have no chance of doing so. It wouldn't cost us much to export the additional types I think. Perhaps it's not needed 🤷‍♂️

kark (approved):
Great work 👏
Thanks for your patience 😅🙏

.changeset/pink-falcons-shout.md
```diff
@@ -0,0 +1,73 @@
+---
+'@commercetools-uikit/localized-multiline-text-field': minor
```
kark, line 2: One more thing: perhaps `patch` is enough 🤔

packages/components/inputs/select-input/src/select-input.tsx
ddouglasz: Thanks again Kacper 🙌🏾
Updated here: https://github.com/commercetools/ui-kit/pull/2429/commits/08aaffd670d8714e901916ef400811f18e97ea64

chloe0592 (approved):
This was a big one! Great work 🎉 👏

PR #2432, opened 2023-02-24
fix(flat-button): use `font-size` consistent with `<Text.Body>`

kark:
#### Summary

This PR applies `<FlatButton>`'s `font-size` consistent with `<Text.body>` in the new theme.

### DIFF
--- a/packages/components/buttons/flat-button/src/flat-button.tsx
+++ b/packages/components/buttons/flat-button/src/flat-button.tsx
@@ -11,6 +11,7 @@ import omit from 'lodash/omit';
 import { designTokens } from '@commercetools-uikit/design-system';
 import { filterInvalidAttributes } from '@commercetools-uikit/utils';
 import AccessibleButton from '@commercetools-uikit/accessible-button';
+import Text from '@commercetools-uikit/text';
 import { getTextColor, getButtonIconColor } from './flat-button.styles';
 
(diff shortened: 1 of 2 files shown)


tdeekens (approved):

packages/components/buttons/flat-button/src/flat-button.tsx
```diff
@@ -121,6 +121,7 @@ const FlatButton = <TStringOrComponent extends ElementType = 'button'>(
       css={css`
         min-height: initial;
         align-items: center;
+        font-size: ${designTokens.fontSizeForTextAsBody};
```
CarlosCortizasCT, line 124: Instead of having a new CSS rule here we could change the `span` which is wrapping button's text with a `Text.Body` component for better composability.

```
// BEFORE
<span>{props.label}</span>

// AFTER
<Text.Body as="span">{props.label}</Text.Body>
```

kark: Thanks. SO much better! 🧘‍♂️

tdeekens (approved):

kark:
This change is no longer needed.
The solution in the reported case was an UX decision of replacing `<Text.Body>` with `<Text.Detail>` in the price mode indicator widget.

PR #2438, opened 2023-02-27
feat(search-text-input): add `forwardRef`

### LINKED ISSUE #2431: SearchTextInput: Add forward ref to input
Problem:
In Audit Log there is an input field used to gather a search string. We are converting this input over to using SearchTextInput from ui kit. We currently use a ref on the input element so that when the user clears the value via the clear button, we can put user focus back on the input element.

Solution:
Add a forwardRef to the input element so we can maintain this functionality without having to hack around it.

Usage in Audit log for reference:
image

kark:
#### Summary

closes #2431

### DIFF
--- a/packages/components/inputs/search-text-input/src/search-text-input.story.js
+++ b/packages/components/inputs/search-text-input/src/search-text-input.story.js
@@ -1,5 +1,4 @@
 import { storiesOf } from '@storybook/react';
-import { useCallback, useRef } from 'react';
 import { action } from '@storybook/addon-actions';
 import { Value } from 'react-value';
 import { withKnobs, boolean, text, select } from '@storybook/addon-knobs/react';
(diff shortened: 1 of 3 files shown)


packages/components/inputs/search-text-input/src/search-text-input.story.js
```diff
@@ -16,6 +17,11 @@ storiesOf('Components|Inputs', module)
     },
   })
   .add('SearchTextInput', () => {
+    const ref = useRef(null);
+
+    const handleReset = useCallback(() => {
+      ref.current?.focus();
+    }, []);
```
kark, line 24: I added this to the story, but perhaps some users may find it misleading? 🤔 WDYT

CarlosCortizasCT: Personally I wouldn't change anything in the Story because of the `ref` change, but it's up to you.

CarlosCortizasCT (approved):
👏

packages/components/inputs/search-text-input/src/search-text-input.story.js
kark: Yes, that makes sense. Reverted here 70558d239e9db12d0349664b39d6703303aa5efe

tdeekens (approved):

Rhotimee (approved):
💯

ddouglasz (approved):
🚀

PR #2440, opened 2023-02-27
chore(icon-button): deprecate `shape` prop

### LINKED ISSUE #2437: Deprecate `shape` prop in `IconButton` for the new design
We want to have all icon buttons with square shape in the new design.

For the new design, we want to just ignore the value received in the `shape` prop of the `IconButton` and always use square internally.
We would be still supporting this prop for the current design but mark it as deprecated.

None.

kark:
#### Summary

closes #2437
Screenshot 2023-02-27 at 14 52 29
Screenshot 2023-02-27 at 14 52 51


packages/components/buttons/icon-button/src/icon-button.tsx
```diff
@@ -59,7 +59,7 @@ export type TIconButtonProps<
     event: MouseEvent<HTMLButtonElement> | KeyboardEvent<HTMLButtonElement>
   ) => void;
   /**
-   * The container shape of the button.
```
kark, line 62: I removed this prop description as I noticed that none of the deprecated prop have one.

CarlosCortizasCT (approved):
Many thanks! 👏

tdeekens (approved):
Nice approach!

FilPob (approved):

packages/components/buttons/icon-button/src/icon-button.tsx
```diff
@@ -59,7 +59,7 @@ export type TIconButtonProps<
...
+   * @deprecated
```
emmenko, line 62: Wondering if we should mark it as deprecated only when we turn on the new design. In any case, it's useful to add a message to inform the user what to do (or what the consequence is).

emmenko (approved):
Nice! 🤗

packages/components/buttons/icon-button/src/icon-button.tsx
emmenko: I meant to add the annotation only after we turn on the new design.

Let's keep it simple and keep the annotation but provide a useful message to the user.

kark: Oh, yes I think I got it, thanks. Would that work for you?
Screenshot 2023-02-28 at 10 49 37
```ts
/**
 * @deprecated This prop is only used in the old theme. For the new theme this prop will not be taken into account and `square` is used by default
 */
```

ddouglasz (approved):

packages/components/buttons/icon-button/src/icon-button.tsx
tdeekens: I like that. Seems like we're otherwise overdoing it a bit.

PR #2441, opened 2023-02-27
Fix `Link` component `font-size` value in the new theme

CarlosCortizasCT:
#### Summary

Fix `Link` component `font-size` value in the new theme

## Description

We want the `font-size` of the `Link` component to be inherited from its parent in the new theme but it's currently not wroking.

We decided to use `font-size: inherit` for the new theme, but that's making the size to be inherited from the `body` element, not its parent.

### DIFF
--- a/design-system/materials/internals/definition.yaml
+++ b/design-system/materials/internals/definition.yaml
@@ -1438,38 +1438,25 @@ decisionGroupsByTheme:
       prefix: font-size
       decisions:
         font-size-for-input:
-          description: ''
           choice: font-size-30
         font-size-for-text-as-h1:
-          description: ''
           choice: font-size-60
         font-size-for-text-as-h2:
-          description: ''
           choice: font-size-50
         font-size-for-text-as-h3:
-          description: ''
           choice: font-size-40
         font-size-for-text-as-h4:
-          description: ''
           choice: font-size-30
         font-size-for-text-as-h5:
-          description: ''
           choice: font-size-30
         font-size-for-text-as-body:
-          description: ''
           choice: font-size-30
         font-size-for-text-as-detail:
-          description: ''
           choice: font-size-20
         font-size-for-body:
-          description: ''
           choice: '16px'
         font-size-for-button:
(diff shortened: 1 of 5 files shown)


design-system/materials/internals/definition.yaml
```diff
@@ -1438,38 +1438,25 @@ decisionGroupsByTheme:
...
         font-size-for-text-as-detail:
-          description: ''
           choice: font-size-20
         font-size-for-body:
-          description: ''
           choice: '16px'
         font-size-for-button:
-          description: ''
           choice: font-size-20
-        font-size-for-link:
```
CarlosCortizasCT, line 1470: Just removed this token as we are not going to be using it and kept the one in the default theme just in case we want to have component tokens after the redesign.

design-system/materials/internals/definition.yaml
```diff
@@ -1438,38 +1438,25 @@ decisionGroupsByTheme:
       prefix: font-size
       decisions:
         font-size-for-input:
-          description: ''
```
CarlosCortizasCT, line 1441: Just removing some useless descriptions

tdeekens (approved):

packages/components/link/src/link.tsx
```diff
@@ -104,21 +104,28 @@ const getActiveColorValue = (tone: string = 'primary') => {
...
+        color: ${hoverColor};
+      }
+      text-decoration: underline;
+    `,
+    !props.isNewTheme
+      ? css`
+          font-size: ${designTokens.fontSizeForLink};
+        `
+      : null,
+  ];
```
CarlosCortizasCT, line 128: I tried using different values for the new theme, but the one that would work (`unset`) can't be used in our context because of the css rule that gets created:
```
var(--font-size-for-link, 1rem)
```

When the value is `unset`, the `var` call uses the default value (`1rem`)

emmenko (approved):
Thanks!

kark (approved):

PR #2442, opened 2023-02-28
refactor(toggle-input): new theme styles

kark:


kark:
Hey @FilPob, 
could you help me out here? How should I approach the `focus` indicator? 
Screenshot 2023-02-28 at 16 39 27

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -0,0 +1,209 @@
...
+    & + span::after {
+      background: ${designTokens.backgroundColorForToggleInputThumbWhenCheckedAndDisabled};
+    }
+  }
+
+  :not(:disabled)&:hover + span::after,
+  :not(:disabled)&:focus + span::after {
+    box-shadow: ${props.isNewTheme ? 'none' : designTokens.shadow16};
+    outline: ${props.isNewTheme
+      ? `${thumb.test[props.size].hoverAreaWidth} solid rgba(0, 0, 0, 0.1)`
```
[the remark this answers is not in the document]
emmenko: What color is this?

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -0,0 +1,209 @@
...
+  isNewTheme: boolean;
+  theme: ThemeName;
+};
+
+type TStyledLabelProps = Pick<TToggleInputProps, 'isDisabled' | 'size'> &
+  NewThemeProps;
+type TStyledSpanProps = Pick<TToggleInputProps, 'size'> & NewThemeProps;
+
+const track = {
+  test: {
```
emmenko, line 20: Hmm not sure if hardcoding the theme here should be a concern of this component.

I would map the values but use the `themedValue()` function to render one value or another.

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -0,0 +1,209 @@
...
+const getThumbSize = (props: TStyledSpanProps) =>
+  thumb[props.theme][props.size].diameter;
+
+const getThumbShift = (props: TStyledSpanProps) =>
+  `${props.isNewTheme ? '-' : ''}${thumb[props.theme][props.size].shift}`;
+
+const getNewThemeTransformation = (props: TStyledSpanProps) =>
+  `calc(${thumb.test[props.size].shift} * 2 + ${
+    track.test[props.size].width
+  } - ${thumb.test[props.size].diameter})`;
```
emmenko, line 81: Nit: this is a bit hard to read. Can you extract the values into variables and assign the variables in the template string? The string then should be rendered in one line (more readable).

emmenko (commented):
Love the new styles!

Just some minor suggestions, otherwise seems good to me 💯

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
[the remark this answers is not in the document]
kark: It's the hover area of the thumb. In the checkbox input this color was also used, you replaced it with `color-neutral-90` (`background-color-for-checkbox-input-icon-when-hovered` token) (and I did the same for the radio input), but in here opacity is crucial)

FilPob (approved):

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
kark: Sure, great point, done here 3222e06cb56ad3fc1f4df9d38856320a0f3e4867

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
kark: Thanks, TBH that was my starting point but I thought you wouldn't want that many `themedValue` calls. 
Wondering if  this notation `thumb(props.themedValue)[props.size].hoverAreaWidth` is too messy...

ddouglasz (approved):
Looks good! tested and works as seen in design  🙌🏾

design-system/materials/custom-properties.css
```diff
@@ -215,6 +216,21 @@
   --background-color-for-checkbox-input-icon-when-readonly: #fff;
   --background-color-for-checkbox-input-icon-when-error: #fff;
   --background-color-for-checkbox-input-icon-when-hovered: #fff;
+  --background-color-for-toggle-input-track: hsl(0, 0%, 60%);
+  --background-color-for-toggle-input-track-when-disabled: #ccc;
+  --background-color-for-toggle-input-thumb-when-disabled: hsl(
+    195,
+    35.2941176471%,
```
tdeekens, line 223: Could be a bit more precise 🤗

tdeekens (approved):
💯

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
emmenko: > Wondering if this notation `thumb(props.themedValue)[props.size].hoverAreaWidth` is too messy...

Up to you to judge. 🤗

In general you should ask yourself if someone would be comfortable with this and understand the logic when reading it (even without a context). That often times includes you as well as the author (for example, _in a week time, if you read this, do you know what it does?_).

Not saying you need to change this, just suggesting how you can think about it and make a decision.

kark: Right, thanks 🙂 Well, I suggested this so I'm ok with that 😅
So the question was rather if anyone else finds this hard to read.

emmenko (approved):
Looks great! 🥳

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -0,0 +1,231 @@
...
+      },
+    }
+  );
+
+const getThemedThumb = (themedValue: TUseThemeResult['themedValue']) =>
+  themedValue(
+    {
+      small: {
+        diameter: '13px',
+        shift: '2px',
```
CarlosCortizasCT, line 49: I'm not sure what `shift` means here. Is it like a margin?

kark: Shift (the way I thought of it) is the relative position difference of thumb against the track. It's only needed for the calculations of the final position of the thumb when on and off (to make it easier to follow the way it's calculated).
Screenshot 2023-03-09 at 10 51 01
Now I think it's definitely worth explaining in a comment.

Issue #2435, opened 2023-02-27
`SecondayIconButton`: Add `size` prop


CarlosCortizasCT:
**Is your feature request related to a problem? Please describe.**
We currently have a `size` prop in the [`IconButton`](https://uikit.commercetools.com/?path=/story/components-buttons--iconbutton) component which allows for having different sizes of the icon but we're not having that property in the [`SecondaryIconButton`](https://uikit.commercetools.com/?path=/story/components-buttons--secondaryiconbutton) component which would also be useful.

**Describe the solution you'd like**
I would like to have the `size` property in the `SecondayIconButton` with the same possible values than `IconButton` component (`small`, `medium` and `big`).

**Describe alternatives you've considered**
None.

PR #2443, opened 2023-03-01
chore(secondary-icon-button): adding size prop

chloe0592:
This PR closes: [2435](https://github.com/commercetools/ui-kit/issues/2435)

Unifying button sizes


chloe0592:
Working on the Percy failed test

tdeekens (approved):

chloe0592 (commented):
@CarlosCortizasCT just to double check with you. Are we sure we want to have **default size** of the `SecondaryIconButton` (-medium) different from the `IconButton`(-big)(? 
I worry it might be confusing since I would also expect it the be the same, but then be puzzled why it's not.

emmenko:
Let's sync about it next week, we might want to consistently apply these changes to all buttons that render an icon.

chloe0592:
> Let's sync about it next week, we might want to consistently apply these changes to all buttons that render an icon.

Yes, sounds good! Carlos just created a meeting for that with Filip [here](https://commercetools.remeet.com/?m=6400cbd257903c79e07068a8&open_m=true).

ddouglasz (approved):
Looks good 🙌🏾

chloe0592:
The conclusion after the meeting: 

> `SecondaryIconButton` will also have a new size property with the same values as the `IconButton` component. The default value will be “big”. We need to be aware that some components will need to be adjusted in the MC frontend when this change is released.
> Also, we might write a release note where we will explain this change from the fixing point of view as we want to have a standard sizing handling for all the buttons.

FilPob:
Looks good from my side. Thanks!

FilPob:
Oooh I would have one thing... Not related to the size prop but it would be great to increase the padding left& right from 8 to 16px in PrimaryButton when size: Medium and SecondaryButton when size: medium

chloe0592:
> Oooh I would have one thing... Not related to the size prop but it would be great to increase the padding left& right from 8 to 16px in PrimaryButton when size: Medium and SecondaryButton when size: medium

Sure, no problem! Paddings added [8f60a5d](https://github.com/commercetools/ui-kit/pull/2443/commits/8f60a5da15098b9263b39586c6169a187ae8162a).

FilPob:
Thanks, all correct :)

.changeset/funny-spies-type.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-uikit/secondary-icon-button': patch
+'@commercetools-uikit/secondary-button': patch
+'@commercetools-uikit/primary-button': patch
+'@commercetools-uikit/design-system': patch
```
kark, line 5: I think some packages are missing, could you please re-generate the changeset?

kark (approved):
Thank you! Looks good to me 👍

.changeset/funny-spies-type.md
chloe0592: Good catch, thanks! Re-added again 😄

packages/components/buttons/primary-button/src/primary-button.tsx
```diff
@@ -59,9 +59,9 @@ export type TPrimaryButtonProps<
     event: MouseEvent<HTMLButtonElement> | KeyboardEvent<HTMLButtonElement>
   ) => void;
   /**
-   * Indicates the size of the icon.
+   * Indicates the size of the button.
    */
-  size?: 'small' | 'big';
+  size?: 'small' | 'medium' | 'big';
```
CarlosCortizasCT, line 64: I guess we are deprecating the `small` value so perhaps we need to add a warning to explain about it.
We could use [useWarnDeprecatedProp](https://github.com/commercetools/ui-kit/blob/50ccbfc83c2f01978fa279c2d395bd6967a11c86/packages/utils/src/warn-deprecated-prop.ts#L21) for this.

packages/components/buttons/secondary-button/src/secondary-button.tsx
```diff
@@ -55,6 +59,10 @@ export type TSecondaryButtonProps<
   onClick?: (
     event: MouseEvent<HTMLButtonElement> | KeyboardEvent<HTMLButtonElement>
   ) => void;
+  /**
+   * Indicates the size of the button.
+   */
+  size?: 'small' | 'medium' | 'big';
```
CarlosCortizasCT, line 65: Since we are already deprecating the `small` value int he `PrimaryButton` I think we could just not use it in this component.

packages/components/buttons/primary-button/src/primary-button.tsx
chloe0592: Thanks for the suggestion! 🙇‍♀️ Here's the added warning - [93b4501](https://github.com/commercetools/ui-kit/pull/2443/commits/93b45011dc06a1efcb078fa61823968cb7182eb7).

packages/components/buttons/secondary-button/src/secondary-button.tsx
chloe0592: That makes more sense, thanks! [93b4501](https://github.com/commercetools/ui-kit/pull/2443/commits/93b45011dc06a1efcb078fa61823968cb7182eb7).

packages/components/buttons/primary-button/src/primary-button.tsx
```diff
@@ -93,6 +94,13 @@ const PrimaryButton = <TStringOrComponent extends ElementType = 'button'>(
     disabled: props.isDisabled,
   };
 
+  useWarnDeprecatedProp(
+    !Boolean(props.size === 'small'),
+    'props.size = small',
+    'PrimaryButton',
+    'Please use `medium` or `big` size properties instead.'
+  );
```
CarlosCortizasCT, line 102: Not sure if this hook is the best one to use here. I guess the message it will log would be something like:
```
"props.size = small" property of "PrimaryButton" component has been deprecated and will be removed in the next major version. Please use `medium` or `big` size properties instead.
```

Perhaps we can use the `useWarning` hook in this case for a tailored message:
```
useWarning(
  !Boolean(props.size === 'small'),
  'Warning: PrimaryButton `small` value for `size` property has been renamed to `medium`. Please update that value when using this component'
);
```

chloe0592: Ah, yes definitely true it looks better now -  https://github.com/commercetools/ui-kit/pull/2443/commits/055ea70fcd3d718c977bf48f1ba760007b6ed0c0

Thank you! 🙇‍♀️ 
Screenshot 2023-03-15 at 08 53 19

CarlosCortizasCT (approved):
Awesome! 🚀

Rhotimee (approved):
👏🏽

PR #2444, opened 2023-03-02
fix(radio-input): radio input default hover area background color

kark:
#### Summary

Before:
Screenshot 2023-03-02 at 12 00 32

After
Screenshot 2023-03-02 at 11 54 16


emmenko (approved):

ddouglasz (approved):
👍🏾

PR #2450, opened 2023-03-08
Adding tokens for PrimaryActionDropdown

chloe0592:
Implement new design for `PrimaryActionDropdown` component

**Before:** 
Screenshot 2023-03-08 at 10 53 54

**Now:**
Screenshot 2023-03-08 at 10 53 46


FilPob:
@chloe0592 thanks! Few points:

**The button:**
- the height of the component should be 40px
- the border radius in that dropdown button should be 4px instead of 6
- the padding left & right in the dropdown button should be 8px instead of 4

**The dropdown menu:**
- border radius of the container should be 4 instead of 6

chloe0592:
> @chloe0592 thanks! Few points:
> 
> **The button:**
> 
> * the height of the component should be 40px
> * the border radius in that dropdown button should be 4px instead of 6
> * the padding left & right in the dropdown button should be 8px instead of 4
> 
> **The dropdown menu:**
> 
> * border radius of the container should be 4 instead of 6

@FilPob thank you! ❤️ On it...

emmenko (approved):
🥳 Looks good!

packages/components/primary-action-dropdown/src/primary-action-dropdown.tsx
```diff
@@ -125,12 +132,15 @@ const DropdownChevron = forwardRef<HTMLButtonElement, TDropdownChevron>(
...
-        ...getButtonStyles(props.isDisabled),
+        ...getButtonStyles(props.isDisabled, props.isNewTheme),
         css`
           padding: 0 ${designTokens.spacing10};
           border-left: 1px solid ${designTokens.colorNeutral};
           border-radius: 0 ${designTokens.borderRadius6}
             ${designTokens.borderRadius6} 0;
+          border: ${props.isNewTheme &&
+          `1px solid ${designTokens.colorNeutral}`};
+          margin-left: ${props.isNewTheme && '-1px'};
```
CarlosCortizasCT, line 143: **nit** An alternative would be maybe to use a new design token here.
Example:
```
// default theme
border-for-primary-action-dropdown-chevron: 'none'
margin-left-for-primary-action-dropdown-chevron: '0'

// new theme
border-for-primary-action-dropdown-chevron: '1px solid var(--color-neutral)'
margin-left-for-primary-action-dropdown-chevron: '-1px'
```

chloe0592:
> @chloe0592 thanks! Few points:
> 
> **The button:**
> 
> * the height of the component should be 40px
> * the border radius in that dropdown button should be 4px instead of 6
> * the padding left & right in the dropdown button should be 8px instead of 4
> 
> **The dropdown menu:**
> 
> * border radius of the container should be 4 instead of 6

@FilPob Should be now fixed here - [8f01c0c](https://github.com/commercetools/ui-kit/pull/2450/commits/8f01c0c6723cf8df6808ee258a4e47aa439693fc)

FilPob:
Cool looks all good now... I just found another bug, which is also present in the default theme. It is that if one of the option is too long, it overflows the container. Better would be if it breaks into another line. But it's not related to the design tokens... so we can also skip in in this PR. Up to you

image

chloe0592:
> Cool looks all good now... I just found another bug, which is also present in the default theme. It is that if one of the option is too long, it overflows the container. Better would be if it breaks into another line. But it's not related to the design tokens... so we can also skip in in this PR. Up to you
> 
> image

Good catch! I'll fix it immediately. It's just adding a `white-space` property.
Screenshot 2023-03-08 at 14 41 30

tdeekens (approved):
🙇🏼

packages/components/primary-action-dropdown/src/primary-action-dropdown.tsx
chloe0592: @CarlosCortizasCT Thank you for this! 🙏 I tried to write even a bit more cleaner these few lines, you can check them here, and let me know what you think - https://github.com/commercetools/ui-kit/pull/2450/commits/3800a9d7995a91b2745c15b667c7ff8ecc04eed1

FilPob (approved):

FilPob:
thanks, looks good now :)

CarlosCortizasCT (approved):
👏

PR #2451, opened 2023-03-09
fix(radio-input): option label alignment

kark:
#### Summary

Before:
Screenshot 2023-03-09 at 16 49 32

After:
Screenshot 2023-03-09 at 16 03 29

In `RadioInput` and `RadioField` components' stories we define `Option` label alignment in the stories ([ref](https://github.com/commercetools/ui-kit/blob/29964960c78206aa7b766b2a173f46f0743d8b5c/packages/components/fields/radio-field/src/radio-field.story.js#L112), [ref2](https://github.com/commercetools/ui-kit/blob/29964960c78206aa7b766b2a173f46f0743d8b5c/packages/components/inputs/radio-input/src/radio-input.story.js#L78)). Recent changes in styles and adding `Flexbox` to the label container without specifying the vertical alignment caused some misalignment of the label for the new theme. We missed that unfortunately.

Perhaps adding a ` <Spacings.Inline scale="xs" alignItems="center">` as the `Option`'s children is not the most fortunate decision. We don't show the label normally with an icon in the MC, and here we are forcing the alignment. This is prone to regressions in the future.

So, although it's less visually appealing, my suggestion would be to change our stories to render just text:
```tsx
<RadioInput.Option
 // ...
>
  {text('label #1', 'Banana', radioOption)}
</RadioInput.Option>
```
instead of:
```
<RadioInput.Option
  // ...
>
    <div>{'🍌'}</div>
    {text('label #1', 'Banana', radioOption1)}
  </Spacings.Inline>
</RadioInput.Option>
```

### DIFF
--- a/.changeset/chilled-boxes-yell.md
+++ b/.changeset/chilled-boxes-yell.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/radio-input': patch
+---
+
+Fix `RadioInput.Option` label alignment
--- a/packages/components/fields/radio-field/src/radio-field.story.js
+++ b/packages/components/fields/radio-field/src/radio-field.story.js
@@ -10,7 +10,6 @@ import {
 } from '@storybook/addon-knobs/react';
 import Constraints from '@commercetools-uikit/constraints';
 import { Value } from 'react-value';
-import Spacings from '@commercetools-uikit/spacings';
 import RadioInput from '@commercetools-uikit/radio-input';
 import Section from '../../../../../docs/.storybook/decorators/section';
 import Readme from '../README.md';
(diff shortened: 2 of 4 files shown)


CarlosCortizasCT:
I would suggest to maybe just remove the `Spacing.Inline` wrapper but keep the icon:
```
<RadioInput.Option
  // ...
>
  <div>{'🍌'}</div>
  {text('label #1', 'Banana', radioOption1)}
</RadioInput.Option>
```

CarlosCortizasCT (approved):
Thanks! 🙇

tdeekens (approved):

FilPob (approved):

PR #2453, opened 2023-03-13
fix(toggle-input):  overflow when in focused or hovered state

kark:
#### Summary

`<ToggleInput>` used within `<Spacings.Inline>` in the MC:
Screenshot 2023-03-13 at 14 12 45

### DIFF
--- a/packages/components/inputs/toggle-input/src/toggle-input.styles.ts
+++ b/packages/components/inputs/toggle-input/src/toggle-input.styles.ts
@@ -126,9 +126,14 @@ const getNewThemeFocusIndicator = (props: TStyledSpanProps) => css`
 `;
 
 const getNewThemeMargin = (props: TStyledLabelProps) =>
+  `calc(2px + ${props.themedThumb[props.size].hoverAreaWidth} + (${
+    props.themedThumb[props.size].diameter
+  } - ${props.themedTrack[props.size].height}) / 2)`;
+
+const getMargin = (props: TStyledLabelProps) =>
   css`
     margin: ${props.isNewTheme
-      ? `${props.size === 'big' ? '14px' : '9px'} !important`
+      ? `${getNewThemeMargin(props)} !important`
       : 'unset'};
   `;
 
@@ -137,7 +142,7 @@ const Label = styled.label<TStyledLabelProps>`
   display: inline-block;
   cursor: ${(props) => (props.isDisabled ? 'not-allowed' : 'pointer')};
   align-self: center;
-  ${getNewThemeMargin}
+  ${getMargin}
 
   ${labelSizeStyles}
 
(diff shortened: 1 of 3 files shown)


packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -129,6 +129,8 @@ const Label = styled.label<TStyledLabelProps>`
   position: relative;
   display: inline-block;
   cursor: ${(props) => (props.isDisabled ? 'not-allowed' : 'pointer')};
+  align-self: center;
```
kark, line 132: @Filpob As discussed, I added the margin to take up the whole focus area. 
Screenshot 2023-03-13 at 13 35 51

My only concern in that the wrapping `<Spacing.Inline>` has a different height now. Should the top bottom overflow also be covered. As an alternative we could do:
 
Screenshot 2023-03-13 at 14 10 27

You may notice a slight overflow on top/bottom, but the wrapping `<Spacing.Inline>` doesn’t change the size.

packages/components/inputs/toggle-input/src/toggle.story.js
```diff
@@ -19,17 +25,29 @@ storiesOf('Components|Inputs', module)
...
+                }}
+              />
+              <IconButton
+                icon={<InformationIcon />}
+                label="A label text"
+                onClick={() => {}}
+              />
+            </Spacings.Inline>
+            <Text.Detail>{'This is some content 👇'}</Text.Detail>
+          </>
```
kark, line 50: Added for the time being (for the sake of the UI review)

FilPob:
@kark thanks, looks great as it is. The different overall height of the inline container doesn't matter as long as the children are aligned to the center

kark:
> @kark thanks, looks great as it is. The different overall height of the inline container doesn't matter as long as the children are aligned to the center

@FilPob Just one thing to consider, since the `<Spacings.Inline>` height will increase, and (for some reason) there are places in the MC where we use `<Spacings.Inline alignItems="flexEnd">` that will end up with:
Screenshot 2023-03-13 at 16 28 19
What I mean by that, some changes in the MC will be needed too and we won't be able to cover them here.

packages/components/inputs/toggle-input/src/toggle-input.styles.ts
```diff
@@ -125,10 +125,19 @@ const getNewThemeFocusIndicator = (props: TStyledSpanProps) => css`
   }
 `;
 
+const getNewThemeMargin = (props: TStyledLabelProps) =>
+  css`
+    margin: ${props.isNewTheme
+      ? `${props.size === 'big' ? '14px' : '9px'} !important`
```
emmenko, line 131: What's the rationale behind these "magic" numbers? How are they being computed?

kark: Oh yes, thank you, it's definitely worth explaining. Updated in 4a0a22bcb667bb9521ede104a9a6fbbf72146fb4

CarlosCortizasCT (approved):
👍

chloe0592 (approved):
How you nailed that calculation is beyond me 👏 🚀

ddouglasz (approved):
🙌🏾

PR #2455, opened 2023-03-16
chore: updating README files

chloe0592:
#### Summary

I wanted to update README files for buttons, but when ran the command there were also 3 more files that were out of date.


emmenko (approved):

packages/components/buttons/primary-button/README.md
```diff
@@ -57,7 +57,7 @@ export default Example;
...
 | `onClick`        | `Function`<br/>[See signature.](#signature-onClick)              |          |             | Handler when the button is clicked.&#xA;<br />&#xA;Required when `as` is `undefined`                                                                                                                                                                                        |
-| `size`           | `union`<br/>Possible values:<br/>`'medium' , 'big'`              |          | `'big'`     | Indicates the size of the button.                                                                                                                                                                                                                                           |
+| `size`           | `union`<br/>Possible values:<br/>`'small' , 'medium' , 'big'`    |          | `'big'`     | Indicates the size of the button.                                                                                                                                                                                                                                           |
```
CarlosCortizasCT, line 60: If we want to introduce the `small` value here, we should update the description to state it is deprecated which can be done with the JSDoc of the `size` property in the `TPrimaryButtonProps` type.

kark (approved):

packages/components/buttons/primary-button/README.md
chloe0592: @CarlosCortizasCT I'm not sure I follow, sorry. Could you explain some more, please?

chloe0592: I guess we don't want the 'small' size to be available here. But how do I do that?

kark: Based on [what we announce in the release note](https://github.com/commercetools/merchant-center-application-kit/blob/595d65ae947ed38531f3b7f3b698b7b96931da2f/website/src/releases/2023-03/button-sizes.mdx?plain=1#L14-L17) and what Carlos suggested we could edit JSDoc [here](https://github.com/commercetools/ui-kit/blob/12d05bc502490f271e4255463f2e226af24387bd/packages/components/buttons/primary-button/src/primary-button.tsx#L62-L65) and add something like:

```ts
/**
 * Indicates the size of the button.
 * <br />
 * @deprecated `small` is deprecated. Use `medium` instead.
 *
 */
size?: 'small' | 'medium' | 'big';
```
and re-generate the readme.

chloe0592: Yes @kark that was exactly that. Thank you! 🙏 Carlos jumped in the huddle with me to quickly explain it to me how the README files are working.

chloe0592: Here is the change - [4cf13fe](https://github.com/commercetools/ui-kit/pull/2455/commits/4cf13fe90d5751e354918b7fb63c56b0480d82b3).

packages/components/buttons/primary-button/src/primary-button.tsx
```diff
@@ -61,6 +61,7 @@ export type TPrimaryButtonProps<
   ) => void;
   /**
    * Indicates the size of the button.
+   * @deprecated `small` value is deprecated. Use `medium` instead.
```
CarlosCortizasCT, line 64: I would maybe not add the `@deprecated` keyword as it's like we are deprecating the `size` property which we're not:
```suggestion
   * Indicates the size of the button.
   * <br />
   * `small` value has been deprecated. Please use `medium` value instead as a replacement.
```

chloe0592: Done here - [c6c0c68](https://github.com/commercetools/ui-kit/pull/2455/commits/c6c0c682e028d0edd0a3a2034bb189c4e3dd50fe) ✅

CarlosCortizasCT (approved):
Cool 😎

PR #2456, opened 2023-03-17
Fix and refactor design tokens story

CarlosCortizasCT:
#### Summary

This PR fixes a couple of errors in the Design Tokens story and also refactors its code is an attempt to make it more readable and maintainable.

## Description

We were told about a couple of errors in the Design Tokens story related to not all choices and decision groups being rendered and also another error when trying to use the filter input to narrow the rendered content.

As I was trying to debug the issues I found myself spending too much time trying to understand the way the story works since we had only one big file (700 LOC) with little structure.
I decided to try to refactor the whole story to apply some modularization aiming for an easy to follow code and I also added a couple of new samplers for fonts related values.

The main _controversial_ change is that we don't render both theme values in tokens tables and, instead, the values change when we change the current theme.
For choices, it was useless (at least currently) as 99% of the values are the same in both themes.
For decisions, it was lacking info as both the decision choice and its value can vary between themes but we were only rendering the values.

This PR is a proposal of how I think we could improve this story not only for now but also in case we want to rethink its UI, but of course it's open for debate.


design-system/materials/internals/definition.yaml
```diff
@@ -220,12 +220,6 @@ choiceGroupsByTheme:
         break-point-jumbodesktop: 1920px
 
   test:
-    colors:
```
CarlosCortizasCT, line 223: This is unrelated but it's not needed as it is the same value as in the default theme.

design-system/materials/internals/story/samplers.js
```diff
@@ -0,0 +1,82 @@
+import styled from '@emotion/styled';
+
+const BasicSample = styled.div``;
```
CarlosCortizasCT, line 3: This file contains the small components we use to render some examples of different design token values.

design-system/materials/internals/story/shared-components.js
```diff
@@ -0,0 +1,270 @@
...
+  <>
+    <Token>{props.tokenName}</Token>
+    {getIsDeprecated(props.tokenName) && <DeprecationBadge />}
+  </>
+);
+TokenBodyCell.propTypes = {
+  tokenName: PropTypes.string.isRequired,
+};
+
+export const TokenGroupLinks = (props) => {
```
CarlosCortizasCT, line 55: This is a shared component to render the links we have in the header of the story for every design tokens group.

design-system/materials/internals/story/shared-components.js
```diff
@@ -0,0 +1,270 @@
...
+      label: PropTypes.string,
+    })
+  ).isRequired,
+  cellRenderer: PropTypes.func.isRequired,
+  themeName: PropTypes.string.isRequired,
+  tokensGroupPrefix: PropTypes.string,
+  tokensGroupData: PropTypes.array.isRequired,
+};
+
+export function SingleTokensGroupDetails(props) {
```
CarlosCortizasCT, line 184: This component is used for simple design token groups: states, component groups and variants.

design-system/materials/internals/story/shared-components.js
```diff
@@ -0,0 +1,270 @@
...
+    PropTypes.shape({
+      key: PropTypes.string.isRequired,
+      label: PropTypes.string,
+    })
+  ).isRequired,
+  cellRenderer: PropTypes.func.isRequired,
+  themeName: PropTypes.string.isRequired,
+};
+
+export const MultiTokensGroupDetails = (props) => (
```
CarlosCortizasCT, line 222: This component is used for complex design token groups: choices and decisions.

CarlosCortizasCT:
I don't understand why stylelint complains about the styled component syntax only in this files when we're using that notation in all the repository.

Any suggestion would be much appreciated.

image

emmenko:
>  don't understand why stylelint complains about the styled component syntax only in this files when we're using that notation in all the repository.
> Any suggestion would be much appreciated.

Actually we should remove stylelint, it doesn't really work with CSS-in-JS. We did that in App Kit as well.
https://github.com/commercetools/merchant-center-application-kit/pull/2465#issuecomment-1026901803

The package [@stylelint/postcss-css-in-js](https://www.npmjs.com/package/@stylelint/postcss-css-in-js) is also being deprecated.

.github/workflows/main.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         run: yarn typecheck
 
       - name: Running linters and tests
-        run: yarn run jest --projects jest.{eslint,stylelint,test,ts-test,bundle}.config.js --reporters github-actions
+        run: yarn run jest --projects jest.{eslint,test,ts-test,bundle}.config.js --reporters github-actions
```
CarlosCortizasCT, line 50: I disabled `stylelint` following @emmenko's  suggestion.

Since it currently is only lintin `js` files and we migrated to typescript, I will remove it from the repository as a follow up.

We can have a separate discussion about how we want to lint CSS rules after we talk about its future in `ui-kit`.

emmenko (approved):
Nice work, thanks! 💯

kark (approved):
Thank you!

PR #2457, opened 2023-03-20
Update TS config for better dependencies navigation

CarlosCortizasCT:
#### Summary

When a user tries to follow one of `ui-kit` dependency from their IDE, it will be redirected to the Typescript types declaration file. We think most of the time this user would like to be redirected to the actual source code file.

## Description

The idea comes from [this tweet](https://twitter.com/kuizinas/status/1636641120477384705?s=20) where it's suggested to add the `declarationMap` configuration attribute to **true**.


chloe0592 (approved):
Thank you for this change 🎉

emmenko:
Interesting find, thanks!

Did you check how the build is affected by this?

CarlosCortizasCT:
> I'm wondering how is this supposed to work. When we build the packages we generate bundle files and declaration files. The `declarationMap` works like a source map from what I understand but the package does not have the original source `.ts` files, so how is this supposed to work? 🤔

Ah! I see. Yes, I didn't realize we don't include source files in the published packages.

The ways this works is exactly as you mentioned, mapping files are created for every compiled one.
Example:
```
{
  "version":3,
  "file":"warning.d.ts",
  "sourceRoot":"../../../src",
  "sources":["warning.ts"],
  "names":[],
  "mappings":"AAKA,KAAK,gBAAgB,GAAG,CACtB,SAAS,EAAE,OAAO,EAClB,OAAO,CAAC,EAAE,MAAM,EAChB,MAAM,CAAC,EAAE,MAAM,KACZ,OAAO,CAAC,SAAS,IAAI,OAAO,CAAC;AAOlC,eAAO,MAAM,OAAO,EAAE,gBAUrB,CAAC;AAEF,eAAO,MAAM,UAAU,cAAe,OAAO,WAAW,MAAM,SAK7D,CAAC"
}
```

From the tests I did locally within the `ui-kit`, it seems if the mapped file does not exist, you are redirected to the definitions file (same as what we have now).
I would like to test this from another package though.

So, worst case scenario, if we add this configuration we will be able to navigate to source from within our packages but that will not work from a host package (eg: appkit using uikit), but the experience in this use case would be the same as we currently have (navigate to declarations).

Another option would be for us to include source files in the published package.
I believe it could help `ui-kit` consumers and since we are already providing the compiled bundles, those source files won't have any affect in runtime bundles size for consumers.
Is this something we can consider?

emmenko:
Including source files could be an option yes. Should we give this a try?

PR #2462, opened 2023-03-22
refactor(VRTs): enable showcasing components in the new theme

kark:
#### Summary

The aim of this PR is to showcase ui-kit components in both old- and the new theme in the ui-kit visual regression tests.
The idea is to leave the old theme vrts as they were, where css variables are read from the `:root` element of the visual testing app. On top of that, as part of each of the `<Spec>`s a local `<ThemeProvider>` context is created to showcase the component also in the new theme.

Remarks: 
- since we use `rem` font-size units, all font-size values (even the ones wrapped with local `<ThemeProvider>`) reference the value set in `:root`. This means that the screenshots do not 100% reflect how components look in the new theme. ([more on that in the thread](https://github.com/commercetools/ui-kit/pull/2462#discussion_r1145888671))
- I also considered going with the `new` theme first approach, but then we're experiencing the same issue as 👆. That means the components look a bit differently in the old theme scope and I thought this old theme will be our stable reference to how we are currently rendering components. I'm not sure about it thought 🤷‍♂️  
Screenshot 2023-03-24 at 10 59 29

- Last remark is about the use of `useTheme` hook within the components. Since we don't allow passing a `parentSelector` function to ui-kit components as a prop to use it like `useTheme(props.partentSelector)`, each use of `themedValue` or `isNewTheme` will not reflect the actual state for the nested theme context. Only `designTokens` values are reflected properly.
For instance: 
Screenshot 2023-03-27 at 11 00 11

### DIFF
--- a/packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
+++ b/packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
@@ -30,7 +30,7 @@ const InteractiveRoute = () => {
 
   return (
     <Suite>
-      <Spec label="minimal" omitPropsList testedTheme="old">
+      <Spec label="minimal" omitPropsList testedThemes={['new']}>
         <div>
           <label htmlFor="reset-button">Reset value to Hello World</label>
           <button
--- a/packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
+++ b/packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
@@ -30,7 +30,7 @@ const InteractiveRoute = () => {
 
   return (
     <Suite>
-      <Spec label="minimal" omitPropsList>
+      <Spec label="minimal" omitPropsList testedTheme="old">
         <div>
           <label htmlFor="reset-button">Reset value to Hello World</label>
           <button
--- a/test/percy/spec.jsx
+++ b/test/percy/spec.jsx
@@ -3,9 +3,8 @@ import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { css } from '@emotion/react';
 import pick from 'lodash/pick';
-import { createSequentialId } from '@commercetools-uikit/utils';
-import { useFieldId } from '@commercetools-uikit/hooks';
-import { designTokens, ThemeProvider, useTheme } from '../../design-system';
+import { designTokens } from '../../design-system';
+import { LocalThemeProvider } from './local-theme-provider';
 
 const SpecContainer = styled.div`
   display: flex;
@@ -52,22 +51,12 @@ const PropValue = styled.span`
 `;
 
 const Box = (props) => {
-  const { isNewTheme } = useTheme(
-    props.boxId ? () => document.getElementById(props.boxId) : undefined
-  );
   return (
     <div
       css={css`
         background-color: ${props.backgroundColor ?? designTokens.colorSurface};
         margin-bottom: 20px;
-
-        ${isNewTheme &&
-        css`
-          font-family: 'Inter', system-ui;
-          // Updating 'font-size' here won't make difference since we use 'rem' units in components based on 'font-size' set in the <html> element
-        `}
       `}
-      {...(props.boxId ? { id: props.boxId } : {})}
     >
       <p
         css={css`
@@ -80,7 +69,11 @@ const Box = (props) => {
       >
         {`${props.themeName} theme`}
       </p>
-      {props.children}
+      {props.themeName === 'new' ? (
+        <LocalThemeProvider theme="test">{props.children}</LocalThemeProvider>
+      ) : (
+        props.children
+      )}
     </div>
   );
 };
@@ -88,7 +81,6 @@ Box.propTypes = {
   children: PropTypes.node.isRequired,
   backgroundColor: PropTypes.string,
   themeName: PropTypes.string.isRequired,
-  boxId: PropTypes.string,
 };
 
 const Pill = (props) => {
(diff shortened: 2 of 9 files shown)


test/percy/spec.jsx
```diff
@@ -48,11 +51,45 @@ const PropValue = styled.span`
...
+    <div
+      css={css`
+        background-color: ${props.backgroundColor ?? designTokens.colorSurface};
+        margin-bottom: 20px;
+
+        ${isNewTheme &&
+        css`
+          font-family: 'Inter', system-ui;
+          font-size: 16px; // This won't ever work since we use 'rem' units in components, so the font size will be calculated based on the font-size set in the <html> element
+        `}
```
kark, line 68: Hey @commercetools/shield-team-fe 

I have a question to you. Apparently this setup won’t ever work (at least I can’t think of how I can make it work). We are changing `font-size` in the new theme in the root element and all components use `rem` units. Using this approach with nested <ThemeProvider> contexts in the VRTs makes no sense. Even if we change `font-size`  in the local <ThemeProvider> context it won’t ever be taken into account in the components. 

Screenshot 2023-03-23 at 10 01 52

Would you be ok with using only the new theme in the VRTs?

CarlosCortizasCT: Thanks a lot for looking into this, @kark 🙇 

My suggestion here would be to move forward with this approach and let me explain why: I think VRTs are actually used as a baseline to check we don't break something. Breaking something means a new screenshot **looks different** than the previous stored one, but it actually does not matter too much **how** it looks like. I mean, yes we want the screenshots to look how we expect but for the check sake, it just matters it a new one looks different than the previous one.

Given this is a temporary solution until we finally go back to having just one theme, I think we could tolerate the new one to not use the expected font sizes.

If we ever support more themes, now we now this approach would not be suitable and we might need to have more _credits_ in Percy to run different sets of screenshots.

packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
```diff
@@ -30,7 +30,7 @@ const InteractiveRoute = () => {
 
   return (
     <Suite>
-      <Spec label="minimal" omitPropsList>
+      <Spec label="minimal" omitPropsList testedThemes={['new']}>
```
kark, line 33: Hey @commercetools/shield-team-fe 👋

Let me ask one more question before I invest more time into that.
We have some interactive Puppeteer tests where series of actions run against a component. This applies to components:
- [`<AsyncCreatableSelectInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.visualspec.js#L11-L20)
- [`<AsyncSelectInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/async-select-input/src/async-select-input.visualspec.js#L11-L20)
- [`<DateInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/date-input/src/date-input.visualspec.js#L10-L35)
- [`<LocalizedRichTextInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js#L55-L145)
- [`<RichTextInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/rich-text-input/src/rich-text-input.visualspec.js#L55-L370)
- [`<SearchSelectInput>`](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/search-select-input/src/search-select-input.visualspec.js#L11-L20)

Those tests were intended to run against a single component existing in a document (they use isolated paths and query functions that are used in those tests expect only one piece of tested component in the document). Therefore, if we render the component in both themes in a document - the tests break without a refactor.

The [snapshots are already disabled](https://github.com/commercetools/ui-kit/blob/cb0a756478ff7a54c3ffaa17c4d3329d1c6fa3aa/packages/components/inputs/async-creatable-select-input/src/async-creatable-select-input.visualspec.js#L18-L19) in all of those tests due to issues with `percy`. 

Would you say we need a run-through of all user actions for both themes or using the new theme is enough for us?

test/percy/spec.jsx
```diff
@@ -133,11 +175,13 @@ Spec.propTypes = {
   propsToList: PropTypes.arrayOf(PropTypes.string),
   omitPropsList: PropTypes.bool,
   backgroundColor: PropTypes.string,
+  testedThemes: PropTypes.arrayOf(PropTypes.string),
 };
 
 Spec.defaultProps = {
   omitPropsList: false,
   listPropsOfNestedChild: false,
+  testedThemes: ['new', 'old'],
```
kark, line 184: I also considered using `themingContexts={[“global”, “local”]}` prop instead of this API (`testedThemes={["old”, “new”]}`) used in the PR but for me this approach seemed to be less natural from the DX point of view.

packages/components/inputs/rich-text-input/src/rich-text-input.visualroute.jsx
emmenko: > Would you say we need a run-through of all user actions for both themes or using the new theme is enough for us?

The new theme only is enough for me.

kark:
Based on our the findings from the discussion, the solution does not give enough confidence to follow this path. 
It's been decided to not to show components in the VRTs in the "new" theme for now.

PR #2466, opened 2023-03-27
Select input conditional render value

ChristianMoll:
Adds the `controlShouldRenderValue` prop to `<SelectField>` and `<SelectInput>`


Rhotimee (approved):
Thank you 🚀

kark (commented):
Hi 👋
Thank you, it looks good to me.
Could you please add this new prop to the components' stories [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/fields/select-field/src/select-field.story.js) and [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/select-input/src/select-input.story.js)?

packages/components/fields/select-field/src/select-field.tsx
```diff
@@ -113,6 +113,10 @@ export type TSelectFieldProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   components?: ReactSelectProps['components'];
+  /**
+   * Control whether the selected values should be rendered in the control
+   */
+  controlShouldRenderValue?: boolean;
```
kark, line 119: nit:
```suggestion
  controlShouldRenderValue?: ReactSelectProps['controlShouldRenderValue'];
```

kark, line 119: nit:
```suggestion
 /**
   * Control whether the selected values should be rendered in the control
   * <br>
   * [Props from React select was used](https://react-select.com/props)
   */
  controlShouldRenderValue?: ReactSelectProps['controlShouldRenderValue'];
```

ChristianMoll:
All changes applied, thank you for the feedback ❤️

kark (approved):
Again, many thanks for working on this 🎉
I noticed a little regression caused by this PR, but from my side I don't want to block your work. So I would merge it and we could figure it out intrenally as a follow up. I'm wondering what you think @commercetools/shield-team-fe

CarlosCortizasCT:
> Again, many thanks for working on this 🎉 I noticed a little regression caused by this PR, but from my side I don't want to block your work. So I would merge it and we could figure it out intrenally as a follow up. I'm wondering what you think @commercetools/shield-team-fe
> 
>  [  ](https://user-images.githubusercontent.com/49066275/228228617-d5069462-c52e-4b99-93af-7ad173be925b.gif) [ ](https://user-images.githubusercontent.com/49066275/228228617-d5069462-c52e-4b99-93af-7ad173be925b.gif)

If I got it right, to reproduce the issue you need to be using `isMulti` prop, `controlShouldRenderValue `, select several items and dynamically change the  `controlShouldRenderValue` value.
It seems to me quite an edge case so I wouldn't block this PR either.

PR #2467, opened 2023-03-28
Fix typescript declaration exports for generic types

CarlosCortizasCT:
#### Summary

Fix typescript declaration exports for generic types

## Description

With our current setup about exporting TS types, consumers are finding problems when using generic types:
image

I found almos all our exports are using some kind of bridge that seems to be problematic for this context.
`/packages/components/data-table/src/export-types.ts`:
```ts
import type {
  TRow as Row,
  TColumn as Column,
  TDataTableProps as DataTableProps,
} from './data-table';

export type TRow = Row;
export type TColumn = Column;
export type TDataTableProps = DataTableProps;
```

When changing this to export directly seems to work:
```ts
export type {
  TRow,
  TColumn,
  TDataTableProps,
} from './data-table';
```

Actually, we are already using this pattern in a few components:
`/packages/components/inputs/money-input/src/export-types.ts`:
```ts
export type {
  TCurrencyCode,
  TMoneyValue,
  TValue
} from './money-input';

```

I don't actually understand why, so if anyone has a better understanding, please share your thoughts.


tdeekens (approved):

kark (approved):
Thank you 💯🙏
I also couldn't figure out why we were using this pattern

CarlosCortizasCT:
> If I remember correctly, exporting type directly didn't work well, in particular with Babel.
> 
> Maybe now there is better support for it and it just works. Let's give it a try!

Thanks for your comment 🙏 

I tried this locally and it seems to work. Also, the built bundles seem to have the right types exported.

I would also use the canary release locally to test it out.

CarlosCortizasCT:
> > If I remember correctly, exporting type directly didn't work well, in particular with Babel.
> > Maybe now there is better support for it and it just works. Let's give it a try!
> 
> Thanks for your comment 🙏
> 
> I tried this locally and it seems to work. Also, the built bundles seem to have the right types exported.
> 
> I would also use the canary release locally to test it out.

I tried importing the canary release into a local project and I could confirm the issue in no longer exists 👍.

PR #2478, opened 2023-04-04
chore: upgrade `puppeteer` to `v19`

emmenko:


tdeekens (approved):

kark (approved):

Rhotimee (approved):
Looks good. 

I think a changeset is needed, though.

emmenko:
> I think a changeset is needed, though.

No need in this case, it only affects the VRTs which are only used within the repo.

PR #2479, opened 2023-04-04
fix: types

emmenko:


design-system/src/index.ts
```diff
@@ -5,7 +5,7 @@ export {
   themes,
 } from './design-tokens';
 export { ThemeProvider, useTheme, withThemeContext } from './theme-provider';
-export type { ThemeName, TUseThemeResult } from './theme-provider';
```
emmenko, line 8: Exporting types directly from the package entry point causes issues with the preconstruct dev symlinks.

That's why we use the `./export-types` module trick.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -320,7 +320,7 @@ const Tooltip = (props: TTooltipProps) => {
             // ref accepts `LegacyRef`, which is a union of `RefObject` and `string`
             // propper.ref returns `RefObject`
             ref={popper.ref as LegacyRef<HTMLDivElement>}
-            css={{
+            style={{
```
emmenko, line 323: Getting a weird error with `css`. Since we are passing a CSS object, I think we can use `style`.

image

kark (approved):
Thank you!

design-system/src/index.ts
CarlosCortizasCT: Thanks! 🙇

PR #2484, opened 2023-04-05
docs: `menuPortalZIndex` docs improvements

### LINKED ISSUE #2477: SelectableSearchInput menuPortalZIndex not being applied properly
**Describe the issue**
When I am using the SelectableSearchInput the menuPortalZIndex seems to not be working and the dropdown is hiding behind items that are below it, as seen in the screenshot below. I have made the z-index 3 as well as the max option but neither works. For this particular case, the z-index of the "Actions" dropdown is only 2. 
Screenshot 2023-04-03 at 10 40 16

```
<SelectableSearchInput
            value={initialValueObject}
            placeholder={getBasicInputPlaceholder()}
            onChange={handleChange}
            options={getSearchOptions()}
            onSubmit={handleBasicSearchSubmit}
            onReset={props.clearSearchTerm}
            horizontalConstraint={16}
            menuHorizontalConstraint={3}
            menuPortalZIndex={10000000}
            isDisabled={props.isDisabled}
          />
```

**To Reproduce**

1. Go to 'Products list'
2. Click on 'All fields' dropdown
3. See error

**Expected behavior**
The dropdown will go over the Actions dropdown 

**Screenshots**

**Additional context**

Maybe this will give you a hint but my previous solution to solve this was to wrap a div around SelectableSearchInput that had z-index: 3.

kark:
#### Summary

closes #2477

Perhaps there is some misunderstanding on my part and those improvements are not needed at all.


packages/components/inputs/select-input/src/select-input.story.js
```diff
@@ -106,47 +114,85 @@ storiesOf('Components|Inputs/SelectInputs', module)
...
+                          zIndex: 2,
+                        }}
+                      >
+                        <Card theme="dark">
+                          Stacking context with <code>z-index: 2</code>
+                        </Card>
+                      </div>
+                    )
+                  );
+                })()}
```
kark, line 194: I wanted to keep those knobs close to each other, I don’t know any other way of providing a certain order of knobs. This is the way I thought of.

Also, would you like me to copy this implementation in all other stories where we test components that have this prop?

packages/components/fields/async-creatable-select-field/README.md
```diff
@@ -76,7 +76,7 @@ export default Example;
...
-| `menuPortalZIndex`        | `number`                                                                                              |          |           | z-index value for the menu portal                                                                                                                                                                                                                                                                                                                                                         |
+| `menuPortalZIndex`        | `number`                                                                                              |          |           | z-index value for the menu portal&#xA;<br>&#xA;Use in conjunction with `menuPortalTarget`                                                                                                                                                                                                                                                                                                 |
```
kark, line 79: I am not adding a changeset as this PR is mainly updating READMEs and a story(ies).

packages/components/inputs/select-input/src/select-input.story.js
Rhotimee: I believe It will also be helpful in other stories.

Rhotimee:
Nice work @kark. What do you think about warning the user when `menuPortalZIndex` is used without `menuPortalTarget`?

Rhotimee:
> Thanks, yes, I think it makes sense to me. I'm a little concerned that this might blow up in several places in the MC-FE 😅 But unless anyone is strongly against, let's go for it and try to fix all the occurrences in the MC.

It will only blow up if `menuPortalTarget` isn't used which means, the zIndex wasn't working properly anyway. 
Let's get other opinions though.

kark:
> It will only blow up if `menuPortalTarget` isn't used which means, the zIndex wasn't working properly anyway. Let's get other opinions though.

I added the warnings [here](324d72416b271c901b6912c018a626278c481233) but now I think we shouldn’t do that. 

For some reason we pass `menuPortalZIndex: 1` in `defaultProps` for all of those components without passing `menuPortalTarget`. Now we must also provide `menuPortalTarget`, otherwise our tests fail. I’m not sure but I think this is technically a breaking change as `menuPortalZIndex` will start to take effect in all the components. And if we delete `menuPortalZIndex` from `defaultProps` I guess it is also a breaking change, because client code may already be relying on that value.

emmenko (commented):
Thanks a lot @kark for working on it and improving the developer experience 🤗

packages/components/inputs/select-utils/src/warning.ts
```diff
@@ -0,0 +1,22 @@
...
+const getMessage = (componentName: string) =>
+  `${componentName}: use \`menuPortalZIndex\` in conjunction with \`menuPortalTarget\``;
+
+type TWarnAgainstMisuseOfMenuPortalProps = {
+  menuPortalZIndex?: number;
+  menuPortalTarget?: ReactSelectProps['menuPortalTarget'];
+  componentName: string;
+};
+
+export const warnAgainstMisuseOfMenuPortalProps = (
```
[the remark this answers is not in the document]
kark: Thank you, sounds good!

emmenko (approved):
💯

CarlosCortizasCT (approved):
Super! 💪

PR #2486, opened 2023-04-06
fix: proper border color when localized input in both disabled and read-only state

### LINKED ISSUE #2483: LocalizedMultilineTextInput and LocalizedTextInput border around language disappears when both isDisabled and isReadOnly is toggled
**Describe the bug**
LocalizedMultilineTextInput and LocalizedTextInput border around language disappears when both isDisabled and isReadOnly is toggled. We would like it to behave like isDisabled. However we need both isDisabled and isReadOnly to be used for accessibility purposes. 
Screenshot 2023-04-05 at 12 08 35

When used practically the disabled language doesn't have a border around it but the text field next to it does.
Screenshot 2023-04-05 at 12 00 40

**To Reproduce**

1. Go to 'Reviewed modified products'
2. Scroll and find a localizedTextInput or LocalizedMultilineTextInput

**Expected behavior**
Both the language and the text field have a border around them, like the editable language field

kark:
#### Summary

closes #2483

Before:
|  LocalizedTextInput |  LocalizedMultilineTextInput |  
|---|---|
|  Screenshot 2023-04-06 at 14 57 48 |   Screenshot 2023-04-06 at 14 59 24 |

After:
|  LocalizedTextInput | LocalizedMultilineTextInput  |  
|---|---|
|  Screenshot 2023-04-06 at 14 58 44 |  Screenshot 2023-04-06 at 15 00 52 |

### DIFF
--- a/packages/components/inputs/localized-text-input/src/localized-text-input.styles.ts
+++ b/packages/components/inputs/localized-text-input/src/localized-text-input.styles.ts
@@ -25,6 +25,15 @@ const getLanguageLabelBackgroundColor = (props: TLocalizedInputProps) => {
   return designTokens.backgroundColorForLocalizedInputLabel;
 };
 
+const getLanguageLabelBorderColor = (props: TLocalizedInputProps) => {
+  if (props.isDisabled) {
+    return designTokens.borderColorForInputWhenDisabled;
+  }
+  return props.isReadOnly
+    ? designTokens.borderColorForLocalizedInputLabelWhenReadonly
+    : designTokens.borderColorForLocalizedInputLabel;
+};
+
 const getLanguageLabelStyles = (
   props: TLocalizedInputProps & { isNewTheme: boolean }
 ) => {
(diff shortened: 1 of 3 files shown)


tdeekens (approved):
💯 Thanks.

Side note: to me the underlying challenge is a bit the boolean 👿. Why do we need readonly and disabled. Does the visual difference have a reason and value? Then at times thinking enums over booleans can help.

kark:
> 💯 Thanks.
> 
> Side note: to me the underlying challenge is a bit the boolean 👿. Why do we need readonly and disabled. Does the visual difference have a reason and value? Then at times thinking enums over booleans can help.

Sure, I feel you. I'm not sure if the visual difference have a reason and value. (As a side note, I think it was directed more to Filip than myself, right? 😅). But I suppose we need to differentiate those states somehow as more than the visuals comes along with those states (a11y, copy-pasting, evaluation by form processors). Did I get your point?

packages/components/inputs/localized-text-input/src/localized-text-input.styles.ts
```diff
@@ -25,6 +25,15 @@ const getLanguageLabelBackgroundColor = (props: TLocalizedInputProps) => {
   return designTokens.backgroundColorForLocalizedInputLabel;
 };
 
+const getLanguageLabelBorderColor = (props: TLocalizedInputProps) => {
+  if (props.isDisabled) {
+    return designTokens.borderColorForInputWhenDisabled;
+  }
+  return props.isReadOnly
```
CarlosCortizasCT, line 32: In the previous implementation we had the `isNewTheme` prop as part of the condition. Did you check we don't actually need it?

kark: Thanks Carlos, 
we don't need it as we're now using different design tokens.

CarlosCortizasCT (approved):
💯

Rhotimee (approved):
Thanks for fixing this.

PR #2487, opened 2023-04-07
fix: vertical alignment of the external Link's icon

kark:
#### Summary

For some of the redesigned components `line-height` has been increased. If those components render `<Link isExternal>` as their child, the Link's icon and text are currently misaligned.
Screenshot 2023-04-07 at 13 16 27

Credit to @bradd123 💪

### DIFF
--- a/packages/components/link/src/link.story.js
+++ b/packages/components/link/src/link.story.js
@@ -1,5 +1,12 @@
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text, boolean, select } from '@storybook/addon-knobs/react';
+import {
+  Text,
+  Card,
+  ContentNotification,
+  Spacings,
+  CheckboxInput,
+} from '@commercetools-frontend/ui-kit';
 import { BrowserRouter as Router } from 'react-router-dom';
 import Section from '../../../../docs/.storybook/decorators/section';
 import Readme from '../README.md';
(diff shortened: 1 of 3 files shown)


packages/components/link/src/link.story.js
```diff
@@ -16,13 +23,39 @@ storiesOf('Components|Links', module)
...
+          </ContentNotification>
+          <CheckboxInput
+            value="foo-radio-value"
+            onChange={(event) => alert(event.target.value)}
+            isChecked={true}
+          >
+            <Link tone="primary" to="/foo/bar" isExternal>
+              text
+            </Link>
+          </CheckboxInput>
```
kark, line 57: This is just a temporary change to showcase `<Link>` wrapped with various components. Will be reverted.

packages/components/link/src/link.tsx
```diff
@@ -127,7 +127,7 @@ const getLinkStyles = (props: TLinkProps & { isNewTheme: boolean }) => {
 const Wrapper = styled.span`
   > svg {
     margin: 0 0 0 ${designTokens.spacing10} !important;
-    vertical-align: bottom;
+    vertical-align: ${designTokens.verticalAlignForLinkIcon};
```
CarlosCortizasCT, line 130: I was wondering whether we actually need a token. When I check the updated story in the preview deployment I see using `middle` in the default theme also seems to work, but maybe I'm missing something.

kark: Thanks, that's a solid point 👍
I actually wanted to bring it up tomorrow. It seems fine from my side as well but I also wanted to double check with everyone.

FilPob (approved):

bradd123 (approved):

Issue #2445, opened 2023-03-02
Adding TagList component to the UI Kit


chloe0592:
**Is your feature request related to a problem? Please describe.**
We were trying to understand how can `Tags` be rendered next to each other in the list.
![Screenshot 2023-03-02 at 12 12 02](https://user-images.githubusercontent.com/52276952/222412770-52d7e09c-dc66-4c9f-8fed-1cf77f1015fd.png)

Initially we thought that it's related to the `HorizontalConstraints` ([issue](https://github.com/commercetools/ui-kit/issues/2434)) or `Spacings`, but as it turns out in the MC it's being wrapped with the `TagList` component or with the `flex` custom styling. 

**Describe the solution you'd like**
Since `TagList` is being used in multiple places in the MC repo, it's time we introduce it in the UI Kit.

**References in MC repo:**
`TagList` component: tag-list.jsx
Custom styling: selected-categories.jsx

PR #2489, opened 2023-04-11
Adding `TagList` component

chloe0592:
Since `TagList` is being used in multiple places in the MC repo, it's time we introduce it in the UI Kit.

Closes [this issue.](https://github.com/commercetools/ui-kit/issues/2445)


packages/components/tag/src/tag-list/index.ts
chloe0592, line 1: @CarlosCortizasCT what do you think, is it alright that this file is inside the `Tag` folder and it has its own subfolder or should it be separate? 
I thought it might be like this since the whole purpose of the component is basically just a helper for the tags. It won't be used for other components. I've seen something similar in `Pagination`.

CarlosCortizasCT: That's a good point.
From my side, I see this implementation fine although there's something a little off since the `@commercetools-uikit/tag` currently exports the `Tag` component as default:
```jsx
import Tag from '@commercetools-frontend/ui-kit';
```

Now we would be exporting also the `TagList` component so if you would use both together, the import would look like this:
```jsx
import Tag, { TagList } from '@commercetools-frontend/ui-kit';
```
which it's ok, but I tiny awkward to me.

We could add `Tag` as a named export also and keep the default one for compatibility purposes:
```jsx
import { Tag, TagList } from '@commercetools-frontend/ui-kit';
```

Or just leave it like you have it now.
Let's see other's opinion.

As a side note, I believe we should focus on providing just our main single preset package to our customers (`@commercetools-frontend/ui-kit`) and make sure we bundle it so they can selective include the packages they use in their bundles.
Whatever we decide, please do not forget to include the new component in [that preset](https://github.com/commercetools/ui-kit/blob/main/presets/ui-kit/src/index.ts).

chloe0592: @CarlosCortizasCT Thank you for your remark! I've updated it in this change - [b1df854](https://github.com/commercetools/ui-kit/pull/2489/commits/b1df854096d92c5a224fd1d3f5becbe85c35a255). Let me know if there's something else off about it.

packages/components/tag/src/tag.story.js
```diff
@@ -26,7 +29,23 @@ const Story = () => (
     </Tag>
   </Section>
 );
-Story.displayName = 'Story';
+TagStory.displayName = 'TagStory';
+
+const TagListStory = () => (
+  <Section>
+    <TagList>
```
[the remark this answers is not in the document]
CarlosCortizasCT: Maybe wrapping the `TagList` into a `Constraints.Horizontal` component will reduce the width and make it easier to see the different rows.

packages/components/tag/src/index.ts
```diff
@@ -1,3 +1,8 @@
+/** @deprecated use new way below of exporting the Tag component*/
```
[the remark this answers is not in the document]
chloe0592: Yes, of course I should’ve added the context to the warning 🤦‍♀️ Thanks! [ce0895a](https://github.com/commercetools/ui-kit/pull/2489/commits/ce0895a84f65ec64d74b3c0c53c4210838be1309)

packages/components/tag/docs/additional-info.md
```diff
@@ -0,0 +1,23 @@
...
+<TagList>
+  {examples.map((tag, index) => (
+    <Tag
+      key={index}
+    >
+      {tag}
+    </Tag>
+  ))}
+</TagList>
+/>
```
kark, line 22: My suggestion would be to use a bit more tangible example (so that everyone knows it should be an array of `ReactNode`s):
```suggestion
<TagList>
  {["apple", "orange"].map((tag, index) => (
    <Tag
      key={index}
    >
      {tag}
    </Tag>
  ))}
</TagList>
```

packages/components/tag/src/tag.visualroute.jsx
```diff
@@ -194,5 +199,43 @@ export const component = () => (
         {longText}
       </Tag>
     </Spec>
+    <Spec
+      label="Tag list - with remove"
+      propsToList={[...basePropsToList, 'to']}
```
kark, line 204:
```suggestion
```
(as those props to list relate more to the `<Tag>`)

kark (commented):
Thank you, implementation-wise looks great :sparkles:
I'm only leaving some nit-picky comments :neckbeard:

chloe0592:
> Thank you, implementation-wise looks great ✨ I'm only leaving some nit-picky comments :neckbeard:

Thank you for all the remarks! Really appreciate it 🍫

packages/components/tag/src/tag-list/tag-list.tsx
```diff
@@ -0,0 +1,32 @@
+import { ReactNode, Children } from 'react';
+import { designTokens } from '@commercetools-uikit/design-system';
+import styled from '@emotion/styled';
+
+export type TTagListProps = {
+  children: ReactNode;
+  tag: string;
```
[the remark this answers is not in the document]
chloe0592: Initially thought I'm gonna need it, but then forgot to remove it later. 🙈 Great catch, thanks! 🦅 [0f9d895](https://github.com/commercetools/ui-kit/pull/2489/commits/0f9d895221e096764842eb483c0eb6e1e1d68f02)

packages/components/tag/src/tag-list/tag-list.tsx
```diff
@@ -0,0 +1,32 @@
+import { ReactNode, Children } from 'react';
```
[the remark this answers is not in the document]
chloe0592: Added the `type` here - [a14e3cf](https://github.com/commercetools/ui-kit/pull/2489/commits/a14e3cf869a705397fd43d237c80de623e4b4051).

kark (approved):
Thank you, looks great from my end 🙌

CarlosCortizasCT (approved):
Awesome! 🙌

chloe0592:
@FilPob could you also please, when you have a bit of time, just quickly take a look if it looks alright from the design point of view 🙏

FilPob:
looks great! thanks :)

FilPob (approved):

packages/components/tag/README.md
```diff
@@ -82,3 +82,28 @@ export default Example;
...
+# TagList
+
+## Description
+
+A `TagList` component is used as wrapper for the list of tags. It has predefined spacings and flex display so that tags are grouped next to each other.
+
+## Usage
+
+The `<TagList>` component accepts `<Tag>` element as `children` props.
+Here's the example how to use them:
```
Rhotimee, line 95: NIT

```suggestion
Here's an example of how to use them:
```

packages/components/tag/src/tag-list/tag-list.tsx
```diff
@@ -0,0 +1,31 @@
...
+`;
+
+const TagListItem = styled.div`
+  margin: ${designTokens.marginForTagList};
+`;
+
+const TagList = (props: TTagListProps) => {
+  return (
+    <TagListContainer>
+      {Children.map(props.children, (tag) => (
```
Rhotimee, line 22: Do we want to allow any child that is not a Tag component? If not, it will be nice to warn the user to use only the Tag component with `TagList`.

Rhotimee (approved):
Nice work @chloe0592 🙌🏽

packages/components/tag/src/tag-list/tag-list.tsx
chloe0592: Screenshot 2023-04-17 at 15 21 21

Rhotimee: Valid Point @CarlosCortizasCT

PR #2490, opened 2023-04-11
fix: `<SelectInput>` styles when `isMulti` prop is used together with `controlShouldRenderValue`

kark:
#### Summary

Issue originating from [the discussion](https://github.com/commercetools/ui-kit/pull/2466#issuecomment-1486896859).

### DIFF
--- a/packages/components/inputs/select-input/src/select-input.tsx
+++ b/packages/components/inputs/select-input/src/select-input.tsx
@@ -402,6 +402,7 @@ const SelectInput = (props: TSelectInputProps) => {
               iconLeft: props.iconLeft,
               isMulti: props.isMulti,
               hasValue: !isEmpty(selectedOptions),
+              controlShouldRenderValue: props.controlShouldRenderValue,
               isNewTheme,
             }) as ReactSelectProps['styles']
           }
--- a/packages/components/inputs/select-input/src/select-input.tsx
+++ b/packages/components/inputs/select-input/src/select-input.tsx
@@ -322,10 +322,11 @@ export type TSelectInputProps = {
 
 const defaultProps: Pick<
   TSelectInputProps,
-  'maxMenuHeight' | 'menuPortalZIndex'
+  'maxMenuHeight' | 'menuPortalZIndex' | 'controlShouldRenderValue'
 > = {
   maxMenuHeight: 220,
   menuPortalZIndex: 1,
+  controlShouldRenderValue: true,
 };
 
 const SelectInput = (props: TSelectInputProps) => {
--- a/.changeset/fast-ducks-run.md
+++ b/.changeset/fast-ducks-run.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/select-input': patch
+'@commercetools-uikit/select-utils': patch
+---
+
+Fix `<SelectInput>` styles when `isMulti` prop is used together with `controlShouldRenderValue`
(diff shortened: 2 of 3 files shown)


packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -322,10 +322,11 @@ export type TSelectInputProps = {
 
 const defaultProps: Pick<
   TSelectInputProps,
-  'maxMenuHeight' | 'menuPortalZIndex'
+  'maxMenuHeight' | 'menuPortalZIndex' | 'controlShouldRenderValue'
 > = {
   maxMenuHeight: 220,
   menuPortalZIndex: 1,
+  controlShouldRenderValue: true,
```
kark, line 329: I'm not sure if you'll like this change. This is [by default done by `react-select` under the hood](https://github.com/JedWatson/react-select/blob/1bbbc47f5bdd3bd81f6b95ada5554cdb8578ced5/packages/react-select/src/Select.tsx#L289), so I thought we might mimic that here as well and pass the prop to the `<Select>` component and to the `createSelectStyles` helper function.

This was the easier way of fixing this bug. Another I think would require adding a few more cases here:

https://github.com/commercetools/ui-kit/blob/ed3f0c1985b590ecdaf2b248edd8cb6db4252f6a/packages/components/inputs/select-utils/src/create-select-styles.ts#L255-L259

kark:
> Just sharing my first thought regarding this issue: Do we actually want to update our code to fix it?
> 
> I'd like to quote what I understood in the original PR:
> 
> > If I got it right, to reproduce the issue you need to be using `isMulti` prop, `controlShouldRenderValue` , select several items and dynamically change the `controlShouldRenderValue` value.
> > It seems to me quite an edge case so I wouldn't block this PR either.
> 
> I'm trying to think in a use case where a developer would need to change the value of the `controlShouldRenderValue` dynamically and I don't find it. Moreover, it would need to be done in a very specific situation.
> 
> My point here is that I believe we should aim for fully functional components but, at the same time, the simple implementation we will maintain. It's just finding the balance can be tricky. In this case, I feel we're adding some complexity that is not worthy.
> 
> I'm interested in hear about other teammates opinion.

Thank you Carlos,

Just one word from my side. This issue doesn't only appear only when `controlShouldRenderValue` is changed dynamically. It's visible always when `isMulti` is truthy and `controlShouldRenderValue` is falsy.
Screenshot 2023-04-11 at 15 59 23
I can update our vrt for this case if you think it's needed.

CarlosCortizasCT:
> Thank you Carlos,
> 
> Just to make everyone aware. This issue doesn't only appear only when `controlShouldRenderValue` is changed dynamically. It's visible always when `isMulti` is truthy and `controlShouldRenderValue` is falsy. Screenshot 2023-04-11 at 15 59 23 I can update our vrt for this case if you think it's needed.

Ok, sorry. Based on the comments in the previous PR, I assumed we didn't have any other use case, but this one definitely needs to be addressed.

Thanks for the explanation.

packages/components/inputs/select-input/src/select-input.tsx
CarlosCortizasCT: Yes, I believe it makes sense 👍

CarlosCortizasCT (approved):
Thanks Kacper 🙇

PR #2491, opened 2023-04-13
Fix `LoadingSpinner` early rendering

### LINKED ISSUE #2425: SSR support: Render loading spinner without timeout if delay=0
Setting the loading spinner delay to 0 does still only render it after a zero ms timeout.  I tried to render it server side and my guess is that that may be the cause why it does only render after client side hydration (not an MC app kit use case).  

The following change (patched in node_modules to try out) resolved the issue: 

https://github.com/commercetools/ui-kit/blob/6fb466f4d59667ccb5731a46692bf54e3e4f19c8/packages/components/loading-spinner/src/loading-spinner.tsx#L62

```js
if (!showSpinner && props.maxDelayDuration > 0 ) return null; 
```

Worth including?

CarlosCortizasCT:
#### Summary

Fix `LoadingSpinner` early rendering.

closes #2425

## Description

When providing a `0` value to `maxDelayDuration` a consumer would expect this component to render right away but that's not technically the case since the component will only start actually render its contents after its first initial render kicks a `useEffect` in, which in turn fires a `setTimeout` which (again) technically can last more that 0 milliseconds as it will add a task to the JS engine event loop and it will depend on how many tasks are already in the loop.

Also, when using the component in a SSR environment, the component would never render since `useEffect` does not run in that environment.

### DIFF
--- a/.changeset/wild-ways-hunt.md
+++ b/.changeset/wild-ways-hunt.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/loading-spinner': patch
+---
+
+Fix early rendering when providing `0` as the value for `maxDelayDuration` prop.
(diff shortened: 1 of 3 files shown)


emmenko (approved):
Thanks

Rhotimee (approved):
💯

Issue #2474, opened 2023-04-04
Support white-space CSS property in Text components


ahmehri:
**Is your feature request related to a problem? Please describe.**

When using `Spacings.Inline` to lay out two elements side-by-side, if one of the elements is a text element and the other is large enough, the text may wrap, which is undesirable in some cases.

For example, consider the following code. The desired behavior is for the text "Product type" to be on a single line.

```js
<Spacings.Inline alignItems="baseline">
  <TextComponent intlMessage={messages.productTypeName} />
  <FlatButton label={productType.obj.name || ''} />
</Spacings.Inline>
```

<img width="379" alt="CleanShot 2023-04-04 at 15 01 20@2x" src="https://user-images.githubusercontent.com/3668245/229799878-2baa9bc2-282a-4a72-abfc-b52f0ffc59e1.png">

**Describe the solution you'd like**

The solution is to support the `white-space` CSS property in Text components.

```js
<Spacings.Inline alignItems="baseline">
  <TextComponent intlMessage={messages.productTypeName} whiteSpace="nowrap" />
  <FlatButton label={productType.obj.name || ''} />
</Spacings.Inline>
```

<img width="379" alt="CleanShot 2023-04-04 at 15 04 22@2x" src="https://user-images.githubusercontent.com/3668245/229800651-68052376-f3ec-4d56-9281-e67b61e1c800.png">

**Describe alternatives you've considered**

The current workaround is to use the Grid layout, which is not ideal for simple one-dimensional layout use cases.

kark:
Hey @ahmehri,
let me thank you again for creating the issue.
We'll look into it.

PR #2492, opened 2023-04-13
Adding `nowrap` prop to the Text component

chloe0592:
When using `Spacings.Inline` to lay out two elements side-by-side, if one of the elements is a text element and the other is large enough, the text may wrap, which is undesirable in some cases.

That's why we've introduced the `nowrap` prop for the Text components. 
It renders the `white-space` CSS prop and then the element will not be wrapped to a new line.

Closes [this issue](https://github.com/commercetools/ui-kit/issues/2474).

**Before the change:**
before_nowrap

**After the change:** 
after_nowrap

### DIFF
--- a/.changeset/serious-peas-run.md
+++ b/.changeset/serious-peas-run.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/text': minor
+---
+
+adding new prop `nowrap` to the Text components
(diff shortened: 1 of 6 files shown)


emmenko (approved):
💯 🙌

Rhotimee (approved):

PR #2493, opened 2023-04-14
Export more Typescript types from Select components

CarlosCortizasCT:
#### Summary

Export more Typescript types from Select components

## Description

In order to help consumers using `SelectInput` and `SelectField` components, I propose to expose more TS types from them.

### DIFF
--- a/packages/components/fields/select-field/src/select-field.tsx
+++ b/packages/components/fields/select-field/src/select-field.tsx
@@ -20,15 +20,15 @@ import FieldErrors from '@commercetools-uikit/field-errors';
 import type { Props as ReactSelectProps } from 'react-select';
 
 type TErrorRenderer = (key: string, error?: boolean) => ReactNode;
-type TOption = {
+export type TOption = {
   value: string;
   label?: ReactNode;
 };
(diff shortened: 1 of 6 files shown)


packages/components/fields/select-field/src/select-field.tsx
```diff
@@ -196,7 +196,7 @@ export type TSelectFieldProps = {
    * <br/>
    * The value will be the selected option, or an array of options in case isMulti is true.
    */
-  onChange?: (event?: TCustomEvent) => void;
+  onChange?: (event: TCustomEvent) => void;
```
CarlosCortizasCT, line 199: The event was marked as optional but this callback is just passed internally to the `SelectInput` component where the event is not optional.

kark (approved):
Thank you 🙌🏅
Just a kind reminder about a changeset.

Rhotimee (approved):
Nice work. Just changeset missing. Thanks

CarlosCortizasCT:
Changesets added here: 43221be4d61c0d510ed2d9a2b6e38adf1c468f35

Thanks for the reminder 🙇

PR #2496, opened 2023-04-18
chore: update typescript to v5

emmenko:


packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.tsx
```diff
@@ -26,14 +26,7 @@ import TranslationInput from './translation-input';
 import RequiredValueErrorMessage from './required-value-error-message';
 import { warning } from '@commercetools-uikit/utils';
 
-type TState = {
-  [key: string]: string;
-};
-
-type TExpandedTranslationsReducerState = {
-  state?: TState;
-};
```
emmenko, line 35: This was apparently wrong. Not sure how it got unnoticed until now.

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -291,7 +290,7 @@ export type TSelectInputProps = {
   /**
    * Array of options that populate the select menu
    */
-  options?: TOptions;
+  options: TOptions;
```
emmenko, line 293: Nothing changed here, the prop is still optional with a default value

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -319,7 +327,7 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
     if (props.onSubmit) {
       props.onSubmit({
         text: searchValue,
-        option: selectedOption?.value,
+        option: selectedOption?.value ?? '',
```
emmenko, line 330: This needs to be double checked. The `option` field is apparently required but the `selectedOption?.value` can be `undefined`.
How should we handle this? Is it fine to return an empty string?

@commercetools/shield-team-fe

CarlosCortizasCT (approved):
Thanks! 🙌

kark (approved):
Thank you! 🎉

PR #2497, opened 2023-04-18
fix: enable passing `data-*` attributes to `TagList`

kark:
#### Summary

I’ve been trying to integrate the new `<TagList>` component in the mc-fe and it turned out that:
- in almost every case we apply custom styles via a `className`. Some styles are quite sophisticated so I didn't want to change anything in the implementation and only pass the classname.
- most of the components are tested via `data-testid` attribute

We accept `className` as a prop in case of several ui-kit components, so I thought about doing it here as well instead of using a wrapper with custom styles in the client code. But I'm open to suggestions if you see it differently.

Another thing is the changeset. I've enclosed it but I'm not sure we need it if the component hasn’t been released yet 🤔, any ideas?

### DIFF
--- a/.changeset/cuddly-months-give.md
+++ b/.changeset/cuddly-months-give.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/tag': patch
 ---
 
-Enable passing a class name and `data-*` attributes to the `<TagList>` component
+Enable passing `data-*` attributes to the `<TagList>` component
--- a/.changeset/cuddly-months-give.md
+++ b/.changeset/cuddly-months-give.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tag': patch
+---
+
+Enable passing a class name and `data-*` attributes to the `<TagList>` component
--- a/packages/components/tag/src/tag-list/tag-list.tsx
+++ b/packages/components/tag/src/tag-list/tag-list.tsx
@@ -6,6 +6,8 @@ import styled from '@emotion/styled';
 export type TTagListProps = {
   children: ReactNode;
   /**
+   * @deprecated
+   *
    * Allow to override the styles by passing a `className` prop.
    * <br/>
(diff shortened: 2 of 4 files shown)


packages/components/tag/src/tag-list/tag-list.tsx
```diff
@@ -18,7 +25,10 @@ const TagListItem = styled.div`
 
 const TagList = (props: TTagListProps) => {
   return (
-    <TagListContainer>
+    <TagListContainer
+      className={props.className}
```
CarlosCortizasCT, line 29: From my point of view, `className` is something to avoid from a design system perspective by default and only to be used with a solid reason. The goal os a design system is to create a visual representation of a brand and consumers should only be able to tailor the experience in a controlled manner so the brand can make sure its guidelines are followed. Otherwise, you could use a set of styleless components (eg: [headlessui](https://headlessui.com/), [radix](https://www.radix-ui.com/)).

That said, I totally get your point about the current state of the MC and (temporary) introducing this prop could be the best worst solution. 
Maybe don't mention it in the docs and also mark it already as depretated.

chloe0592 (approved):
@kark good catch! I agree that we are trying to avoid classnames with this component. But you're right it should be backwards compatible with the TagLists we already have in the MC. So nothing further to add here from my side. 
Thank you for working on it! 👏

.changeset/cuddly-months-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/tag': patch
+---
+
+Enable passing a class name and `data-*` attributes to the `<TagList>` component
```
CarlosCortizasCT, line 5: Sorry to be so picky, but I would also not mention `class name` here as it's mostly intended for our internal applications use case and we don't want consumers to use it.
If we don't mention it, we will have more chances that no-one uses it.

CarlosCortizasCT (approved):
Thanks a lot 🙌

PR #2498, opened 2023-04-18
Disable flaky visual regression tests

CarlosCortizasCT:
#### Summary

Disable flaky visual regression tests

## Description

We have some visual regression tests that are constantly and randomly failing because the components they tests execute some ui adjustments after being mounted and the current Percy implementation does not address them reliable in terms of timing.

We're disabling them until we (or Percy) find a solid solution.

image


kark (approved):
Thank you! Great idea!

Rhotimee (approved):

PR #2504, opened 2023-04-25
fix: ui regression with `controlShouldRenderValue`

### LINKED ISSUE #2503: SearchSelectInput UI seems different when redesign is toggled ON
**Describe the bug**
The `SearchSelectInput` UI seems different when isMulti is passed as true. eg, The width of each tag is more than the actual content.

**To Reproduce**

1. Go to SearchSelectInput in storybook
2. Enable `isMulti` prop
3. Try to add select options from the SearchSelectInput

**Screenshots**
UI on left is how it is supposed to look. 
UI on right is a bug. (Confirmed by filip [here](https://commercetools.slack.com/archives/C04F6CF7NGK/p1682407662315729?thread_ts=1682407522.739989&cid=C04F6CF7NGK) ) 

Screenshot 2023-04-25 at 09 22 09

kark:
#### Summary

Closes #2503
Regression resulting from https://github.com/commercetools/ui-kit/pull/2490/

Appearing in all select fields and inputs (apart from `SelectField` and `SelectInput` where the default `controlShouldRenderValue` prop value was added).

### DIFF
--- a/packages/components/inputs/select-input/src/select-input.tsx
+++ b/packages/components/inputs/select-input/src/select-input.tsx
@@ -324,11 +324,10 @@ export type TSelectInputProps = {
 
 const defaultProps: Pick<
   TSelectInputProps,
-  'maxMenuHeight' | 'menuPortalZIndex' | 'controlShouldRenderValue' | 'options'
+  'maxMenuHeight' | 'menuPortalZIndex' | 'options'
 > = {
   maxMenuHeight: 220,
   menuPortalZIndex: 1,
-  controlShouldRenderValue: true,
   options: [],
 };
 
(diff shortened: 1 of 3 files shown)


packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -324,11 +324,10 @@ export type TSelectInputProps = {
 
 const defaultProps: Pick<
   TSelectInputProps,
-  'maxMenuHeight' | 'menuPortalZIndex' | 'controlShouldRenderValue' | 'options'
+  'maxMenuHeight' | 'menuPortalZIndex' | 'options'
 > = {
   maxMenuHeight: 220,
   menuPortalZIndex: 1,
-  controlShouldRenderValue: true,
```
kark, line 331: I guess technically this is a breaking change, but this was added only to fix the ui bug.

emmenko (approved):
Thanks

CarlosCortizasCT (approved):
Thanks! 🙌

PR #2507, opened 2023-04-28
fix: extend TDataTableProps type

ragafus:
#### Summary
This PR extends `TDataTableProps` type by using the generic row type in its `columns` property

## Description
While typing column definitions in one of Discounts web app features, the following error appeared. The type of the `columns` prop of the `DataTable` component can be improved.
Screenshot 2023-04-28 at 12 59 47


emmenko (approved):
All good, thanks for the contribution! ✅

PR #2515, opened 2023-05-08
Update docs yarn version

CarlosCortizasCT:
#### Summary

Update docs yarn version

## Description

I was experimenting this weird errors when trying to run the docs locally:
Screenshot 2023-05-08 at 17 18 20

According to what I found, it might be due to an outdated version of `yarn` so I updated the one we have in the docs folder (`3.1.0`) to be the same of the root folder (`3.5.0`) and it works like charm.

I also tried removing the docs yarn version altogether and also the `yarnPath` from the configuration file (docs folder `.yarnrc.yml`) and it also works but since I don't why it was there in the first place, I opted for just updating the file and config.


emmenko (approved):
Thanks

emmenko:
> I also tried removing the docs yarn version altogether and also the yarnPath from the configuration file (docs folder .yarnrc.yml) and it also works but since I don't why it was there in the first place, I opted for just updating the file and config.

The `yarnPath` explicitly instructs yarn to use that binary, thus using the version that we have bundled in the repo.
If not, it will use of course the version of yarn that you have installed in your machine but that might be a different version.

So the `yarnPath` is necessary to ensure that everyone working on this repo uses the same yarn version, regardless of the version of yarn installed in our machines.

emmenko:
Worth a try yes

Rhotimee (approved):

emmenko (approved):

CarlosCortizasCT:
> Worth a try yes

It works as well so I updated it here: a7ec7daacb94cae067065cbb226f0cc24f8b5239

PR #2516, opened 2023-05-10
Remove theme selector from Storybook

CarlosCortizasCT:
#### Summary

Remove theme selector from Storybook.

## Description

Now that we released the new theme in the Merchant Center, we can start cleaning up the previous theme.
As a first quick step, we're changing Storybook to only show the new theme so developers can see it by default.


emmenko (approved):
💯

PR #2521, opened 2023-05-15
Old look and field removal

CarlosCortizasCT:
#### Summary

Old look and field removal

## Description

During the last months we've been working on implementing a new look and feel for `ui-kit` components. The process involved including some code to support both the old and the new styles during the transition.

Now that the new look and feel has finally been released, we are removing all that support code and keep just the new styles.


design-system/materials/internals/definition.yaml
```diff
@@ -99,7 +99,7 @@ choiceGroupsByTheme:
       label: Font families
       prefix: font-family
       choices:
-        font-family: "'Open Sans', sans-serif"
+        font-family: "'Inter', system-ui"
```
CarlosCortizasCT, line 102: The changes in this file consist in overwriting the default theme values with the ones we had in the test one and remove the latter so we just keep the `default` one with the new styles values.

We introduced many new design tokens for the transition from one look and feel to the other. We're keeping those ones by now since we're not sure how much granularity we want to have with the tokens. That would be an upcoming project.

design-system/src/theme-provider.tsx
```diff
@@ -112,44 +109,18 @@ type TUseThemeResult = {
...
-  }, []);
-
   return {
-    theme,
+    theme: 'default',
+    // @deprecated
     themedValue,
-    isNewTheme: theme === 'test',
+    // @deprecated
+    isNewTheme: true,
```
CarlosCortizasCT, line 123: We will be returning a static theme value since we now have just one and the other two properties will be deprecated.
Another option is to directly removing: since we were only using them internally, I think it would be save, but I want to know your thoughts about it. It would mean that teams relying on them will be forced to update their code to be able to get any new `ui-kit` versions. That's why I think deprecating them first would be helpful.

CarlosCortizasCT:
Not really sure about the VRTs conflicts but new screenshots look good to me.

emmenko (approved):
🚀 🙌 Awesome!

packages/components/collapsible-panel/src/collapsible-panel.styles.tsx
```diff
@@ -19,14 +19,13 @@ const getHeaderContainerStyles = (
...
+  >,
   isOpen: boolean
 ) => {
   const baseStyles = css`
     background-color: ${designTokens.colorSurface};
-    border-bottom: ${isOpen || props.isNewTheme
-      ? '1px solid ' + designTokens.borderColorForCollapsiblePanelHeader
-      : 'none'};
+    border-bottom: ${'1px solid ' +
+    designTokens.borderColorForCollapsiblePanelHeader};
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 70b52798bf8fed53eacc136a1d4ad54d154b0008

ddouglasz (approved):
Looks Great 🦾🦾. Thank you! 
Only some few observations: 
`
  packages/components/data-table/src/header-cell.styles.tsx 
  packages/components/inputs/select-utils/src/create-select-styles.ts
  packages/components/pagination/src/page-navigator/page-navigator.tsx 
` are still having the comments `// TODO: @redesign cleanup` when it seems like you have handled them already.

kark (approved):
Thanks 🙌

design-system/src/theme-provider.tsx
CarlosCortizasCT: Updated in this commit: a20220d2b0974e85a814750710b1608f9b57f855

Thanks!

Issue #2522, opened 2023-05-15
LocalizedRichTextInput adds <p> tag to entered value


kterry1:
**Describe the bug**
The `LocalizedRichTextInput` component adds a `<p>` tag to the value entered in the input.

**To Reproduce**
1. Go to https://uikit.commercetools.com/?path=/story/components-inputs--localizedrichtextinput
2. Click on the input
3. Open up the console and check the value added in data.args[0].target.value
4. See that the value entered has a `<p>` tag around it

**Expected behavior**
The value to just be what is entered in the input without the `<p>` tag. 

**Screenshots**
<img src="https://p66.tr3.n0.cdn.getcloudapp.com/items/Jrup0LvZ/32f73faf-717a-4f2c-864a-ce0604c0e007.jpg?source=viewer&v=cd19d63d102d602291ce0cddbcbdef77"/>

**Additional context**
It's causing the Audit Log to show `<p>` when a user makes changes to the Product description under `General Info` as the `LocalizedRichTextInput` component is used there.

jaikumar-tj:
Plus one on this issue. We face the same problem in products. Because of this, we couldn't go to the public beta with the rich text editor feature on product detail page. It would be super helpful if the default text didn't have a `<p>` tag enclosing it.

emmenko:
Hi, thanks for the feedback.

From my understanding this is actually the normal behavior for richt-text inputs, to include the `<p>` tag, as it's supposed to emit HTML.

@kark Do you know if it's possible to omit it?

kark:
Thanks, 
I'd have refresh my memory and look it up. Off the top of my head, that was required.
I remember we had a regression that was caused by not wrapping text nodes with `<p>` in the initial implementation
https://github.com/commercetools/ui-kit/pull/2236/files

emmenko:
@kterry1 @jaikumar-tj @kark Let's try keeping the discussion in the issue instead of the PR.

As [mentioned here](https://github.com/commercetools/ui-kit/pull/2523#discussion_r1195013675) I believe a richt-text input always emits HTML and always wraps the content into an element by default. For instance, typing just some text is considered a paragraph thus wrapping it into a `<p>` is in theory correct.

What we would like to understand is what "problems" exactly are caused by this behavior.

Is it just an expectation problem from your side or is this causing like bugs or more technical issues?

jaikumar-tj:
@emmenko Sure 👍. 
[...]

PR #2523, opened 2023-05-16
fix(RichTextInput): deserialization of `<sup>` and `<sub>`

kark:
#### Summary

This PR is meant to fix:
- `<sup>` and `<sub>` deserialization

- issue with rich text state resetting
- removes unnecessary wrapping every text node with paragraph in the deserialization function
- adds a more complex story to enable manual tests of `<RichTextInput>` and `LocalizedRichTextInput>` 

⚠️ ⚠️ ⚠️ it does not cover #2522 ⚠️ ⚠️ ⚠️

### DIFF
--- a/packages/components/inputs/rich-text-utils/src/html/html.spec.js
+++ b/packages/components/inputs/rich-text-utils/src/html/html.spec.js
@@ -106,9 +106,19 @@ describe('html', () => {
             '<ol><li><span style="font-weight: bold; font-family: &quot;Comic Sans MS&quot;;">Computermouse for <span style="text-decoration-line: underline;">controlling</span></span></li></ol><table class="table table-bordered"><tbody><tr><td>hello</td></tr><tr><td><p>world<img src="https://www.rollingstone.com/wp-content/uploads/2019/01/shutterstock_10010937aj.jpg" style="width: 100%; float: right;" class="pull-right img-circle"></p></td></tr></tbody></table><ol><li><span style="font-weight: bold; font-family: &quot;Comic Sans MS&quot;;"><span style="text-decoration-line: underline;"><br></span></span></li></ol>';
 
           expect(html.serialize(html.deserialize(htmlValue))).toEqual(
-            '<ol><li><strong>Computermouse for </strong><u><strong>controlling</strong></u></li></ol><p>hello</p><p>world</p><ol><li><u><strong></strong></u></li></ol>'
+            '<ol><li><strong>Computermouse for </strong><u><strong>controlling</strong></u></li></ol>hello<p>world</p><ol><li><u><strong></strong></u></li></ol>'
           );
         });
+      });
+    });
+    describe('multiple nested text nodes within parent text node', () => {
+      it('should properly serialize', () => {
+        const htmlValue =
+          '<p><del>first</del><del><sup>second</sup></del><del><sub><sup>third</sup></sub></del></p>';
+
+        expect(html.serialize(html.deserialize(htmlValue))).toEqual(
+          '<p><del>first</del><del><sup>second</sup></del><del><sub><sup>third</sup></sub></del></p>'
+        );
       });
     });
   });
--- a/packages/components/inputs/rich-text-utils/src/html/html.tsx
+++ b/packages/components/inputs/rich-text-utils/src/html/html.tsx
@@ -149,6 +149,8 @@ const TEXT_TAGS = {
   EM: () => ({ italic: true }),
   I: () => ({ italic: true }),
   S: () => ({ strikethrough: true }),
+  SUP: () => ({ superscript: true }),
+  SUB: () => ({ subscript: true }),
   STRONG: () => ({ bold: true }),
   U: () => ({ underline: true }),
 };
@@ -279,10 +281,7 @@ const deserializeElement = (
     return children.map((child) => jsx('text', attrs, child));
   }
 
-  // each non-empty text node must be wrapped with a paragraph
-  return children.map((child) =>
-    Text.isText(child) && child.text ? wrapWithParagraph(child) : child
-  );
+  return children;
 };
 const deserialize = (html: Html) => {
   const document = new DOMParser().parseFromString(
--- a/packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
+++ b/packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
@@ -190,14 +190,24 @@ const toggleBlock = (editor: TEditor, format: Format) => {
   }
 };
 
+function nonNullable<T>(value: T): value is NonNullable<T> {
+  return value !== null && value !== undefined;
+}
+
 const validSlateStateAdapter = (
   value: Deserialized | Deserialized[]
 ): Descendant[] => {
   if (SlateElement.isElement(value) || Text.isText(value)) {
     return [value];
   }
-  if (SlateElement.isElementList(value) || Text.isTextList(value)) {
-    return value;
+  if (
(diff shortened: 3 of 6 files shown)


packages/components/inputs/rich-text-utils/src/html/html.tsx
```diff
@@ -179,12 +179,16 @@ const wrapWithParagraph = (
...
+  textContent: TElement | TText | (TElement | TText)[]
+) => jsx('fragment', {}, textContent);
+
+const wrapWithFragmentIfRootElement = (
   el: HTMLElement | ChildNode,
   textContent: TElement | TText | (TElement | TText)[]
 ) =>
   el.parentNode?.nodeName === 'BODY' // root element, because body is eventually turned to React fragment
-    ? wrapWithParagraph(textContent)
+    ? wrapWithFragment(textContent)
```
jaikumar-tj, line 191: Would this be a problem in case of new line? 

In the below screenshot, there are two individual elements and above condition would be true since the parentNode is 'BODY' in this case, isn't it?

This would make the text to be in the single line without any space between them if the `<p>` tag is removed. 

Screenshot 2023-05-16 at 10 47 04

kark: Thanks,
I'm not sure I 100% understand the question but let me try to answer.
You're right that in this case we'll have:
```
Body                                          
  |                                            
  |                                            
  |----P                                       
  |    |                                       
  |    |                                       
  |    |                                       
  |    +------ #text (this is first line)                        
  |                                            
  |                                            
  +-----P                                      
        |                                      
        |                                      
        +------#text (this is second line)                           
```
This function only replaces `<body>` with `<></>`.
Therefore, if you provide:
```tsx
<LocalizedRichTextInput
        onChange={() => {}}
        value={{
          en: '<p>this is the first line</p><p>this is the second line</p>',
        }}
        selectedLanguage="en"
        horizontalConstraint={7}
        defaultExpandMultilineText={true}
      />
```
the wrapping paragraphs are there.
I don't get the part how do you expect the `<p>`s to be removed? They will always be there if this content is typed in the input by the user.

jaikumar-tj: Thanks for the detailed explanation. I understand what you mean. 
However I am confused since you linked this [issue](https://github.com/commercetools/ui-kit/issues/2522) which talks about removing `<p>` tags from unformatted text content. So does this PR removes `<p>` only if there is only one `<p>` tag at the top level?  

This was my overall assumption. 
If users type anything in rich text editor without formatting (i.e, not clicking bold or italic, etc) then text will not be wrapped by any elements like `<p>`. 
If the users type anything and format their content, then respective HTML elements will be wrapped.
Is my understanding or expectations correct or is this something not possible to achieve it?

kark: Thanks again,
> If users type anything in rich text editor without formatting (i.e, not clicking bold or italic, etc) then text will not be wrapped by any elements like <p>.

Got it. Not sure, I need to check that.

emmenko: > If users type anything in rich text editor without formatting (i.e, not clicking bold or italic, etc) then text will not be wrapped by any elements like `<p>`.
> If the users type anything and format their content, then respective HTML elements will be wrapped.
> Is my understanding or expectations correct or is this something not possible to achieve it?

To my understanding, a richt-text input always emits HTML and always wraps the content into an element by default. For instance, typing just some text is considered a paragraph thus wrapping it into a `<p>` is in theory correct.

CarlosCortizasCT: I’m not sure about this one, but since the component is formatting text, I believe it needs to output some kind of rich information where we don’t only get the content but also its structure and styles. I mean, we need to know if a text is part of a list, if it’s a header, etc.

For what I can see in other libraries, the output uses to be HTML (which includes content, structure and styles) or JSON where all of those can be represented ([example](https://editorjs.io/)).

If the first element in the output is a list, I guess we wouldn’t be questioning if we want it to be wrapped in a `ul/ol`.

I’m not sure I understand the underlying problem.

packages/components/inputs/rich-text-utils/src/html/html.tsx
```diff
@@ -279,10 +281,7 @@ const deserializeElement = (
     return children.map((child) => jsx('text', attrs, child));
   }
 
-  // each non-empty text node must be wrapped with a paragraph
-  return children.map((child) =>
-    Text.isText(child) && child.text ? wrapWithParagraph(child) : child
-  );
```
kark, line 285: This is not needed, it was added as a hot fix of a bug, but text nodes can in fact be added without a wrapping paragraph

packages/components/inputs/rich-text-utils/src/slate-helpers.tsx
```diff
@@ -190,31 +190,49 @@ const toggleBlock = (editor: TEditor, format: Format) => {
...
     return [value];
   }
-  if (SlateElement.isElementList(value) || Text.isTextList(value)) {
-    return value;
+  if (
+    SlateElement.isElementList(value) ||
+    Text.isTextList(value) ||
+    // in case of an array of mixed text and element nodes
+    (Array.isArray(value) &&
+      value.every((node) => SlateElement.isElement(node) || Text.isText(node)))
```
kark, line 208: This is a valid condition now, since text nodes don't need a wrapping paragraph

packages/components/inputs/rich-text-utils/src/html/html.tsx
```diff
@@ -149,6 +149,8 @@ const TEXT_TAGS = {
   EM: () => ({ italic: true }),
   I: () => ({ italic: true }),
   S: () => ({ strikethrough: true }),
+  SUP: () => ({ superscript: true }),
+  SUB: () => ({ subscript: true }),
```
kark, line 153: This is actually the missing piece that is a root cause of the bug

packages/components/inputs/rich-text-utils/src/html/html.spec.js
```diff
@@ -106,10 +106,20 @@ describe('html', () => {
             '<ol><li><span style="font-weight: bold; font-family: &quot;Comic Sans MS&quot;;">Computermouse for <span style="text-decoration-line: underline;">controlling</span></span></li></ol><table class="table table-bordered"><tbody><tr><td>hello</td></tr><tr><td><p>world<img src="https://www.rollingstone.com/wp-content/uploads/2019/01/shutterstock_10010937aj.jpg" style="width: 100%; float: right;" class="pull-right img-circle"></p></td></tr></tbody></table><ol><li><span style="font-weight: bold; font-family: &quot;Comic Sans MS&quot;;"><span style="text-decoration-line: underline;"><br></span></span></li></ol>';
 
           expect(html.serialize(html.deserialize(htmlValue))).toEqual(
-            '<ol><li><strong>Computermouse for </strong><u><strong>controlling</strong></u></li></ol><p>hello</p><p>world</p><ol><li><u><strong></strong></u></li></ol>'
+            '<ol><li><strong>Computermouse for </strong><u><strong>controlling</strong></u></li></ol>hello<p>world</p><ol><li><u><strong></strong></u></li></ol>'
           );
         });
       });
     });
+    describe('multiple nested text nodes within parent text node', () => {
```
CarlosCortizasCT, line 114: ❤️

CarlosCortizasCT (approved):
Awesome work! 💪

chloe0592 (approved):
Great work! 😎 👏

PR #2524, opened 2023-05-18
Icons accessibility

chloe0592:
The icons we export from the ui-kit package are not currently accessible since they don't provide a text based alternative for users using assistive technologies (eg: screen readers) to have any context about them.
These icons are exported as React components which render a `svg` element with no accessible context (as, for example img elements would use with the `alt`.)

Changes made in this PR mean that icons will from now on have generated accessible information when rendered (`title` and `titleId`).


packages/components/inputs/checkbox-input/src/checkbox-input.spec.tsx
chloe0592, line 1: The `exact` was added here because tests were failing. When set to `false` it matches substrings and it's not case-sensitive.

One potential concern here is that this change might mean that some tests (like this one) might be failing in other repos as well, therefore it might be a breaking change. So we need to have that in mind if we decide to go with it.

Wanted to know what the team thinks about it? Or maybe if there are some other ideas? 🤓 
@commercetools/shield-team-fe

CarlosCortizasCT: Just to maybe add some more context.
This component renders a label which contains some text (`Accept Terms` in these tests) and also an icon representing the actual checkbox input (which is accessible hidden).
Now that the icons have a corresponding accessible title, this one is also part of the label content.

In the case a test with the checkbox checked, when calling `getByLabelText('Accept Terms')` we're telling the testing library we want to find a label with that exact string, but now the label we're testing have this content: `AcceptTermsChecked icon` (=> `<children><icon_label>`).
That's why we need to now search for a partial match: this can be done using the `exact` optional param or changing the test to be a regular expression: `expect(screen.getByLabelText(/Accept Terms/)).toBeInTheDocument();`.

Hope this helps to better understand the problem.

kark (commented):
Thank you for working on this! That's a great improvement and nicely figured out 🤓! 
I'm leaving some minor remarks

packages/components/inputs/checkbox-input/src/checkbox-input.spec.tsx
kark: Thanks for the explanation 👍 Perhaps making it consistent and use **either** `exact: false` **or** regex in all of the tests would help to not cause any confusion in the future (now it's a mix of both)

packages/components/icons/src/generated/AngleDownReact.tsx
```diff
@@ -1,6 +1,8 @@
 // THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 // This file is created with 'yarn generate-icons'.
 // Original SVG file: src/svg/angle-down.react.svg
+import { useMemo } from 'react';
+import { createSequentialId } from '@commercetools-uikit/utils';
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import { warning } from '@commercetools-uikit/utils';
```
emmenko, line 7: Nit: It looks like we're importing from the same package twice.

emmenko (approved):
Super nice, thanks a lot! 🥳

packages/components/inputs/checkbox-input/src/checkbox-input.spec.tsx
chloe0592: Yes thank you, you're right. I replaced it with regex here - [cc91671](https://github.com/commercetools/ui-kit/pull/2524/commits/cc9167151e00a0ee777d5059f74b2a998096cc62)

svgr.config.js
```diff
@@ -30,14 +30,31 @@ const annotationsToRemove = [
...
+    .join('');
+};
+const getSvgFileName = (filePathname) =>
+  filePathname.match(/\/svg\/(.*)\.react\.svg/)[1];
+
 // https://react-svgr.com/docs/custom-templates/
 // https://github.com/gregberge/svgr/blob/c57ee04b19c15a76ae4caf40d1bb82c210d6c398/packages/babel-plugin-transform-svg-component/src/types.ts#L18
 const styledIconsTemplate = (variables, context) => {
+  console.log('Variables:', JSON.stringify(variables, null, 4));
+  console.log('Context:', JSON.stringify(context, null, 4));
```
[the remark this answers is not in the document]
chloe0592: Yes, good catch! 🦅 [8d72088](https://github.com/commercetools/ui-kit/pull/2524/commits/8d72088ddc79425a18cec795b798b60a20b045f0)

svgr.config.js
```diff
@@ -79,6 +106,8 @@ export default ${reactComponentName};
 
 module.exports = {
   icon: false,
+  expandProps: 'end',
```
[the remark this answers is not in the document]
chloe0592: I suppose that's true. I haven't seen any effects either so it should be safe to remove it. [8d72088](https://github.com/commercetools/ui-kit/pull/2524/commits/8d72088ddc79425a18cec795b798b60a20b045f0)
Unless @CarlosCortizasCT thinks otherwise?

packages/components/inputs/input-utils/src/accessible-hidden-input.styles.ts
```diff
@@ -1,13 +1,9 @@
...
 const accessibleHiddenInputStyles = css`
-  pointer-events: none;
-  height: 100%;
-  left: 0;
-  opacity: 0.0001;
-  position: absolute;
+  left: -9999px;
+  position: fixed;
   top: 0;
-  width: 100%;
```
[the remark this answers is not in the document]
chloe0592: Ah, forgot to mention this. It's not related to this PR, but when we were looking into why tests were failing, these stylings were looking weird. Then the OCD kicked in. 😄 
Is it better if I removed this since it's not related?

ddouglasz (approved):
Wow, very cool. Thank you 🙌🏾

Rhotimee (approved):

packages/components/icons/src/generated/AngleDownReact.tsx
chloe0592: @emmenko I moved the imports to `icons.styles.tsx` and that seems to fix it. You can see it here - [3846b48](https://github.com/commercetools/ui-kit/pull/2524/commits/3846b489475daa819e0b5075233aad38ac75cc7d).
Let me know if it all looks good now 😄

emmenko (approved):
Thanks

svgr.config.js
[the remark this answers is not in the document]
CarlosCortizasCT: Maybe this was something I added when I was playing with the config at the very beginning.

I think we don't actually needed now ([reference](https://commercetools.atlassian.net/browse/SHIELD-782)).

CarlosCortizasCT (approved):
Many thanks! 💯

PR #2525, opened 2023-05-19
chore: use cac instead of mri in all ui-kit CLIs

YahiaElTai:
## Summary

This PR relates to:
- https://github.com/commercetools/fe-chapter-notes/issues/403

## Description
This PR migrates all ui-kit CLIs to use `cac` instead of `mri` for parsing of CLI flags

#### Affected CLIs:
- generate-readme
- generate-package-json
- check-ts-migration
- version

Note: After running `generate-readme`, there was some changes but I think they are only styling changes

image

image

image


generators/package-json/package.json
```diff
@@ -14,9 +14,8 @@
   "dependencies": {
     "@manypkg/find-root": "1.1.0",
     "@manypkg/get-packages": "1.1.3",
-    "@types/mri": "1.1.1",
     "@types/node": "16.18.31",
-    "mri": "1.2.0",
+    "cac": "^6.7.14",
```
CarlosCortizasCT, line 18: I was wondering if we should use a fixed version as we do with the rest of the dependencies 🤔

generators/readme/package.json
```diff
@@ -16,15 +16,14 @@
     "@types/listr": "0.14.4",
     "@types/lodash": "4.14.194",
     "@types/mdast": "3.0.11",
-    "@types/mri": "1.1.1",
     "@types/node": "16.18.31",
     "@types/shelljs": "0.8.12",
     "@types/unist": "2.0.6",
     "@types/vfile": "4.0.0",
+    "cac": "^6.7.14",
```
CarlosCortizasCT, line 23: I was wondering if we should use a fixed version as we do with the rest of the dependencies 🤔

package.json
```diff
@@ -96,6 +96,7 @@
     "babel-plugin-typescript-to-proptypes": "1.4.2",
     "browserslist": "4.21.5",
     "bundlesize": "0.18.1",
+    "cac": "^6.7.14",
```
CarlosCortizasCT, line 99: I was wondering if we should use a fixed version as we do with the rest of the dependencies 🤔

packages/components/data-table/README.md
```diff
@@ -59,25 +59,25 @@ export default Example;
...
+| `columns`                 | `Array: TColumn<Row>[]`<br/>[See signature.](#signature-columns) |          | `[]`                               | Each object requires a unique `key` which should correspond to property key of&#xA;the items of `rows` that you want to render under this column, and a `label`&#xA;which defines the name shown on the header.&#xA;The list of columns to be rendered.&#xA;Each column can be customized (see properties below).                                                                                                                                                                                                                                 |
```
CarlosCortizasCT, line 65: 🚀

generators/package-json/package.json
YahiaElTai: Ah sure, I can change that. Thanks 🙏

scripts/check-ts-migration.mjs
emmenko, line 1: I guess this file can also be removed 😅

We'll do that as a follow up

emmenko (approved):
🥳

emmenko (approved):
🥳

PR #2532, opened 2023-05-31
`DataTable` UI improvements

chloe0592:
Applying smaller spacings for the `DataTable` and setting `isCondensed` to `true` as default.


FilPob (approved):

CarlosCortizasCT (commented):
Is this a UI breaking change? Is seems all `DataTable` consumers not using a `condensed` property will have the look of their tables changed.
Are we ok with that? @FilPob 

@chloe0592 Please remember to add changesets 🙏

FilPob:
@CarlosCortizasCT  yes I can confirm that it's expected that all tables in MC that are not specifically set to the `default` row density, should be changed to `condensed` ...If that answers your question :D

CarlosCortizasCT (approved):
Many thanks! 🙌

kark (approved):
😎
Just one nit-picky remark after inspecting the vrts - there are a couple of tests labeled `condensed mode` (but all are condensed now really). Perhaps we could make them test the not condensed version instead.  Up to you.

FilPob:
I will mention this better late than never :D But there was also the wish from multiple sides for renaming the density `Default` to something like `Comfortable` The prop name and also the label in the UI. I think that definitely makes sense, but doesn't need to be part of this PR. What do you think @chloe0592  @CarlosCortizasCT ?

chloe0592:
> 😎 Just one nit-picky remark after inspecting the vrts - there are a couple of tests labeled `condensed mode` (but all are condensed now really). Perhaps we could make them test the not condensed version instead. Up to you.

@kark that's a good point, thanks! I'll update the tests. 🤓

chloe0592:
> I will mention this better late than never :D But there was also the wish from multiple sides for renaming the density `Default` to something like `Comfortable` The prop name and also the label in the UI. I think that definitely makes sense, but doesn't need to be part of this PR. What do you think @chloe0592 @CarlosCortizasCT ?

@FilPob I think it can be easily done. 
@CarlosCortizasCT if I understood correctly I just update the values in `.json` files in `packages/i18n/data/`, right?
As for the translations, do we ourselves use the translators for each of the language or someone else needs to check them?

CarlosCortizasCT:
> @FilPob I think it can be fairly easily done. @CarlosCortizasCT if I understood correctly I just update the values in .json files in packages/i18n/data/, right? As for the translations, do we ourselves use the translators for each of the language or someone else needs to check them?

What I would do is to update the default message [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/data-table-manager/src/display-settings-manager/messages.ts#L32) and then run yarn extract-intl but I think we would prefer to create a new key in the messages file, run the command and then we would need to wait for the translations. We would keep using the "old" text until we get all the translations.
But I'd like to ask [@emmenko](https://github.com/emmenko) for confirmation here as how to proceed.

ddouglasz (approved):
💯

chloe0592:
> > @FilPob I think it can be fairly easily done. @CarlosCortizasCT if I understood correctly I just update the values in .json files in packages/i18n/data/, right? As for the translations, do we ourselves use the translators for each of the language or someone else needs to check them?
> 
> What I would do is to update the default message [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/data-table-manager/src/display-settings-manager/messages.ts#L32) and then run yarn extract-intl but I think we would prefer to create a new key in the messages file, run the command and then we would need to wait for the translations. We would keep using the "old" text until we get all the translations. But I'd like to ask [@emmenko](https://github.com/emmenko) for confirmation here as how to proceed.

@emmenko did you maybe had a change to look at this? 🙏

packages/i18n/data/es.json
```diff
@@ -15,7 +15,7 @@
   "UIKit.DataTableManager.ColumnSelectorGroups.visible": "Columnas visibles",
   "UIKit.DataTableManager.DensityManager.closeButtonLabel": "Cerrar",
   "UIKit.DataTableManager.DensityManager.densityDisplayCompactOption": "Compacta",
-  "UIKit.DataTableManager.DensityManager.densityDisplayDefaultOption": "Predeterminada",
+  "UIKit.DataTableManager.DensityManager.densityDisplayComfortableOption": "Confortable",
```
emmenko, line 18: What's the spanish translation for this? @CarlosCortizasCT

CarlosCortizasCT: That's the one: "Confortable"

chloe0592: Cool, then we are all set? Could I merge this PR?

Rhotimee (approved):

packages/i18n/data/es.json
CarlosCortizasCT: I think we can, yes.

Many thanks!

PR #2533, opened 2023-06-02
fix(Tag): make `TagBody` render as `span` rather than `p`

kark:
#### Summary

This PR fixes an issue with rendering `<div>`s as children of `<Tag>` component


CarlosCortizasCT (approved):
Many thanks! 🙇

emmenko (approved):
💯

PR #2537, opened 2023-06-09
Add `CheckbBoxField` component to uikit

ddouglasz:
#### Summary

Add CheckboxField component to the UIkit

## Description

CheckboxInput and Field labels exist in the UIkit, but a combination of both is not available as it is for any other field in the UIkit. When consumer need to build forms with such fields, they need to use those two components separately and that lead to inconsistencies in both code and UI in our platform.


chloe0592 (approved):
Great work! Thank you 👏 
I'll just leave few minor comments

packages/components/fields/checkbox-field/src/checkbox-field.story.js
```diff
@@ -0,0 +1,168 @@
...
+                    hint={text('hint', 'Select an option', checkBoxField)}
+                    description={text('description', '', checkBoxField)}
+                    onInfoButtonClick={
+                      boolean('show info button', false, checkBoxField)
+                        ? action('onInfoButtonClick')
+                        : undefined
+                    }
+                    isRequired={boolean('isRequired', false, checkBoxField)}
+                    hintIcon={hintIcon}
+                    badge={text('badge', '', checkBoxField)}
```
chloe0592, line 157: For the `badge` prop, maybe we could instead of _text_ add a _boolean_ value, what do you think?
```
badge={
   boolean('show badge', false, checkBoxField) ? (
       <FlatButton
            tone="primary"
             icon={<icons.BoxIcon />}
             label="label"
             onClick={() => {}}
      />
    ) : undefined
 }
```

packages/components/fields/checkbox-field/docs/usage-example.js
```diff
@@ -0,0 +1,27 @@
+import CheckBoxField from '@commercetools-uikit/checkbox-field';
+
+const Example = () => (
+  <div>
+    <CheckBoxField
+      value="foo-checkbox-value"
+      onChange={(event) => alert(event.target.value)}
+      isChecked={true}
```
chloe0592, line 8: Nit: might be nice to have some examples of the `FieldLabel` props.
Maybe just adding something like: 
```
    title="Welcome Text"
    hintIcon={<WarningIcon />}
    description={<FormattedMessage {...messages.description} />}
    badge={<FlatButton tone="primary" label="label" />}
```

CarlosCortizasCT:
nit: I think that, instead of having a matrix of different versions plus the knobs for further tailor the rendering, we could have just one render component and use the knobs to control how it renders as we do in the other fields ([example](https://ui-3syih84xs-commercetools.vercel.app/?path=/story/components-fields--datefield)).

CarlosCortizasCT:
Do we have a Figma design for this new field we can use to check the expected result?

packages/components/fields/checkbox-field/src/checkbox-field.story.js
ddouglasz: Oh, this makes sense. Nice catch! 🙌🏾 
Thank you.

ddouglasz:
> Do we have a Figma design for this new field we can use to check the expected result?

No design. I pretty much went with the approach used for the [radio field](https://github.com/commercetools/ui-kit/blob/4c2501b92dc41777b45bb5c99b4f15cc4e08f224/packages/components/fields/radio-field/src/radio-field.tsx)

ddouglasz:
All feedback taken care of here: https://github.com/commercetools/ui-kit/pull/2537/commits/fa0d26409ab9837b3173e68e7c30a1c5f64b6d82
Thank you 🙌🏾

packages/components/fields/checkbox-field/src/checkbox-field.spec.js
```diff
@@ -0,0 +1,165 @@
...
+it('should have an HTML name', () => {
+  const { container } = renderCheckBoxField({ name: 'foo' });
+  expect(container.querySelector('[name="foo"]')).toBeInTheDocument();
+});
+
+it('should call onChange when changing the value', () => {
+  const { getByLabelText, onChange } = renderCheckBoxField({ isChecked: true });
+  fireEvent.click(getByLabelText(/CheckBoxField/));
+  expect(onChange).toHaveBeenCalled();
+});
```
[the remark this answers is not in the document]
ddouglasz: handled here: https://github.com/commercetools/ui-kit/pull/2537/commits/fa0d26409ab9837b3173e68e7c30a1c5f64b6d82
Thank you

packages/components/fields/checkbox-field/src/checkbox-field.story.js
ddouglasz: handled here: https://github.com/commercetools/ui-kit/pull/2537/commits/fa0d26409ab9837b3173e68e7c30a1c5f64b6d82
Thank you

ddouglasz:
> Since we have no design to compare with, I just mention here some things that are weird to me and maybe @FilPob can verify:
> 
> image
> 
> `hint` and `description` look very similar (although they have different font size and weight). Is this intended? `badge` seems out of place to me. Is it supposed to look like that?

So I think we can wait for filip to guide on the hint and description similarity. Maybe a suggestion from him to differentiate them.
As regards the badge, I considered how it appears in the field-label at the moment. since we may be merging attributes from both components.

CarlosCortizasCT:
> So I think we can wait for filip to guide on the hint and description similarity. Maybe a suggestion from him to differentiate them. As regards the badge, I considered how it appears in the field-label at the moment. since we may be merging attributes from both components.

Yes, I understand, but I'd like to know what's the use case for having the badge in such a displaced position.
Maybe Filip can also help here.

ddouglasz:
> Also, within the Knobs tab there are another two tabs (Fruits and Other) and both render the same content. What's the difference between them? 🤔

Updated here: https://github.com/commercetools/ui-kit/pull/2537/commits/075810a5c9937ce9745c5ba99bb3cb96ed179cd9

Actually, what I was thinking is not valid 😅

packages/components/fields/checkbox-field/src/checkbox-field.story.js
```diff
@@ -0,0 +1,90 @@
...
+                switch (key) {
+                  case 'customError':
+                    return 'A custom error.';
+                  default:
+                    return null;
+                }
+              }}
+              isRequired={boolean('isRequired', false, '')}
+              touched={boolean('touched', false, '')}
+              hintIcon={hintIcon}
```
kark, line 72: I'm not sure, but for some reason selecting a hint icon doesn't work in the preview (it does for other field components though)

packages/components/fields/checkbox-field/src/checkbox-field.tsx
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Makes the CheckboxInput readonly
+   */
+  isReadOnly?: boolean;
+  /**
+   * Indicates that the checkbox has an error
+   */
+  hasError?: boolean;
+  /**
+   * The descriptive text of the CheckboxInput, used as its label.
```
kark, line 96: I suppose we should be describing `CheckboxField` props (not `CheckboxInput`)

kark (commented):
Looks good! Thanks for working on this 🙌
+1 to the badge location, other than that just leaving some minor remarks

FilPob:
@ddouglasz @CarlosCortizasCT thanks for the good comments :)

Badge: I though this comes from the `Fieldlabel` component and therefore is just available inherently. But i agree it doesnt make too much sense in the `Radio` and `Checkbox` Field. We can disable it our in these two components as it looks quite displaced.

Hint: Should be styled the same way as in other Inputfields.

Generally I though following the same structure for the storybook page, the props as in the `Radiofield` component would be the best approach. Right now it is a bit different. What I miss in the  `Checkboxfield` the most would be:

- to see the 3 checkbox options instead of just 1  
- the possibility to switch between the direction of `Stack`, and `Inline` for the options.
- the `directionProps` property

ddouglasz:
> * directionProps

Thank you @FilPob
So for the Radio option, we have the `directionProps` because we can group radio buttons mostly and that prop already exists in the radio-group component. We do not have it normally on the checkbox or `FieldLabel` component. Just to double-check, do you mean I should implement it in the storybook and add the props as a new prop in the `FieldLabel` component ?

packages/components/fields/checkbox-field/src/checkbox-field.tsx
ddouglasz: Thank you for all the feedback. Updated here: https://github.com/commercetools/ui-kit/pull/2537/commits/0da451d5d93ec16a223a40ee9f6d27f940682a8b. 
There are still some design decisions follow-ups we are concluding on. So the PR is on wip for now until we conclude 🙌🏾

PR #2541, opened 2023-06-20
fix: table sticky header

chloe0592:
The issue which we are trying to solve with this PR is the `DataTable` header when we have many rows and when we're not setting up `maxHeight`. The sticky header was supposed to ease our way through the long table, but it didn't work. 
Two main issues:

1. `overflow` property of the table and its parent, blocks out the header's `position: sticky`. But we need the `overflow` prop when the height is being set so we can scroll though the table. In order to have the sticky header for cases with large tables we are making `overflow` prop conditional. 
From the changes in this PR - if the `max-height` is being defined the `overflow` is then unset.

2. some undesirable spacings are visible in some of the tables in Merchant Center when we have a sticky header. This is due to the spacings of the page content. Therefore, we will have a follow up PR on this one, with the changes in App Kit and MC to fix this.

### DIFF
--- a/packages/components/data-table/src/header-cell.styles.tsx
+++ b/packages/components/data-table/src/header-cell.styles.tsx
@@ -87,7 +87,6 @@ const HeaderCellInner = styled.div<THeaderCellInner>`
 type TBaseHeaderCell = {
   disableHeaderStickiness?: boolean;
   shouldClipContent?: boolean;
-  maxHeight?: string | number;
 };
 const BaseHeaderCell = styled.th<TBaseHeaderCell>`
   color: ${designTokens.fontColorForTableHeader};


emmenko:
Hey @chloe0592 , thanks for proposing a solution!

The changes here look good, I'm just not 100% sure about this custom `top` property.
As you know, the main issue is with the "overflow behind the header" due to the page `padding` 

image

However, if we change the page `padding` to `margin` we fix the issue so we don't need the `top` prop workaround.

image

I checked with @Shecki and he's fine with this change (in our App Kit page components).
This change also will fix another weird issue with tables as with `padding` the right-hand side of the screen has no "space".

Before

image

After

image

chloe0592:
> Hey @chloe0592 , thanks for proposing a solution!
> 
> The changes here look good, I'm just not 100% sure about this custom `top` property. As you know, the main issue is with the "overflow behind the header" due to the page `padding`
> 
> image
> However, if we change the page `padding` to `margin` we fix the issue so we don't need the `top` prop workaround.
> 
> image
> I checked with @Shecki and he's fine with this change (in our App Kit page components). This change also will fix another weird issue with tables as with `padding` the right-hand side of the screen has no "space".
> 
> Before
> 
> image
> After
> 
> image

Thank you @emmenko for this! 🙏 I tested it out myself in App Kit and I fully agree that this is much better solution. So I've updated the PR accordingly and have set up a quick meeting for the two of us just to make sure we are on the same page.

emmenko (approved):
🎉

kark (approved):
Thanks, looks good 🚀

Rhotimee (approved):
💯

tdeekens (approved):
💯

PR #2542, opened 2023-06-21
Align siblings of radio input options

ddouglasz:
Enable proper alignment of <RadioInput.Option> sibling components

#### Summary

n the current implementation of the <RadioInput.Option> there is no possibility of proper aligning of sibling components out of the box. This leads to misalignments like in the picture below:

image

### DIFF
--- a/.changeset/funny-melons-design.md
+++ b/.changeset/funny-melons-design.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/radio-input': minor
+---
+
+align radio options siblings
--- a/.changeset/funny-melons-design.md
+++ b/.changeset/funny-melons-design.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/radio-input': minor
 ---
 
-align radio options siblings
+Align additional content.
--- a/packages/components/inputs/radio-input/src/radio-option.styles.ts
+++ b/packages/components/inputs/radio-input/src/radio-option.styles.ts
@@ -14,7 +14,6 @@ const LabelTextWrapper = styled.div<TStylesProps>`
 
 const AdditionalTextWrapper = styled.div<TStylesProps>`
   grid-area: content;
-  margin-left: ${designTokens.spacing10};
   font-size: 1rem;
   font-family: inherit;
 `;
@@ -159,6 +158,7 @@ const RadioOptionLabel = styled.label<TStylesProps>`
       ? `:focus-within ${LabelTextWrapper} {
       outline: auto 2px ${designTokens.borderColorForInputWhenFocused};
       outline-offset: 2px;
+      width: fit-content;
     }`
       : ''}
 


kark (approved):
Thank you for the research and the proposal!
For me this works great 🙌
Jun-21-2023 13-42-32
And it will fix the focus state indicator and option additional content alignment e.g. here:
Screenshot 2023-06-21 at 13 48 00
After change:
Screenshot 2023-06-21 at 13 55 19

I'd only edit the changeset note to e.g.: `Align additional content`

FilPob (approved):

kark:
> Hey, sorry but I'm trying to understand the issue and the change.
> 
> I went to [staging](https://mc.europe-west1.gcp.escemo.com/test-project-with-sample-data/settings/project/misc) to see the error but I don't see it: image
> 
> I don't also see a way to compare the before/after in Storybook.
> 
> @ddouglasz Could you maybe share a little more context? 🙏

Actually, I can provide a bit more context (if I may).

The view that was brought up looks identical, but the implementation is different. Right now the additional content is rendered as a sibling to radio option with a wrapper with `radioOptionSibling` css class (with hardcoded `padding-left`) which is an ugly workaround (but we didn't have a way to do it differently yet follow the designs when this view was re-designed in the MC, so we decided to tackle it as a follow-up).

What @ddouglasz came up with in this PR is a `ui-kit` native solution (with `additionalContent` prop).

CarlosCortizasCT:
> > Actually, I can provide a bit more context (if I may).
> > The view that was brought up looks identical, but the implementation is different. Right now the additional content is rendered as a sibling to radio option with a wrapper with `radioOptionSibling` css class (with hardcoded `padding-left`) which is an ugly workaround (but we didn't have a way to do it differently yet follow the designs when this view was re-designed in the MC, so we decided to tackle it as a follow-up).
> > What @ddouglasz came up with in this PR is a `ui-kit` native solution (with `additionalContent` prop).
> 
> Thank you for the explanation @kark 💯 So when this is done, we will take out the hardcoded styling from MC-FE and use this directly. I created is a follow-up ticket for that too.

@kark, @ddouglasz  Thanks a lot for the explanations 🙇 

However, I still don't understand the need for this change given the example you provided.
In the "recommended categories" example, if we currently change the implementation to use the already available `aditionalContent` property it seems to work fine.

Would it be ok to have a sync to clarify this point? 🙏

kark:
> @kark, @ddouglasz Thanks a lot for the explanations 🙇
> 
> However, I still don't understand the need for this change given the example you provided. In the "recommended categories" example, if we currently change the implementation to use the already available `aditionalContent` property it seems to work fine.
> 
> Would it be ok to have a sync to clarify this point? 🙏

It works OK, but not in my opinion it's not 100% there.
The difference lies in the focus indicator and the alignment of the additional content (and these were the re-design requirements).
Screenshot 2023-06-22 at 09 41 55

But I'm up for the sync 👍

CarlosCortizasCT:
> > @kark, @ddouglasz Thanks a lot for the explanations 🙇
> > However, I still don't understand the need for this change given the example you provided. In the "recommended categories" example, if we currently change the implementation to use the already available `aditionalContent` property it seems to work fine.
> > Would it be ok to have a sync to clarify this point? 🙏
> 
> In my opinion it works OK, but it's not 100% there. The difference lies in the focus indicator and the alignment of the additional content (and these were the re-design requirements). Screenshot 2023-06-22 at 09 41 55 And these issues are solved in this PR. I'm up for the sync 👍

Thanks for the explanation @kark 

I think that misalignment could be adjusted by removing this [left margin](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/radio-input/src/radio-option.styles.ts#L17).

So when I look at the `fit-content` rule value [docs](https://devdocs.io/css/fit-content), I fail to see how it solves this alignment issue. To me it seems the root problem is that left margin.

CarlosCortizasCT (approved):
💯

PR #2544, opened 2023-06-26
ci: enable aggregated release notes

VictorAvelar:
#### Summary

Use dotansimha/changesets-action as it includes the aggregate feature.

## Description

- Added step to get the version number to publish the release notes
- Changed the changeset/action to use a fork with the aggregate feature
- Enabled the aggregate release notes feature.
- Included script to get the latest version from one of the packages (took it from app-kit)


CarlosCortizasCT:
Is there a way we can test this before merging so we check if it will work in this repo and how the release note would look like?

.github/workflows/release.yml
```diff
@@ -57,13 +57,24 @@ jobs:
         env:
           CI: true
 
+      - name: Storing release version for changeset
+        id: release_version
+        run: echo "VALUE=$(./scripts/print_release_version.sh)" >> $GITHUB_OUTPUT
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
```
CarlosCortizasCT, line 65: I was wondering what do we need this env for? 🤔

scripts/print_release_version.sh
emmenko, line 1: This file needs to be executable.

```
chmod 755 scripts/print_release_version.sh
```

.github/workflows/release.yml
emmenko: It's needed for changeset (we execute `changeset version` in the script) as far as I know.

emmenko (commented):
Thanks! 🙌

emmenko:
> Is there a way we can test this before merging so we check if it will work in this repo and how the release note would look like?

Not really, we need to trigger a normal release to really find out.

emmenko (approved):
🥳

CarlosCortizasCT (approved):
👍

.github/workflows/release.yml
```diff
@@ -57,13 +57,24 @@ jobs:
...
+
       - name: Create Release Pull Request or Publish to npm
         id: changesets
-        uses: changesets/action@v1.4.4
+        # uses: changesets/action@v1.4.4
```
kark, line 69: nit: would that still be needed?

kark (approved):
Thanks for the first contribution here and for researching the changesets action fork 🎉 🙌

.github/workflows/release.yml
VictorAvelar: If it works, then no. 😄

PR #2548, opened 2023-07-03
Fix for the heading issue of the `CollapsiblePanel'

chloe0592:
We've found an inconsistency in out `CollapsiblePanel` component manages the `header` prop.

The issue is that we are rendering an invalid HTML structure when passing a `<CollapsiblePanelHeader>` component as a `header` prop and when `condensed` prop is set as true in `<CollapsiblePanel>`. 
In that case it would wrap `<h2>` tag in the `<h4>` which would cause the error - `<h2> cannot appear as a child of <h4>`.

Screenshot 2023-07-03 at 14 11 14

This is a temporary fix, in order to unblock us and until we have a chance to discuss it with Filip on how to proceed further.

### DIFF
--- a/packages/components/collapsible-panel/src/collapsible-panel.spec.js
+++ b/packages/components/collapsible-panel/src/collapsible-panel.spec.js
@@ -124,7 +124,7 @@ describe('when onToggle is provided without isClosed', () => {
   });
 });
 
-describe('hen using CollapsiblePanelHeader component as header', () => {
+describe('condensed and when header prop has <CollapsiblePanelHeader>', () => {
   it('should render an H4 title when condensed', () => {
     render(
       <CollapsiblePanel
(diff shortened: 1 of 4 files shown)


packages/components/collapsible-panel/src/collapsible-panel.spec.js
```diff
@@ -123,6 +124,38 @@ describe('when onToggle is provided without isClosed', () => {
   });
 });
 
+describe('condensed and when header prop has <CollapsiblePanelHeader>', () => {
```
CarlosCortizasCT, line 127: **nit**: perhaps we can use a test title like this:
`when using CollapsiblePanelHeader component as header`

and the test could be:
`should render an H4 title when condensed`
`should render an H2 title when not condensed`

chloe0592: Thanks @CarlosCortizasCT! Makes more sense now. Updated here - [ba42b83](https://github.com/commercetools/ui-kit/pull/2548/commits/ba42b835ce0d135fd9401d3154ebdda2ea19eade).

CarlosCortizasCT (approved):
Thanks! 💯

Rhotimee (approved):
Looks good. Don't forget the changeset. 
Thanks.

ddouglasz (approved):
Nice workaround 🙌🏾

PR #2553, opened 2023-07-06
revert: (#2524) accessible info in the icons

chloe0592:
## Description

After enabling accessible titles for our SVG icons, we're experiencing problems in our `ui-kit` consumers tests.

We think we didn't find the best solution for this area so we want to first revert the initial implementation and later discuss for a better solution based on our findings.


.changeset/eight-glasses-kneel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/rich-text-utils': minor
+'@commercetools-uikit/checkbox-input': minor
+'@commercetools-uikit/icons': minor
+---
+
+Reverting accessible information in the icons
```
CarlosCortizasCT, line 7: Perhaps we can give a little more context to `ui-kit` consumers:
```suggestion
After enabling accessible titles for our SVG icons, we've seen issues in ui-kit consumers tests.
We think we didn't find the best solution for this area so we want to first revert the initial implementation and later discuss for a better solution based on our findings.
```

chloe0592: Yes, way better. Tnx! [740407e](https://github.com/commercetools/ui-kit/pull/2553/commits/740407e6834c26881dd3831204870204293640d8)

CarlosCortizasCT (approved):
Thanks 👍

PR #2561, opened 2023-07-13
Update `puppeteer` config param

CarlosCortizasCT:
#### Summary

Update `puppeteer` config param

## Description

We have some warning logs when running the visual regression tests related to some configuration update of the `puppeteer` package:
image

That package will change the way the headless mode configuration works so I think it's better for use to update consciously.

[Here](https://developer.chrome.com/articles/new-headless/) are more details about this change.

### DIFF
--- a/.changeset/nine-cooks-promise.md
+++ b/.changeset/nine-cooks-promise.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/localized-rich-text-input': minor
+'@commercetools-uikit/rich-text-input': minor
+---
+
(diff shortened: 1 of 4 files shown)


kark (approved):
Thanks 🙌 
seems we have already enabled it in app-kit
https://github.com/commercetools/merchant-center-application-kit/pull/2976#discussion_r1195029691

ddouglasz (approved):
Thank you 👏

chloe0592 (approved):
Thanks! 🙏

PR #2566, opened 2023-07-17
Add show delay and transition effects to `Tooltip` component

### LINKED ISSUE #2550: `Delay` props for <Tooltip> component
The <Tooltip> component has `closeAfter` prop but there is no `openAfter` prop. Currently it feels very disturbing when user moves their cursor and see a tooltip keeps popping up. It would be a better experience if the tooltip can display after few milli seconds

Allow Tooltip to consume a extra prop for delay and display the tooltip after the delay (only if cursor stayed on top of it throught out the delay time). If user quickly hovers over and moves away, we don't display the tooltip to avoid disturbances.

**Additional context**
Right now, I'm creating custom `BodyComponent` with custom css and passing it to tooltip to achieve this. It would be better if the component itself provide such a option.

CarlosCortizasCT:
#### Summary

Add show delay to `Tooltip` component.

closes #2550

## Description

Up until now we had a `closeAfter` property which controlled how much time we wait since the user exits the tooltip until it is remove from the ui.
The default value for this property was 0 (milliseconds) and we changed it to 200.

Also, we added a new property called `showAfter` which now controls how much time the tooltip takes to be displayed when the user hover over the triggering element. This is to avoid showing the tooltip when the user just "passes by" the triggering element with the cursor.
The default value for this property id 300 (milliseconds).

On top of that, we also added a new ui effect so instead of displaying/hiding the tooltip right away, we have a grow transition.
Most of the code of this PR is about this feature as we need to orchestrate the show/hide timings with the ui effect.


packages/components/tooltip/src/tooltip.tsx
```diff
@@ -117,24 +127,28 @@ export type TTooltipProps = {
...
   horizontalConstraint: 'scale',
   off: false,
   placement: 'top',
 };
 
 const Tooltip = (props: TTooltipProps) => {
+  const enterTimer = useRef<ReturnType<typeof setTimeout>>();
   const leaveTimer = useRef<ReturnType<typeof setTimeout>>();
-  const childrenRef = useRef<ReturnType<typeof setTimeout>>();
+  const childrenRef = useRef<HTMLElement>();
```
CarlosCortizasCT, line 151: This was incorrectly used as the ref was not passed to any element.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -156,51 +170,49 @@ const Tooltip = (props: TTooltipProps) => {
...
     modifiers: props.modifiers,
   });
-  const [isOpen, toggle] = useToggleState(false);
-  const closeTooltip = useCallback(() => {
-    toggle(false);
-  }, [toggle]);
-  const openTooltip = useCallback(() => {
-    toggle(true);
-  }, [toggle]);
+  const [state, setState] = useState<TTooltipState>('closed');
```
CarlosCortizasCT, line 186: This helps to handle the different ui states the component goes through.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -212,29 +224,31 @@ const Tooltip = (props: TTooltipProps) => {
           onFocus(event);
         }
 
-        if (!isOpen && !isControlled) {
-          openTooltip();
-        }
+        if (state !== 'opened' && !isControlled) {
+          setState('entering');
+          enterTimer.current = setTimeout(() => {
```
CarlosCortizasCT, line 229: Here's where we add the new display delay.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -244,21 +258,33 @@ const Tooltip = (props: TTooltipProps) => {
         onBlur(event);
       }
 
-      if (closeAfter) {
+      if (closeAfter && state === 'opened') {
         leaveTimer.current = setTimeout(() => {
-          handleClose(event);
+          const tooltipElement = popperInstance?.popper.querySelector(
```
CarlosCortizasCT, line 263: When the close timeout is done, we can't just remove the tooltip but start the exit transition. We do this by setting the state to `exiting`.
We also register an event to be notified when the transition is finished to actually remove the tooltip from the DOM.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -244,21 +258,33 @@ const Tooltip = (props: TTooltipProps) => {
...
+            '[data-testid="tooltip-message-wrapper"]'
+          ) as HTMLElement;
+          tooltipElement.addEventListener(
+            'animationend',
+            (event: AnimationEvent) => {
+              const element = event.target as HTMLElement;
+              element.style.display = 'none';
```
CarlosCortizasCT, line 270: This is something I'm now sure about.
After the transition is done, the tooltip gets visible again and I don't understand why, so I just hide it before it gets removed in the next tick.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -331,7 +358,14 @@ const Tooltip = (props: TTooltipProps) => {
             })}
             data-placement={popper.placement}
           >
-            <BodyComponent>{props.title}</BodyComponent>
+            <div
+              css={css({
+                ...getTooltipStyles(state),
+              })}
+              data-testid="tooltip-message-wrapper"
```
CarlosCortizasCT, line 365: Since `BodyComponent` can be provided by a consumer, we need a wrapper to handle the effect transitions.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -117,24 +127,28 @@ export type TTooltipProps = {
...
   horizontalConstraint: 'scale',
   off: false,
   placement: 'top',
 };
 
 const Tooltip = (props: TTooltipProps) => {
+  const enterTimer = useRef<ReturnType<typeof setTimeout>>();
   const leaveTimer = useRef<ReturnType<typeof setTimeout>>();
-  const childrenRef = useRef<ReturnType<typeof setTimeout>>();
+
```
CarlosCortizasCT, line 151: `childrenRef` was not assigned to any element, so it was doing nothing.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -156,53 +170,44 @@ const Tooltip = (props: TTooltipProps) => {
...
 
   const { onFocus, onMouseOver } = props.children.props;
-  const { onOpen } = props;
+  const { showAfter, onOpen } = props;
   const handleEnter = useCallback(
     (event?: ChangeEvent | FocusEvent) => {
-      // Remove the title ahead of time.
-      // We don't want to wait for the next render commit.
-      // We would risk displaying two tooltips at the same time (native + this one).
-      if (childrenRef && typeof childrenRef === 'function') {
```
CarlosCortizasCT, line 201: `childrenRef` was not assigned to any element so this wasn't working.
Handling the children title is done down below.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -283,15 +302,17 @@ const Tooltip = (props: TTooltipProps) => {
         // for seo and accessibility, we add the tooltip's title
         // as a native title when the title is hidden
         title:
-          !tooltipIsOpen && typeof props.title === 'string'
+          !tooltipIsOpen &&
+          state !== 'entering' &&
```
CarlosCortizasCT, line 306: Adding the new `entering` state here helps with correctly handling children's `title`.

FilPob:
Thanks! Looks awesome. I would suggest to decrease the grow animation duration to 80ms. @CarlosCortizasCT . But this duration should then not be adjustable by user, as discussed previously.

CarlosCortizasCT:
> Thanks! Looks awesome. I would suggest to decrease the grow animation duration to 80ms. @CarlosCortizasCT . But this duration should then not be adjustable by user, as discussed previously.

Updated b9f3259080edc1cc9afb2b9f955db5636c6f4545.

FilPob (approved):

packages/components/tooltip/src/tooltip.tsx
kark: if we add some basic logging of the current state on each rerender
```js
useEffect(() => {
    console.log(state);
  });
```
it seems that there is some inconsistency of state changes.
Screenshot 2023-07-24 at 15 17 05
I think changing `setState('exiting');` to `setState('closed');` solves the issue with the tooltip popping up again (but I don't think that's what we expect).

CarlosCortizasCT: Thanks for looking into it, Kacper.

I don't seem to be able to reproduce that behaviour.

If you can share any context I can use that would be helpful.

Also, if you can see why the events are in that unexpected order, that would be great.

kark: Thanks Carlos, and sorry for not being precise.

Perhaps this brings no value, it's just my observation derived from your comment. 

The incorrect order (and the effect of the tooltip showing up 2 times) happens when we eliminate a part of the code, where you added the `popperInstance` to hide the tooltip with css.

If we omit the `exiting` state in favor of `closed` like in the diff below it works without any manual changes in the css (the sequence is `entering` -> `opened` -> `closed`).

Screenshot 2023-07-25 at 08 48 04

Please let me know if you need any more info.

CarlosCortizasCT: Thanks for the explanation, Kacper 👍 
But if we get rid of the "exiting" state, I think we lose the new required close transition.
We now need to show and hide the tooltip with a transition. If we don't wait for the transition to end before setting the state to "close", the tooltip will be removed from the DOM before the transition ends.

packages/components/tooltip/src/tooltip.spec.js
```diff
@@ -1,3 +1,4 @@
+import { merge } from 'lodash';
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 8e9850115194386c1d0271c8868361106a1e75cc

packages/components/tooltip/src/tooltip.spec.js
```diff
@@ -6,10 +7,93 @@ import {
...
+  );
+  // Let's focus the button to trigger the tooltip
+  fireEvent[options.eventType](options.triggerElement);
+
+  // First we need to wait for the tooltip to be visible
+  // after the 'showAfter' delay
+  await screen.findByText('What kind of bear is best?');
+
+  // Now we can verify the callbacks
+  // expect(onMouseOver).toHaveBeenCalled();
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 8e9850115194386c1d0271c8868361106a1e75cc

packages/components/tooltip/src/tooltip.tsx
CarlosCortizasCT: Anyway, I removed (9687276a119f9408be21c6f1acac0ce7622c6744) the `display` _hack_ and everything is working fine.

Maybe it was something I added while I was working in the feature and with the final implementation is not needed.

kark (approved):
Thank you Carlos! 🎉
It looks good to me

packages/components/tooltip/src/tooltip.styles.ts
```diff
@@ -1,10 +1,44 @@
...
+const growOut = keyframes`
+  from {
+    transform: scale(1);
+  }
+  to {
+    transform: scale(0);
+  }
+`;
+
+const getAnimation = (state: TTooltipState) => {
```
ddouglasz, line 27: 💯

ddouglasz (approved):
Transition looks smooth. Thank you 🙌🏾

PR #2572, opened 2023-07-31
Fix `auto` horizontal constraint behaviour in select components

CarlosCortizasCT:
#### Summary

Fix `auto` horizontal constraint behaviour in select components

## Description

All `ui-kit`'s **select** components accept a property called `horizontalConstraint` which allows to control the width of the rendered element.
We have several numeric values available and also two string ones:
* **scale**: This makes the element to use 100% of available width
* **auto**: Currently this behaves as the previous one but, according to what has been discussed with UX, this should have the minimum width required by the element (as per its contents).

In this PR we're updating all inputs so the `auto` value for the `horizontalConstraint` property behaves as expected by UX.


FilPob:
@CarlosCortizasCT thanks! From a quick look I see that the Select Inputs work exactly as expected. Other Input types like Text or Number Input behave differently though. E.g setting the width to h. constraint `Auto` makes the Input even wider than h. constraint `3` 
image

FilPob:
Another question. How can we actually check if we dont break some UI's when making this change? Would each team need to go trough their pages and check for places where they used the `auto` value? And eventually set it to `scale`, if that's what they want?

CarlosCortizasCT:
> Another question. How can we actually check if we dont break some UI's when making this change? Would each team need to go trough their pages and check for places where they used the `auto` value? And eventually set it to `scale`, if that's what they want?

That's a good question.
The way I see it, we're fixing a bug so we won't be breaking anything.
Maybe the way you see it is that we would be changing the UI if there are consumers using the `auto` value, but I still see that this change would be for fixing it.

I checked in the MC repository and I only found one usage of the `auto` value (in the account application), so I don't see a big deal with that.

For Custom Applications, we will be changing the behaviour, but again I think we want to do this because it is a fix.
If we don't see it as a fix, then maybe we can discuss nexts steps.

FilPob (approved):

kark (approved):
Thank you. I think it works great 🙌

Rhotimee (approved):
Nice 🙌🏽

ddouglasz (approved):
Looks good to me 👏
Thanks!

PR #2577, opened 2023-08-09
fix(pagination): prevent searching inside page size selector

ChristianMoll:
Turns the `SelectInput` inside `PageSizeSelector` non searchable.
[Relevant shield board task](https://commercetools.atlassian.net/browse/SHIELD-874)


kark (approved):
Thanks for the contribution 🎉
Looks good to me!

FilPob (approved):

CarlosCortizasCT:
**Thanks a lot for** your contribution @ChristianMoll 🙇 

We had two issues pending with this component and it seems we addressed just one of them here.

Did you have the chance to look into the other one?
image

I'm asking because, since we are already fixing this component, it might be a good time to check both issues (maybe the second one is more a consumer one, but we need to know).

kark:
> **Thanks a lot for** your contribution @ChristianMoll 🙇
> 
> We had two issues pending with this component and it seems we addressed just one of them here.
> 
> Did you have the chance to look into the other one? image
> 
> I'm asking because, since we are already fixing this component, it might be a good time to check both issues (maybe the second one is more a consumer one, but we need to know).

If you agree, considering that this PR resolves another (GH) issue, could we consider merging it as is to prevent any progress blocks? We could address the other issue separately through a different PR, as I believe our team should dedicate time to that.

CarlosCortizasCT:
> If you agree, considering that this PR resolves another (GH) issue, could we consider merging it as is to prevent any progress blocks? We could address the other issue separately through a different PR, as I believe our team should dedicate time to that.

We can totally do that.

I was suggesting to also tackle the other issue since we are already working on this component and @ChristianMoll has already volunteered to help here.

We (Shield) team also have another priority topics to work on, so every help is much appreciated.

However, if @ChristianMoll does not have time to take a look at the other issue, then I would agree in merging this PR as it is right now.

ChristianMoll:
@CarlosCortizasCT I will take a look at the other problem and see if I can find a solution until tomorrow.

ChristianMoll:
@CarlosCortizasCT @FilPob @kark
The problem with the shadow is that the left and bottom part is cut off on the left and bottom because it overflows. For the left side a reasonable fix would be to simply add horizontal margins to the pagination.
Screenshot 2023-08-11 at 09 16 39

For the bottom on the other hand the problem is that the dropdown only expands enough space for the content + border if there is no space already available to display it. (it works fine if a bottom spacer is present for example). So the only solution I can come up with is to add another element or pseudo element to move the shadow into the expanded space. Here is how it looks if I add an after pseudo-element to move the shadow up into the space reserved by the dropdown as an example.
Screenshot 2023-08-11 at 11 17 22

CarlosCortizasCT:
@ChristianMoll thanks a lot for your investigation 🙇 

My suggestion would be to not add anything else to this component yet.
To me it seems more a problem of the context where it is used and, since we are already thinking in having a new component for all list pages, I believe we can address this issues when we start implementing that component.

@FilPob What do you think?

FilPob:
@ChristianMoll  thanks for the suggestion! I am not a developer but it does sound a bit like a workaround. It would solve the issue, so from my side it would be fine, but if tackling it differently (like Carlos suggested) would be a better approach code-wise then we should probably go for that.

CarlosCortizasCT:
> @ChristianMoll thanks for the suggestion! I am not a developer but it does sound a bit like a workaround. It would solve the issue, so from my side it would be fine, but if tackling it differently (like Carlos suggested) would be a better approach code-wise then we should probably go for that.

So let's merge this as it is and leave the other issues for the upcoming new component.

CarlosCortizasCT (approved):
Thanks a lot for your contribution! 🙇

PR #2581, opened 2023-08-22
Navigation new icons and color

chloe0592:
We need 3 new icons and one color (`primary-30: '#009987'`) for the Main navigation redesign.

1. Collapse icon
Screenshot 2023-08-22 at 10 18 59

2. Expand icon
Screenshot 2023-08-22 at 10 18 51

3. CT logo
Screenshot 2023-08-22 at 10 19 11


FilPob:
Looks good! Just one thing: Can we define the color-primary-30 as hsl instead of hex? The value would be `hsl(172.9608938547486, 100%, 30%)`

emmenko:
I would also suggest to have the icons in App Kit only, no need to include them in UI Kit.

CarlosCortizasCT:
> I would also suggest to have the icons in App Kit only, no need to include them in UI Kit.

I believe the collapse/expand icons represent a generic action which can be used in different context/applications where `ui-kit` would be consumed so maybe having them here would make sense.

If we want to move them to `app-kit` I would ask what would be the best place for them. 
I see we currently have this [assets](https://github.com/commercetools/merchant-center-application-kit/tree/main/packages/assets) package but none of the subdirectories seems to be appropriate for these new icons. Perhaps we can add a new one to accommodate them.

I know we will be using these new icons only in the navbar now, but I feel like they could be reused.

Anyway, that's just my suggestion.

emmenko:
> I believe the collapse/expand icons represent a generic action which can be used in different context/applications where ui-kit would be consumed so maybe having them here would make sense.

True, if it's fine by @FilPob we can keep them here in UI Kit.

The logo can be moved to App Kit then.

chloe0592:
> Looks good! Just one thing: Can we define the color-primary-30 as hsl instead of hex? The value would be `hsl(172.9608938547486, 100%, 30%)`

Thanks @FilPob! Changed here - [839567a](https://github.com/commercetools/ui-kit/pull/2581/commits/839567a43f4cf0c2d50a6f4872b7211d3c053434). ✅

chloe0592:
Thanks @emmenko and @CarlosCortizasCT! Removed the icon here - [7fc894e](https://github.com/commercetools/ui-kit/pull/2581/commits/7fc894e2556d2b7f4e98e8873006fc833df88583). And will include it in the App-Kit instead (if Filip won't object).

FilPob:
I see a potential usecase for the collapse / Expand icons for the filter sidebar. I would keep them in UI Kit

emmenko (approved):
Thanks!

.changeset/gold-guests-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/icons': patch
+'@commercetools-uikit/design-system': patch
+---
+
+Adding new color (primary-30) and icons for main navigation
```
CarlosCortizasCT, line 6:
```suggestion
Adding new color (primary-30) and icons for representing expand/collapse actions.
```

CarlosCortizasCT (approved):
Thanks 👍

PR #2582, opened 2023-08-23
Make page size select input depend on isSearchable prop

ddouglasz:
#### Summary

Make page size select input depend on `isSearchable` prop

## Description

Currently, for a small sized list with pagination, the search select input in the pagination size allows users to input a value that is even larger than the size of the list by default, this should be dependent on usage and allow users to turn on and off this feature with the `isSearchable` prop.

Kapture 2023-08-01 at 14 17 48-20230801-121803 (1)


ddouglasz:
This was supposed to be a second part of a reported pagination required feature, but it seems this has been handled on a recently merged PR. So I will be closing this PR.

PR #2584, opened 2023-08-23
Adding color `primary-20`

chloe0592:
We are supposed to use this new color in the main navigation redesign.


chloe0592:
@FilPob yeah you're right, thanks. Updated here - [ca961e3](https://github.com/commercetools/ui-kit/pull/2584/commits/ca961e3ccc37287bdb908ec9a17dc305fe809b1f). ✔️

CarlosCortizasCT (approved):
Thanks! 👍

PR #2593, opened 2023-08-31
Fix edge case with managing end transition in `Tooltip` component

CarlosCortizasCT:
#### Summary

Fix edge case with managing end transition in `Tooltip` component

## Description

We've seen some end-to-end test having random problems that involves the `Tooltip` component.

The error we see in the browser console in those cases is like this:
image

Code referenced in the stack trace:
image

It means, for some unknown reason, when we want to add a listener for the animation ending, the expected element does not exist anymore.
I don't understand this behaviour. Maybe it's due to the e2e test environment being slow (just a guess).

In this  PR I just check for that potential condition and, if there is no element to be animated, I assume is not in the DOM anymore and apply the side effect (calling the `onClose` props callback) right away.

I'm not adding any test because I don't know how to reproduce this error.

Any suggestion about how to handle this in any other way would be much appreciated.

### DIFF
--- a/.changeset/khaki-mirrors-refuse.md
+++ b/.changeset/khaki-mirrors-refuse.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Fix an edge case when managing end transition side effects.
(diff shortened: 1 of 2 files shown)


ddouglasz (approved):
Thank you,  Carlos 👏

emmenko (approved):
💯

PR #2595, opened 2023-09-08
chore: adding new design tokens

chloe0592:
We need these 2 design tokens for the main navigation work and they will be used further in the future.


CarlosCortizasCT (approved):
Thanks! 👍

kark (approved):
💯

ddouglasz (approved):
🙌🏽

Issue #2601, opened 2023-09-20
Missing peer dependencies in the uikit packages are causing yarn warnings


timonrey:
**Describe the bug**
In the docs-kit, we currently see a lot of yarn warnings due to missing peer dependencies in the uikit packages. Most of the warnings show react dependencies that probably need to be added to the packages.
<img width="1626" alt="Bildschirmfoto 2023-09-20 um 10 29 07" src="https://github.com/commercetools/ui-kit/assets/31767369/4e01358c-1703-47fa-baa5-648cee695089">

**To Reproduce**
Install the uikit packages in any project and run yarn to see the console output.

**Expected behavior**
It would be great if we could solve these yarn warnings in the console output to have a better overview for other dependency warnings or issues.

Thank you!

PR #2603, opened 2023-09-20
chore: add missing peer dependencies

kark:
#### Summary

Fixes https://github.com/commercetools/ui-kit/issues/2601

This PR covers the missing peer dependencies that seem to cause yarn warnings.
This are the logs after changes:
 
Screenshot 2023-09-20 at 12 02 21

> NOTE: `yarn explain peer-requirements` show a lot more missing packages.

### DIFF
--- a/packages/calendar-utils/package.json
+++ b/packages/calendar-utils/package.json
@@ -40,11 +40,13 @@
   "devDependencies": {
     "moment": "2.29.4",
     "react": "17.0.2",
+    "react-dom": "17.0.2",
     "react-intl": "^6.3.2"
   },
   "peerDependencies": {
     "moment": "2.x",
     "react": "17.x",
+    "react-dom": "17.x",
     "react-intl": "6.x"
   }
 }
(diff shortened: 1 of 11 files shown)


packages/calendar-utils/package.json
```diff
@@ -40,11 +40,13 @@
   "devDependencies": {
     "moment": "2.29.4",
     "react": "17.0.2",
+    "react-dom": "17.0.2",
     "react-intl": "^6.3.2"
   },
   "peerDependencies": {
     "moment": "2.x",
     "react": "17.x",
+    "react-dom": "17.x",
```
emmenko, line 49: As we are "changing" the peer dependencies, hopefully changeset won't assume that this is a breaking change.

emmenko (approved):
Thanks for looking into this!

packages/calendar-utils/package.json
kark: Thanks for the heads-up, I had no idea about this. I'm not sure, but seems it's been fixed https://github.com/changesets/changesets/issues/380

PR #2607, opened 2023-09-25
Tooltip long title issue

chloe0592:
Currently, when a tooltip is contained within a parent element with a `white-space` property set to `nowrap`, it will inherit the property and that will create an unwanted layout if the tooltip label is long.

Screenshot 2023-09-25 at 12 10 51

We can address this issue by explicitly setting up the `white-space` to `break-spaces` and move to the next line when the text exceeds a certain length.

Screenshot 2023-09-25 at 12 11 19

### DIFF
--- a/.changeset/violet-mice-change.md
+++ b/.changeset/violet-mice-change.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/tooltip': patch
 ---
 
-Adding `break-space` prop to tooltip for long titles
+Fix tooltip content not completely visible when parent component has `white-space: nowrap` style.
--- a/.changeset/violet-mice-change.md
+++ b/.changeset/violet-mice-change.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Adding `break-space` prop to tooltip for long titles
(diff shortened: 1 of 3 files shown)


packages/components/tooltip/src/tooltip.visualroute.jsx
```diff
@@ -45,6 +46,31 @@ export const component = () => {
           </Tooltip>
         </ContainerWithPadding>
       </Spec>
+      <Spec
+        label="CollapsiblePanel as a parent"
+        listPropsOfNestedChild={true}
+      >
+        <CollapsiblePanel
```
chloe0592, line 53: The `CollapsablePanel` component has css property `white-space: nowrap` so in this test we can make sure that it doesn't override the tooltip's `white-space: break-spaces`. 

Screenshot 2023-09-25 at 14 48 03

CarlosCortizasCT (approved):
Looks brilliant to me 🙇 

Thanks!

.changeset/violet-mice-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Adding `break-space` prop to tooltip for long titles
```
CarlosCortizasCT, line 5: **nit**
```suggestion
Fix tooltip content not completely visible when parent component has `white-space: nowrap` style.
```

emmenko (approved):
Nice! 🙌

ddouglasz (approved):
🙌🏾

PR #2613, opened 2023-10-05
feat(Typography): update Text.Body/Detail components and create Caption component

jmcreasman:
Summary
- Updated the Text.Body/Detail components to use a new fontWeight prop instead of isBold which will soon be deprecated
- Created a new Text.Caption component
- Added new prop to every case where isBold was used or passed down so it can replace it when ready

[Design](https://www.figma.com/file/v6TLyh6Egyh5NJxWOb3ixK/New-Audit-Log?type=design&node-id=200-5936&mode=design)
The new prop has three possible font weight selections; regular, medium or bold.
Screen Shot 2023-10-10 at 11 10 25 AM

Here is an example of how the new prop looks:
Screen Shot 2023-10-10 at 11 18 45 AM
Screen Shot 2023-10-10 at 11 18 50 AM
Screen Shot 2023-10-10 at 11 18 56 AM

Here is the new Caption component:
Screen Shot 2023-10-10 at 11 19 05 AM


packages/components/text/src/text.tsx
```diff
@@ -176,10 +179,16 @@ Wrap.displayName = 'TextWrap';
...
+  /**
+   * @deprecated
+   *
+   * `isBold` prop is being deprecatied in favor of the new `fontWeight` prop.
+   */
   isBold?: boolean;
   isItalic?: boolean;
   isStrikethrough?: boolean;
   tone?: TTone | 'inverted';
+  fontWeight?: TFontWeight | 'regular' | 'medium' | 'bold';
```
emmenko, line 191:
```suggestion
  fontWeight?: TFontWeight;
```

packages/components/text/src/text.tsx
```diff
@@ -217,11 +226,17 @@ const Body = (props: TBodyProps) => {
...
+   * @deprecated
+   *
+   * `isBold` prop is being deprecatied in favor of the new `fontWeight` prop.
+   */
   isBold?: boolean;
   isItalic?: boolean;
   isStrikethrough?: boolean;
   as?: 'span' | 'small';
   tone?: TTone | 'warning' | 'inverted';
+  fontWeight?: TFontWeight | 'regular' | 'medium' | 'bold';
```
emmenko, line 239:
```suggestion
  fontWeight?: TFontWeight;
```

packages/components/text/src/text.tsx
```diff
@@ -176,10 +179,16 @@ Wrap.displayName = 'TextWrap';
 
 export type TBodyProps = {
   as?: 'span' | 'p';
+  /**
+   * @deprecated
+   *
+   * `isBold` prop is being deprecatied in favor of the new `fontWeight` prop.
```
emmenko, line 185:
```suggestion
   * @deprecated: Use the new `fontWeight` prop.
```

packages/components/text/src/text.tsx
```diff
@@ -217,11 +226,17 @@ const Body = (props: TBodyProps) => {
 Body.displayName = 'TextBody';
 
 export type TDetailProps = {
+  /**
+   * @deprecated
+   *
+   * `isBold` prop is being deprecatied in favor of the new `fontWeight` prop.
```
emmenko, line 232:
```suggestion
   * @deprecated: Use the new `fontWeight` prop.
```

packages/components/text/src/text.tsx
```diff
@@ -259,4 +274,31 @@ const Detail = (props: TDetailProps) => {
 };
 Detail.displayName = 'TextDetail';
 
-export default { Headline, Wrap, Subheadline, Detail, Body };
+export type TCaptionProps = {
+  isItalic?: boolean;
+  isStrikethrough?: boolean;
+  tone?: TTone | 'warning' | 'inverted';
+  fontWeight?: TFontWeight | 'regular' | 'medium' | 'bold';
```
emmenko, line 281:
```suggestion
  fontWeight?: TFontWeight;
```

packages/components/text/src/text.styles.ts
```diff
@@ -122,6 +135,7 @@ export const bodyStyles = (props: TBodyProps) => css`
   ${props.isItalic && italic}
   ${props.isStrikethrough && strikethrough}
   ${props.tone && getTone(props.tone)}
+  ${props.fontWeight && getFontWeight(props.fontWeight)}
```
emmenko, line 138: I would suggest to merge the two props and remove the `${props.isBold && bold}` part

```suggestion
  ${getFontWeight(props)}
```

In the `getFontWeight` function then you can check for the `isBold` prop as well for backwards compatibility.

packages/calendar-utils/src/calendar-header/calendar-header.tsx
```diff
@@ -82,7 +82,7 @@ const CalendarHeader = (props: TCalendarHeader) => {
             size="medium"
           />
         </Tooltip>
-        <Text.Body as="span" isBold={true}>
+        <Text.Body as="span" isBold={true} fontWeight="bold">
```
emmenko, line 85: I would suggest to already remove the `isBold` usage and keep only the `fontWeight` prop. Having both does not make much sense.

Thanks

emmenko (commented):
Thanks for the contribution! 🙌

Overall looks good, I left some suggestions to improve things a bit.

packages/components/label/src/label.tsx
```diff
@@ -38,6 +38,7 @@ export type TLabelProps = {
   htmlFor?: string;
   // Indicates if the label title should be in bold text
   isBold?: boolean;
+  fontWeight?: 'regular' | 'medium' | 'bold';
```
Sarah4VT, line 41: There is no comment for fontWeight prop. Not sure if that's a consistency thing but it looks like all the others have a comment. Not sure if these comments are used in documentation somewhere.

Sarah4VT (changes requested):
Biggest thing I'd like to see cleaned up is removing internal usages of the isBold prop so we don't pass both isBold and fontWeight everywhere.

packages/calendar-utils/src/calendar-header/calendar-header.tsx
jmcreasman: @emmenko I might have misunderstood but I thought I needed to support both?

packages/components/text/src/text.styles.ts
jmcreasman: @emmenko I get `Argument of type 'TBodyProps' is not assignable to parameter of type 'string'` when I try this. Am I doing something wrong?

packages/components/label/README.md
```diff
@@ -25,14 +25,15 @@ export default Example;
...
+| `tone`                       | `union`<br/>Possible values:<br/>`'primary' , 'inverted'`        |    -     | -      | -         | Indicates the tone to be applied to the label                                                                   |
+| `children`                   | `ReactNode`                                                      | ✅ (\*)  | -      | -         | Value of the label                                                                                              |
+| `intlMessage`                | `MessageDescriptor`                                              | ✅ (\*)  | -      | -         | An intl message object that will be rendered with `FormattedMessage`                                            |
+| `isBold`                     | `boolean`                                                        |    -     | -      | `false`   | Indicates if the label title should be in bold text                                                             |
```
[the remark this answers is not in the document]
jmcreasman: Can't hurt, I'll add something in!

packages/components/label/src/label.tsx
jmcreasman: I'll add one just to be safe

jmcreasman:
> Biggest thing I'd like to see cleaned up is removing internal usages of the isBold prop so we don't pass both isBold and fontWeight everywhere.

From what I was told I shouldn't remove isBold but just mark it as deprecated for now. Please let me know if I misunderstood that though :)

packages/components/text/src/text.styles.ts
tdeekens: Prior the function is `const getFontWeight = (fontWeight: string) => string` so the argument `fontWeight` is a `string`. With the suggested change it's whatever props come in depending on what component it's called I presume. So the type needs to account for that as it's `TBodyProps`.

Sarah4VT:
> > Biggest thing I'd like to see cleaned up is removing internal usages of the isBold prop so we don't pass both isBold and fontWeight everywhere.
> 
> From what I was told I shouldn't remove isBold but just mark it as deprecated for now. Please let me know if I misunderstood that though :)

We shouldn't remove support for isBold, but we should go ahead and convert internal usages of isBold to using fontWeight. That's what I meant and I *think* what the other's meant as well.

packages/calendar-utils/src/calendar-header/calendar-header.tsx
jmcreasman: Makes sense thanks I'll update

packages/components/text/src/text.styles.ts
jmcreasman: @emmenko / @tdeekens 

Even when I fix the TS error it still breaks when I try and make this change. I tried doing something similar with getTone on line 137 and it broke too. Is there a reason to implement this prop differently than the pattern the others are using? Just trying to better understand :)

packages/components/label/src/label.tsx
```diff
@@ -36,8 +36,10 @@ export type TLabelProps = {
    * ````
    */
   htmlFor?: string;
-  // Indicates if the label title should be in bold text
+  // Indicates if the label title should be in bold text. This is deprecated, use `fontWeight` instead.
   isBold?: boolean;
+  // Indicates if the label title should be in bold text
```
Sarah4VT, line 41: Nit picky: Really this is indicating the level of "boldness", not if it's bold or not.

packages/components/text/src/text.visualroute.jsx
```diff
@@ -100,7 +100,9 @@ export const component = () => (
       <Text.Body>Body text</Text.Body>
     </Spec>
     <Spec label="Body - bold">
-      <Text.Body isBold={true}>Body text bold</Text.Body>
+      <Text.Body isBold={true} fontWeight="bold">
```
CarlosCortizasCT, line 103: I think we should only use `fontWeight` prop in these tests.

packages/components/text/src/text.styles.ts
CarlosCortizasCT: I think the reasoning behind Nicola's proposal is to abstract away from consumers of the `getFontWeight` function the logic to know whether to use `isBold` or `fontWeight` properties.

Perhaps we can use an implementation like this one:
```ts
const getFontWeight = (props: { isBold?: boolean; fontWeight?: string; }) => {
  switch (props.fontWeight) {
    case 'regular':
      return `font-weight: ${designTokens.fontWeight400};`;
    case 'medium':
      return `font-weight: ${designTokens.fontWeight500};`;
    case 'bold':
      return `font-weight: ${designTokens.fontWeight600};`;
    default:
      if (props.isBold) {
        return `font-weight: ${designTokens.fontWeight600};`;
      }
      return ``;
  }
};
```

Then, we can update text components styles definitions like this:

**BEFORE**
```ts
export const bodyStyles = (props: TBodyProps) => css`
  ${getBaseStyles()}
  margin: 0;
  font-size: ${designTokens.fontSizeForTextAsBody};
  line-height: ${designTokens.lineHeightForTextAsBody};
  font-weight: ${designTokens.fontWeightForTextAsBody};
  ${props.isBold && bold}
  ${props.isItalic && italic}
  ${props.isStrikethrough && strikethrough}
  ${props.tone && getTone(props.tone)}
  ${props.fontWeight && getFontWeight(props)}
  ${props.truncate && truncate}
  ${props.nowrap && nowrap}
`;
```

**AFTER**
```ts
export const bodyStyles = (props: TBodyProps) => css`
  ${getBaseStyles()}
  margin: 0;
  font-size: ${designTokens.fontSizeForTextAsBody};
  line-height: ${designTokens.lineHeightForTextAsBody};
  font-weight: ${designTokens.fontWeightForTextAsBody};
  ${props.isItalic && italic}
  ${props.isStrikethrough && strikethrough}
  ${props.tone && getTone(props.tone)}
  ${getFontWeight(props)}
  ${props.truncate && truncate}
  ${props.nowrap && nowrap}
`;
```

.changeset/eight-rules-clap.md
```diff
@@ -0,0 +1,12 @@
...
+'@commercetools-uikit/select-utils': minor
+'@commercetools-uikit/icons': minor
+'@commercetools-uikit/label': minor
+'@commercetools-uikit/card': minor
+'@commercetools-uikit/text': minor
+'@commercetools-uikit/calendar-utils': minor
+'@commercetools-uikit/design-system': minor
+---
+
+feat(Typography): update Text.Body/Detail components and create Caption component
```
CarlosCortizasCT, line 12: Perhaps we need to explain a little more what we're introducing since this is what `ui-kit` users will read when learning about what's new in this version.

----------

Introducing a new `Text` component named `Caption` which will render the smallest text available.
Example:
```jsx
<Text.Caption>your small text</Text.Caption>
```

Also, we've deprecated the `isBold` property in these components in favour of a new property called `fontWeight`.
The former property only supported two values whereas the new one supports three: `regular`, `medium` and `bold`.

packages/components/text/src/text.styles.ts
jmcreasman: @CarlosCortizasCT makes sense, thanks for the explanation!

packages/components/label/src/label.tsx
jmcreasman: True, I'll correct it

packages/components/text/src/text.visualroute.jsx
jmcreasman: Good catch, I missed this in the first pass through. Thanks!

.changeset/eight-rules-clap.md
jmcreasman: Edited :)

jmcreasman:
@culotta just confirming we're not removing isBold from subheadlineStyles?

ByronDWall (approved):
From what I can tell, this looks good and makes it so that `isBold` is handled by `getFontWeight`.

FilPob:
UI wise looks good :) thanks

FilPob (approved):

.changeset/eight-rules-clap.md
```diff
@@ -0,0 +1,19 @@
...
```
CarlosCortizasCT, line 12: Please remove this line 🙏 

```suggestion
```

.changeset/eight-rules-clap.md
```diff
@@ -0,0 +1,19 @@
...
+
+Introducing a new Text component named Caption which will render the smallest text available.
+Example:
+
+<Text.Caption>your small text</Text.Caption>
+Also, we've deprecated the isBold property in these components in favour of a new property called fontWeight.
+The former property only supported two values whereas the new one supports three: regular, medium and bold.
```
CarlosCortizasCT, line 19: Please update this section to highlight the code: 🙏 

--------

Introducing a new `Text` component named `Caption` which will render the smallest text available.

Example:
```jsx
<Text.Caption>your small text</Text.Caption>
```

Also, we've deprecated the `isBold` property in these components in favour of a new property called `fontWeight`.
The former property only supported two values whereas the new one supports three: `regular`, `medium` and `bold`.

packages/components/text/src/text.tsx
CarlosCortizasCT: I think this change is pending here.

packages/components/text/README.md
```diff
@@ -100,19 +100,20 @@ import Text from '@commercetools-uikit/text';
...
+| Props             | Type             | Required | Values                                                                        | Default   |                                                                                           |
+| ----------------- | ---------------- | :------: | ----------------------------------------------------------------------------- | --------- | ----------------------------------------------------------------------------------------- |
+| `as`              | `String`         |    -     | `['p', 'span']`                                                               | -         |                                                                                           |
+| `id`              | `String`         |    -     | -                                                                             | -         | Used as HTML id property                                                                  |
+| `isBold`          | `Boolean`        |    -     | -                                                                             | `false`   |
```
CarlosCortizasCT, line 107: We should also add a warning here about `isBold` is deprecated in favour of `fontWeight`.

packages/components/text/README.md
```diff
@@ -138,20 +139,21 @@ import Text from '@commercetools-uikit/text';
...
+| Props             | Type             | Required | Values                                                                        | Default   |                                                                                           |
+| ----------------- | ---------------- | :------: | ----------------------------------------------------------------------------- | --------- | ----------------------------------------------------------------------------------------- |
+| `as`              | `string`         |    -     | `['small', 'span']` `[^]`                                                     | `false`   |                                                                                           |
+| `id`              | `String`         |    -     | -                                                                             | -         | Used as HTML id property                                                                  |
+| `isBold`          | `Boolean`        |    -     | -                                                                             | `false`   |
```
CarlosCortizasCT, line 146: We should also add a warning here about `isBold` is deprecated in favour of `fontWeight`.

CarlosCortizasCT: and also for the `Subheadline` component.

CarlosCortizasCT:
I'm also missing some new use cases for `Label` and `Text` components in the VRTs.

Please, ping me if you need help with those.

packages/components/text/README.md
jmcreasman: @CarlosCortizasCT is it being deprecated in Subheadline? The ticket told me only to remove it for Body/Detail components.

CarlosCortizasCT: Oh! My bad. Sorry 🤦 

Forget about `Subhealine`.

ByronDWall:
> I'm also missing some new use cases for Label and Text components in the VRTs.

@CarlosCortizasCT I believe I've added the necessary VRT use cases for Label and Text, and have made updates based on your other comments.  Please let me know if there's anything further necessary for this PR, as Jon will be out for the next 2 days.  Thanks!

culotta:
> @culotta just confirming we're not removing isBold from subheadlineStyles?

We had originally spec'd only updating typography critical to the AuLo redesign, but this case is a bit unique. From a design perspective, I don't think either approach would have a huge impact, but I could see an inconsistency on Subheadline (`isBold` vs `fontWeight`) leading to a poorer dev experience. Curious your thoughts here @FilPob @Sarah4VT @ByronDWall

packages/components/label/src/label.tsx
```diff
@@ -80,14 +79,14 @@ const Label = (props: TLabelProps) => {
     <label
       css={css`
         > div {
-          font-weight: ${!props.isBold && designTokens.fontWeight500};
+          font-weight: ${props.fontWeight};
```
CarlosCortizasCT, line 82: I think this change should be different since now we can see a lot of changes in Percy related to `Label` component using a different `font-weight`.

If I'm not mistaken, the idea here is that, if the consumer didn't want to use `bold`, the `font-weight` to be used should be `500` instead of the default value `400`.
Also, now we're passing one of these values to this rule: `regular`, `medium` and `bold`, and only the latter is a supported value for that rule.

Also, I noticed we're removing the `isBold` property from this component, which would be a breaking change so we should avoid it.

I think we should try to keep the previous behaviour:
```
font-weight: ${(props.isBold !== true && props.fontWeight !== 'bold') && designTokens.fontWeight500}
```

@FilPob did you take a look at Percy? Does this make sense to you?

FilPob:
@CarlosCortizasCT I looked at the VRT and in fact I found 2 issues. The `label` and `pagination` components now fall back to font-weight: `400`, whereas before it has been `500`.

FilPob:
> > @culotta just confirming we're not removing isBold from subheadlineStyles?
> 
> We had originally spec'd only updating typography critical to the AuLo redesign, but this case is a bit unique. From a design perspective, I don't think either approach would have a huge impact, but I could see an inconsistency on Subheadline (`isBold` vs `fontWeight`) leading to a poorer dev experience. Curious your thoughts here @FilPob @Sarah4VT @ByronDWall

Good catch, I was not even aware we have the `is bold` prop for the subheadlines. Right now we dont have intentions to start altering between different font-weights for headlines and subheadlines, so I think adding the font-weight prop there wouldn't make sense.

packages/components/label/src/label.tsx
jmcreasman: @CarlosCortizasCT so in a previous comment you suggested to remove isBold from the `label.visualroute.jsx` file and I think I took that to mean it needed to be removed from the `label` component too. It looked like `isBold` there was being passed into the label component where it ultimately was used on a `Text.Detail` within that component. So you're saying `isBold` needs to be restored at the top level usage?

jmcreasman: > Also, I noticed we're removing the `isBold` property from this component, which would be a breaking change so we should avoid it.

@CarlosCortizasCT so in a previous comment you suggested to remove isBold from the label.visualroute.jsx file and I think I took that to mean it needed to be removed from the label component too. It looked like isBold there was being passed into the label component where it ultimately was used on a Text.Detail within that component. So you're saying isBold needs to be restored at the top level usage?

CarlosCortizasCT: I'm sorry for the misunderstanding.

We need to bear in mind the `Level` component is a public one; it can be consumed in isolation requiring [its own package](https://uikit.commercetools.com/?path=/story/components-label--label). So if we remove a property from it, all current consumers would be affected and that's why I think we should not do that.

It makes sense to me to not use the deprecated property (**isBold**) internally, but maybe it would make sense to keep one VRT where we use it exclusively to resemble current consumers expected behaviour.

Sorry for the back and forth. 😢 

We can have a chat about this anytime you want if you feel like we need more clarification.

ByronDWall: I am seeing two conflicting messages here.  

That we should keep `isBold` and not accept the `fontWeight` prop in the component api for `Label`:

> Also, I noticed we're removing the isBold property from this component, which would be a breaking change so we should avoid it.

And that we should accept `fontWeight`, but override the default to `500` if there isn't a value specified for the `fontWeight` prop on the `Label`.

> `font-weight: ${(props.isBold !== true && props.fontWeight !== 'bold') && designTokens.fontWeight500}`

So, here is my solution - which adds `isBold` back to the `Label` component api as a deprecated prop and sets the default `font-weight` to `500` if neither `isBold` or `fontWeight` is passed to `Label`

```js
font-weight: ${props.isBold !== true && props.fontWeight !== undefined && designTokens.fontWeight500};
```

@CarlosCortizasCT @FilPob let me know if this is the correct solution, or if you'd like to not deprecate `isBold` and remove the `fontWeight` prop from the `Label` api altogether.

CarlosCortizasCT: Agreed and updated here: 8797858fefbf364f1a510b3e96caeb4087fb145b

ByronDWall: Just leaving this here for posterity:

After talking with @CarlosCortizasCT, we have:

- deprecated the `isBold` prop, but not removed it from `Label`
- only allowed `medium` (`500`) or `bold` (`600`) as values for `fontWeight` for `Label`

Instead of implementing `font-weight: 500` by using a ternary in a css child selector that targets `Text.Detail`, I've refactored the component so that `fontWeight='medium'` is the default `fontWeight` prop for `label`, which will insure that the label's font weight is `500` unless the user specifies `bold`

CarlosCortizasCT (approved):
Looks good to me know 💯 

Thanks a lot for your work 🙇

CarlosCortizasCT:
@FilPob could you take a last look at Percy and approve the changes?

ByronDWall:
> We had originally spec'd only updating typography critical to the AuLo redesign, but this case is a bit unique. From a design perspective, I don't think either approach would have a huge impact, but I could see an inconsistency on Subheadline (`isBold` vs `fontWeight`) leading to a poorer dev experience. Curious your thoughts here @FilPob @Sarah4VT @ByronDWall

@culotta I don't think we need to update `Subheadline`, especially since it is very straightforward to see what the prop is in the storybook documentation

Sarah4VT (approved):

emmenko:
Great job everyone. Team work! 🙌

PR #2617, opened 2023-10-10
feat(DataTable): update styling for data table container

Sarah4VT:
#### Summary

Updates the styling for the DataTable container and BaseCell inside a table row.

## Description
[DataTable Requirements in Figma](https://www.figma.com/file/v6TLyh6Egyh5NJxWOb3ixK/New-Audit-Log?type=design&node-id=200-5568&mode=design&t=wNHHZQ5Xa0cwJw1X-0)
image

[BaseCell Requirements in Figma](https://www.figma.com/file/v6TLyh6Egyh5NJxWOb3ixK/New-Audit-Log?type=design&node-id=278-11838&mode=design&t=wNHHZQ5Xa0cwJw1X-4)
image

After screenshot:
image

image
image


FilPob:
@Sarah4VT thanks! all looks good, just he shadow is not correct. I see we have some naming mistakes around shadows in Figma... What we need here is the shadow that is used in the token:  `shadow-for-card-when-raised` Can we reuse that?

Sarah4VT:
> @Sarah4VT thanks! all looks good, just he shadow is not correct. I see we have some naming mistakes around shadows in Figma... What we need here is the shadow that is used in the token: `shadow-for-card-when-raised` Can we reuse that?

@FilPob I changed that value in the code. Haven't updated the screenshots yet, but I will.

FilPob:
awesome! all good now :)

FilPob (approved):

CarlosCortizasCT:
> > @Sarah4VT thanks! all looks good, just he shadow is not correct. I see we have some naming mistakes around shadows in Figma... What we need here is the shadow that is used in the token: `shadow-for-card-when-raised` Can we reuse that?
> 
> @FilPob I changed that value in the code and updated the screenshots.

Hey, I'm sorry to be be picky here, but I don't think decision tokens ([more info](https://uikit.commercetools.com/?path=/story/basics-tokens--all-tokens)) should be reused in different components as they are a decision for a specific context (which uses to be a component).

Even though we want the value of the shadows for `Cards` and `DataTable` components to be the same now, it would be difficult to reason about if we want to change the `Cards` in the future and, unexpectedly, we find out the `DataTable` is also changing.

I believe it would be better to create a new choice token (maybe `shadow-17`) and reference that value in the `shadow-for-card-when-raised` decision token. Also, create a new decision token (something like `shadow-for-data-table`) and pointing it ti the same choice token.

In the end it will be the same, but I believe this setup will help us maintaining the code.

FilPob:
@CarlosCortizasCT yes that's probably even better

Sarah4VT:
> @CarlosCortizasCT yes that's probably even better

That sounds reasonable to me, and I agree you don't want to have a change to `shadow-for-card-when-raised` affect the data table as well.

I can create 2 new tokens:
- `shadow17: 'var(--shadow-17, 0 1px 5px 0 rgba(0, 0, 0, 0.05))',`
- `shadowForDataTable: 'var(--shadow-for-data-table, var(--shadow-17))'`

Please confirm these are the right values.

Should `shadow-for-card-when-raised` also point to shadow-17?
```
shadowForCardWhenRaised:
    'var(--shadow-for-card-when-raised, var(--shadow-17))',
```

CarlosCortizasCT:
> > @CarlosCortizasCT yes that's probably even better
> 
> That sounds reasonable to me, and I agree you don't want to have a change to `shadow-for-card-when-raised` affect the data table as well.
> 
> I can create 2 new tokens:
> 
> * `shadow17: 'var(--shadow-17, 0 1px 5px 0 rgba(0, 0, 0, 0.05))',`
> * `shadowForDataTable: 'var(--shadow-for-data-table, var(--shadow-17))'`
> 
> Please confirm these are the right values.
> 
> Should `shadow-for-card-when-raised` also point to shadow-17?
> 
> ```
> shadowForCardWhenRaised:
>     'var(--shadow-for-card-when-raised, var(--shadow-17))',
> ```

Yes, that's my proposal.

Bear in mind you need to change the [definitions.yml](https://github.com/commercetools/ui-kit/blob/main/design-system/materials/internals/definition.yaml) file in the `design-system` package.
After that, you will run the `design-tokens:build` which will update the `css`, `json` and `ts` files of that package.

Please let me know in case you find something not clear enough. I'll be happy to help.

ByronDWall (approved):

Sarah4VT:
> > > @CarlosCortizasCT yes that's probably even better
> > 
> > 
> > That sounds reasonable to me, and I agree you don't want to have a change to `shadow-for-card-when-raised` affect the data table as well.
> > I can create 2 new tokens:
> > 
> > * `shadow17: 'var(--shadow-17, 0 1px 5px 0 rgba(0, 0, 0, 0.05))',`
> > * `shadowForDataTable: 'var(--shadow-for-data-table, var(--shadow-17))'`
> > 
> > Please confirm these are the right values.
> > Should `shadow-for-card-when-raised` also point to shadow-17?
> > ```
> > shadowForCardWhenRaised:
> >     'var(--shadow-for-card-when-raised, var(--shadow-17))',
> > ```
> 
> Yes, that's my proposal.
> 
> Bear in mind you need to change the [definitions.yml](https://github.com/commercetools/ui-kit/blob/main/design-system/materials/internals/definition.yaml) file in the `design-system` package. After that, you will run the `design-tokens:build` which will update the `css`, `json` and `ts` files of that package.
> 
> Please let me know in case you find something not clear enough. I'll be happy to help.

I made one tweak to what I did, instead of `shadowForDataTable` I put it under box shadows as `boxShadowForTable` (it wouldn't let me use DataTable anyways) since we are using this for a css 'box-shadow' property.

tdeekens (approved):

Rhotimee (approved):
Thank you!

.changeset/orange-suits-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/data-table': minor
```
Rhotimee, line 2:
```suggestion
'@commercetools-uikit/data-table': patch
```

PR #2621, opened 2023-10-17
Pangolin-3075: Avatar component update

kterry1:
#### Summary

`<Avatar/>` component update:
- Adds `color` prop
- Adds `icon` prop
- Updates initials font size per character length
- Updates background colors(w/isHighlighted) and font/fill colors

#### Adds `color` prop
**Design**

**Storybook**

#### Adds `icon` prop
**Design**

**Storybook**

#### Updates initials font size per character length
**Design**

#### Updates background colors(w/isHighlighted) and font/fill colors
**Design**

**Storybook**


packages/components/avatar/package.json
```diff
@@ -25,6 +25,7 @@
     "@commercetools-uikit/utils": "16.7.5",
     "@emotion/react": "^11.10.5",
     "@emotion/styled": "^11.10.5",
+    "lodash": "4.17.21",
```
kterry1, line 28: Adding this dependency is causing the build to fail. Would you rather I just write a similar function than pulling in `startCase` from lodash?

CarlosCortizasCT: I think this might be because the updated lockfile has not been committed.

CarlosCortizasCT:
Hi @kterry1 

First of all, thanks for your contribution to `ui-kit` 🙇

After an initial review, I found some thing that we might consider updating based on some conventions we usually use in this repository.

I've opened [this PR](https://github.com/commercetools/ui-kit/pull/2628) with a proposal to refactor things a bit which I believe could improve the maintainability in the long term.

Please take a look and let me know what you think.

We can also have a call to talk about it if you want to.

kterry1:
> Hi @kterry1
> 
> First of all, thanks for your contribution to `ui-kit` 🙇
> 
> After an initial review, I found some thing that we might consider updating based on some conventions we usually use in this repository.
> 
> I've opened [this PR](https://github.com/commercetools/ui-kit/pull/2628) with a proposal to refactor things a bit which I believe could improve the maintainability in the long term.
> 
> Please take a look and let me know what you think.
> 
> We can also have a call to talk about it if you want to.

@CarlosCortizasCT Thanks for making those changes. We can merge your proposal in but will need to make a slight change to the logic because I spoke with Filip and the accent's `colorAccent50` needs to be `colorAccent40` but the rest of the colors should be `color...50` as we've changed in these PRs. I'll approve that PR but would you like me to merge it or would you like to?

CarlosCortizasCT:
> @CarlosCortizasCT Thanks for making those changes. We can merge your proposal in but will need to make a slight change to the logic because I spoke with Filip and the accent's `colorAccent50` needs to be `colorAccent40` but the rest of the colors should be `color...50` as we've changed in these PRs. I'll approve that PR but would you like me to merge it or would you like to?

@kterry1 
I've just merged the proposal.

Please feel free to adjust whatever it's needed.

Thanks.

packages/components/avatar/src/avatar.styles.ts
```diff
@@ -0,0 +1,93 @@
...
+    wide: designTokens.fontSize10, // '12px',
+  },
+  m: {
+    narrow: designTokens.fontSize50, // '20px',
+    wide: designTokens.fontSize30, // '16px',
+  },
+  l: {
+    narrow: designTokens.fontSize90, //'48px',
+    wide: designTokens.fontSize80, // '40px',
+  },
```
CarlosCortizasCT, line 18: I left those commented pixels I used in my branch we should remove them. 🙏

CarlosCortizasCT (approved):
Looks good from my side 👍 

Thanks a lot for your contribution, Kevin 🙇

FilPob:
Awesome!, looks good from my side

packages/components/avatar/src/avatar.styles.ts
kterry1: I'll do this now

culotta (approved):
Nice, this is an exciting one 🙌

jmcreasman (approved):
Looks good to me :D

kterry1:
@CarlosCortizasCT I just rebased this branch. When you have a chance could you merge?

PR #2623, opened 2023-10-18
fix(LocalizedRichTextInput): reset callback

kark:
#### Summary

Fixes https://github.com/commercetools/ui-kit/issues/2626

##### More context about the issue:

In the `<LocalizedRichTextEditor>` component, when `<RichTextEditors>` for certain languages are first rendered, the refs are created. When the editors are hidden later on, the refs are null, so calling the reset callback for those hidden editors ends up with:
275848398-a6c7601a-3fe8-4090-a547-29b52d6f9b97-1

### DIFF
--- a/.changeset/early-peas-return.md
+++ b/.changeset/early-peas-return.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/localized-rich-text-input': patch
+---
+
+Fix reset callback
--- a/packages/components/inputs/localized-rich-text-input/src/editor.styles.ts
+++ b/packages/components/inputs/localized-rich-text-input/src/editor.styles.ts
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled';
+import { css } from '@emotion/react';
 import { designTokens } from '@commercetools-uikit/design-system';
 import type { TEditorProps } from './editor';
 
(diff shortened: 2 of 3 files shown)


kark:
> So I think this change prevents the reported error from happening but does not fix the underlying issue, which is imperatively resetting the component would not work for hidden languages.
> 
> Interestingly, I only found one usage of the imperative API in MC center.

Thank you. That's a valid concern. 

I suggest we differentiate between the part that demands immediate attention—resetting editors in the current form might risk blowing up the page—and the part that allows resetting hidden editors. 

I explored a potential solution with some CSS adjustments, but it falls short. Hence, it might be worthwhile to consider if using the imperative API is a viable option. However, this would need further investigation. 

So at this point I'd just suggest fixing the first issue.

CarlosCortizasCT:
> > So I think this change prevents the reported error from happening but does not fix the underlying issue, which is imperatively resetting the component would not work for hidden languages.
> > Interestingly, I only found one usage of the imperative API in MC center.
> 
> Thank you. That's a valid concern.
> 
> I suggest we differentiate between the part that demands immediate attention—resetting editors in the current form might risk blowing up the page—and the part that allows resetting hidden editors.
> 
> I explored a potential solution with some CSS adjustments, but it falls short. Hence, it might be worthwhile to consider if using the imperative API is a viable option. However, this would need further investigation.
> 
> So at this point I'd just suggest fixing the first issue.

I agree on having a two steps solution where we first just focus on preventing the uncontrolled error and later we try to make the API correctly work.

Thanks for you work and the explanation, Kacper 🙇

CarlosCortizasCT (approved):
Thanks for working on this, Kacper 🙇

chloe0592 (approved):
Thanks Kacper for fixing it! 🔧

PR #2624, opened 2023-10-18
Add full height to card inset property

ddouglasz:
#### Summary

Inset component should have a height of 100% to inherit the height of the parent.

Context: 
In a number of places where the cards component is used, we want to be able to always have some of its child components at a certain fixed position, just at the bottom, example is a button, this is sometimes difficult to attain if the card has an inset and the inset component it's using is not taking the actual height of it's parent container, thereby displacing the content not as low as we would like it to be.
This forces some custom styling to the card which should not be the case. 
So instead of styling the card and it's inner component, every time we want to use it this way, we simply just make sure the inset component is having a height of 100%.

current:
image

Expected:
image

### DIFF
--- a/.changeset/brown-lizards-remain.md
+++ b/.changeset/brown-lizards-remain.md
@@ -1,5 +1,0 @@
----
-'@commercetools-uikit/card': patch
----
-
-Inset property should have a height of 100% to inherit the height of the parent.
--- a/.changeset/brown-lizards-remain.md
+++ b/.changeset/brown-lizards-remain.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/card': patch
 ---
 
-card inset property should have a height of 100% to inherit the height of the parent.
+Inset property should have a height of 100% to inherit the height of the parent.
--- a/.changeset/brown-lizards-remain.md
+++ b/.changeset/brown-lizards-remain.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/card': patch
+---
+
+card inset property should have a height of 100% to inherit the height of the parent.
--- a/.changeset/wild-drinks-push.md
+++ b/.changeset/wild-drinks-push.md
@@ -3,4 +3,4 @@
 '@commercetools-uikit/spacings-inset': patch
 ---
 
-Updated Inset and InsetSquish components styles so they both use all available vertical space from their parent components.
+Updated `Inset` and `InsetSquish` components styles so they both use all available vertical space from their parent components.
--- a/.changeset/wild-drinks-push.md
+++ b/.changeset/wild-drinks-push.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/spacings-inset-squish': patch
+'@commercetools-uikit/spacings-inset': patch
+---
+
+Inset and Inset Squish property should have a height of 100% to inherit the height of the parent.
--- a/.changeset/wild-drinks-push.md
+++ b/.changeset/wild-drinks-push.md
@@ -3,4 +3,4 @@
 '@commercetools-uikit/spacings-inset': patch
 ---
 
-Inset and Inset Squish property should have a height of 100% to inherit the height of the parent.
+Updated Inset and InsetSquish components styles so they both use all available vertical space from their parent components.
--- a/packages/components/card/src/card.tsx
+++ b/packages/components/card/src/card.tsx
@@ -55,7 +55,15 @@ const Card = (props: TCardProps) => (
       // This is mostly useful in case custom styles are targeting this element.
       <div>{props.children}</div>
     ) : (
-      <Inset scale={props.insetScale}>{props.children}</Inset>
+      <div
+        css={css`
+          > div {
+            height: 100%;
+          }
+        `}
+      >
+        <Inset scale={props.insetScale}>{props.children}</Inset>
+      </div>
     )}
   </div>
 );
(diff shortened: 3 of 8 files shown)


packages/components/card/src/card.tsx
```diff
@@ -55,7 +55,15 @@ const Card = (props: TCardProps) => (
       // This is mostly useful in case custom styles are targeting this element.
       <div>{props.children}</div>
     ) : (
-      <Inset scale={props.insetScale}>{props.children}</Inset>
+      <div
+        css={css`
+          > div {
```
CarlosCortizasCT, line 60: I was thinking that maybe this rule should be applied to the `Inset` component instead.
What do you think?

ddouglasz: Thinking about it now, I think that should work fine too, I was not so sure of how this would be of effect to other use cases of the `Inset` component. That's why I tried to keep it within the scope of the cards use case. 
updated here: https://github.com/commercetools/ui-kit/pull/2624/commits/b54dd8e22fd387d7a9ddb8885b97386dbf9023ed

Thank you

.changeset/brown-lizards-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/card': patch
```
CarlosCortizasCT, line 2: Maybe we need to recreate the changeset since now we're changing a different component.

packages/components/card/src/card.tsx
CarlosCortizasCT: Let's see what our teammates think about it.

.changeset/brown-lizards-remain.md
ddouglasz: Ah true, updated here: https://github.com/commercetools/ui-kit/pull/2624/commits/ba411e4fc09d409573abad10eed42803dc69e915

Thank you

packages/components/card/src/card.tsx
kark: I'm not aware of any downsides. Perhaps we should consider updating `<InsetSquish>` as well for consistency 🤷‍♂️

ddouglasz: Good idea @kark, looking at it, I think we should add it there too. 👍🏾

ddouglasz: updated here since I also have checked and don't see any reason why we should not update` Inset squish`  also: https://github.com/commercetools/ui-kit/pull/2624/commits/8af1368ae3411f08b3b5e7c92f042856e19c4c45

.changeset/wild-drinks-push.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/spacings-inset-squish': patch
+'@commercetools-uikit/spacings-inset': patch
+---
+
+Inset and Inset Squish property should have a height of 100% to inherit the height of the parent.
```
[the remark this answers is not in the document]
ddouglasz: Updated: https://github.com/commercetools/ui-kit/pull/2624/commits/348880edd723586f25cf2be460c5e4063a7dc67a
Thank you

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks for taking care, Douglas 🙇

Rhotimee (approved):

kark (approved):

chloe0592 (approved):
Thank you @ddouglasz!

emmenko:
Shouldn't we apply the styles to the card component? We don't always render the inset component and it's anyway and implementation detail of the card.

https://github.com/commercetools/ui-kit/blob/14b4d9406c82d8994da19b6133d42febf58ae085/packages/components/card/src/card.tsx#L53-L59

CarlosCortizasCT:
> Shouldn't we apply the styles to the card component? We don't always render the inset component and it's anyway and implementation detail of the card.
> 
> https://github.com/commercetools/ui-kit/blob/14b4d9406c82d8994da19b6133d42febf58ae085/packages/components/card/src/card.tsx#L53-L59

That's the first idea Douglas implemented (ea8c97168c0e40f09d526270e066faf8d467a883) and surely that's a valid option. However, I suggested to add the css rule to the inset components as it seems to me this is something we always want for this kind of components. I believe they should be using all parents vertical space no matter what since that enable its children to implement different use cases (like the one with having content top and bottom).

Whenever I see a style trying to change a child component styles, I think there's something we need to reconsider"
```css
> div {
  height: 100%;
}
```

Pelase, let me know if I'm missing something.

emmenko:
Yes but the `Inset` component is not rendered if `insetScale === 'none'`. So the style is not applied.

How should that work?

> Whenever I see a style trying to change a child component styles, I think there's something we need to reconsider

Yes, 💯 totally agree. However, it seems weird to me to apply a style meant for the `Card` component to another component. What if some day we replace the `Inset` component from the `Card` with something else?

CarlosCortizasCT:
> Yes but the `Inset` component is not rendered if `insetScale === 'none'`. So the style is not applied.
> 
> How should that work?
> 
> > Whenever I see a style trying to change a child component styles, I think there's something we need to reconsider
> 
> Yes, 💯 totally agree. However, it seems weird to me to apply a style meant for the `Card` component to another component. What if some day we replace the `Inset` component from the `Card` with something else?

I think it would be easier if we have a chat about this. It would probably be faster so I'll schedule a call.

emmenko (approved):
Thanks @CarlosCortizasCT for the clarification. All good then 👌

PR #2628, opened 2023-10-19
Avatar styles refactor proposal

CarlosCortizasCT:
#### Summary

Avatar styles refactor proposal

## Description

This is a proposal for refactoring the ongoing work on [this PR](https://github.com/commercetools/ui-kit/pull/2621) with these goals in mind:
* Use less decision tokens
* Move styles logic to its own file
* Update visual regression tests to include more use cases
* Adjust the large size of the component


design-system/materials/internals/definition.yaml
```diff
@@ -561,22 +560,6 @@ decisionGroupsByTheme:
           choice: color-surface
         background-color-for-select-input-option-when-hovered:
           choice: color-neutral-98
-        background-color-accent-for-avatar:
```
ByronDWall, line 564: @CarlosCortizasCT Looking at this update, I think it makes a lot of sense, but it brings up a topic that I think a number of the people on the Pangolin team, myself included, might be unclear on.

Specifically - when is it best to specify a value as a decision token, and when is it best to use a design(choice?) token directly in the code?

CarlosCortizasCT: I totally understand you have this question because we don't have an official statement about them yet (although we are working on it).

Just to give you some more context, we introduced a lot of decision tokens during the migration to the new look and feel we applied to the Merchant Center because we needed to support two themes during the transition.
We realized managing that much tokens would be difficult in the long term and we also didn't think they will actually be needed since we don't plan to support custom themes in the mid term (apart from `dark theme` maybe).

We already have some tasks in our backlog (not high priority by now) to get rid of most of the decision tokens.

Hope this helps.

PR #2629, opened 2023-10-19
Pangolin 3072: Header cell updates

ByronDWall:
#### Summary

Updating the styling of the `DataTable` `HeaderCell`.

## Description

This PR updates the styling of the `HeaderCell` based on [this design](https://www.figma.com/file/v6TLyh6Egyh5NJxWOb3ixK/New-Audit-Log?type=design&node-id=200-5568&mode=design&t=mxErMfltHSqL99Uq-0):
Screenshot 2023-10-19 at 2 28 23 PM

In addition to specific css selectors being updated,

- the icons used to indicate whether a column's sort direction is ascending or descending were changed
- Animations were added to the sort icons, as shown in this screen recording:
https://github.com/commercetools/ui-kit/assets/87667330/758ed0d4-9431-442d-95d6-fc1bdc72c09b

- The header's label will be a maximum of 2 lines, after which the label text is truncated, as shown in this screen recording:

https://github.com/commercetools/ui-kit/assets/87667330/9324e1fb-a036-47f9-94b3-4f49281b8349

**_PLEASE NOTE:_** In order to achieve truncating the label text, it was necessary to make it so that the `headerIcon` (the `information` icon in the screen recording) is centered in the header cell instead of it being justified on the same line as the text.  In order to justify the icon along with the text, it would be necessary to position the icon inline with the text at the end of the string.  As shown in the screen recording, when the text overflows past the second line, it is still positioned - if the icon were justified along with the text, it would be hidden at the end of the text.  There is no way to both truncate the text at the end of the second line and justify the icon along with the text.  The solution I have implemented centers the icon in the header cell - when the label is 1 line, the icon appears to be justified with the text, it is only when the label has a height of 2 lines that it becomes apparent that the icon is centered.  Because the icon is specified by the user, there is no way to know what size it will be, and thus there is no way to reliably place it next to the text on the second line.  Therefore the icon is centered.


FilPob:
@ByronDWall Awesome, looks all good from my side!

FilPob (approved):

packages/components/data-table/src/header-cell.styles.tsx
```diff
@@ -100,12 +152,15 @@ const BaseHeaderCell = styled.th<TBaseHeaderCell>`
...
-  font-weight: ${designTokens.fontWeightForTableHeader};
-  font-size: ${designTokens.fontSizeForTable};
+  font-weight: ${designTokens.fontWeight500};
+  font-size: ${designTokens.fontSize10};
 
   /* right border that doesn't count towards the column width */
   box-shadow: inset -1px 0 ${designTokens.borderColorForTableHeader};
 
+  border-bottom: ${designTokens.borderWidth1} solid
+    ${designTokens.colorNeutral95};
```
kark, line 162: I've observed this effect:
Screenshot 2023-10-20 at 10 48 50
Just thought I'd suggest it as a possible alternative:
```suggestion
  /**
   * bottom border that doesn't overshadow the resize indicator
   */
  :after {
    content: '';
    position: absolute;
    z-index: -1;
    width: 100%;
    height: ${designTokens.borderWidth1};
    bottom: 0;
    left: 0;
    background-color: ${designTokens.colorNeutral95};
  }
```

kark (approved):
Thank you, it looks really great 🙌

culotta (approved):

jmcreasman (approved):
LGTM 🎸

Sarah4VT (approved):

PR #2630, opened 2023-10-19
Pangolin 3077: Header updates

ByronDWall:
#### Summary

Updates styles for `DataTable` `Header` and `HeaderCell`

## Description

Adds styles to the `DataTable` `Header` based on [this design](https://www.figma.com/file/v6TLyh6Egyh5NJxWOb3ixK/New-Audit-Log?type=design&node-id=200-5568&mode=design&t=mxErMfltHSqL99Uq-0):
Screenshot 2023-10-19 at 3 10 18 PM

Removes the white `border-right` from each header cell, and adds a `color-neutral` `border-right` to each header cell when the user hovers over the header component, as demonstrated in this screen recording:

https://github.com/commercetools/ui-kit/assets/87667330/7f7a2d6e-9b20-43ab-9fff-dd03bfd79980


packages/components/data-table/src/data-table.styles.tsx
```diff
@@ -103,6 +104,23 @@ const TableGrid = styled.table<TTableGrid>`
...
   display: contents;
+  /* Adds right border that doesn't affect column width
+  *  to each header cell when table header is hovered
+  */
+  :hover {
+    ${BaseHeaderCell}:before {
+      content: '';
+      position: absolute;
+      border-right: 1px solid ${designTokens.colorNeutral};
+      top: ${designTokens.spacing20};
```
ByronDWall, line 115: It was unclear what the borders top/bottom offset should be.  In the figma file it was 10px, but that is not a spacing value in the design system, so I went with 8px.  This can be changed if necessary.

CarlosCortizasCT:
Based on the designs, I would say sorting icons don't look as expected (color, placement and animations).
Is this something that will be addressed as a follow-up?

ByronDWall:
> Based on the designs, I would say sorting icons don't look as expected (color, placement and animations).
Is this something that will be addressed as a follow-up?

Hey @CarlosCortizasCT, those updates are being handled in [this PR](https://github.com/commercetools/ui-kit/pull/2630) since they are specific to the `HeaderCell` and not the `Header` overall

CarlosCortizasCT:
> Hey @CarlosCortizasCT, those updates are being handled in [this PR](https://github.com/commercetools/ui-kit/pull/2630) since they are specific to the `HeaderCell` and not the `Header` overall

Oh! Sorry.

I was confused because the cell is also mentioned in this PR's summary.

> Updates styles for DataTable Header and HeaderCell

ByronDWall:
sorry for the confusion, I just mentioned the cell because the styling here affects it

culotta (approved):
All looks good to me from design side 👍

kark (approved):
Thank you for this one as well! Assuming the 8px top/bottom offset of the column separator is correct, everything looks good from my end. 🎉

CarlosCortizasCT (approved):
LGTM 🙌 

Thanks!

PR #2635, opened 2023-11-02
Refactor `Inset` components vertical space usage

CarlosCortizasCT:
#### Summary

Refactor `Inset` components vertical space usage

## Description

We found a problem with the vertical space of the `Card` component contents which we tried to address in [this PR](https://github.com/commercetools/ui-kit/pull/2624).

In that one, we thought it made sense to address the `Inset` and `InsetSquish` components (as the `Card` is using the former`) since, as container components, they should always use its parent available vertical space but we were wrong.
We noticed several use cases where this is not true so we're now proposing another approach.

The idea is that the vertical space used by `Inset` and `InsetSquish` components can be controlled with a new prop (`useAllAvailableHeight`) defaulting to the previous behaviour and only changing it for the `Card` component use case.

We suggest to use a boolean property instead of an open height so we can still control what we allow those component to render, but I'm happy to hear another suggestions.

### DIFF
--- a/.changeset/stale-radios-punch.md
+++ b/.changeset/stale-radios-punch.md
@@ -5,5 +5,4 @@
 
 Refactor vertical space usage so consumers can control it.
 
-There's a new property named `height` (available values: `min-content` and `max-content`; the former is the default) that will allow consumers to control the vertical space usage of the component.
-When using the `max-content` value, it will force the component main container to use `100%` of the available height.
+There's a new property named `height` (available values: `auto` and `100%`; the former is the default) that will allow consumers to control the vertical space usage of the component.
--- a/.changeset/stale-radios-punch.md
+++ b/.changeset/stale-radios-punch.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-uikit/spacings-inset-squish': minor
+'@commercetools-uikit/spacings-inset': minor
+---
+
+Refactor vertical space usage so consumers can control it.
+
+There's a new property named `useAllAvailableHeight` that will force the component main container to use `100%` of the available height.
--- a/.changeset/stale-radios-punch.md
+++ b/.changeset/stale-radios-punch.md
@@ -5,5 +5,5 @@
 
 Refactor vertical space usage so consumers can control it.
 
-There's a new property named `height` (available values: `collapsed` and `expanded`) that will allow consumers to control the vertical space usage of the component.
-When using the `expanded` value, it will force the component main container to use `100%` of the available height.
+There's a new property named `height` (available values: `min-content` and `max-content`; the former is the default) that will allow consumers to control the vertical space usage of the component.
+When using the `max-content` value, it will force the component main container to use `100%` of the available height.
--- a/.changeset/stale-radios-punch.md
+++ b/.changeset/stale-radios-punch.md
@@ -5,4 +5,5 @@
 
 Refactor vertical space usage so consumers can control it.
 
-There's a new property named `useAllAvailableHeight` that will force the component main container to use `100%` of the available height.
+There's a new property named `height` (available values: `collapsed` and `expanded`) that will allow consumers to control the vertical space usage of the component.
+When using the `expanded` value, it will force the component main container to use `100%` of the available height.
--- a/packages/components/spacings/spacings-inset-squish/README.md
+++ b/packages/components/spacings/spacings-inset-squish/README.md
@@ -36,11 +36,11 @@ import Spacings from '@commercetools-uikit/spacings';
 
 ## Properties
 
-| Props      | Type             | Required | Values                       | Default       |
-| ---------- | ---------------- | :------: | ---------------------------- | ------------- |
-| `scale`    | `String`         |    -     | `['s', 'm', 'l']`            | `m`           |
-| `height`   | `String`         |    -     | `min-content`, `max-content` | `min-content` |
-| `children` | `PropTypes.node` |    -     | -                            | -             |
+| Props      | Type             | Required | Values            | Default |
+| ---------- | ---------------- | :------: | ----------------- | ------- |
+| `scale`    | `String`         |    -     | `['s', 'm', 'l']` | `m`     |
+| `height`   | `String`         |    -     | `auto`, `100%`    | `auto`  |
+| `children` | `PropTypes.node` |    -     | -                 | -       |
 
 ## Scales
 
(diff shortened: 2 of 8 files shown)


packages/components/spacings/spacings-inset-squish/src/inset-squish.tsx
```diff
@@ -26,16 +27,18 @@ const InsetSquish = (props: TInsetSquishProps) => (
   <div
     css={css`
       padding: ${getPadding(props.scale)};
-      height: 100%;
+      height: ${props.useAllAvailableHeight ? '100%' : 'auto'};
```
tdeekens, line 30: 💯

packages/components/spacings/spacings-inset-squish/README.md
```diff
@@ -36,10 +36,11 @@ import Spacings from '@commercetools-uikit/spacings';
...
 ## Properties
 
-| Props      | Type             | Required | Values            | Default |
-| ---------- | ---------------- | :------: | ----------------- | ------- |
-| `scale`    | `String`         |    -     | `['s', 'm', 'l']` | `m`     |
-| `children` | `PropTypes.node` |    -     | -                 | -       |
+| Props                   | Type             | Required | Values            | Default |
+| ----------------------- | ---------------- | :------: | ----------------- | ------- |
+| `scale`                 | `String`         |    -     | `['s', 'm', 'l']` | `m`     |
+| `useAllAvailableHeight` | `boolean`        |    -     | `true`, `false`   | `false` |
```
tdeekens, line 42: Nice. This is very explicit and nicely constrained as an API. Did you think about going with e.g. `height` being a enum of `expand` or `collapse` to avoid boolean props and be more open towards future changes?

CarlosCortizasCT: Yes, I was hesitant about using a `boolean` here, but I didn't think we would need to have more values for the new property. However, I like the property name better so let me quickly change it.

tdeekens: Totally agree. I also don't see us adding more values (famous last works 😉). I like that booleans are slightly shorter (no value in positive case) also. Thanks for taking on the improvement to the same though. Maybe it's clearer on that front and hence over all better.

CarlosCortizasCT: Updated here: ce034c3fba22c83026e6dc72f71a21e05df063f8

.changeset/stale-radios-punch.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-uikit/spacings-inset-squish': minor
+'@commercetools-uikit/spacings-inset': minor
+---
+
+Refactor vertical space usage so consumers can control it.
+
+There's a new property named `useAllAvailableHeight` that will force the component main container to use `100%` of the available height.
```
tdeekens, line 8: 👋🏼 this should change 🤗

CarlosCortizasCT: Thanks 👍 

Updated here: fd1624cf18d6da2cce7ddc72eedee273311450f7

tdeekens (approved):
💯 💘

packages/components/spacings/spacings-inset/src/inset.tsx
```diff
@@ -24,22 +24,24 @@ const getPadding = (scale?: TScale) => {
 
 export type TInsetProps = {
   scale: TScale;
+  height: 'collapsed' | 'expanded';
```
emmenko, line 27: Nit: sorry to be picky but I feel like the naming is a bit confusing 🤔

On one hand we don't accept any "normal" CSS values for `height` but the prop is named `height`.
On the other hand, it seems that `collapsed` is not really used and the prop is meant to be more of a boolean.

Could you clarify these points and the reasoning behind the naming?

emmenko (commented):
Thanks for the improvements!

Apologies for the naming questions, I might be missing some background...

packages/components/spacings/spacings-inset/src/inset.tsx
tdeekens: Please make a counter suggestion as a point for discussion. I feel this is the best current naming we have. The prior boolean prop was abandoned to avoid "boolean prop hell" and to have an enum for scalability.

emmenko: I'm just trying to understand the background 🙂

On the top of my head I can think of `heightMode` or something similar.

tdeekens: What about `verticalSpacing` or `verticalSpace` with `collapse` and `expand`? I like `height` more than `*Mode` but a careful of the naming rabbit whole. @CarlosCortizasCT you can choose. I am happy to make an adjustment so you don't have to again 🤗

CarlosCortizasCT: What if we keep the `height` property but we use known CSS values?
* [max-content](https://developer.mozilla.org/en-US/docs/Web/CSS/height#max-content): The intrinsic preferred height.
* [min-content](https://developer.mozilla.org/en-US/docs/Web/CSS/height#min-content): The intrinsic minimum height.

tdeekens: Good idea. I like that. ➕➕

CarlosCortizasCT: Updated here: c0256a1628f3319d03fd9a5d59307b7b62f0042e

packages/components/spacings/spacings-inset/src/inset.tsx
```diff
@@ -24,22 +24,24 @@ const getPadding = (scale?: TScale) => {
...
+  height: 'min-content' | 'max-content';
   children?: ReactNode;
 };
 
 const Inset = (props: TInsetProps) => (
   <div
     css={css`
       padding: ${getPadding(props.scale)};
-      height: 100%;
+      height: ${props.height === 'max-content' ? '100%' : 'auto'};
```
emmenko, line 35: Ah I thought we will be using the actual value `max-content` or `min-content` instead of `100%` or `auto` 😅

Maybe we can use `100%` and `auto` as the acceptable prop values?

emmenko (approved):
Thanks and sorry for the back and forth! ❤️

packages/components/spacings/spacings-inset/src/inset.tsx
CarlosCortizasCT: Updated here: 374ddc4eaf315cc80fc71b34e06eee4d560f1035

Please let me know if this is enough for you or if there's anything else you want me to update.

Issue #2633, opened 2023-10-24
Support for iconLeft on Search Select Input


ismaelocaramelo:
**Is your feature request related to a problem? Please describe.**
The search select input does not have support for icon left as the rest of the selectors. To keep consistency with the rest of the selectors we have in place.

**Describe the solution you'd like**
Add the support for leftIcon property

<img width="452" alt="Screenshot 2023-10-23 at 13 01 37" src="https://github.com/commercetools/ui-kit/assets/21059277/9d9b2ee1-be82-4f7d-ab29-1eedb3b195fb">

References:
[Async select](https://uikit.commercetools.com/?path=/story/components-inputs-selectinputs--asyncselectinput)
[Select input](https://uikit.commercetools.com/?path=/story/components-inputs-selectinputs--selectinput)

chloe0592:
Thank you @ismaelocaramelo for creating this issue!
I just wanted to check first with @FilPob if there could be some potential design concerns about this before we implement it in the code?

FilPob:
Hey together, thanks for the contribution @ismaelocaramelo . I dont see any concerns for the addition from the design perspective :)

PR #2636, opened 2023-11-03
Support for iconLeft in Search Select Input

chloe0592:
We aim to enable the selection of the `iconLeft` for the `SearchSelectInput`, just like other inputs already allow.

This PR closes - https://github.com/commercetools/ui-kit/issues/2633

Screenshot 2023-11-03 at 15 55 59


.changeset/loud-paws-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/search-select-input': patch
```
tdeekens, line 2: I think it's `minor` if we want to follow Semantic Versioning as it's a feature.

chloe0592:
> I think we should also include the new property in the [SearchSelectField](https://ui-n28yshqbg-commercetools.vercel.app/?path=/story/components-fields-selectfields--searchselectfield) since its just a field wrapper around the SearchSelectInput

@CarlosCortizasCT good point! Thank you 🙏  [60739d7](https://github.com/commercetools/ui-kit/pull/2636/commits/60739d7ca2247f6617766ebfc305c5c2a100bb57)

.changeset/loud-paws-impress.md
chloe0592: Yes, makes more sense. Thanks for the heads-up! [60739d7](https://github.com/commercetools/ui-kit/pull/2636/commits/60739d7ca2247f6617766ebfc305c5c2a100bb57)

CarlosCortizasCT (approved):
Many thanks! 🙇

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -258,6 +258,10 @@ export type TSearchSelectInputProps = {
    * The style of the an option in the dropdown menu. It could be single lined option or an option with more and custom info
    */
   optionType?: 'single-property' | 'double-property' | 'multiple-properties';
+  /**
+   * Icon to display on the left of the placeholder text and selected value. Has no effect when `isMulti` is enabled.
```
emmenko, line 262: Is the `Has no effect when `isMulti` is enabled.` expected? 🤔

```suggestion
   * Icon to display on the left of the placeholder text and selected value.
```

emmenko (commented):
🤗

packages/components/inputs/search-select-input/src/search-select-input.tsx
chloe0592: @emmenko I would expect so, since all other `selectInputs` and `selectFields` behave the same way. So if the input has set `isMulti` to `true`, then we hide the `leftIcon`.

FilPob: oh yeah good catch, let's follow the same approach as we do in other select inputs

emmenko: Alright, thanks of the clarification 🤗

emmenko (approved):

PR #2638, opened 2023-11-07
refactor: height prop values to be descriptive

tdeekens:
Follow up of #2635 

As per discussion we move back to names which are more descriptive.

### DIFF
--- a/packages/components/card/src/card.visualroute.jsx
+++ b/packages/components/card/src/card.visualroute.jsx
@@ -110,12 +110,8 @@ export const component = () => (
             height: 100%;
           `}
         >
-          <div>
-            {text}
-          </div>
-          <div>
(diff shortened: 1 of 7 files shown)


packages/components/card/src/card.visualroute.jsx
```diff
@@ -110,12 +110,8 @@ export const component = () => (
             height: 100%;
           `}
         >
-          <div>
-            {text}
-          </div>
-          <div>
-            {text}
-          </div>
+          <div>{text}</div>
```
tdeekens, line 113: Prettier.

CarlosCortizasCT (approved):
Many thanks

emmenko (approved):
Thanks! ❤️

PR #2639, opened 2023-11-07
feat: add menuIsOpen prop to select inputs

LucasGrahn:
#### Summary

Add `menuIsOpen` property from [react-select library ](https://react-select.com/props#:~:text=number-,menuIsOpen,-Whether%20the%20menu)

## Description

I want in future to be able to keep an select input open for a case where an info icon inside a option group is clicked which opens up an info modal. With the current setup the select closes after a modal opens, so the user has to click the select again. That is bad UX and should be avoidable.

This implementation respects the `isReadOnly` parameter which also utilizes the native `menuIsOpen` property.

### Affected components
- SearchSelectInput 
- SelectInput 
- AsyncSelectInput


packages/components/inputs/select-input/src/select-input.spec.js
```diff
@@ -77,6 +77,26 @@ it('should have focus automatically when isAutofocussed is passed', () => {
...
+it('should have a open menu if menuIsOpen is true', () => {
+  const { getByLabelText, getByText } = renderInput({
+    menuIsOpen: true,
+  });
+  const input = getByLabelText('Fruit');
+
+  expect(getByText('Mango')).toBeInTheDocument();
+});
+
+it('should not have a open menu if menuIsOpen is true and isReadOnly is true', () => {
```
Rhotimee, line 89: NIT 
```suggestion
it('should not have an open menu if menuIsOpen is true and isReadOnly is true', () => {
```

This also applies to the other places in the code.

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -403,7 +409,13 @@ const SelectInput = (props: TSelectInputProps) => {
...
             } as ReactSelectProps['components']
           }
-          menuIsOpen={props.isReadOnly ? false : undefined}
+          menuIsOpen={
+            props?.menuIsOpen && !props.isReadOnly
+              ? props.menuIsOpen
+              : props.isReadOnly
+              ? false
+              : undefined
+          }
```
Rhotimee, line 418: NIT: Here's probably a cleaner version:
```suggestion
          menuIsOpen={
            props.isReadOnly ? false : !!props.menuIsOpen
          }
```

Rhotimee (commented):
Thanks for the PR @LucasGrahn. 
Please, Include a changeset by running `yarn changeset`

packages/components/inputs/select-input/src/select-input.spec.js
LucasGrahn: Good catch

packages/components/inputs/select-input/src/select-input.tsx
LucasGrahn: Looks way cleaner, will change that

LucasGrahn: I tried that and it breaks the behaviour as the undefined state is missing

Rhotimee: Actually, something like below should also work. 

`props.menuIsOpen && !props.isReadOnly`

.changeset/polite-bags-collect.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/search-select-input': major
+'@commercetools-uikit/async-select-input': major
+'@commercetools-uikit/select-input': major
+---
+
+Added menuIsOpen property
```
Rhotimee, line 7:
```suggestion
'@commercetools-uikit/search-select-input': minor
'@commercetools-uikit/async-select-input': minor
'@commercetools-uikit/select-input': minor
---

Added menuIsOpen property to SelectInputs
```

CarlosCortizasCT:
Hi @LucasGrahn 

> I want in future to be able to keep an select input open for a case where an info icon inside a option group is clicked which opens up an info modal. 

Would the current `closeMenuOnSelect` property would help with your use case?

.changeset/polite-bags-collect.md
CarlosCortizasCT: Maybe we can also explain about the new property. Perhaps something like:
```
We introduced a new (_boolean_) property (**menuIsOpen**) which allows to render the input with the dropdown initially unfolded.
```

packages/components/inputs/async-select-input/src/async-select-input.tsx
```diff
@@ -370,7 +377,13 @@ const AsyncSelectInput = (props: TAsyncSelectInputProps) => {
...
             } as ReactSelectAsyncProps['components']
           }
-          menuIsOpen={props.isReadOnly ? false : undefined}
+          menuIsOpen={
+            props?.menuIsOpen && !props.isReadOnly
+              ? props.menuIsOpen
+              : props.isReadOnly
+              ? false
+              : undefined
+          }
```
CarlosCortizasCT, line 386: I would suggest a simpler condition:

```suggestion
          menuIsOpen={
            props.isReadOnly ? false : props.menuIsOpen
          }
```

packages/components/inputs/search-select-input/src/search-select-input.tsx
```diff
@@ -164,6 +164,13 @@ export type TSearchSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   noOptionsMessage?: ReactSelectAsyncProps['noOptionsMessage'];
+  // menuIsOpen: PropTypes.bool,
```
CarlosCortizasCT, line 167: Can we maybe remove this commented line? 🙏

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -207,6 +207,13 @@ export type TSelectInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   isSearchable?: ReactSelectProps['isSearchable'];
+  // menuIsOpen: PropTypes.bool,
```
CarlosCortizasCT, line 210: Can we maybe remove this commented line? 🙏

packages/components/inputs/select-input/src/select-input.tsx
CarlosCortizasCT: Just wanted to mention that boolean casting is not needed:
```jsx
menuIsOpen={props.isReadOnly ? false : props.menuIsOpen}
```

Rhotimee: yeah. it's not. the second condition is even better. 

`props.menuIsOpen && !props.isReadOnly`

LucasGrahn:
> Hi @LucasGrahn
> 
> > I want in future to be able to keep an select input open for a case where an info icon inside a option group is clicked which opens up an info modal.
> 
> Would the current `closeMenuOnSelect` property would help with your use case?

Hey @CarlosCortizasCT , no sadly not, as I use an disabled option with an icon button, so basically nothing is selected. But the modal which appears after clicking the icon button takes the focus and the select closes. Without the modal the select stays open when no follow up modal is shown. This is fairly said just an convenience prop for edge cases.

.changeset/kind-moons-drive.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/search-select-input': major
+'@commercetools-uikit/async-select-input': major
+'@commercetools-uikit/select-input': major
+---
+
+We introduced a new (_boolean_) property (**menuIsOpen**) to Select Inputs which allows to render the input with the dropdown initially unfolded.
```
CarlosCortizasCT, line 7: We now have a duplicated changeset.

We should keep just one and change the versions to `minor`. (`major` is only for breaking changes).

CarlosCortizasCT:
> > Hi @LucasGrahn
> > > I want in future to be able to keep an select input open for a case where an info icon inside a option group is clicked which opens up an info modal.
> > 
> > 
> > Would the current `closeMenuOnSelect` property would help with your use case?
> 
> Hey @CarlosCortizasCT , no sadly not, as I use an disabled option with an icon button, so basically nothing is selected. But the modal which appears after clicking the icon button takes the focus and the select closes. Without the modal the select stays open when no follow up modal is shown. This is fairly said just an convenience prop for edge cases.

Sorry @LucasGrahn but I don't understand the use case.
Perhaps we can have a meeting as a follow up so we can really get it and think about what we can do about it from `ui-kit`.

.changeset/polite-bags-collect.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/search-select-input': minor
+'@commercetools-uikit/async-select-input': minor
+'@commercetools-uikit/select-input': minor
+---
+
+Added menuIsOpen property to SelectInputs
```
CarlosCortizasCT, line 7: I guess it would help if we explain a little bit the new property:

```suggestion
We introduced a new (_boolean_) property (**menuIsOpen**) to Select Inputs which allows to render the input with the dropdown initially unfolded.
```

LucasGrahn: I would describe it more as a way to enforce the dropdown to be opened. As you can open it programatically even after initialization. Wdyt?

Rhotimee (approved):
Thanks for working on this @LucasGrahn 🙌🏽

CarlosCortizasCT (approved):
Thanks a lot for your contribution, Lucas! 💯 🙇

PR #2645, opened 2023-11-14
Decrease default font size for `Tooltip` component

CarlosCortizasCT:
#### Summary

Decrease default font size for `Tooltip` component.

## Description

We're changing the default font size for the tooltip text from `0.875rem` to `0.75rem`. 

Bear in mind that consumers can customize Tooltip contents, so this change only applies when they don't customize them.

### DIFF
--- a/.changeset/cool-countries-brake.md
+++ b/.changeset/cool-countries-brake.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/tooltip': minor
 ---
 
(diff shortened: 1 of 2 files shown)


Rhotimee (approved):
🙌🏽

chloe0592 (approved):
Thank you 🤗

FilPob (approved):

FilPob:
Thanks :)

PR #2648, opened 2023-11-22
`MenuIsOpen` prop in Storybook

chloe0592:
There was a [PR](https://github.com/commercetools/ui-kit/pull/2639) few days ago where MenuIsOpen prop was introduce in `AsyncSelectInput`, `SearchSelectInput` and `SelectInput`. 

Currently, there's a possible undesired behavior in Storybook when we use the `MenuIsOpen` as a knob. Although we can set it to a boolean value, it restricts further interaction with the selector.

Underlying, `react-select` offers three potential values for `MenuIsOpen`:

- `true`: The menu is consistently open, no matter the interaction.
- `false`: The menu remains closed at all times, no matter the interaction.
- `undefined`: The menu's state depends on the interaction, open or closed.

Since we pass a boolean value in Storybook, we only have a selector that is either always open or always closed.

One consideration is to eliminate the prop entirely from Storybook. Since the usage of the knob is not very straightforward, it creates a unwanted behaviour and it is written in the README as an available prop, which will ensure that users know it exists.

Another option involves presenting the knob in Storybook as a `select` (with options: true, false, undefined) rather than a `boolean`. However, this might introduce confusion regarding the functionality of undefined.

I welcome any additional suggestions or ideas!

https://github.com/commercetools/ui-kit/assets/52276952/bf0dc1f3-2f13-4a8f-84af-27a37cd87073


CarlosCortizasCT (commented):
Fine from my side 👍

kark (approved):
That's fine by me as well, however, we have a case (`menuPortalTarget` prop knob) where we do the other thing you suggested

PR #2649, opened 2023-11-23
Update `preconstruct` dependency

CarlosCortizasCT:
## Summary

This PR updates the `preconstruct` dependency version we're using from `2.6.4` to `2.8.1`.

_(this is a follow up of #2569)_ 

## Description

There's a problem when updating the dependency which makes no sense to me, but it is this one:

Screenshot 2023-11-23 at 15 37 05

```
🎁 error Generating TypeScript declarations for packages/components/inputs/money-input/src/currencies.json failed:
🎁 error
🎁 info If you want to learn more about the above error, check https://preconstruct.tools/errors
🎁 info If the error is not there and you want to learn more about it, open an issue at https://github.com/preconstruct/preconstruct/issues/new
```

I've been trying to find information on how to prevent the error while generating the TS declarations but found nothing helpful.

My proposal is to change that file to be a TS file exporting the currencies list. It solves the issue but I'm not properly addressing the error.


packages/components/inputs/money-input/src/currencies.ts
```diff
@@ -0,0 +1,224 @@
...
+  YER: { fractionDigits: 2 },
+  YUM: { fractionDigits: 2 },
+  ZAR: { fractionDigits: 2 },
+  ZMK: { fractionDigits: 2 },
+  ZMW: { fractionDigits: 2 },
+  ZWD: { fractionDigits: 2 },
+  ZWL: { fractionDigits: 2 },
+  ZWN: { fractionDigits: 2 },
+  ZWR: { fractionDigits: 2 },
+};
```
kark, line 222: nit:
```suggestion
} as const;
```

kark (approved):
Thank you 🙇‍♂️
Just a reminder about pushing the changeset before merging

CarlosCortizasCT:
> Thank you 🙇‍♂️ Just a reminder about pushing the changeset before merging

Thanks for the reminder, Kacper 🙇 

Changeset added here: b20ea009d6cb116a0a382065867f5afba3f9dd25

emmenko (approved):
Thanks 💯

PR #2654, opened 2023-11-28
Set min-width property to `SelectInput`

chloe0592:
When the `horizontalConstraint` is configured as `auto` and the `selectInput` menu is narrow (e.g.locale selector), it appears somewhat awkward. 
Therefore, we have implemented a minimum width to address this issue.

**Before:**
Screenshot 2023-11-28 at 10 00 54

**After:**
Screenshot 2023-11-28 at 09 58 55


.changeset/seven-owls-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/select-utils': patch
+---
+
+Added min-width property to `SelectInput` component
```
CarlosCortizasCT, line 5: Sorry to be picky but maybe we can use another working as this might give the impression that we're adding a new property to the component when we're just adding a style rule.

Suggestion:

Updated `SelectInput` based components so the dropdown layer has a minimum width ([`constraint 3`](https://uikit.commercetools.com/?path=/story/components-constraints--horizontal))

chloe0592: @CarlosCortizasCT good point thanks! Updated here - [270db7d](https://github.com/commercetools/ui-kit/pull/2654/commits/270db7d7def8f7d5982c14fac2b4aa76861e5c52).

FilPob (approved):

CarlosCortizasCT (approved):
Thanks 👍

PR #2656, opened 2023-11-28
chore: add window eye icon

kark:


packages/components/icons/src/generated/CustomViewsSelectorReact.tsx
kark, line 1: Any suggestions for a more generic name for this icon?

emmenko: We usually name the icon to what it represents, not for what is meant to be used for.

In this case something like `window-eye.react.svg`

CarlosCortizasCT (approved):
Many thanks! 👍

emmenko (approved):

chloe0592 (approved):
Thank you!

PR #2658, opened 2023-12-01
Padding Quiet select input

chloe0592:
To more seamlessly integrate Quiet `selectInput` into sentences, which was a primary motivation for developing this component, we aim to modify the input paddings to `8px` instead of `16px`.

Screenshot 2023-11-28 at 13 57 15

-----------------------------------------
**16px:**

16px

-----------------------------------------
**8px:**
8px

### DIFF
--- a/.changeset/lemon-keys-double.md
+++ b/.changeset/lemon-keys-double.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/select-utils': patch
+'@commercetools-uikit/design-system': patch
+---
+
+Use smaller padding for `selectInput` component when the appearance is `quiet`.
--- a/packages/components/inputs/select-utils/src/create-select-styles.ts
+++ b/packages/components/inputs/select-utils/src/create-select-styles.ts
@@ -136,7 +136,10 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
       if (props.isReadOnly) return 'default';
       return 'pointer';
     })(),
-    padding: `0 ${designTokens.paddingForInput}`,
+    padding:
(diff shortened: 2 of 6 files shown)


packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -136,7 +136,10 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
       if (props.isReadOnly) return 'default';
       return 'pointer';
     })(),
-    padding: `0 ${designTokens.paddingForInput}`,
+    padding:
+      props.appearance === 'quiet'
+        ? `0 ${designTokens.paddingForInputAsQuiet}`
+        : `0 ${designTokens.paddingForInput}`,
```
kark, line 142: nit:
```suggestion
    padding: `0 ${
      props.appearance === 'quiet'
        ? designTokens.paddingForInputAsQuiet
        : designTokens.paddingForInput
    }`,
```

CarlosCortizasCT: nit: I personally think the current code version is easier to read 📖

.changeset/lemon-keys-double.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/select-utils': patch
+'@commercetools-uikit/design-system': patch
+---
+
+Use smaller padding for `selectInput` component when the appearance is `quiet`.
```
CarlosCortizasCT, line 6:
```suggestion
Use smaller padding for `SelectInput` component when the appearance is `quiet`.
```

Issue #2632, opened 2023-10-23
<DateRageInput> & <DateInput> -  Incorrect calendar display based off of region


valoriecarli:
[Initial notification](https://commercetools.slack.com/archives/CSL88HXQV/p1698023027316699)
[Official Support ticket
](https://commercetools.atlassian.net/browse/SUPPORT-23624)

**Describe the bug**

Many countries consider Monday to be the starting day of the calendar week. 
Once profiles are set with the preferred user settings, the days adjust to the local calendar, but the days remain static, causing them to be one off. 

**To Reproduce within an application **
1.) My Profile--> Personal Settings → Language & Region:  `English (United States) (en-US)`
2.) Go to any date range within the application. Here is an example using a Dashboard widget.
 
This display is **correct**
<img width="1878" alt="Screenshot 2023-10-23 at 12 11 47" src="https://github.com/commercetools/ui-kit/assets/47182700/6de4122f-bd98-49f6-a879-348ada3bafac">

1.) My Profile--> Personal Settings → Language & Region:  `English (United Kingdom) (en-GB)`
2.) Go to any date range within the application. 

This display is **incorrect**. October 23rd is a Monday, _not_ Tuesday. The calendar correctly begins with Monday, but the days themselves remain in the fixed ‘starting on Sunday’ position.
<img width="1834" alt="Screenshot 2023-10-23 at 12 12 48" src="https://github.com/commercetools/ui-kit/assets/47182700/fceddd04-f5f0-496e-9d49-5c56fe52d737">

With respect to the calendar select input showing the incorrect date when the week begins on a Monday, we have found that the issue is present in both the `DateInput` and `DateRangeInput` components residing in the commercetools-wide `ui-kit `library (as shown in the attached screen recording).
This was detected with GB, but this is a reoccurring theme for all countries with Monday-starting calendars.

**To reproduce within Storybook**

https://github.com/commercetools/ui-kit/assets/47182700/d30f46fa-d12c-4c4d-9240-85bbf3b19279

*******
**Additional Notes:**
The correct start day of the week based for the UK `en-GB`, Ireland `en-IE`, and Australia `en-AU` is Monday, according to the [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601#Adoption_as_national_standards) standard.

The US `en-US` and the fallback en locale both use Sunday as the first day of the week.

[...]

PR #2598, opened 2023-09-18
fix: DateInput when providing invalid calendar date input

kark:
#### Summary

Fixes [the reported issue](https://commercetools.atlassian.net/browse/SUPPORT-22938?atlOrigin=eyJpIjoiNDcwZTc0MWY5NzRjNDJlYjg5ZjQ1NDNmZjIyMGZkZTciLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

[Sentry error log](https://commercetools.sentry.io/issues/4477369881/events/3755b7e3471b478b88e632d15539ed52/?project=1549825&referrer=issue-list&statsPeriod=14d) 

The error:
Screenshot 2023-09-18 at 12 20 43

Please mind all contexts in which `getPaddingDayCount` helper function is used ([1](https://github.com/commercetools/ui-kit/blob/73bc8841022ea3b46dc9e2b75d3f9fd3e1d4fff6/packages/components/inputs/date-input/src/date-input.tsx#L253-L254), [2](https://github.com/commercetools/ui-kit/blob/73bc8841022ea3b46dc9e2b75d3f9fd3e1d4fff6/packages/components/inputs/date-range-input/src/date-range-input.tsx#L430-L434), [3](https://github.com/commercetools/ui-kit/blob/73bc8841022ea3b46dc9e2b75d3f9fd3e1d4fff6/packages/components/inputs/date-time-input/src/date-time-input.tsx#L373-L378)) - it's always passed to the Array function (constructor) call

PR #2669, opened 2023-12-15
fix(calendar-utils): `getPaddingDayCount` helper function

kark:
#### Summary
Closes https://github.com/commercetools/ui-kit/issues/2632
Fixes a bug introduced [here](https://github.com/commercetools/ui-kit/pull/2598)

### DIFF
--- a/packages/calendar-utils/src/calendar.ts
+++ b/packages/calendar-utils/src/calendar.ts
@@ -1,5 +1,4 @@
 import moment from 'moment';
-import inRange from 'lodash/inRange';
 import type { MomentInput, DurationInputArg1 } from 'moment';
 
 export const getDaysInMonth = (day: MomentInput) => moment(day).daysInMonth();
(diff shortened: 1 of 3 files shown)


packages/calendar-utils/src/calendar.ts
```diff
@@ -21,7 +20,7 @@ export const getPaddingDayCount = (day: MomentInput, locale: string) => {
   const paddingDayCount = (firstDayOfMonth - firstDayOfWeek + 7) % 7;
 
   // ensure number is always positive
```
CarlosCortizasCT, line 22: **nit**: now this message is a bit misleading

CarlosCortizasCT (approved):
Thanks 👍

packages/calendar-utils/src/calendar.ts
kark: Thank you for pointing it out, I think it's still true - we always return a number in range 0-6.

CarlosCortizasCT: Yes, you're right. 
What I was trying to highlight is that the comment, which refers to the next line, does not match the next instruction: we previously made sure that assumption with the call to `inRange(paddingDayCount, 0, 6)` but the the next call just makes sure `paddingDayCount` is a number, it does not validate what type of number is.

Anyway, this is just a tiny recommendation from my part.

kark: Ah, I understand now. Thanks for the explanation; it makes perfect sense to me!

kark:
I'd like to bring to your attention that I couldn't find any explanation of the change in visual regression tests for the `<Tag>` component. This is consistent across all new branches, and I haven't noticed any recent alterations in the baseline. AFAIK, the current modification appears to be correct. I'm just raising this for awareness.

PR #2672, opened 2023-12-20
Remove unwanted cursor css property

### LINKED ISSUE #2663: Tooltip wrapper uses `cursor: not-allowed`. Should we remove it?
**Describe the bug**

It appears that the `<Tooltip>` component uses `cursor: not-allowed` on the tooltip container.

https://github.com/commercetools/ui-kit/blob/e7981d960919a5de7e59470ae5ecc33d308c1e9b/packages/components/tooltip/src/tooltip.styles.ts#L100-L106

When using the tooltip for example around buttons, this is not noticeable because the button itself has it's own `cursor` style override. However, for normal elements like text the cursor shows the `not-allowed` symbol.

This was introduced in #1369, but I don't recall exactly the reasoning behind this.

**Expected behavior**

It seems the behaviour is a bit unexpected and we should probably remove the `cursor` style.

ddouglasz:
#### Summary
Remove unnecessary cursor: not-allowed property from tooltip container on the <Tooltip> component.

Closes #2663

### DIFF
--- a/.changeset/tricky-deers-sparkle.md
+++ b/.changeset/tricky-deers-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Remove unnecessary cursor css property from tooltip wrapper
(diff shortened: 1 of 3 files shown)


chloe0592 (approved):
Thank you @ddouglasz!

CarlosCortizasCT (approved):
Thanks! 🙇

PR #2677, opened 2023-12-29
fix(PasswordFiled): fix password readonly visibility

### LINKED ISSUE #2676: [fields/password-field] Disable "show" when isReadOnly props is true
**Description:**
If the show password was selected and then the PasswordField is set to readonly, the password remains visible, however, the "hide" button is not available to the end-user.

**Steps to reproduce the behavior:**

1. Enter some text into the PasswordField
2. Click "show" button
3. Add isReadOnly=true props to the PasswordField 
4. The PasswordField is now readonly, but the text is still visible, but the "hide" button is not displayed

**Expected behavior:**
For such a sensitive input as a PasswordField it seems wrong to have the password visible after the field is set to readonly. At least the user needs to have controls to hide the password even when the field itself is no longer editable.

**Additional context**

https://github.com/commercetools/ui-kit/assets/58188597/7e34d0a7-ba5e-4fd0-9caf-fb63f95553ed

serge-st:
#### Summary

This change can resolve #2676 by disabling password visibility every time the password field is set to readonly and or disabled

### DIFF
--- a/packages/components/fields/password-field/src/password-field.tsx
+++ b/packages/components/fields/password-field/src/password-field.tsx
@@ -176,6 +176,7 @@ const PasswordField = (props: TPasswordField) => {
   const [isPasswordVisible, togglePasswordVisibility] = useToggleState(false);
   const id = useFieldId(props.id, sequentialId);
   const hasError = props.touched && hasErrors(props.errors);
+  const canInteract = !props.isDisabled && !props.isReadOnly;
 
   if (!props.isReadOnly) {
     warning(
@@ -205,7 +206,7 @@ const PasswordField = (props: TPasswordField) => {
             onInfoButtonClick={props.onInfoButtonClick}
             hasRequiredIndicator={props.isRequired}
           />
-          {!props.isDisabled && !props.isReadOnly && (
+          {canInteract && (
             <FlatButton
               icon={isPasswordVisible ? <EyeCrossedIcon /> : <EyeIcon />}
               label={
@@ -228,7 +229,7 @@ const PasswordField = (props: TPasswordField) => {
           onBlur={props.onBlur}
           onFocus={props.onFocus}
           isAutofocussed={props.isAutofocussed}
-          isPasswordVisible={isPasswordVisible}
+          isPasswordVisible={canInteract ? isPasswordVisible : false}
           isDisabled={props.isDisabled}
           isReadOnly={props.isReadOnly}
           hasError={hasError}


ddouglasz:
Hello @serge-st 
Thank you for catching this and also providing a solution. Please kindly add a test and changeset, otherwise, the work is 👌🏽

serge-st:
Hello @ddouglasz!

Thanks for the prompt reply.

I added changeset and updated the tests in my latest commit.

packages/components/fields/password-field/src/password-field.spec.js
```diff
@@ -141,13 +141,27 @@ describe('when disabled', () => {
     const { getByLabelText } = renderPasswordField({ isDisabled: true });
     expect(getByLabelText('PasswordField')).toBeDisabled();
   });
+  it('should set the input type to password', () => {
+    const { container } = renderPasswordField({ isDisabled: true });
```
ddouglasz, line 145: We still need to give it some value, otherwise, the test will pass just in all cases since we are testing a particular scenario:
1. when there is value
2. when readOnly and disabled
3. when input type is text

```suggestion
    const { getByLabelText } = renderPasswordField({ value: 'foo', isDisabled: true });
```

packages/components/fields/password-field/src/password-field.spec.js
```diff
@@ -141,13 +141,27 @@ describe('when disabled', () => {
...
+    expect(container.querySelector('input')).toHaveAttribute(
```
ddouglasz, line 146:
```suggestion
   expect(getByLabelText('PasswordField')).toHaveAttribute(
```

packages/components/fields/password-field/src/password-field.spec.js
```diff
@@ -141,13 +141,27 @@ describe('when disabled', () => {
...
+  });
 });
 
 describe('when readOnly', () => {
   it('should disable the input', () => {
     const { getByLabelText } = renderPasswordField({ isReadOnly: true });
     expect(getByLabelText('PasswordField')).toHaveAttribute('readonly');
   });
+  it('should set the input type to password', () => {
+    const { container } = renderPasswordField({ isReadOnly: true });
```
ddouglasz, line 159:
```suggestion
    const { getByLabelText } = renderPasswordField({value: 'foo', isReadOnly: true });
```

packages/components/fields/password-field/src/password-field.spec.js
```diff
@@ -141,13 +141,27 @@ describe('when disabled', () => {
...
+    expect(container.querySelector('input')).toHaveAttribute(
```
ddouglasz, line 160:
```suggestion
    expect(getByLabelText('PasswordField')).toHaveAttribute(
```

ddouglasz (commented):
Thank you for the tests 🙌🏾. 
I just have some few observation to make it uniform with the other test.

serge-st:
@ddouglasz Hello!

Thank you for your patience.

I followed your suggestion and made the test more uniform.

I also covered:

1. scenario with/without value
2. scenario when both disabled and readOnly

about "when input type is text", since it is not possible to force input type via props, do you think it is needed to cover this as well, and how?

ddouglasz:
> @ddouglasz Hello!
> 
> Thank you for your patience.
> 
> I followed your suggestion and made the test more uniform.
> 
> I also covered:
> 
> 1. scenario with/without value
> 2. scenario when both disabled and readOnly
> 
> about "when input type is text", since it is not possible to force input type via props, do you think it is needed to cover this as well, and how?

Thank you for the changes. Honestly, I think these scenarios covered are enough.
Thank you for working on this 🙌🏾

packages/components/fields/password-field/src/password-field.spec.js
```diff
@@ -141,13 +141,64 @@ describe('when disabled', () => {
...
     expect(getByLabelText('PasswordField')).toBeDisabled();
   });
+  it('should set the input type to password', () => {
+    const { getByLabelText } = renderPasswordField({ isDisabled: true });
+    expect(getByLabelText('PasswordField')).toHaveAttribute(
+      'type',
+      'password'
+    );
+  });
+  descrbe('when has value', () => {
```
ddouglasz, line 151:
```suggestion
  describe('when has value', () => {
```

ddouglasz (commented):
@serge-st please we have some tiny last typos to fix 👌🏽

serge-st:
@ddouglasz my bad, I didn't setup the working environment correctly.

Now I ran the format and lint scripts, checks should pass.

Issue #2673, opened 2023-12-22
SelectableSearchInput: Allow choice for data-* props to be routed to search input or option dropdown


jocelyn-dunkley:
**Is your feature request related to a problem? Please describe.**
Recently we unmasked some components on the product list in FullStory in order to better track what users are doing. This does not touch real user data. It is recommended to use semantic selectors to select relevant component sections. However I ran into an issue with properly selecting the option dropdown value of SelectableSearchInput. All data-* props are automatically attached to the search input. While we can still get the necessary data with this method, it is visually confusing to attach a data prop with the option value to the search input. 

<img width="1383" alt="Screenshot 2023-12-22 at 11 31 33" src="https://github.com/commercetools/ui-kit/assets/115018694/d19fccbb-5911-43e3-8f20-788a6fe9ff5b">

**Describe the solution you'd like**
Allow an option for the data prop to be routed to the search input or the option dropdown

ddouglasz:
Hey @jocelyn-dunkley 
Thank you for pointing this out and suggesting a solution. We are currently taking a look at it.

ddouglasz:
Hello @jocelyn-dunkley 
Please I would like to confirm if these "data attributes" we want to make available are expected to be provided to each option individually or to all options?

jocelyn-dunkley:
@ddouglasz I think it should just have 1 data attribute attached to the dropdown component itself. On the product list side I am able to get what option was selected since we have a useState that keeps track of the "search mode" used, ie. search by all props, search only by product name, etc. We use this within the onChange function that is passed to SelectableSearchInput. So if the data-attribute is attached to the dropdown component itself I will have the up-to-date option that is selected at any given time. 

Let me know if that makes sense.

ddouglasz:
This makes sense. 
Thank you!

PR #2678, opened 2023-12-29
Data prop provision for select dropdown option

ddouglasz:
Fixes https://github.com/commercetools/ui-kit/issues/2673

#### Summary

SelectableSearchInput: Allow choice for data-* props to be routed to search input or option dropdown. At the moment, any selector passed as a prop only appears in the search input element tree and no way to semantically capture the dropdown options using a selector. 

This PR provides a prop to pass selectors to the dropdown option.


emmenko:
Thanks @ddouglasz for looking into this!

I have one main question first: from the issue request it is not clear to me if these "data attributes" are expected to be provided to each option individually or to all options.
Do you have more information about it or do we need to clarify this further?

Additionally, we might want to restrict the attributes that can be defined to only `data-` attributes, as otherwise any type of HTML attribute can be potentially provided. I don't think that's what we want here.
Btw, we have some utility function to do the filtering of `data-` attributes.

ddouglasz:
> Thanks @ddouglasz for looking into this!
> 
> I have one main question first: from the issue request it is not clear to me if these "data attributes" are expected to be provided to each option individually or to all options. Do you have more information about it or do we need to clarify this further?
> 
> Additionally, we might want to restrict the attributes that can be defined to only `data-` attributes, as otherwise any type of HTML attribute can be potentially provided. I don't think that's what we want here. Btw, we have some utility function to do the filtering of `data-` attributes.

Very important observations, I will get more clarity on these. 
Thank you 🙏🏾

ddouglasz:
> I have one main question first: from the issue request it is not clear to me if these "data attributes" are expected to be provided to each option individually or to all options. Do you have more information about it or do we need to clarify this further?

So I was able to get a [response](https://github.com/commercetools/ui-kit/issues/2673#issuecomment-1880712440) the requirement is to have 1 data attribute attached to the dropdown component itself.

ddouglasz:
@emmenko  Updated all feedback here: https://github.com/commercetools/ui-kit/pull/2678/commits/158e3f212ab003973a195a88f17a83bad8bd30fd

Thank you for the pointers

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -206,6 +206,10 @@ export type TSelectableSearchInputProps = {
    * Show submit button in the input
    */
   showSubmitButton?: boolean;
+  /**
+   * array of selector options that can be used to pass `data-*` props selectors to the options dropdown
+   */
+  optionSelector?: TOptionSelector[];
```
CarlosCortizasCT, line 212: I think the name of this prop might be misleading as you should be able to pass any `data-*` attribute and also would be difficult, from the consumer point of view, to understand how those attributes can be forwarded to the select or to the input.

My proposal would be to create two new props for this purpose:
* inputDataProps
* selectDataProps

We can also keep the raw `data-*` props we're currently receiving and forwarding to the `input` element for backwards compatibility.

This way, consumers who want to provide `data` attributes would need to explicitly indicate where they want them to be forwarded.

Example:
```ts
<SelectableSearchInput
  ...
  selectDataProps={{
    'tracking': 'products-search-type'
  }}
  inputDataProps={{
    'tracking': 'product-search-value' 
  }}
  ...
>
```

Then we would transform those values to HTML `data` attributes in the corresponding elements.

@emmenko would that work for you?

ddouglasz: This makes so much sense. I will refactor once we conclude on this approach

Thank you 🙏🏾

emmenko: Makes sense to me, thanks for the suggestion! 🤗

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2678/commits/79399cc59a7f211ce9b969875c4b515f8b6c24fe

Thank you

packages/components/inputs/selectable-search-input/docs/usage-example.js
```diff
@@ -15,6 +15,14 @@ const Example = () => (
       { value: 'one', label: 'One' },
       { value: 'two', label: 'Two' },
     ]}
+    selectDataProps={[
+      { 'data-select-1': 'value-1' },
+      { 'data-select-2': 'value-2' },
+    ]}
+    inputDataProps={[
+      { 'data-input-1': 'value-1' },
```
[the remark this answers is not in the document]
ddouglasz: Thank you for the feedback. So if I get this clearly, if a user passes in `inputDataProps={[{ 'prop-1': 'value-1' },]}` then we should internally convert it to something like `<input data-prop-1="value-1" />` ? so user does not necessarily need to add the `data-` prefix ?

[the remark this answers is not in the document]
CarlosCortizasCT: Yes, that's what I would expect.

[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2678/commits/7bdd6af91ee5c5052752dc28fb049b10585e2818

Thank you

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -239,6 +247,28 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
   const containerRef = useRef<HTMLDivElement>(null);
   const textInputRef = useRef<HTMLInputElement>(null);
 
+  const getDataProps = (data?: TDataProps[]) => {
+    const addPrefixToKeys = (obj: Record<string, string>) => {
```
ddouglasz, line 251: Thinking too make this a util since its used in 2 places 🤔

CarlosCortizasCT:
Thanks for all the work in this PR, @ddouglasz 🙇 

I think that maybe we weren't aligned on the path to take in terms of what I was suggesting so I created [this alternative PR](https://github.com/commercetools/ui-kit/pull/2684) where I tried to show an alternative that might help use to get to a common ground.

Please let me know your thoughts 🙏 

We can have a sync if you prefer to 👍

ddouglasz:
> Thanks for all the work in this PR, @ddouglasz 🙇
> 
> I think that maybe we weren't aligned on the path to take in terms of what I was suggesting so I created [this alternative PR](https://github.com/commercetools/ui-kit/pull/2684) where I tried to show an alternative that might help use to get to a common ground.
> 
> Please let me know your thoughts 🙏
> 
> We can have a sync if you prefer to 👍

I have gone through the work now, I see what you mean. It makes sense to me. 
Thank you very much 🙏🏾

chloe0592 (approved):
Makes sense to me. Thanks for the work and effort @ddouglasz 🙏

.changeset/spicy-books-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/selectable-search-input': patch
+---
+
+provide option to pass selector prop to the dropdown option in selectable search input.
```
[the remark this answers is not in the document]
ddouglasz: I will raise a small follow up PR to clean these up. 
Thank you

PR #2684, opened 2024-01-12
Data prop provision for select dropdown option [alternative proposal]

CarlosCortizasCT:
#### Summary

This PR implements a different approach than the one used in #2678 that I think could be simpler and fulfil the same requirement.

### DIFF
--- a/packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
+++ b/packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
@@ -22,7 +22,7 @@ import {
   getSelectableSearchInputContainerStyles,
   getSelectableSearchInputStyles,
 } from './selectable-search-input.styles';
-import SelectableSelect, { type TDataProps } from './selectable-select';
+import SelectableSelect from './selectable-select';
 import { useFieldId, useToggleState } from '@commercetools-uikit/hooks';
 import styled from '@emotion/styled';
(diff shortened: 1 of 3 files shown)


packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -241,33 +241,23 @@ const isOptionObject = (
   option: TOption | TOptionObject
 ): option is TOptionObject => (option as TOptionObject).options !== undefined;
 
+const transformDataProps = (dataProps?: Record<string, string>) =>
```
ddouglasz, line 244: This makes sense too. Thank you.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -460,7 +450,8 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
             aria-readonly={props.isReadOnly}
             contentEditable={!props.isReadOnly}
             css={getSelectableSearchInputStyles(props)}
-            {...getDataProps(props.inputDataProps)}
+            {...transformedInputDataProps}
+            {...legacyDataProps}
```
ddouglasz, line 454: This is clear. I see what you mean now 🙏🏾

ddouglasz (approved):

PR #2685, opened 2024-01-15
Add min width and max width prop to select input

ddouglasz:
#### Summary

Add min width and max width prop to select input


packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -22,6 +22,8 @@ type TProps = {
   hasValue?: boolean;
   controlShouldRenderValue?: boolean;
   appearance?: 'default' | 'quiet';
+  minWidth?: number | string;
+  maxWidth?: number | string;
```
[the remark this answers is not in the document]
ddouglasz: All updated here: https://github.com/commercetools/ui-kit/pull/2685/commits/09207ccb2542602c2b2edc325006c967523db4e7

Thank you.

packages/components/inputs/select-input/src/select-input.story.js
```diff
@@ -163,6 +163,8 @@ storiesOf('Components|Inputs/SelectInputs', module)
                   value={value}
                   showOptionGroupDivider={showOptionGroupDivider}
                   iconLeft={iconLeft ? createElement(iconLeft) : undefined}
+                  minWidth={number('minWidth', 120)}
```
chloe0592, line 166: I believe you missed to change the props name here to `minMenuWidth` and `maxMenuWidth`.

But other than that we wanted to have the `horizontalContraints` options for user to choose from to set the min and max width, instead of the pixels, if you remember yesterday's conversation?
But of course let me know if you want to sync some more about it.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -195,7 +197,10 @@ const menuStyles = (props: TProps) => (base: TBase) => {
       return base.borderColorForInput;
     })(),
     width: props.horizontalConstraint === 'auto' ? 'auto' : '100%',
-    minWidth: designTokens.constraint3,
+    minMenuWidth: props.minMenuWidth
```
chloe0592, line 200: It's a small typo here, should stay `minWidth` instead `minMenuWidth` 🤓

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -195,7 +197,10 @@ const menuStyles = (props: TProps) => (base: TBase) => {
...
+      ? props.minMenuWidth
+      : designTokens.constraint3,
+    maxMenuWidth: props.maxMenuWidth ?? props.maxMenuWidth,
```
chloe0592, line 203: Here as well - `maxWidth`.

packages/components/inputs/select-input/src/select-input.story.js
chloe0592: An example can be `FieldLabel` that uses `horizontalContraints` values for setting the available width of the component.

Screenshot 2024-01-16 at 14 10 57

ddouglasz: Oh! very true. 
Thanks for catching this.

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: All updated here: https://github.com/commercetools/ui-kit/pull/2685/commits/469dc51a1509da1549c73f2fb934cfe51103575c

Thank you.

FilPob (approved):

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -10,6 +10,26 @@
 import { ReactNode } from 'react';
 import { designTokens } from '@commercetools-uikit/design-system';
 
+export type THorizontalConstraint =
```
CarlosCortizasCT, line 13: Do we actually want to support all horizontal constraints for both `min` and `max` props?
I doesn't seem the `1` value, for instance to be allowed for the `max` prop.

.changeset/polite-colts-joke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/select-input': minor
+'@commercetools-uikit/select-utils': minor
+---
+
+provide min and max width props to the select input component
```
CarlosCortizasCT, line 6: We should enumerate the new props here and explain they control the `min` and `max` width of the select dropdown menu.

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -44,29 +44,33 @@ export type TCustomEvent = {
   persist: () => void;
 };
 
+export type THorizontalConstraint =
```
CarlosCortizasCT, line 47: Do we actually want to support all horizontal constraints for both min and max props?
I doesn't seem the 1 value, for instance to be allowed for the max prop.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -58,6 +64,17 @@ type TState = {
...
+type TDesignTokenName = keyof typeof designTokens;
+
+const getHorrizontalConstraintValue = (value?: THorizontalConstraint) => {
+  if (!value) return null;
+  const tokenName = `constraint${value}`;
+  if (tokenName in designTokens) {
+    return designTokens[tokenName as TDesignTokenName];
+  }
+  return null;
+};
```
CarlosCortizasCT, line 76: If we return `null`, then we will have that as the value of a CSS property, which is invalid.

I would suggest to keep this simple and return a default value in case a null param is provided (which should not happen as it's checked in this function consumers).

```suggestion
const getHorizontalConstraintValue = (value?: THorizontalConstraint) => {
  return designTokens[`constraint${value}` as TDesignTokenName] || designTokens.constraint3;
};
```

packages/components/inputs/selectable-search-input/README.md
```diff
@@ -95,8 +89,8 @@ export default Example;
...
-| `inputDataProps`           | `Array: TDataProps[]`                                                                  |          |           | used to pass `data-*` props to the input element                                                                                                                                                                                                                                                                                                                                         |
+| `selectDataProps`          | `Record`                                                                               |          |           | used to pass `data-*` props to the select component                                                                                                                                                                                                                                                                                                                                      |
```
CarlosCortizasCT, line 92: Not sure why this type has changed and the previous one is more accurate.
Also, I think it would help if we include an example of the expected object in the prop description.

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: @FilPob, maybe we can have a discussion on a range of values we want for then min and max horizontal values.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -58,6 +92,15 @@ type TState = {
   isSelected?: boolean;
 };
 
+type TDesignTokenName = keyof typeof designTokens;
+
+const getHorizontalConstraintValue = (value?: string | number) => {
```
CarlosCortizasCT, line 97: I think the type of this value should be `TProps['minMenuWidth'] | TProps['maxMenuWidth']`, right?

.changeset/polite-colts-joke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/select-input': minor
+'@commercetools-uikit/select-utils': minor
+---
+
+provide min and max width props to the select input component. These props are a range of values from the set of horizontal constraints that control the min and max width of the select dropdown menu.
```
CarlosCortizasCT, line 6: I'm sorry but I still think we should name the new props here.

What about something like this?
```suggestion
We're introducing two new props (`minMenuWidth` and `maxMenuWidth`) with allow to control the minimum and maximum widths of the select dropdown menu.
```

packages/components/inputs/select-utils/src/create-select-styles.ts
CarlosCortizasCT: @ddouglasz has this already been addressed?

packages/components/inputs/select-input/src/select-input.tsx
CarlosCortizasCT: @ddouglasz has this already been addressed?

CarlosCortizasCT (approved):
Thanks! 👍

PR #2689, opened 2024-01-16
refactor(storybook): migrate to v7 - part 1

kark:
#### Summary

This pull request migrates the initial set of Storybook stories for the `Tooltip`, `ViewSwitcher`, `ErrorMessage`, and `WarningMessage` components

### DIFF
--- a/.storybook/helpers/horizontal-constraint-arg-type.ts
+++ b/.storybook/helpers/horizontal-constraint-arg-type.ts
@@ -9,7 +9,7 @@ type ExtractNumericValues<T> = T extends infer U
 type TMinMax = ExtractNumericValues<TMaxProp>;
 
 type THorizontalConstraintArgType = {
-  defaultValue: TMaxProp;
+  defaultValue?: TMaxProp;
   min?: TMinMax;
   max?: TMinMax;
 };
--- a/.storybook/helpers/horizontal-constraint-arg-type.ts
+++ b/.storybook/helpers/horizontal-constraint-arg-type.ts
@@ -9,17 +9,13 @@ type ExtractNumericValues<T> = T extends infer U
 type TMinMax = ExtractNumericValues<TMaxProp>;
 
 type THorizontalConstraintArgType = {
-  defaultValue?: TMaxProp;
   min?: TMinMax;
   max?: TMinMax;
 };
 
-const horizontalConstraintArgType = (
-  { defaultValue, min, max }: THorizontalConstraintArgType = { defaultValue: 7 }
-) => ({
+const horizontalConstraintArgType = (args?: THorizontalConstraintArgType) => ({
   control: 'select',
-  options: Constraints.getAcceptedMaxPropValues(min, max),
-  defaultValue: defaultValue,
+  options: Constraints.getAcceptedMaxPropValues(args?.min, args?.max),
 });
 
 export default horizontalConstraintArgType;
--- a/.storybook/helpers/horizontal-constraint-arg-type.ts
+++ b/.storybook/helpers/horizontal-constraint-arg-type.ts
@@ -1,0 +1,25 @@
+import Constraints, { type TMaxProp } from '@commercetools-uikit/constraints';
+
+type ExtractNumericValues<T> = T extends infer U
+  ? U extends number
+    ? U
+    : never
+  : never;
+
+type TMinMax = ExtractNumericValues<TMaxProp>;
+
+type THorizontalConstraintArgType = {
+  defaultValue: TMaxProp;
+  min?: TMinMax;
+  max?: TMinMax;
+};
+
+const horizontalConstraintArgType = (
+  { defaultValue, min, max }: THorizontalConstraintArgType = { defaultValue: 7 }
+) => ({
+  control: 'select',
+  options: Constraints.getAcceptedMaxPropValues(min, max),
+  defaultValue: defaultValue,
+});
+
+export default horizontalConstraintArgType;
--- a/packages/components/messages/src/error-message/README.mdx
+++ b/packages/components/messages/src/error-message/README.mdx
@@ -1,0 +1,21 @@
+import { Canvas, Meta } from '@storybook/blocks';
+
+import * as ErrorMessageStories from './error-message.stories';
+
+<Meta of={ErrorMessageStories} />
+
+# Messages: ErrorMessage
+
+## Description
+
+Represents an error message.
+
+## Usage
+
+```js
+import { ErrorMessage } from '@commercetools-uikit/messages';
+
+<ErrorMessage>Something went wrong</ErrorMessage>;
+```
+
+<Canvas of={ErrorMessageStories.Default} />
--- a/packages/components/tooltip/src/tooltip.stories.tsx
+++ b/packages/components/tooltip/src/tooltip.stories.tsx
@@ -53,6 +53,7 @@ export const Default: Story = {
     fullWidth: false,
     customBodyWrapper: false,
     horizontalConstraint: 'scale',
+    placement: 'top',
   },
   argTypes: {
     placement: {
@@ -71,7 +72,6 @@ export const Default: Story = {
         'left-start',
         'left-end',
       ],
-      defaultValue: 'top',
     },
     horizontalConstraint: horizontalConstraintArgType(),
   },
--- a/packages/components/tooltip/src/tooltip.stories.tsx
+++ b/packages/components/tooltip/src/tooltip.stories.tsx
@@ -1,7 +1,7 @@
 import type { ComponentProps } from 'react';
 import styled from '@emotion/styled';
 import { css } from '@emotion/react';
-import Constraints from '@commercetools-uikit/constraints';
+import { horizontalConstraintArgType } from '@/storybook-helpers';
 import { PrimaryButton } from '@commercetools-uikit/buttons';
 import type { Meta, StoryObj } from '@storybook/react';
 import Tooltip from './tooltip';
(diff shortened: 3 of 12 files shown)


packages/components/messages/src/error-message/README.mdx
kark, line 1: The Readmes for ErrorMessage and WarningMessage have been split to accommodate the constraint of having only one `<Meta>`.

packages/components/view-switcher/src/view-switcher.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+  iconViewSwitcherButton4: string;
+  iconOnlyButton1: boolean;
+  iconOnlyButton2: boolean;
+  iconOnlyButton3: boolean;
+  iconOnlyButton4: boolean;
+  isDisabledButton1: boolean;
+  isDisabledButton2: boolean;
+  isDisabledButton3: boolean;
+  isDisabledButton4: boolean;
+};
```
kark, line 25: This adheres to the pattern outlined [here](https://storybook.js.org/docs/writing-stories/typescript#typing-custom-args)
To the best of my knowledge, there is no method for grouping storybook v7 controls in a manner similar to what we are achieving with storybook v5 knobs e.g. [here](https://github.com/commercetools/ui-kit/blob/699175cc0e8d5b1b11419bde7e8c8fb48c753d47/packages/components/view-switcher/src/view-switcher-with-icons.story.js#L45)

packages/components/tooltip/src/tooltip.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+        'left-start',
+        'left-end',
+      ],
+      defaultValue: 'top',
+    },
+    horizontalConstraint: {
+      control: 'select',
+      options: Constraints.getAcceptedMaxPropValues(),
+      defaultValue: 'scale',
+    },
```
CarlosCortizasCT, line 79: I know this is a small definition but I think we will be using it in many other components.
What do you think about creating a shared helper similar to what we have for icons?

packages/components/view-switcher/README.mdx
```diff
@@ -0,0 +1,115 @@
...
+export { UncontrolledExample, ControlledExample };
+```
+
+## Signatures
+
+### Signature `onChange`
+
+```ts
+(value: string) => void
+```
```
CarlosCortizasCT, line 109: We should avoid including the signatures, as that's part of the types table docs which is automatically included in the docs page.

packages/components/view-switcher/src/view-switcher.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+    children: {
+      table: {
+        disable: true,
+      },
+    },
+    defaultSelected: {
+      table: {
+        disable: true,
+      },
+    },
```
CarlosCortizasCT, line 64: These two are repeated in both stories so we might want to move them to the meta definition.

packages/components/view-switcher/src/view-switcher.stories.tsx
CarlosCortizasCT: I think this is something we might try to improve after the migration.

packages/components/view-switcher/src/view-switcher.stories.tsx
CarlosCortizasCT: We may also consider using the [compact alternative](https://storybook.js.org/docs/essentials/controls#filtering-controls):
```jsx
parameters: {
  controls: { exclude: ['children', 'defaultSelected'] },
},
```

packages/components/view-switcher/src/view-switcher.stories.tsx
CarlosCortizasCT: Something that might help here is to visually group the controls related to the items in the props table using this config:
```js
argTypes: {
  iconViewSwitcherButton1: {
    table: {
      subcategory: 'Items controls',
    }
  }
}
```


image

kark: Thank you for researching it 🙏 I pushed an update 8f2315e1d21d7a91bf31ed46a7150477bbe97b4c

.storybook/helpers/horizontal-constraint-arg-type.ts
```diff
@@ -0,0 +1,25 @@
...
+  max?: TMinMax;
+};
+
+const horizontalConstraintArgType = (
+  { defaultValue, min, max }: THorizontalConstraintArgType = { defaultValue: 7 }
+) => ({
+  control: 'select',
+  options: Constraints.getAcceptedMaxPropValues(min, max),
+  defaultValue: defaultValue,
+});
```
kark, line 23: Setting the `defaultValue` like this doesn't seem to work
https://github.com/storybookjs/storybook/issues/21016

CarlosCortizasCT: But we don't need to provide the default value here, right.
To me this is more like the definition of the control and the default value should be passed as part of the args to the default story.
You can take a look at the `PrimaryButton` stories to see what I mean.

kark: Sure, no need to provide it here 👍. Just thought it would be a nice-to-have, considering that the API technically allows it. Additionally, for the `iconArgType` helper, making it a function with a default icon option could be a nice addition

CarlosCortizasCT: I get it, but I still think it would be beneficial to keep the default value out of the control definition.

I believe it's better to define the default values in the props of the default story so it's easier to reuse them in alternative variants (I saw this pattern in Storybook docs) or even in the args value section of the meta definition, as they serve that purpose.

kark: 👍 Removed 4c97db5de73c40558da473f4d905c5c01c36bd64

packages/components/view-switcher/src/view-switcher.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+          },
+          [`isDisabledButton${i}`]: {
+            table: {
+              subcategory: `View Switcher Button ${i}`,
+            },
+          },
+        },
+      };
+      return temp;
+    }, {}),
```
CarlosCortizasCT, line 89: Sometimes I think a little repetition makes code easier to understand and thus, to maintain.

What do you think about something like this instead?
```js
   const categorize = (index: number) => ({ table: { subcategory: `View Switcher Button ${i}` } });
  
  ...

    iconViewSwitcherButton1: categorize(1),
    iconOnlyButton1: categorize(1),
    isDisabledButton1: categorize(1),
    iconViewSwitcherButton2: categorize(2),
    iconOnlyButton2: categorize(2),
    isDisabledButton2: categorize(2),
    iconViewSwitcherButton3: categorize(3),
    iconOnlyButton3: categorize(3),
    isDisabledButton3: categorize(3),
    iconViewSwitcherButton4: categorize(4),
    iconOnlyButton4: categorize(4),
    isDisabledButton4: categorize(4),
```

packages/components/view-switcher/src/view-switcher.stories.tsx
CarlosCortizasCT: Do you finally prefer to keep it like this over the compact alternative?

packages/components/view-switcher/src/view-switcher.stories.tsx
kark: Agreed. Done in 12591b3b82a56d88b5c5c9023d165b005ac0f6fa

packages/components/view-switcher/src/view-switcher.stories.tsx
kark: Thanks, not really, I wanted to clean it up later. Anyways, I moved the implementation of `hideControls` from the other PR

packages/components/view-switcher/src/view-switcher.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+            value={`Button ${i}`}
+            icon={
+              args[
+                `iconViewSwitcherButton${i}` as keyof typeof args
+              ] as TViewSwitcherButtonProps['icon']
+            }
+          >
+            {!args[`iconOnlyButton${i}` as keyof typeof args]
+              ? `View ${i}`
+              : undefined}
```
CarlosCortizasCT, line 105: I was wondering if there could be a way for us to better type the `args` so we don't need the castings and I came up with [this alternative proposal](https://github.com/commercetools/ui-kit/pull/2715).

Please take a look when you have some time and let me know what you think about it. 🙏

packages/components/tooltip/src/tooltip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+type Story = StoryObj<TooltipPropsAndCustomArgs>;
+
+export const Default: Story = {
+  args: {
+    off: false,
+    title: 'Tooltip text.',
+    showAfter: 300,
+    closeAfter: 200,
+    children: (
```
CarlosCortizasCT, line 72: I think it would be easier to read if we keep the children in the `render` function and just add this and the `Tooltip` inner content.

CarlosCortizasCT (approved):
Thanks a lot for all the work and the discussions around this first chunk of stories migration 👏 👏 👏 👏 

I think we're setting some good best practices for the next migration work.

Super excited to see this moving forward 🔝

PR #2690, opened 2024-01-17
Add border bottom to select input group header menu

ddouglasz:
#### Summary

Add border bottom to select input group header menu

Screenshots:

image

Design: 
image


CarlosCortizasCT:
In general, it would help reviewers if we can provide a little more context about the change. For instance, its motivation and some instructions how to validate the changes (or even example screenshots).

I just wanted to double check if this is something we want for all use cases and if we are aligned with the design team.

Thanks.

CarlosCortizasCT:
Comparing the design screenshot with what the component actually renders, I can see the header text styles do not match.

Also, did we have any confirmation about applying this change for every select-type component?

Thanks.

ddouglasz:
> Comparing the design screenshot with what the component actually renders, I can see the header text styles do not match.
> 
> Also, did we have any confirmation about applying this change for every select-type component?
> 

The header text style only has one change: the font-weight is changed to 500. I had the same observation too but I confirmed the only change was the font-weight. 

The change should be for all other select-type.

.changeset/sour-donkeys-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/select-utils': minor
+---
+
+add border bottom to the select input group header menu
```
CarlosCortizasCT, line 5: Since we are updating both the border and the font weight, perhaps we can be more generic here:

```suggestion
Small styles adjustments to dropdown menu groups headers.
```

CarlosCortizasCT (approved):
Many thanks! 👏

.changeset/sour-donkeys-cough.md
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2690/commits/2e134ee172617b000a9bbca77df9a58706a2eb6c

Thank you.

chloe0592 (approved):
Thanks @ddouglasz! Looks good to me ✨

FilPob (approved):

PR #2691, opened 2024-01-17
refactor(storybook): migrate to v7 - part 2

kark:

### DIFF
--- a/packages/components/pagination/README.mdx
+++ b/packages/components/pagination/README.mdx
@@ -49,5 +49,3 @@ const Example = () => {
 
 export default Example;
 ```
-
-<Canvas of={PaginationStories.Default} />
--- a/packages/components/pagination/README.mdx
+++ b/packages/components/pagination/README.mdx
@@ -1,4 +1,4 @@
-import { Canvas, Meta } from '@storybook/blocks';
+import { Canvas, Meta, DocsContainer, Story } from '@storybook/blocks';
 
 import * as PaginationStories from './src/pagination.stories';
 
@@ -49,3 +49,7 @@ const Example = () => {
 
 export default Example;
 ```
+
+<Canvas>
+  <Story of={PaginationStories.Default} height="100vh" />
+</Canvas>
--- a/packages/components/pagination/README.mdx
+++ b/packages/components/pagination/README.mdx
@@ -1,0 +1,53 @@
+import { Canvas, Meta } from '@storybook/blocks';
+
+import * as PaginationStories from './src/pagination.stories';
+
+<Meta of={PaginationStories} />
+
+# Pagination
+
+## Description
+
+Components for navigating through pages of items
+
+## Installation
+
+```
+yarn add @commercetools-uikit/pagination
+```
+
+```
+npm --save install @commercetools-uikit/pagination
+```
+
+Additionally install the peer dependencies (if not present)
+
+```
+yarn add react react-intl
+```
+
+```
+npm --save install react react-intl
+```
+
+## Usage
+
+```jsx
+import { Pagination } from '@commercetools-uikit/pagination';
+
+const Example = () => {
+  const items = [{ id: '1' }, { id: '2' }];
+  return (
+    <Pagination
+      totalItems={items.length}
+      page={1}
+      onPageChange={() => {}}
+      onPerPageChange={() => {}}
+    />
+  );
+};
+
+export default Example;
+```
+
+<Canvas of={PaginationStories.Default} />
--- a/packages/components/pagination/src/page-navigator/page-navigator.stories.tsx
+++ b/packages/components/pagination/src/page-navigator/page-navigator.stories.tsx
@@ -1,4 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react';
+import { Value } from 'react-value';
 import PageNavigator from './page-navigator';
 
 const meta = {
@@ -12,8 +13,21 @@ type Story = StoryObj<typeof meta>;
 
 export const Default: Story = {
   args: {
-    totalPages: 3,
+    totalPages: 10,
     page: 1,
-    onPageChange: () => {},
+  },
+  render: (args) => {
+    return (
+      <Value
+        defaultValue={args.page}
+        render={(page, onPageChange) => (
+          <PageNavigator
+            totalPages={args.totalPages}
+            page={page}
+            onPageChange={onPageChange}
+          />
+        )}
+      />
+    );
   },
 };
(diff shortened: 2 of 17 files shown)


packages/components/pagination/src/page-navigator/page-navigator.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    totalPages: 10,
+    page: 1,
+  },
+  render: (args) => {
```
CarlosCortizasCT, line 19: Could we maybe use the helper here so we keep a compact code?

```ts
render: withControlledValue<TPageNavigatorProps>(PageNavigator)
```

packages/components/pagination/src/page-size-selector/page-size-selector.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+const getMinimumPageSizeFromRange = (selectedRange: TPageRangeSize) => {
+  switch (selectedRange) {
+    case 'l':
+      return 200;
+    default:
+      return 20;
+  }
+};
+
+const meta = {
```
CarlosCortizasCT, line 13: This story is not working correctly locally for me.
I'd like to have a sync about it. 🙏

packages/components/primary-action-dropdown/src/primary-action-dropdown.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+  },
+} satisfies Meta<PrimaryActionDropdownPropsAndCustomArgs>;
+
+export default meta;
+type Story = StoryObj<PrimaryActionDropdownPropsAndCustomArgs>;
+
+export const Default: Story = {
+  args: {
+    isDisabledOption1: false,
```
CarlosCortizasCT, line 37: We can also group this one to differentiate them from the actual component props.

packages/components/stamp/src/stamp.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+} satisfies Meta<typeof Stamp>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    isCondensed: false,
+  },
+  parameters: { controls: { exclude: ['tone', 'children', 'icon', 'label'] } },
```
CarlosCortizasCT, line 25: If we exclude these props they won't be displayed in the props table so users won't know about them, right?

packages/components/pagination/src/page-size-selector/page-size-selector.stories.tsx
kark: Ok, thank you, I'll schedule a meeting. I just wonder, it any different from https://uikit.commercetools.com/?path=/story/components-pagination--pagesizeselector?

packages/components/pagination/src/page-navigator/page-navigator.stories.tsx
kark: I pushed a refactor of the HOC to be able to use it here 7d8f97d48da932198d4a3145e874dd503250a9b1

packages/components/stamp/src/stamp.stories.tsx
kark: Well, that's true, but keeping in mind how this story works, exposing those controls and allowing users to tweak them might create the impression that something is broken, since changing the values won't have any effect. We can leave them out, but then the story must be changed

CarlosCortizasCT: I think we should include this topic in our upcoming meeting.

packages/components/pagination/src/page-size-selector/page-size-selector.stories.tsx
CarlosCortizasCT: Something I image is that using `inline: false` will make the controls to not work.
Perhaps we can wrap the story in a component which applies that height.

image

CarlosCortizasCT: I found out that maybe using the `height` property can work:
image

kark: Thank you, I updated the story to use this parameter. However, I noticed a down side of using it. It works well in the `Docs` page and the story page, but it doesn't work in the "README" page. That is why I removed the  `<Canvas of={PaginationStories.Default} />` from the pagination's readme.mdx.

packages/components/stamp/src/stamp.stories.tsx
kark: As discussed during our sync I used [this pattern](https://storybook.js.org/docs/essentials/controls#disable-controls-for-specific-properties) in the new `hideControls` helper. Now the controls are visible in the table but the value is not passed to the story

packages/components/pagination/src/page-size-selector/page-size-selector.stories.tsx
CarlosCortizasCT: According to [the docs](https://storybook.js.org/docs/api/doc-block-canvas#story), it seems we can adjust the height of the story the `Canvas` component renders.

Testing it locally, this example seems to work:
```jsx
<Canvas of={PaginationStories.Default} story={{ height: '560px' }}/>
```

image

The downside is that I had to use a high value for the height, but I think it's a fair compromise right now and maybe can take a deeper look in the future.

kark: Thanks. From my observations, the story within `<Canvas>` treats the viewport as its lower boundary. 
Whether using high or low values, the story may be perceived as not working as expected, depending on the current scroll position on the page.

CarlosCortizasCT: Maybe then we can just add a wrapper for the component in the story?

packages/components/primary-action-dropdown/src/primary-action-dropdown.stories.tsx
CarlosCortizasCT: Sometimes I think a little repetition makes code easier to understand and thus, to maintain.

What do you think about something like this instead?
```js
   const categorize = (index: number) => ({ table: { subcategory: `Option ${i}` } });
  
  ...

    isDisabledOption1: categorize(1),
    isDisabledOption2: categorize(2),
    isDisabledOption3: categorize(3),
    labelOption1: categorize(1),
    labelOption1: categorize(2),
    labelOption1: categorize(3),
```

kark: Sure, let's focus on getting the first PR merged and then I'll take care of the rest after rebasing.

packages/components/pagination/README.mdx
```diff
@@ -0,0 +1,55 @@
...
+    />
+  );
+};
+
+export default Example;
+```
+
+<Canvas>
+  <Story of={PaginationStories.Default} height="100vh" />
+</Canvas>
```
kark, line 55: This was the only way I could figure out to make this story work on the docs page.

PR #2693, opened 2024-01-19
Fix styled-components integration in new Storybook

CarlosCortizasCT:
#### Summary

Fix styled-components integration in new Storybook

## Description

We currently have a problem when trying to render a story which includes a component using styled component [component selectors](https://styled-components.com/docs/advanced#referring-to-other-components):

```
Error: Component selectors can only be used in conjunction with @emotion/babel-plugin, the swc Emotion plugin, or another Emotion-aware compiler transform.
```

Screenshot 2024-01-19 at 12 25 26

After some research I found [here](https://github.com/storybookjs/builder-vite/issues/210) something helpful which states we need to remove the default Storybook react babel plugin and add it again include the `babel-emotion` plugin.

I also took the opportunity to update Storybook to the latest version.


kark (approved):
Thanks 🙌 Super cool you made it work!

Issue #2671, opened 2023-12-19
`hasWarning` prop is not supported in many`<*Input>` components


jaikumar-tj:
**Is your feature request related to a problem? Please describe.**
Not all the Field components support the orange border when we want to display warning

**Describe the solution you'd like**
Some components like `SelectField` accepts the `hasWarning` props but the other field components like `TextField, MoneyField, etc` don't have the same prop
<img width="959" alt="Screenshot 2023-12-19 at 16 05 42" src="https://github.com/commercetools/ui-kit/assets/83634106/45331e97-6428-405c-963e-c9c601159368">

Pacman Team and Context Team have an upcoming feature that requires this feature in order to be implemented it in the MC. Otherwise, we have to get rid of the `Field` component and use the `Input` & `Text` components (i.e, replacing `<TextField>` from product detail page and use `<TextInput>` and `<Text.Body>` to achieve the same result)  which seems like going against the idea of using the existing components.

@AnjaBuerger , @mariabarrena and @FilPob already discussed and agreed on this.

Ideally all the `<*Field>` components should have `hasWarning` feature
<img width="217" alt="Screenshot 2023-12-19 at 16 06 12" src="https://github.com/commercetools/ui-kit/assets/83634106/3ce1f7d1-98c4-4cce-88b8-b2c1b6018a0b">

**Note: We(@commercetools/pacman-team-fe) were originally planning to contribute to the UI-kit repo in the next Q during our feature development stage to enable this feature. If we get any early support from @commercetools/shield-team-fe or @commercetools/context-team-fe  that would be much appreciated as well**

jaikumar-tj:
Just for transparency, I'll write this here. 

It was decided that
- All `<*Input>` components should have `hasWarning` prop to display orange border
- All `<*Field>` component should have `warning` prop to display orange border + warning text below

@commercetools/pacman-team-fe will work on implementing this.

chloe0592:
Hi @jaikumar-tj, I see your team has started working on this issue. Will your team be handling it all the way through? Thanks!

obulaworld:
> Hi @jaikumar-tj, I see your team has started working on this issue. Will your team be handling it all the way through? Thanks!

Hi @chloe0592 Yes, the final PR for this is here: https://github.com/commercetools/ui-kit/pull/2698

jaikumar-tj:
Hi @chloe0592, Please let us know if we need to do anything else other than what Chisom commented. We are not sure about all the procedures.

[...]

PR #2698, opened 2024-01-22
FEAT: Added field warning component to all fields

obulaworld:
#### Summary

Update all Field component to handle warning texts.
It seeks to fully fix this UI-KIT [issue](https://github.com/commercetools/ui-kit/issues/2671)

## Description

This PR seeks to update all Field component to handle warning texts by:
- Adding the new `FieldWarnings` component to all Fields
- Updating all fields dependencies to include FieldWarnings package
- Updating all fields story, spec, visual route to reflect the new changes.

## Screenshots
Screenshot 2024-01-23 at 10 56 28
Screenshot 2024-01-23 at 10 56 38
Screenshot 2024-01-23 at 10 56 50


obulaworld:
I had to replace all select fields' hasWarning prop to only be true if and only if the field is touched and the newly added warnings object prop has at least one warning. Please let me know if this will affect any team that might have been using the hasWarning prop as a simple truthy or falsy prop.

.changeset/fair-cooks-press.md
```diff
@@ -0,0 +1,23 @@
...
+'@commercetools-uikit/select-field': minor
+'@commercetools-uikit/money-field': minor
+'@commercetools-uikit/radio-field': minor
+'@commercetools-uikit/collapsible-panel': minor
+'@commercetools-uikit/date-field': minor
+'@commercetools-uikit/text-field': minor
+'@commercetools-uikit/time-field': minor
+---
+
+Added field warning component to all fields
```
CarlosCortizasCT, line 23: Perhaps we can elaborate a little more this message so consumers get a better idea about this change:

```suggestion
We've included a new property in all `*Field` components so consumers can not only render errors (below the input) but also warnings.
The main difference between them is the color used for the text.

Please take a look at the RADME files of the fields components to learn how to use this new property.
```

packages/components/fields/async-creatable-select-field/README.md
```diff
@@ -58,6 +58,9 @@ export default Example;
...
+| `renderWarning`           | `Function`<br/>[See signature.](#signature-renderWarning)                                             |          |           | Called with custom warnings. This function can return a message which will be wrapped in an WarningMessage. It can also return null to show no warning.                                                                                                                                                                                                                                   |
+| `renderDefaultWarning`    | `Function`<br/>[See signature.](#signature-renderDefaultWarning)                                      |          |           | Called with default warnings. This function can return a message which will be wrapped in an WarningMessage. It can also return null to show no warning.                                                                                                                                                                                                                                  |
```
CarlosCortizasCT, line 63: I'm not sure this is a property we want to expose in the `Field` components.
I know we have this property in the `FieldWarnings` component (similar to what we have in the `FieldErrors` one), but I don't see we are using it for errors so I think we shouldn't neither for warnings.

@emmenko do you maybe remember why is this prop used for (in the `FieldErrors` component?
I don't find any component in this repository using the `FieldErrors` component and passing it the `renderDefaultWarning` prop.

emmenko: I have no idea what the `renderDefaultError` is meant for.

Maybe we should just deprecate it and remove it, the `renderError` should be good enough.

CarlosCortizasCT: Thanks @emmenko 

@obulaworld so in this PR we can at least remove the `renderDefaultWarning` from the `FieldWarnings` component (as we were not using it anywhere yet) and don't use it in the fields components.

Does it make sense to you?

obulaworld: Yes @CarlosCortizasCT It does. I will remove it.

CarlosCortizasCT:
It seems we're missing something in the `DateRageField` component as it's not rendering the orange border when it has a warning message.

image

obulaworld:
> It seems we're missing something in the `DateRageField` component as it's not rendering the orange border when it has a warning message.
> 
> image

@CarlosCortizasCT Fixed

packages/components/fields/date-field/src/date-field.visualroute.jsx
```diff
@@ -79,5 +79,26 @@ export const component = () => (
...
+      <DateField
+        title="Release Date"
+        horizontalConstraint={7}
+        value=""
+        onChange={() => {}}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 92:
```suggestion
    <Spec label="with warning when touched">
```

packages/components/fields/date-time-field/src/date-time-field.visualroute.jsx
```diff
@@ -87,5 +87,28 @@ export const component = () => (
...
+        timeZone="UTC"
+        title="Release Date"
+        horizontalConstraint={7}
+        value=""
+        onChange={() => {}}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 101:
```suggestion
    <Spec label="with warning when touched">
```

packages/components/fields/localized-multiline-text-field/src/localized-multiline-text-field.visualroute.jsx
```diff
@@ -127,5 +127,28 @@ export const component = () => (
...
+        title="Welcome Text"
+        value={value}
+        onChange={() => {}}
+        selectedLanguage="en"
+        horizontalConstraint={7}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 141:
```suggestion
    <Spec label="with warning when touched">
```

packages/components/fields/localized-text-field/src/localized-text-field.visualroute.jsx
```diff
@@ -103,5 +103,28 @@ export const component = () => (
...
+        title="Welcome Text"
+        value={value}
+        onChange={() => {}}
+        selectedLanguage="en"
+        horizontalConstraint={7}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 117:
```suggestion
    <Spec label="with warning when touched">
```

packages/components/fields/money-field/src/money-field.visualroute.jsx
```diff
@@ -117,5 +117,28 @@ export const component = () => (
...
+        title="Price"
+        horizontalConstraint={7}
+        value={emptyValue}
+        onChange={() => {}}
+        currencies={currencies}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 131:
```suggestion
    <Spec label="with warning when touched">
```

packages/components/fields/multiline-text-field/src/multiline-text-field.visualroute.jsx
```diff
@@ -72,6 +72,27 @@ export const component = () => (
...
+      <MultilineTextField
+        title="Welcome Text"
+        value={value}
+        onChange={() => {}}
+        horizontalConstraint={7}
+        warnings={{ customWarning: true }}
+        renderWarning={() => 'Custom warning'}
+      />
+    </Spec>
+    <Spec label="with error when touched">
```
CarlosCortizasCT, line 85:
```suggestion
    <Spec label="with warning when touched">
```

CarlosCortizasCT (commented):
I think we just to adjust some test names and we will be good to go

CarlosCortizasCT (approved):
Many thanks for your work in this PR, Chisom 🙇 

Please, wait for Filip and another Shield engineer approval 🙏

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
```diff
@@ -467,7 +489,7 @@ export default class AsyncCreatableSelectField extends Component<
             isOptionDisabled={this.props.isOptionDisabled}
             isMulti={this.props.isMulti}
             isSearchable={this.props.isSearchable}
-            hasWarning={this.props.hasWarning}
+            hasWarning={hasWarning}
```
kark, line 492: I'm a bit confused about one aspect in general. Based on my understanding, the original requirements were outlined [here](https://github.com/commercetools/ui-kit/issues/2671#issuecomment-1886715030).

I read [this comment](https://github.com/commercetools/ui-kit/pull/2698#issuecomment-1905713921) but I'm not sure if this was addressed. 

Previously, the `hasWarning` prop alone triggered the border display, but it's no longer considered. Is the `hasWarning` prop deprecated? Shouldn't this be mentioned in the changeset?

kark (commented):
Thank you, I'm leaving the comment to get a better understanding of the changes

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.tsx
CarlosCortizasCT: That's a good catch, Kacper.

I think the `hasWarning` prop we currently have should be deprecated, but that means keeping it so consumers have time to transition from one prop (`hasWarning`) to the other (`warnings`).

We probably need the `hasWarning` constant calculation to be adjusted:
```js
const hasWarning =
     this.props.hasWarnings ||
      (AsyncCreatableSelectInput.isTouched(this.props.touched) &&
      hasWarnings(this.props.warnings));
```

And also update the jsdoc of that property to mention about it.

@kark what do you think?

obulaworld: Ahhh yeahh I think it should. @CarlosCortizasCT Whats your take on this?

obulaworld: Cool stuff. @CarlosCortizasCT So the jsdoc should mention that the field is already deprecated or is their a specific copy you would like to use?

CarlosCortizasCT: Hi @obulaworld 

> @CarlosCortizasCT So the jsdoc should mention that the field is already deprecated or is their a specific copy you would like to use?

I mean both.
For example:
```
  /**
   * Indicates the input field has a warning.
   * @deprecated Please use the `warnings` prop instead so users know the reason why the field is in warning state.
   */
  hasWarning?: boolean;
```

> Also does this mean we keep the hasWarning boolean prop in the story?

We should remove it as we don't want consumers to use it anymore.

obulaworld: Done

CarlosCortizasCT (approved):
Thanks.

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.visualroute.jsx
```diff
@@ -77,15 +77,27 @@ const DefaultRoute = () => (
...
+    </Spec>
+    <Spec label="with warning when touched">
       <AsyncCreatableSelectField
         title="State"
-        name="form-field-name"
         value={value}
         onChange={() => {}}
         loadOptions={loadOptions}
         horizontalConstraint={7}
-        hasWarning={true}
```
kark, line 88: @CarlosCortizasCT Do you think we should delete the `<Spec>`s with the legacy `hasWarning` or leave them?

kark (commented):
Thank you for the contribution 🎉 looks good to me

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.visualroute.jsx
obulaworld: @CarlosCortizasCT ?

obulaworld: @kark I think we can remove them when we are finally removing everything about the hasWarning prop.

CarlosCortizasCT: Yes, I think we should keep them to make sure we don't break that behaviour ahead of time.
We still want to support that prop although we want to encourage consumers to not use it anymore.

obulaworld: Okay. @CarlosCortizasCT I do not have the permission to either approve the percy test or merge this PR. So I believe someone on the UI-Kit will have to do both

CarlosCortizasCT: > Many thanks for your work in this PR, Chisom 🙇
> 
> Please, wait for Filip and another Shield engineer approval 🙏

As I explained before, we need another Shield engineer approval as well as @FilPob's one.

FilPob:
Looking good from my side! I noticed only one point: In the `Timefield` the orange border takes precedence, even though both errors and warnings are shown. In all other fields when both errors and warnings are shown, the Red border takes precedence, which is correct
image

packages/components/fields/async-creatable-select-field/src/async-creatable-select-field.visualroute.jsx
kark: > @kark I think we can remove them when we are finally removing everything about the hasWarning prop.

My point here was that we are removing a bunch of `<Specs>` in this PR e.g.:
Screenshot 2024-01-29 at 16 43 04
and we should probably leave them for now

obulaworld: Ahhhhh Okay. I get it now. Will add them back plus the new ones

obulaworld:
> Looking good from my side! I noticed only one point: In the `Timefield` the orange border takes precedence, even though both errors and warnings are shown. In all other fields when both errors and warnings are shown, the Red border takes precedence, which is correct image

This will be coming from the Input directly. So I guess I will update the Input in this PR.

kark (approved):

PR #2699, opened 2024-01-23
Fix select input components dropdown menu width

CarlosCortizasCT:
#### Summary

Fix select input components dropdown menu width

## Description

We have an issue while calculating the width for the dropdown menu rendered when a select input component is clicked.
The problem is that we are using the minimum width configured also for the maximum width of that element:
image

### DIFF
--- a/packages/components/inputs/select-utils/src/create-select-styles.ts
+++ b/packages/components/inputs/select-utils/src/create-select-styles.ts
@@ -94,13 +94,17 @@ type TState = {
 
 type TDesignTokenName = keyof typeof designTokens;
 
+const upperFirst = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
+
 const getHorizontalConstraintValue = (
   value?: TProps['minMenuWidth'] | TProps['maxMenuWidth']
 ) => {
-  return (
-    designTokens[`constraint${value}` as TDesignTokenName] ||
-    designTokens.constraint3
-  );
+  const designTokenSuffix =
+    typeof value === 'string' ? upperFirst(value) : value;
+  const constraintValue =
+    designTokens[`constraint${designTokenSuffix}` as TDesignTokenName] ||
+    designTokens.constraint3;
+  return constraintValue;
 };
 
 const getInputBackgroundColor = (props: TProps) => {
@@ -243,8 +247,9 @@ const menuStyles = (props: TProps) => (base: TBase) => {
     minWidth: props.minMenuWidth
       ? getHorizontalConstraintValue(props.minMenuWidth)
       : designTokens.constraint3,
-    maxWidth:
-      props.maxMenuWidth ?? getHorizontalConstraintValue(props.maxMenuWidth),
+    maxWidth: props.maxMenuWidth
+      ? getHorizontalConstraintValue(props.maxMenuWidth)
+      : designTokens.constraintScale,
   };
 };
 
--- a/packages/components/inputs/select-utils/src/create-select-styles.ts
+++ b/packages/components/inputs/select-utils/src/create-select-styles.ts
@@ -94,17 +94,18 @@ type TState = {
 
 type TDesignTokenName = keyof typeof designTokens;
 
-const upperFirst = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
-
 const getHorizontalConstraintValue = (
   value?: TProps['minMenuWidth'] | TProps['maxMenuWidth']
 ) => {
-  const designTokenSuffix =
-    typeof value === 'string' ? upperFirst(value) : value;
-  const constraintValue =
-    designTokens[`constraint${designTokenSuffix}` as TDesignTokenName] ||
-    designTokens.constraint3;
-  return constraintValue;
+  switch (value) {
+    case 'auto':
+      return 'initial';
+    default:
+      return (
+        designTokens[`constraint${value}` as TDesignTokenName] ||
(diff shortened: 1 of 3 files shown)


packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -243,8 +247,9 @@ const menuStyles = (props: TProps) => (base: TBase) => {
     minWidth: props.minMenuWidth
       ? getHorizontalConstraintValue(props.minMenuWidth)
       : designTokens.constraint3,
-    maxWidth:
-      props.maxMenuWidth ?? getHorizontalConstraintValue(props.maxMenuWidth),
+    maxWidth: props.maxMenuWidth
+      ? getHorizontalConstraintValue(props.maxMenuWidth)
+      : designTokens.constraintScale,
```
CarlosCortizasCT, line 252: If no `maxMenuWidth` is provided, we use 'scale' as the default value.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -94,13 +94,17 @@ type TState = {
...
+
 const getHorizontalConstraintValue = (
   value?: TProps['minMenuWidth'] | TProps['maxMenuWidth']
 ) => {
-  return (
-    designTokens[`constraint${value}` as TDesignTokenName] ||
-    designTokens.constraint3
-  );
+  const designTokenSuffix =
+    typeof value === 'string' ? upperFirst(value) : value;
```
CarlosCortizasCT, line 103: If the `scale` value is provided, we need a little transformation as its related token name is `constraintScale`.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -94,13 +94,17 @@ type TState = {
...
+  const constraintValue =
+    designTokens[`constraint${designTokenSuffix}` as TDesignTokenName] ||
+    designTokens.constraint3;
```
kark, line 106: I think if the current `value` is `auto`, then `constraintValue` will fall back to `designTokens.constraint3` here. Not sure if this is correct. My point is perhaps we don't need the fallback value.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -94,13 +94,17 @@ type TState = {
 
 type TDesignTokenName = keyof typeof designTokens;
 
+const upperFirst = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
```
kark, line 97: Perhaps we can use `upperFirst` from lodash, since it's already a dependency in this package

kark (commented):
Thanks!

FilPob (approved):

packages/components/inputs/select-utils/src/create-select-styles.ts
chloe0592: Yes, I also have doubts about this. Maybe @FilPob can help us what should the 'auto' prop do in this case?

CarlosCortizasCT: Thanks for pointing that out.

I had another implementation alternative that I discarded at last minute but I think is better.

Please take a look at the changes here: 54bbd00d7ca3185499dc6547e59f4cdfd1e3ea02

packages/components/inputs/select-utils/src/create-select-styles.ts
CarlosCortizasCT: I checked the main `package.json` file and didn't see that dependency so I assumed we were not using it in this repository 🤦 

Anyway, with the new approach we don't need that utility anymore.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -97,10 +97,15 @@ type TDesignTokenName = keyof typeof designTokens;
...
-    designTokens[`constraint${value}` as TDesignTokenName] ||
-    designTokens.constraint3
-  );
+  switch (value) {
+    case 'auto':
+      return 'initial';
+    default:
+      return (
+        designTokens[`constraint${value}` as TDesignTokenName] ||
+        designTokens.constraintScale
```
CarlosCortizasCT, line 106: From this helper point of view, I think it's safer to default to 100%.

Using this approach, even if we don't transform the incoming `scale` value, we will return that value anyway.

packages/components/inputs/select-utils/src/create-select-styles.ts
chloe0592: Thanks, Carlos. I agree, this is a better solution.

chloe0592 (approved):
Thanks so much for the quick fix! 🙏

packages/components/inputs/select-utils/src/create-select-styles.ts
FilPob: Honestly I am also not very sure what that `auto` option is doing :D But maybe it doesnt really make sense to have that option available for min.width and max.width.... I can image if you have just one "width" as a prop that the value gets calculated automatically somehow. But setting both min.width and max.width to auto doesnt make sense does it?

CarlosCortizasCT: Let's try to get this merged so we fix the bug and we can customize the available options as a follow-up 🙏

kark (approved):
Thanks again for the fix!

PR #2701, opened 2024-01-25
FCT 663: add `LeadingIcon` component

ByronDWall:
Adds the `LeadingIcon` component to the ui-kit `Icons` family of components.


FilPob:
@ByronDWall thanks! Looking good so far! 2 points:

- In the `white` version the icon gets a tiny bit downscaled in all sizes. Could be maybe because of the border, that is otherwise not present in other colors?
- I would like to increase the size of the icon in the size `10` as it is too tiny... I would suggest to make the icon 22x22px

CarlosCortizasCT (commented):
Thanks for the contribution, @ByronDWall 

I left some comments with some suggestions in order to keep consistency with how we manage styles in the rest of the components.

Also, I created [this PR](https://github.com/commercetools/ui-kit/pull/2708/) with an alternative approach which would follow those suggestions.

Please let me know if it make sense to me.
I'm also available if you want to have a chat about it.

packages/components/icons/src/icons.visualroute.jsx
```diff
@@ -3,15 +3,23 @@ import { Switch, Route } from 'react-router-dom';
...
 import { Suite, Spec } from '../../../../test/percy';
 
 const IconList = styled.div`
   display: grid;
   grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr;
 `;
 
+const LeadingIconList = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr;
```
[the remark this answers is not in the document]
ByronDWall: it is a good suggestion, I will also make the change to the other `grid-template-columns` selector in this file

packages/components/icons/src/leading-icon/leading-icon.tsx
```diff
@@ -0,0 +1,197 @@
...
+   */
+  icon?: ReactElement;
+  /**
+   * Custom SVG to be displayed, must pass either an svg prop or an icon prop
+   */
+  svg?: string;
+};
+
+const iconSizes = {
+  10: designTokens.spacing50, // 32px
```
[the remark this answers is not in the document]
ByronDWall: I'll remove them, that was really more for me than for future people

packages/components/icons/src/leading-icon/leading-icon.tsx
```diff
@@ -0,0 +1,197 @@
...
+  20: `calc(${designTokens.spacing50} + ${designTokens.spacing20}`, // 40px
+  30: designTokens.spacing60, //48px
+  40: designTokens.spacing70, //64px
+} as const;
+
+function getSizeStyles(size: TLeadingIconProps['size'] = '20') {
```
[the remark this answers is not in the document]
ByronDWall: I'll see what I can get ts to let me do, but Im sure I can accommodate this reasonable request

packages/components/icons/src/leading-icon/leading-icon.tsx
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  const sizeStyles = getSizeStyles(props.size);
+  const colorStyles = getColorStyles({
+    color: props.svg ? 'customSvg' : props.color,
+    isInverted: props.isInverted,
+  });
+  return (
+    <div
+      // https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles
```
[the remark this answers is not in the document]
ByronDWall: This is one place where I am going to push back a little bit. I definitely understand the need for consistency.

That being said, this is a component library that is used extensively, and I am going to argue that doing as much as is possible to reduce the amount of 
1) js that needs to be executed to put styles on the page at runtime, and 
2) the amount of css that gets added to the CSSOM by that js,
is an overall win.  

This implementation makes sure that this component generates a single class, instead of generating up to as many classes as there are color themes * sizes (plus more if theres custom svg).

I think that the benefits, along with the fact that this is the approach that is recommended by the emotion docs for this use-case, makes a pretty compelling argument for allowing this pattern at the cost of being less consistent.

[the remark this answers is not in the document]
CarlosCortizasCT: Thanks for your explanation, @ByronDWall 

I totally understand your point but to me consistency is also a very high priority goal for such a big (and growing) components library in order to keep maintainability in place.

I would buy into taking this as a checkpoint for us to think about how are we are writing the styles and maybe come up with a new best practice and a plan for migrating to it.

My reasoning for not doing so it two folded:
* We are already studying alternatives to `styled-component` as our main library and event if we want to keep using CSS-inJS, so I don't feel this is like the right moment to raise this debate.
* I understand that it's difficult to find the balance between performance and maintainability and also that we all might have different perspectives. Mine, as someone who had to maintain large codebases for many years, I would prefer maintainability to keep codebase manageable and scalable. I'm not saying at all that performance is not important, but I also think we should wight how much benefit we would be gaining and compare it with the code complexity.

Hope this reasoning can make sense to you.

Please feel free to ping me if you prefer to have an online discussion about it 👍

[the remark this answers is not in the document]
ByronDWall: @CarlosCortizasCT Your reasoning certainly makes sense, and I understand your concern.  You work in this library every day, Im just visiting - so I'm not going to push back too hard, I just wanted to explicitly state my concerns.

I will change the component up to be consistent in how it handles styles.

I also want to add a +1 for maybe-not-css-in-js when that conversation is had.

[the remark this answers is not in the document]
CarlosCortizasCT: Thanks for your understanding, @ByronDWall 

Please, do not forget to that a look at [this proposal](https://github.com/commercetools/ui-kit/pull/2708/) to see if it could fit the comments 🙏

ByronDWall:
>     * In the `white` version the icon gets a tiny bit downscaled in all sizes. Could be maybe because of the border, that is otherwise not present in other colors?

>     * I would like to increase the size of the icon in the size `10` as it is too tiny... I would suggest to make the icon 22x22px

@FilPob the most recent commit should address your concerns:
 

https://github.com/commercetools/ui-kit/assets/87667330/e88a353d-d3ac-4d89-ad68-21ef5b835f34

FilPob (approved):

packages/components/icons/README.md
```diff
@@ -72,3 +72,36 @@ The `data` passed to the component is run through a DOM sanitizer to prevent unw
 ### Where to use
 
 This component can be used whenever the icon has to be rendered dynamically on runtime. For example in the Merchant Center this can be the case for the navigation menu icons, etc.
+
+## Leading Icon
+
+For use when you need to render an icon with a themed bounding-box. Icons can be either from the default set of icons, or custom SVG's.
```
CarlosCortizasCT, line 78: @FilPob Do you have any other tip we should include here about what's the purpose of this new component?

packages/components/icons/src/leading-icon/leading-icon.styles.ts
```diff
@@ -0,0 +1,156 @@
...
+
+function getColorThemeStyles({
+  color,
+  isInverted,
+}: {
+  color: TLeadingIconProps['color'] | 'customSvg';
+  isInverted: TLeadingIconProps['isInverted'];
+}): { background: string; fill?: string; border?: string; margin?: string } {
+  return colorThemeMap[color!][isInverted ? 'dark' : 'light'];
+}
```
CarlosCortizasCT, line 106: **nit**: these two functions are one liners, so perhaps we don't need them and can include those two lines in the `getLeadingIconStyles` function below.

CarlosCortizasCT (commented):
When trying the new component in the Storybook preview deployment I noticed that, when a consumer provides a custom SVG, the `color` and `isInverted` properties do nothing.

The way I see it, the should affect the background color of the component but not the icon color (as we don't know what color it will be and should be controlled by the consumer).

@FilPob can you please share what would you expect in this scenario?

FilPob:
> When trying the new component in the Storybook preview deployment I noticed that, when a consumer provides a custom SVG, the `color` and `isInverted` properties do nothing.
> 
> The way I see it, the should affect the background color of the component but not the icon color (as we don't know what color it will be and should be controlled by the consumer).
> 
> @FilPob can you please share what would you expect in this scenario?

Good point. It was not specified in the Figma file... I would suggest following in the custom svg case. Set the background color to `color-surface` and `color-solid` when it's inverted. The svg should stay unaffected as it could be an icon or also a colored graphics of any kind

packages/components/icons/src/leading-icon/leading-icon.styles.ts
```diff
@@ -0,0 +1,156 @@
+import { designTokens } from '@commercetools-uikit/design-system';
+import { css } from '@emotion/react';
+import { type TLeadingIconProps } from './leading-icon';
+
+const sizeMap = {
+  10: designTokens.spacing50,
+  20: `calc(${designTokens.spacing50} + ${designTokens.spacing20})`,
```
chloe0592, line 7: Should we perhaps add a new design token for the spacing that'll be 40px (`designTokens.spacing55`) instead of the calculation? 
@FilPob do you know if we plan in the future to use this spacing in more places?

FilPob: I think @CarlosCortizasCT already added that in another recent PR. Is that right?

packages/components/icons/README.md
FilPob: Sure, I would say something like: The leading icon is a an eye-catchy visual element that should be used when an additional visual prominence is needed for a content section in the UI. The different colours in combination with the icons can be utilised to create certain categorisation of the elements in the UI.

packages/components/icons/src/leading-icon/leading-icon.styles.ts
CarlosCortizasCT: Yes, that's part of [this PR](https://github.com/commercetools/ui-kit/pull/2703).

packages/components/icons/src/leading-icon/leading-icon.styles.ts
ByronDWall: @CarlosCortizasCT these are declared as discrete functions so that the types are more readable and (in my opinion) easier to reason about.  I can change it up if you really prefer, but I'd just as soon leave it as-is.
Screenshot 2024-01-30 at 8 02 11 AM

packages/components/icons/src/leading-icon/leading-icon.styles.ts
ByronDWall: @chloe0592 @CarlosCortizasCT  if `spacing55` has been added recently, it is not available on `main` (screenshot is after a rebase and running `yarn install` within the past 15min).
Screenshot 2024-01-30 at 8 27 30 AM
Screenshot 2024-01-30 at 8 27 43 AM


Let me know what youd like to do here, im fine with whatever yall decide.

packages/components/icons/src/leading-icon/leading-icon.styles.ts
ByronDWall: updated [here](https://github.com/commercetools/ui-kit/pull/2701/commits/ec6296c92de660bc69dbf14453079f7433827243)

packages/components/icons/README.md
ByronDWall: updated to Filip's description [here](https://github.com/commercetools/ui-kit/pull/2701/commits/ec6296c92de660bc69dbf14453079f7433827243)

stephsprinkle (approved):

packages/components/icons/src/leading-icon/leading-icon.styles.ts
CarlosCortizasCT: Don't need to.

If this one is merged before the other one, I'll adjust it over there.

packages/components/icons/src/leading-icon/leading-icon.styles.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const getLeadingIconStyles = (props: TLeadingIconProps) => {
+  const sizeStyles = {
+    height: sizeMap[props.size!],
+    width: sizeMap[props.size!],
+  };
+
+  const colorThemeStyles = colorThemeMap[
+    props.svg ? 'customSvg' : props.color!
+  ][props.isInverted ? 'dark' : 'light'] as {
```
CarlosCortizasCT, line 114: **nit**: perhaps we can extract this to a meaningful named type:
```ts
type TColorThemeConfiguration = {
  background: string;
  fill?: string;
  border?: string;
};
```

CarlosCortizasCT (approved):
Looks good to me 💯 

Thanks a lot for you work, Byron 🙇

chloe0592 (approved):
Thank you @ByronDWall for the work and effort! 🙌

PR #2702, opened 2024-01-25
feat: add progress-bar wip

tylermorrisford:
#### Summary

A not-insignificant number of updates and refactors left, needs more tests... hence very much WIP

[design spec](https://www.figma.com/file/sv5dnCoy76dp0Lo8ULtJ1V/Progress-bar?type=design&node-id=115-542&mode=design&t=4OrxBm3aEAov3YVt-0)
[technical guidelines](https://docs.google.com/document/d/16zRs77B2OCVTbNOims5A74M9VesKZl9xdiVJMtMhC7w/edit?pli=1#heading=h.peiknuubl2k)


FilPob:
@tylermorrisford thanks! Looking quite good sofar!

1. I see an issue with setting the width of the component. When setting the width of the progress bar to `scale`, it doesn't seem to take the full available width of the page

2. It would be great to somehow modify the page in Storybook to also showcase the inverted version. Right now, it's not visible cause the page background is white. Would it be possible to maybe show both default and inverted progress bars on the same page? And the inverted one would be placed in a black container?  Something similar that we have for the **inline icon** Story for `color:surface` [See HERE](https://ui-kit-git-fct-664-progress-bar-commercetools.vercel.app/?path=/story/components-icons--inline-svg)

tylermorrisford:
> @tylermorrisford thanks! Looking quite good sofar!
> 
> 1. I see an issue with setting the width of the component. When setting the width of the progress bar to `scale`, it doesn't seem to take the full available width of the page
> 2. It would be great to somehow modify the page in Storybook to also showcase the inverted version. Right now, it's not visible cause the page background is white. Would it be possible to maybe show both default and inverted progress bars on the same page? And the inverted one would be placed in a black container?  Something similar that we have for the **inline icon** Story for `color:surface` [See HERE](https://ui-kit-git-fct-664-progress-bar-commercetools.vercel.app/?path=/story/components-icons--inline-svg)

Hi @FilPob - Great finds! 
1. Thank you for pointing this out, I've also noticed that setting the bar width to 'auto' causes the bar to disappear completely. Working on fixes for these issues.
2. This should be fixed with [adf453d](https://github.com/commercetools/ui-kit/pull/2702/commits/adf453d002a3f2afb7c4c551179496b8b6ef7cca)

packages/components/progress-bar/src/progress-bar.story.js
```diff
@@ -0,0 +1,61 @@
...
+              labelPosition={select(
+                'labelPosition',
+                ['left', 'right', 'top', 'bottom'],
+                'top'
+              )}
+              height={select('height', ['10', '20'], '20')}
+              labelWidth={select(
+                'labelWidth',
+                ['auto', 'scale', 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16],
+                6
```
CarlosCortizasCT, line 49: label default should be `scale`

packages/components/progress-bar/src/progress-bar.tsx
```diff
@@ -0,0 +1,295 @@
...
+        </SpacingsStack>
+      );
+  }
+};
+
+const defaultProps: TProgressBarProps = {
+  progress: 0,
+  label: null,
+  labelPosition: 'top',
+  labelWidth: 6,
```
CarlosCortizasCT, line 286: label width default value should be `scale`

CarlosCortizasCT (commented):
Something I noticed when testing in the deployment preview is that the vertical spacing between the label and the bar does not change when the user select the small height (`10`).
| height | vertical spacing |
|--------|--------|
| 10 | 8px |
| 20 | 16px |

packages/components/progress-bar/src/progress-bar.tsx
```diff
@@ -0,0 +1,295 @@
...
+            height={props.height}
+          />
+        </SpacingsStack>
+      );
+    case 'left':
+      return (
+        <SpacingsInline scale="m" alignItems="center">
+          <div
+            css={css`
+              text-align: right;
```
CarlosCortizasCT, line 216: I don't get when we need this style.
Could you please share the use case it is here for? 🙏

packages/components/progress-bar/src/progress-bar.tsx
```diff
@@ -0,0 +1,295 @@
...
+              : 'none'};
+            border-radius: ${designTokens.spacingL};
+          `}
+        />
+      </div>
+    </Constraints.Horizontal>
+  );
+};
+
+const ProgressBar = (props: TProgressBarProps) => {
```
CarlosCortizasCT, line 190: Just a refactor idea so we don't repeat the `Bar` and `ProgressBarLabel` components that much:

```tsx
const ProgressBar = (props: TProgressBarProps) => {
  const BarWithProps = (
    <Bar
      progress={props.progress}
      barWidth={props.barWidth}
      isInverted={props.isInverted}
      height={props.height}
      isAnimated={props.isAnimated}
    />
  );

  const LabelWithWithProps = (
    <ProgressBarLabel
      label={props.label}
      labelWidth={props.labelWidth}
      labelPosition={props.labelPosition}
      isInverted={props.isInverted}
      height={props.height}
    />
  );

  switch (props.labelPosition) {
    case 'bottom':
      return (
        <SpacingsStack scale="m" alignItems="center">
          {BarWithProps}
          {LabelWithWithProps}
        </SpacingsStack>
      );
    case 'left':
      return (
        <SpacingsInline scale="m" alignItems="center">
          <div
            css={css`
              text-align: right;
            `}
          >
            <SpacingsInline justifyContent="center">
              {LabelWithWithProps}
            </SpacingsInline>
          </div>
          {BarWithProps}
        </SpacingsInline>
      );
    case 'right':
      return (
        <SpacingsInline scale="m" alignItems="center">
          {BarWithProps}
          <SpacingsInline justifyContent="center">
            {LabelWithWithProps}
          </SpacingsInline>
        </SpacingsInline>
      );
    case 'top':
    default:
      return (
        <SpacingsStack scale="m" alignItems="center">
          {LabelWithWithProps}
          {BarWithProps}
        </SpacingsStack>
      );
  }
};
```

CarlosCortizasCT: @tylermorrisford I kept thinking in this rendering logic and I ended up with an alternative solution which I try to showcase in this alternative PR with the addition of extracting styles to its own file.

Let me see what you think about it 🙏

packages/components/progress-bar/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```
+yarn add @commercetools-uikit/progress-bar
+```
+
+```
+npm --save install @commercetools-uikit/progress-bar
+```
+
+Additionally install the peer dependencies (if not present)
```
ByronDWall, line 20: prolly don't need this bit?

ByronDWall (approved):
so progress, such bar

packages/components/progress-bar/src/progress-bar.tsx
tylermorrisford: I like it! I will test your solution a bit more but should I close this?

packages/components/progress-bar/src/progress-bar.tsx
```diff
@@ -0,0 +1,229 @@
...
+  // We need a special case for 'auto' as a Constraint with `width: auto` will hide the bar
+  if (props.barWidth === 'auto') {
+    return <BarChildren />;
+  }
+  return (
+    <Constraints.Horizontal max={props.barWidth}>
+      <BarChildren />
+    </Constraints.Horizontal>
+  );
+};
```
tylermorrisford, line 151: @CarlosCortizasCT I know this is a bit hacky; hoping you might have a better idea for how to handle this.

packages/components/progress-bar/src/progress-bar.story.js
tylermorrisford: Thank you - completely missed this.

packages/components/progress-bar/src/progress-bar.tsx
CarlosCortizasCT: If you mean this PR, no. You can merge the other PR into this one or pick the bits you agree with.

packages/components/progress-bar/src/progress-bar.tsx
CarlosCortizasCT: To be honest, I don't think the `auto` value makes sense for this component.

For what I've been reading, `auto` (this is default `width` for any block component) will take as much as space possible from the parent, but it's based on the contents con the element.
The progress bar has no content since we're working only with relative (percentage) widths of the two divs (background and foreground bars), therefore using `auto` does not use any space at all.

My suggestion is to not allow the `auto` value for the `barWidth` property.
We can keep it for the `labelWidth` prop as this will affect the label element which will always have content when rendered, although I don't think it's going to be useful.

References:
* [one](https://developer.mozilla.org/en-US/docs/Web/CSS/width#auto)
* [two](https://www.youtube.com/watch?v=-st14lUQD3U)
* [three](https://stackoverflow.com/questions/17468733/difference-between-width-auto-and-width-100-percent)
* [four](https://stackoverflow.com/questions/28353625/why-does-percentage-width-work-even-if-no-explicit-width-value-given-for-contain/28354270#28354270)
* [five](https://ishadeed.com/article/auto-css/)

chloe0592 (approved):
Thanks @tylermorrisford for the work and effort on this! 🙌 
I don't have much of remarks except I agree with Carlos, the `auto` value  seems redundant for width in this case. Everything else looks good to me!

chloe0592:
@FilPob would you say the Percy test is good to be approved? 🕵️

kark (commented):
Thank you for working on this 🙌
I've left just a few minor comments in addition to Carlos's observations. Overall, it looks good to me. Please remember to include a changeset 🙏

packages/components/progress-bar/README.md
```diff
@@ -0,0 +1,51 @@
...
+| `progress`      | `number`                                                                                                                      |          | `0`     | The percentage of the task completion to fill the bar.                                               |
+| `label`         | `union`<br/>Possible values:<br/>`, string, ReactElement, (MessageDescriptor & { values?: Record<string, ReactNode> }), null` |          | `null`  | The text to display alongside the bar.                                                               |
+| `labelPosition` | `union`<br/>Possible values:<br/>`'top' , 'bottom' , 'left' , 'right'`                                                        |          | `'top'` | Location of the text in relation to the bar.                                                         |
+| `labelWidth`    | `union`<br/>Possible values:<br/>`, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 'scale', 'auto'`                                  |          | `6`     | The scale of the width for the label, uses values available in the Constraints.Horizontal component. |
```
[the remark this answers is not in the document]
tylermorrisford: Yes! thank you - good catch 😎

packages/components/progress-bar/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "@babel/runtime-corejs3": "^7.20.13",
+    "@commercetools-uikit/constraints": "17.0.1",
+    "@commercetools-uikit/design-system": "17.0.1",
+    "@commercetools-uikit/spacings-inline": "17.0.1",
+    "@commercetools-uikit/spacings-stack": "17.0.1",
+    "@commercetools-uikit/text": "17.0.1",
+    "@commercetools-uikit/utils": "17.0.1",
+    "@emotion/react": "^11.10.5",
+    "lodash": "4.17.21",
+    "prop-types": "15.8.1",
```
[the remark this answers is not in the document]
tylermorrisford: Thank you, would've left him here forever (fixed in [717bf33](https://github.com/commercetools/ui-kit/pull/2702/commits/717bf33bf9cdb0bd2faef4f79b0a41f20ec55fee))

packages/components/progress-bar/src/progress-bar.tsx
tylermorrisford: Lovely - thank you for considering that. I removed it in [717bf33](https://github.com/commercetools/ui-kit/pull/2702/commits/717bf33bf9cdb0bd2faef4f79b0a41f20ec55fee).

packages/components/progress-bar/package.json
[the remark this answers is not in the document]
tylermorrisford: Hmm - the linting test fails when prop-types is removed. I replace it, guessing I was wrong about which dep you were referring to?

FilPob:
> @FilPob would you say the Percy test is good to be approved? 🕵️

Yes Percy looks good expect that there is the case with the `auto` width, that we decided to remove if I understood it correctly

packages/components/progress-bar/package.json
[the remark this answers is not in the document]
kark: Interesting. My bad.
I've just realized that we're deviating from the monorepo pattern with the `react-intl` dependency

tylermorrisford:
@CarlosCortizasCT This commit ([e6ae103](https://github.com/commercetools/ui-kit/pull/2702/commits/e6ae1033d044002b09abdf256f00b95af7717689)) has max-width instead of a Constraint as discussed! It gets us closer but there will still be some text wrapping if the bar width is set too wide for the parent component.
Screenshot 2024-01-31 at 8 54 08 AM

packages/components/progress-bar/src/progress-bar.styles.tsx
```diff
@@ -0,0 +1,62 @@
...
+  props: TProgressBarProps & { textAlignment: string }
+) => css`
+  min-height: ${heightPerScale[props.height ?? '20']};
+  text-align: ${props.textAlignment};
+  max-width: ${getMaxPropTokenValue(props.labelWidth!)};
+  display: inline-flex;
+  > span {
+    width: max-content;
+  }
+`;
```
tylermorrisford, line 30: @CarlosCortizasCT A solution for not wrapping the label text unless the label text actually exceeds the width of the parent. WDYT?

tylermorrisford:
> > @FilPob would you say the Percy test is good to be approved? 🕵️
> 
> Yes Percy looks good expect that there is the case with the `auto` width, that we decided to remove if I understood it correctly

@FilPob 'auto' was removed, and there's a proposed solution for an issue we had with a small amount of text being wrapped even if it did not exceed the parent width 🥲

packages/components/progress-bar/src/progress-bar.styles.tsx
CarlosCortizasCT: We're getting close, but there's an issue with some extra horizontal space:

image

As you can see, we have some initial extra space we don't want to have.
Think about a use case where we want to render the progress bar beside another element: it would work well if the label text fits in one line, but if there are more, then we have this unwanted extra space.

I'm trying to find a solution for that, but no luck so far.

FilPob (approved):

packages/components/progress-bar/src/progress-bar.styles.tsx
CarlosCortizasCT: I'm still struggling to find a solution for this one.

I'll ask for help to my team on Monday.

tylermorrisford: Moving `max-width` to the child `span` element seems to work but still would require the consumer to pay close attention to their label width choice. 
```
export const getLabelStyles = (
  props: TProgressBarProps & { textAlignment: string }
) => css`
  min-height: ${heightPerScale[props.height ?? '20']};
  text-align: ${props.textAlignment};
  display: inline-flex;
  > span {
    width: max-content;
    max-width: ${getMaxPropTokenValue(props.labelWidth!)};
  }
`;
```

CarlosCortizasCT: Thanks @tylermorrisford 

I still see the same problem.
If you don't can you please share a similar screenshot?

Screenshot 2024-02-05 at 09 52 46

tylermorrisford: I'm still seeing it as well.

packages/components/constraints/src/index.ts
```diff
@@ -1,11 +1,12 @@
 // eslint-disable-next-line import/prefer-default-export
 import Horizontal from './horizontal';
-import { getAcceptedMaxPropValues } from './helpers';
+import { getAcceptedMaxPropValues, getMaxPropTokenValue } from './helpers';
 export * from './horizontal/export-types';
 
 const Constraints = {
   Horizontal,
   getAcceptedMaxPropValues,
+  getMaxPropTokenValue,
```
tylermorrisford, line 9: This is a really nice change!

packages/components/progress-bar/src/progress-bar.styles.tsx
CarlosCortizasCT: We've been trying to find a solution to this problem but were not able to, so we weighed to options:
1. keep trying to find a solution
2. accept this as an edge case where we won't have the exact styles required

We've discussed this options with @FilPob and come to an agreement to accept the behaviour as it is.
We don't expect consumers to use the component in a way such it will reproduce the edge case.

If we ever find an important consumer use case where this is not acceptable, we can come back and research again for a solution.

PR #2703, opened 2024-01-26
Design tokens cleanup

CarlosCortizasCT:
#### Summary

Design tokens cleanup

## Description

Last year we introduced many design tokens with the purpose of our big look and feel migration.

Once that migration is over, we want to remove all unnecessary design tokens to reduce complexity and maintainabllity.

The main work in this PR has been removing as much decision tokens as possible.
I basically only left the ones related to inputs and buttons, since we have several components related and we want them to share styles among them.

After removing the decision tokens, I updated all components to directly use the choice tokens they need for their only required styling needs.

There's a failing VRT regarding the [SearchTextInput](https://github.com/commercetools/ui-kit/pull/2703/files#diff-ebfdedb91d969ec494cc1fadf66f165e83dcbd42c1a46d70d156611b02bc9605) which I don't see the actual visual change and can't spot the difference in the code.
If someone does, please let me know.


design-system/materials/internals/definition.yaml
```diff
@@ -244,6 +244,7 @@ choiceGroupsByTheme:
         spacing-30: 16px
         spacing-40: 24px
         spacing-50: 32px
+        spacing-55: 40px
```
CarlosCortizasCT, line 247: Agreed with @FilPob to add this new choice token.

packages/calendar-utils/src/calendar-body/calendar-body.styles.ts
```diff
@@ -179,7 +179,7 @@ const getInputContainerStyles = (props: TCalendarBody, state: TState) => {
       css`
         &:focus-within {
           border-color: ${designTokens.borderColorForInputWhenFocused};
-          box-shadow: ${designTokens.boxShadowForDatetimeInputWhenHovered}
+          box-shadow: inset 0 0 0 1px
```
emmenko, line 182: Shouldn't box-shadow still have design tokens to avoid hardcoding values?

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -99,50 +96,54 @@ const getBaseStyles = (
...
+        border-color: ${designTokens.colorNeutral};
       }
     `;
   }
 
   switch (theme) {
+    // TODO: These custom colors where itroduced in the Merchant Center redign 2023
+    // We need to check with designers if we want to introduce them in our color palette
+    // #15A390 -> color-primary with 10% black opacity
+    // #057FCC -> color-info with 10% black opacity
```
emmenko, line 114: I think as a basic rule, we should avoid having to hardcode such values. I would expect to use a design token for these, especially colors.

emmenko (commented):
First of all, huge thanks for working on the clean up! ❤️

Before I continue the review maybe we should clarify / discuss a couple of basic points (the hardcoded values) so that we align on how to proceed.

CarlosCortizasCT:
> First of all, huge thanks for working on the clean up! ❤️
> 
> Before I continue the review maybe we should clarify / discuss a couple of basic points (the hardcoded values) so that we align on how to proceed.

Sure.

I created a meeting for us to discuss this topic as I see we need to align on the main goal for design tokens.

packages/components/buttons/icon-button/src/icon-button.styles.ts
CarlosCortizasCT: I have a different point of view (because this is a local value), so let's discuss this in the upcoming meeting.

FilPob: I remember this was kinda of a try to have consistent `hover` and `pressed` states across the MC, without the need to introduce new color-tokens for each use case. The idea was that regardless of what color an interactive UI element has, it should become  e.g 10% darker on hover. We can also revisit this and see if there is a better approach

CarlosCortizasCT: Definitely this is something we will be addressing in our broader discussion about design tokens future.

emmenko:
ℹ️ We had meeting and clarified the points. All good then 🙌

emmenko (commented):
Thank you again for doing the hard work! 🙏 
I trust you mapped the tokens correctly (I haven't double checked them) and Percy for catching any potential regressions 🤗

Regarding the changeset: should we consider this a breaking change (for the `design-system` package) or not?

chloe0592 (approved):
Thank you so much @CarlosCortizasCT for this effort! 🙏

CarlosCortizasCT:
> Regarding the changeset: should we consider this a breaking change (for the `design-system` package) or not?

That's a good catch.

I took a look in the MC frontend repository and I found usages of decision tokens so yes, this would be a breaking change.

We need to think about how to deal with this.

emmenko:
> We need to think about how to deal with this.

We can keep it simple: we release a breaking change and document the potential migrations. No need to worry about it too much.

It won't likely affect "normal users" as I doubt people have been using our specific decision tokens.

FilPob (approved):

kark (approved):

CarlosCortizasCT:
**This PR must not be merged until MC applications are prepared for these changes.**

emmenko (approved):
💯

CarlosCortizasCT:
I've unblocked this PR as all MC applications have already been updated and prepared for the changes in this PR.

PR #2704, opened 2024-01-26
Create CODEOWNERS

tdeekens:
#### Summary

Been flipping a 🪙 about the @commercetools/shield-team-design-system and @commercetools/shield-team-fe for this.

I think the team should decide. I can share my argument for picking the @commercetools/shield-team-fe. Teams are the "entity of communication" towards the outside world. Pods are an inside construct which "the outside" of a team doesn't care about.


CarlosCortizasCT:
> I think the team should decide. I can share my argument for picking the @commercetools/shield-team-fe. Teams are the "entity of communication" towards the outside world. Pods are an inside construct which "the outside" of a team doesn't care about.

Thanks for bringing this up, Tobi.

From my point of view, I thing the codeowner should be the design system pod.
The way we think of the, pods are not something internal to the Shield team but also a public facing entity external teams can interact with, the reason being, since they are both focused groups, we want them to not be "bothered" with topics they should not will be focusing on.

Those are my two cents 🪙 🪙

tdeekens:
> From my point of view, I thing the codeowner should be the design system pod.

Fine by me. Just wanted to share my view. Wasn't aware that Pods are that much of an external facing concept for the team already.

CarlosCortizasCT:
> > From my point of view, I thing the codeowner should be the design system pod.
> 
> Fine by me. Just wanted to share my view. Wasn't aware that Pods are that much of an external facing concept for the team already.

At least that's what I understood.
Please @VictorAvelar, could you confirm? 🙏

CarlosCortizasCT (approved):

PR #2706, opened 2024-01-26
chore(ui-kit): add clickable functionality to `Card`

jaikamat:


packages/components/card/src/card.tsx
```diff
@@ -44,23 +64,63 @@ const Card = (props: TCardProps) => (
...
-  >
-    {props.insetScale === 'none' ? (
-      // Use a `<div>` to ensure that there is always a wrapper container.
-      // This is mostly useful in case custom styles are targeting this element.
+      cursor: ${props.isDisabled
+        ? 'not-allowed'
+        : isClickable
+        ? 'pointer'
+        : 'default'};
+      pointer-events: ${props.isDisabled ? 'none' : 'auto'};
```
jaikamat, line 72: setting `cursor` and `pointer-events` on the same node is mutually exclusive. Are we ok with creating a new `div` container between this one and the content so we can match the design spec (image below)? The tradeoff is, of course, more nodes and a deeper tree.

I can set the cursor here, and attach the pointer events styling and click handler to the first child.

image

packages/components/card/src/card.visualroute.jsx
```diff
@@ -100,6 +100,27 @@ export const component = () => (
       </WrappedCard>
     </Spec>
 
+    <Spec label="Type - Flat, Theme - Dark, Disabled: true">
```
jaikamat, line 103: The only visual difference for these new props is the `disabled` prop (lower the opacity of the card content)

packages/components/card/src/card.tsx
```diff
@@ -25,12 +26,33 @@ export type TCardProps = {
...
-  <div
-    {...filterDataAttributes(props)}
-    css={css`
+const Card = (props: TCardProps) => {
+  const isClickable = Boolean(!props.isDisabled && (props.onClick || props.to));
+
+  const commonProps = {
+    ...filterDataAttributes(props),
+    onClick: isClickable ? props.onClick : undefined,
+    role: isClickable ? 'button' : undefined,
```
kark, line 53: In my opinion we should omit that, this would result in `<a role="button" ...`

packages/components/card/src/card.visualroute.jsx
```diff
@@ -100,6 +100,27 @@ export const component = () => (
...
+    </Spec>
+    <Spec label="Type - Raised, Theme - Dark, Disabled: true">
+      <WrappedCard type="raised" theme="dark" insetScale="s">
+        <Text.Body>{text}</Text.Body>
+      </WrappedCard>
+    </Spec>
+    <Spec label="Type - Raised, Theme - Light, Disabled: true">
+      <WrappedCard type="raised" theme="light" insetScale="s">
+        <Text.Body>{text}</Text.Body>
+      </WrappedCard>
```
kark, line 121: The label should be "Disabled: false" for these 2

packages/components/card/src/card.tsx
```diff
@@ -25,12 +26,33 @@ export type TCardProps = {
...
   className?: string;
   children?: ReactNode;
+  /**
+   * The callback function to be executed when the Card component is clicked. Prefer this for managing side effects rather than navigation.
+   */
+  onClick?: () => void;
+  /**
+   * The URL that the Card should point to. If provided, the Card will be rendered as an anchor element.
+   */
+  to?: string;
```
kark, line 36:
```suggestion
  to?: string | LocationDescriptor;
```

packages/components/card/src/card.tsx
kark: What if we set pointer-events directly for the `<a>`?
```
a {
  pointer-events: ${props.isDisabled ? 'none' : 'auto'};
}
```

kark (commented):
Thank you for your work on this! It's already looking great 🚀 Just a few minor remarks I've noted.

packages/components/card/src/card.tsx
jaikamat: Since the `<a>` tag is possibly the parent node, we will still be unable to view `cursor: not-allowed` styling if `pointer-events: none` is configured

packages/components/card/README.md
```diff
@@ -50,3 +50,7 @@ export default Example;
...
+| `onClick`    | `() => void`                                                       |          |            | The callback function to be executed when the Card component is clicked. Prefer this for managing side effects rather than navigation.                                                             |
+| `to`         | `string`                                                           |          |            | The URL that the Card should point to. When provided, the Card will render as a react-router `<Link>`. Use with `isExternal` to determine if an internal or external link should be used.          |
+| `isExternal` | `boolean`                                                          |          |            | A flag to indicate if the Card points to an external source. When true, the Card will render as an `<a>` tag with appropriate attributes for external links.                                       |
```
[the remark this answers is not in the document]
jaikamat: Just following the tech spec documentation here - is it ok to deviate?

packages/components/card/src/card.tsx
jaikamat: Can you provide an example of how an object of type `LocationDescriptor` is unpacked for use with `<a>`'s `to` prop? Having some difficulty figuring this out 😅

packages/components/card/src/card.tsx
ByronDWall: wouldn't this resolve to `<a role="button"...` only if `to` is passed? 
```suggestion
    role: isClickable && !to ? 'button' : undefined,
```

packages/components/card/src/card.tsx
jaikamat: Ah, I see this [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/link/src/link.tsx#L140), just want to make sure this is the direction we want to go in?

packages/components/card/src/card.tsx
jaikamat: `isClickable` is only `true` if the card is not disabled and we have either an `onClick` handler or `to` (which will resolve to an `<a>`)

`const isClickable = Boolean(!props.isDisabled && (props.onClick || props.to));`

packages/components/card/src/card.visualroute.jsx
jaikamat: Thanks for this, great find! Addressed in e4fac58f908e6949718accc8a0a27b188155c797

packages/components/card/src/card.tsx
jaikamat: Wow, @ByronDWall, I need to learn to read - will make that change

Still waking up 😆

packages/components/card/README.md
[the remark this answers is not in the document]
kark: I think it's exactly like that in the spec file 😅

packages/components/card/src/card.tsx
kark: Yes, that would be great, thanks!

packages/components/card/src/card.tsx
jaikamat: @ByronDWall @kark please see dcca0ca60c27aa725343edcbc35f98e03a27c727, I think this should be correct

packages/components/card/README.md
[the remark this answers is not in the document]
jaikamat: @kark Addressed in 66d0adc8eb087fc3bc66879a2f82eaf9ebe0d169

packages/components/card/src/card.tsx
jaikamat: See 6f6f96fef23f45736ca82dd50b0386a6c3279f27 for the fix - thank you!

packages/components/card/src/card.tsx
kark: ❤️ Thanks, I think it is correct. However, now it seems to me we should also consider [addressing the accessibility aspect](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#:~:text=Note%3A%20If,example%20code.).

jaikamat: >Note: If using role="button" instead of the semantic <button> or <input type="button"> elements, you will need to make the element focusable and define event handlers for [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) and [keydown](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) events. This includes handling the Enter and Space keypresses in order to process all forms of user input. See [the official WAI-ARIA example code](https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/).

Are you referring to this portion?

jaikamat: @FilPob Wondering if you can weigh in here, if we're going to support `button` accessibility for this component, we should have an idea of what a focus state should look like

packages/components/card/src/card.tsx
jaikamat: @FilPob, would you be ok removing the "cursor: not-allowed" portion of the design to improve the semantics of this component and ease development? I imagine the opacity change would be enough to indicate this to the user, mimicking our existing design on disabled buttons

packages/components/card/src/card.tsx
jaikamat: @FilPob this is the default focus state, just for your reference:
image

FilPob: perfect, default is fine

packages/components/card/src/card.tsx
kark: > Since the `<a>` tag is possibly the parent node, we will still be unable to view `cursor: not-allowed` styling if `pointer-events: none` is configured

Understood. Personally, I'm fine with using a conditional wrapper to handle scenarios where the card shouldn't be clickable. I believe this approach should suffice:

```tsx
const DisablingWrapper = (props: TCardProps) => {
  const isNotClickable = props.isDisabled && (props.onClick || props.to);
  return isNotClickable ? (
    <div
      css={css`
        cursor: not-allowed;
        > a,
        div {
          display: block;
          pointer-events: none;
        }
      `}
    >
      {props.children}
    </div>
  ) : (
    <>{props.children}</>
  );
};
```

FilPob: Hmm I dont want to block the progress in order to fulfil this requirement, as the disabled state for a card will not be used very often I suppose. it is definitely rather an edge case. But for the completeness it would be a nice to have. I would like to ask @CarlosCortizasCT if he maybe has an idea, as he originally (and very correctly) made the point about the disabled state... If it turns out to be tricky, or bad practise code-wise, I can imagine to skip this requirement.

packages/components/card/src/card.tsx
```diff
@@ -25,12 +27,32 @@ export type TCardProps = {
...
-  <div
-    {...filterDataAttributes(props)}
-    css={css`
+const Card = (props: TCardProps) => {
+  const isClickable = Boolean(!props.isDisabled && (props.onClick || props.to));
+
+  const commonProps = {
+    ...filterDataAttributes(props),
+    onClick: isClickable ? props.onClick : undefined,
+    ['aria-disabled']: props.isDisabled ? true : undefined,
```
CarlosCortizasCT, line 54:
```suggestion
    'aria-disabled': props.isDisabled ? true : undefined,
```

packages/components/card/src/card.tsx
```diff
@@ -44,23 +66,74 @@ const Card = (props: TCardProps) => (
...
+      if (typeof props.to !== 'string') {
+        throw new Error(
+          '`to` must be a `string` when `isExternal` is provided.'
+        );
+      }
+
+      return (
+        <a
+          {...commonProps}
+          role="link"
```
CarlosCortizasCT, line 114: We don't need the role as this is already the native one for an `a` element.

packages/components/card/src/card.tsx
```diff
@@ -44,23 +66,74 @@ const Card = (props: TCardProps) => (
...
-);
+    );
+
+  if (props.to) {
+    if (props.isExternalLink) {
+      if (typeof props.to !== 'string') {
+        throw new Error(
+          '`to` must be a `string` when `isExternal` is provided.'
+        );
+      }
```
CarlosCortizasCT, line 109: We're used to be using the [`warning`](https://github.com/commercetools/ui-kit/blob/main/packages/utils/src/warning.ts#L17) helper for this cases (don't throw an error as it would very likely break the whole page where this component is used).
```suggestion
warning(
  typeof props.to === 'string',
  'ui-kit/Card: "to" property must be a string when "isExternal" value is true'
);
```

CarlosCortizasCT (commented):
Something I noticed is that the `disabled` UI state is only controlled by the `isDisabled` property.
I think the expectation is that the UI should only be in disabled state if the `isDisabled` property is `true` AND we have received something in either, `onClick` or `to` properties.

ByronDWall (approved):
this card, so clickable, so linkable - delightful

packages/components/card/src/card.tsx
jaikamat: I personally felt `pointer-events: none` was a more succinct way of disabling event handlers as opposed to using JS to conditionally apply them, but perhaps it might be time to start using more branching to do so 🤔

packages/components/card/src/card.tsx
jaikamat: Works for me! See fbf459afc2e4bb78bfbfd40378e82e99a6542355. As long as we're okay with the `ts-ignore` for now-reachable types

packages/components/card/src/card.tsx
jaikamat: Today I learned! Thank you, see 1cd4ef896fc51aeb43ee49dc59e4c5d7caf46cee

packages/components/card/src/card.tsx
jaikamat: Thanks, see 0b4901799de1140d1b4b0000bc8085322fb65382

packages/components/card/src/card.tsx
jaikamat: @CarlosCortizasCT @kark See 4b94814bce1a13c7fdd03fb6bae33964a85f2b7c, thanks for questioning the `pointer-events` strategy I had implemented because I feel this gets us where we need to be, and is more readable. Ultimately, if the card isn't clickable, render the `<div>` parent, otherwise we render a `<a>`

jaikamat: @CarlosCortizasCT @kark added a test in ac80da0bc214b42a5630cb682b691777e0641953 for completeness

jaikamat: Also @FilPob, this change allows us to have the disabled cursor styling - sorry for the wild goose chase here!

packages/components/card/src/card.tsx
```diff
@@ -44,23 +66,74 @@ const Card = (props: TCardProps) => (
...
+        warning(
+          typeof props.to === 'string',
+          'ui-kit/Card: "to" property must be a string when "isExternal" value is true'
+        );
+
+        return (
+          <a
+            {...commonProps}
+            // @ts-ignore
+            href={props.to}
```
CarlosCortizasCT, line 114: Maybe we can use a casting here:
```
href={props.to as string|
```

packages/components/card/src/card.tsx
CarlosCortizasCT: Please, see [this suggestion](https://github.com/commercetools/ui-kit/pull/2706/files#r1471390370).

packages/components/card/src/card.tsx
CarlosCortizasCT: If we have received an object instead of a string, it will be used it's `toString()` under the hood and it will render a string (bad link URL) but it's not the component responsibility and we also rendered a warning.

packages/components/card/src/card.tsx
FilPob: awesome! :) thanks for the effort

packages/components/card/src/card.tsx
jaikamat: Great thinking. Addressed in 007db829991ecf9c421d3195b7aee93d9bf586ba

stephsprinkle (approved):

packages/components/card/src/card.tsx
kark: Thank you for all the improvements! ❤️ I don't think [this](https://github.com/commercetools/ui-kit/pull/2706#discussion_r1469767596) has been addressed. I'm still wondering if the `<div>` with `role="button"` should be accessible. Adding `tabIndex` of `0` and an `onKeyDown` prop in case the `<div>` is clickable should be enough.

packages/components/card/src/card.story.js
```diff
@@ -1,41 +1,61 @@
...
-        </Spacings.Inline>
-      </Spacings.Stack>
-    ))}
-  </Section>
-);
+const CardStoryTemplate = (props) => {
+  const onClickEnabled = boolean('onClick enabled', false);
+  const isDisabled = boolean('isDisabled', false);
+  const to = text('to', '');
+  const isExternalLink = boolean('isExternalLink', false);
```
CarlosCortizasCT, line 21: This property does not change anything in the components UI nor in its behaviour inside the story so I think it's not needed and we can use just a fixed value.

CarlosCortizasCT:
@jaikamat what do you think about this comment?

> Something I noticed is that the disabled UI state is only controlled by the isDisabled property.
I think the expectation is that the UI should only be in disabled state if the isDisabled property is true AND we have received something in either, onClick or to properties.

packages/components/card/src/card.story.js
jaikamat: toggling this as `true` opens a new tab in the story if `to` is set - still ok to remove?

CarlosCortizasCT: The way I see it, it doesn't add any value and it can even lead to thinking the story is broken.

Let's see what @chloe0592 and @kark think of it.

jaikamat: Sounds good - for what it's worth, this behavior exists for the `Link` component as well

jaikamat:
> @jaikamat what do you think about this comment?
> 
> > Something I noticed is that the disabled UI state is only controlled by the isDisabled property.
> > I think the expectation is that the UI should only be in disabled state if the isDisabled property is true AND we have received something in either, onClick or to properties.

Sounds like a plan, thanks for the reminder. Addressed in a8de1eb1ed99bfe7b0578584d94f5967fbef55f5 😄

packages/components/card/src/card.tsx
```diff
@@ -25,12 +27,34 @@ export type TCardProps = {
...
 };
 
-const Card = (props: TCardProps) => (
-  <div
-    {...filterDataAttributes(props)}
-    css={css`
+const Card = (props: TCardProps) => {
+  const isClickable = Boolean(!props.isDisabled && (props.onClick || props.to));
+  // Only disable styling if the card is not clickable
+  const isDisableEnabled = props.isDisabled && (props.onClick || props.to);
```
CarlosCortizasCT, line 51: Maybe a tricky to understand name 🤣 

What about `shouldBeDisabled`?

jaikamat: Hahaha, yeah...not my finest work 😅 

Addressed in 4c8f203565a1e0d9e46c4589b6306a575ee64049

FilPob (approved):

packages/components/card/src/card.tsx
jaikamat: Thanks for that reminder, addressed in 3bac38bdbcb6b25458d7af61d66c1eefa7aed33b 👍

packages/components/card/src/card.story.js
jaikamat: @CarlosCortizasCT @chloe0592 Glad we reached consensus, addressed in 2b97618de31a7ab2ea37d3ee40dba8a5bd2848f5

chloe0592 (approved):
Looks great to me! 🙌  Thanks so much for the effort!

kark (approved):
Thank you! All looks great 💯

CarlosCortizasCT (approved):
Awesome! 💯 

Thanks a lot, @jaikamat 🙇

PR #2710, opened 2024-01-29
feat: add additionalInfo field to LocalizedRichTextInput, LocalizedMultilineTextInput and LocalizedTextInput

LucasGrahn:
#### Summary

Add support for `additionalInfo` field to several localized text inputs

## Description

This PR adds the additionalInfo field which accepts localized text value records. The orientiation for this fields comes from [this document](https://docs.google.com/document/d/1dvP-WNoOvbxIid839zrcSDJlFKqla0FCbxRqD_W5fr0/edit).

Primitive tests are added and storybook files are adjusted.

Please let me know any improvements I could add 🙇‍♂️


packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx
```diff
@@ -205,6 +207,14 @@ const TranslationInput = (props: TranslationInputProps) => {
...
               </LeftColumn>
             );
+          if (props.additionalInfo)
+            return (
+              <LeftColumn>
+                <Text.Detail tone="tertiary">
+                  {props.additionalInfo}
+                </Text.Detail>
+              </LeftColumn>
+            );
```
LucasGrahn, line 217: Not sure if this is the proper way tbh

CarlosCortizasCT (commented):
Please, run the `yarn generate-readmes` command once the props and their jsdocs are updated in order to update the README files of the modified component.

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -330,6 +339,11 @@ const LocalizedTextInput = (props: TLocalizedTextInputProps) => {
                   />
                   {props.errors && props.errors[language]}
                   {props.warnings && props.warnings[language]}
+                  {props.additionalInfo && (
+                    <Text.Detail tone="tertiary">
```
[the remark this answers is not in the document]
LucasGrahn: Will add that. Thank you

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -133,6 +134,17 @@ export type TLocalizedTextInputProps = {
    * A map of warnings.
    */
   warnings?: Record<string, ReactNode>;
+  /**
+   * A map of additional information
```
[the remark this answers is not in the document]
LucasGrahn: Agree

packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx
```diff
@@ -205,6 +207,12 @@ const TranslationInput = (props: TranslationInputProps) => {
                 <div>{props.warning}</div>
               </LeftColumn>
             );
+          if (props.additionalInfo)
+            return (
+              <LeftColumn>
+                <InfoMessage>{props.additionalInfo}</InfoMessage>
+              </LeftColumn>
+            );
```
kark, line 215: I wonder if this accommodates the following scenario
Screenshot 2024-01-31 at 14 42 45
Shouldn't the `additionalInfo` be added to a `<Row>` of its own?

To verify that, perhaps it might be beneficial to include some test cases to verify behavior when `warning`/ `error` is passed alongside `additionalInfo`? Additionally, incorporating some VRTs could be helpful. You can find references in the `*.visualroute.tsx` files.

kark (commented):
Thank you for working on this ❤️ 
Just dropping a comment to get a clearer understanding of the changes.

packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx
LucasGrahn: Screenshot 2024-01-31 at 15 20 03

Thats how it looks in the in the vercel deployment. But I can add a VRT if you want

LucasGrahn: Good catch it doesn't support this case with multiline. Will fix that

LucasGrahn:
Hey @kark I added the VRT's but they seem to create a timeout now for the localized rich text input. Not really sure what could be the reason

packages/components/messages/src/info-message/info-message.tsx
```diff
@@ -0,0 +1,25 @@
...
+  values?: Record<string, ReactNode>;
+};
+
+export type TInfoMessageProps = {
+  id?: string;
+  intlMessage?: TIntlMessageDescriptor;
+  children?: ReactNode;
+};
+
+const InfoMessage = (props: TInfoMessageProps) => (
```
CarlosCortizasCT, line 17: I had suggested to use the name `AdditionaInfoMessage` on purpose.

If we use only `InfoMessage`, we would think this is going to be using a blueish color as that's what we usually use the `info` word to associate that color in some other component.

Could you please rename this files/component to use the `AdditionalInfoMessage`? 🙏

packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx
```diff
@@ -235,6 +237,13 @@ const TranslationInput = (props: TranslationInputProps) => {
           </>
         )}
       </Row>
+      {props.additionalInfo && (
+        <Row>
+          <LeftColumn>
+            <InfoMessage>{props.additionalInfo}</InfoMessage>
+          </LeftColumn>
+        </Row>
+      )}
```
CarlosCortizasCT, line 246: I think rendering the additional info should be placed where the errors and warning are (around line 199) so no new `Row` component is needed.

packages/components/inputs/localized-rich-text-input/src/editor.tsx
```diff
@@ -348,6 +350,11 @@ const Editor = forwardRef((props: TEditorProps, forwardedRef) => {
                 </RightColumn>
               )}
             </Row>
+            {props.additionalInfo && (
+              <LeftColumn>
+                <InfoMessage>{props.additionalInfo}</InfoMessage>
+              </LeftColumn>
+            )}
```
CarlosCortizasCT, line 357: I think rendering the additional info should be placed where the errors and warning are (around line 308) so no new Row component is needed.

LucasGrahn: It was like that before but it would not align with this design
Screenshot 2024-02-01 at 12 35 13

packages/components/inputs/localized-multiline-text-input/src/translation-input.tsx
LucasGrahn: See my other comment about that

LucasGrahn:
> > Hey @kark I added the VRT's but they seem to create a timeout now for the localized rich text input. Not really sure what could be the reason
> 
> Thanks. Occasionally, this step ends up with a failure due to flakes. Typically, re-running the jobs resolves the issue. However, for the LocalizedRichTextInput component, it may be necessary to adjust the [timeout setting](https://github.com/commercetools/ui-kit/blob/a0abf51483dcf26861f6af506b3acf10cb9fcd2c/packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js#L7)
> 
> ```
> FAIL packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js (25.009 s)
> ```

I adjusted it to 30 secs and it still fails. I need some percy token to test it locally right?

kark:
> > > Hey @kark I added the VRT's but they seem to create a timeout now for the localized rich text input. Not really sure what could be the reason
> > 
> > 
> > Thanks. Occasionally, this step ends up with a failure due to flakes. Typically, re-running the jobs resolves the issue. However, for the LocalizedRichTextInput component, it may be necessary to adjust the [timeout setting](https://github.com/commercetools/ui-kit/blob/a0abf51483dcf26861f6af506b3acf10cb9fcd2c/packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js#L7)
> > ```
> > FAIL packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js (25.009 s)
> > ```
> 
> I adjusted it to 30 secs and it still fails. I need some percy token to test it locally right?

Hey, no you don't. The issue with your tests is herre:
```
<Spec label="when there is a additional info set for a locale">
      <LocalizedRichTextInput
        value={value} // 👈 should be initialValue
```

LucasGrahn:
Thank you @kark ! I think now my changes are up for review in Percy. Should I raise this somewhere?

packages/components/inputs/localized-rich-text-input/src/editor.tsx
CarlosCortizasCT: Oh! I missed that one 🤦 

Thanks for the explanation 👍

packages/components/messages/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export { default as ErrorMessage } from './error-message';
 export { default as WarningMessage } from './warning-message';
+export { default as AdditionalInfoMessage } from './additional-info-message';
```
kark, line 3: types should also be exported from here, otherwise they won't be available for direct import from `"@commercetools-uikit/messages"`

packages/components/inputs/localized-text-input/src/localized-text-input.tsx
```diff
@@ -4,10 +4,13 @@ import {
   type ReactNode,
   useCallback,
 } from 'react';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, MessageDescriptor } from 'react-intl';
```
kark, line 7:
```suggestion
import { FormattedMessage, type MessageDescriptor } from 'react-intl';
```

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.visualspec.js
```diff
@@ -4,7 +4,7 @@ import { getDocument, queries } from 'pptr-testing-library';
 let browser;
 let page;
 
-jest.setTimeout(20000);
+jest.setTimeout(30000);
```
kark, line 7: We can revert it

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.tsx
```diff
@@ -12,6 +12,7 @@ import {
   type FocusEventHandler,
   type MutableRefObject,
 } from 'react';
+import { MessageDescriptor } from 'react-intl';
```
kark, line 15:
```suggestion
import type { MessageDescriptor } from 'react-intl';
```

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.tsx
```diff
@@ -5,7 +5,7 @@ import {
   type ChangeEventHandler,
   type FocusEventHandler,
 } from 'react';
-import { useIntl } from 'react-intl';
+import { useIntl, MessageDescriptor } from 'react-intl';
```
kark, line 8:
```suggestion
import { useIntl, type MessageDescriptor } from 'react-intl';
```

kark (commented):
Thanks for working on this 🙌 
From my perspective, everything looks great with the LocalizedRichTextInput, LocalizedMultilineTextInput, and LocalizedTextInput components (AFAIK, other components won't be addressed). I've only got a few minor suggestions for cleanup. Could you please include a changeset? 🙏 Much appreciated!

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
```diff
@@ -35,6 +35,7 @@ type TRichTextInputProps = {
   toggleLanguage: (language: string) => void;
   warning?: ReactNode;
   error?: string;
+  additionalInfo?: ReactNode;
```
CarlosCortizasCT, line 38: There's a type mismatch here.
The `additionalInfo` prop the `LocalizedRichTextInput` component receives has this type:
```ts
Record<
    string,
    | string
    | ReactNode
    | (MessageDescriptor & {
        values: Record<string, ReactNode>;
      })
  >;
```

This means that, if the parent component receives a message descriptor, we won't be rendering the right message.

I don't see specs testing what happens when we pass message descriptions as part of the `additionalInfo` prop so perhaps we can add them and use them to make sure we implement that code path.

packages/components/card/README.md
```diff
@@ -20,11 +20,11 @@ npm --save install @commercetools-uikit/card
 Additionally install the peer dependencies (if not present)
 
 ```
-yarn add react
```
CarlosCortizasCT, line 23: This shouldn't be part of this PR.
After refreshing this branch with `main` we should run the readmes generator again.

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.visualroute.jsx
```diff
@@ -155,5 +155,25 @@ export const component = () => (
         hasWarning={true}
       />
     </Spec>
+    <Spec label="when there is a additional info set for a locale">
```
CarlosCortizasCT, line 158: I think we need new specs to validate all the possible combinations:
image

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
LucasGrahn: I adjusted that and have a question. How do I fix that type mismatch which happens after doing your requested change? I have to admit I barely have done typescript

`Type 'ReactNode | (MessageDescriptor & { values: Record<string, ReactNode>; })' is not assignable to type 'Record<string, ReactNode | (MessageDescriptor & { values: Record<string, ReactNode>; })> | undefined'.
  Type 'null' is not assignable to type 'Record<string, ReactNode | (MessageDescriptor & { values: Record<string, ReactNode>; })> | undefined'.`

CarlosCortizasCT: Hi Lucas,

I've just pushed [this commit](https://github.com/commercetools/ui-kit/pull/2710/commits/85b31144a828f7769027cdcc7b5f7eb3194f05a4) to fix the issue.

The problem is that the outer components receive a `Record` (an object where keys are locales and values are the message in potentially different format) but the underlying components expect only the message.

I also refactored a bit the `AdditionalInfoMessage` so its easier for consumers to provide its value.

Please take a look at the changes in the commit and let me know if there are ok for you.

Also, remember it would be great if we could add a test in the input components to validate they work when a message descriptor or a React element are provided 🙏

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.visualroute.jsx
CarlosCortizasCT: Actually I found what I think is an unexpected behaviour in the `LocalizedRichTextInput`.

If we have additional messages for all the languages but they are collapsed (only english one is rendered), we can see the additional message is right below the text editor:
image

But if we show all the languages, now the first two additional messages are rendered like in the second row, but the third editor shows the additional message right below the editor:
image

LucasGrahn: Good catch! I pushed a fix for that

packages/components/inputs/localized-rich-text-input/src/rich-text-input.tsx
LucasGrahn: I will follow up with those tests 🙇‍♂️

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.spec.js
```diff
@@ -272,3 +283,43 @@ describe('when the error is on the selected language', () => {
...
+      renderLocalizedMultilineTextInput({
+        additionalInfo: {
+          en: 'Some description',
+        },
+      });
+    expect(getByLabelText('EN')).toBeInTheDocument();
+    expect(queryByLabelText('FR')).not.toBeInTheDocument();
+    expect(getByText('Some description')).toBeInTheDocument();
+  });
+  it('when given localized data', () => {
```
CarlosCortizasCT, line 299: Please, avoid repeating test names 🙏 
```suggestion
  it('when given localized data as intl message', () => {
```

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.spec.js
```diff
@@ -272,3 +283,43 @@ describe('when the error is on the selected language', () => {
     expect(getByText('Some error')).toBeInTheDocument();
   });
 });
+
+describe('when the additionalInfo exists on the selected locale should display', () => {
+  it('when given localized data', () => {
```
CarlosCortizasCT, line 288: Please, avoid repeating test names 🙏 
```suggestion
  it('when given localized data as string', () => {
```

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.spec.js
```diff
@@ -272,3 +283,43 @@ describe('when the error is on the selected language', () => {
...
+          en: { id: 'i18en', defaultMessage: 'english i18n message' },
+          fr: { id: 'i18fr', defaultMessage: 'french i18n message' },
+        },
+      });
+    expect(getByLabelText('EN')).toBeInTheDocument();
+    expect(queryByLabelText('FR')).not.toBeInTheDocument();
+    expect(getByText('english i18n message')).toBeInTheDocument();
+    expect(queryByLabelText('french i18n message')).not.toBeInTheDocument();
+  });
+  it('when given localized data', () => {
```
CarlosCortizasCT, line 312: Please, avoid repeating test names 🙏 
```suggestion
  it('when given localized data as react element', () => {
```

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.spec.js
```diff
@@ -136,6 +152,59 @@ describe('LocalizedRichTextInput', () => {
...
+                de: 'eine beschreibung',
+              }}
+            />
+          );
+          expect(screen.getByText('cool description')).toBeInTheDocument();
+          expect(
+            screen.queryByLabelText('eine beschreibung')
+          ).not.toBeInTheDocument();
+        });
+        it('when given localized data', () => {
```
CarlosCortizasCT, line 173: Please, avoid repeating test names 🙏

packages/components/inputs/localized-rich-text-input/src/localized-rich-text-input.spec.js
```diff
@@ -136,6 +152,59 @@ describe('LocalizedRichTextInput', () => {
...
+                fr: { id: 'i18fr', defaultMessage: 'french i18n message' },
+              }}
+            />
+          );
+          expect(screen.getByText('english i18n message')).toBeInTheDocument();
+          expect(
+            screen.queryByLabelText('french i18n message')
+          ).not.toBeInTheDocument();
+        });
+        it('when given localized data', () => {
```
CarlosCortizasCT, line 190: Please, avoid repeating test names 🙏

CarlosCortizasCT (commented):
We're almost there 👍 

I've just left some comments about test names repetition.

Also, I realized we need to update field components (`LocalizedMultilineTextField` and `LocalizedTextField`) to allow the new `addidionaInfo` prop and forward it to the input components so we can use them when a consumer uses fields instead of input components directly.

Finally, we need to generate changesets for this modifications.
You can do that by running the `pnpm changeset` command from the root of the repository.

LucasGrahn:
> We're almost there 👍
> 
> I've just left some comments about test names repetition.
> 
> Also, I realized we need to update field components (`LocalizedMultilineTextField` and `LocalizedTextField`) to allow the new `addidionaInfo` prop and forward it to the input components so we can use them when a consumer uses fields instead of input components directly.
> 
> Finally, we need to generate changesets for this modifications. You can do that by running the `pnpm changeset` command from the root of the repository.

I'm on it

packages/components/messages/src/additional-info-message/additional-info-message.tsx
```diff
@@ -0,0 +1,37 @@
...
+export type TIntlMessageDescriptor = {
+  id: string;
+  description?: string | object;
+  defaultMessage: string;
+  values?: Record<string, ReactNode>;
+};
+
+export type TAdditionalInfoMessageProps = {
+  id?: string;
+  message: string | ReactNode | TIntlMessageDescriptor;
```
LucasGrahn, line 13: This should be renamed to `intlMessage` right?

CarlosCortizasCT: I think it's better to keep it like this.
We use the `intlMessage` name when the prop is expected to receive an i18n message: a message object description.

In this case, the prop is polymorphic as it can receive different types (an already internationalized message string or even an element, for instance) so keeping the current name looks more appropirate to me.

.changeset/clever-chicken-press.md
```diff
@@ -0,0 +1,9 @@
+---
+'@commercetools-uikit/localized-multiline-text-field': minor
+'@commercetools-uikit/localized-multiline-text-input': minor
+'@commercetools-uikit/localized-rich-text-input': minor
+'@commercetools-uikit/localized-text-field': minor
+'@commercetools-uikit/localized-text-input': minor
+---
+
+Add new property called `additionalInfo` that allows the user to render a message below the input. If a warning and/or an error is specified for the field the `additionalInfo` will be rendered below that.
```
CarlosCortizasCT, line 9: Here is a suggestion for the changeset:
```
We've added a new property named `additionalInfo` to the component so consumers can provide some more contextual text.
The content will be rendered below each input using the [Text.Detail](https://uikit.commercetools.com/?path=/story/basics-typography-text--detail) component (`tertiary` tone) by default.
```

CarlosCortizasCT (approved):
Looks good to me now.

Let's see what designer this about it.

These are the components we need to check:
* LocalizedMultilineTextField
* LocalizedTextField
* LocalizedMultilineTextInput
* LocalizedRichTextInput
* LocalizedTextInput

FilPob:
Looks all correct to me! Thank you all for working on this!

FilPob (approved):

FilPob:
Oh I found one weird one there. Instead of the error message, there is a "invalid input" label
image

LucasGrahn:
> Oh I found one weird one there. Instead of the error message, there is a "invalid input" label image

this is intentionally set that way

packages/components/messages/src/index.ts
LucasGrahn: I updated it now

.changeset/clever-chicken-press.md
LucasGrahn: Done!

FilPob:
> > Oh I found one weird one there. Instead of the error message, there is a "invalid input" label image
> 
> this is intentionally set that way

I mean the Input with additional info and error message should not look like in the screenshot. Is there any reason why it is set like that? Cause in Storybook this scenario looks correct.

LucasGrahn:
> > > Oh I found one weird one there. Instead of the error message, there is a "invalid input" label image
> > 
> > 
> > this is intentionally set that way
> 
> I mean the Input with additional info and error message should not look like in the screenshot. Is there any reason why it is set like that? Cause in Storybook this scenario looks correct.

Hey Filip

> > > Oh I found one weird one there. Instead of the error message, there is a "invalid input" label image
> > 
> > 
> > this is intentionally set that way
> 
> I mean the Input with additional info and error message should not look like in the screenshot. Is there any reason why it is set like that? Cause in Storybook this scenario looks correct.

Hey Filip, I adjusted it :)

FilPob:
much better now :) thanks!

kark (approved):
Thank you for the contribution 🙌

mariabarrena (approved):
Passed UX review

PR #2711, opened 2024-01-30
New ProgressBar component implementation alternative

CarlosCortizasCT:
#### Summary

New ProgressBar component implementation alternative

## Description

In this PR I showcase a couple of changes to [the original one](https://github.com/commercetools/ui-kit/pull/2702) so we better manage styling and structure independently and I also try to suggest a cleaner rendering logic.


tylermorrisford (approved):
Brilliant, and much more readable 😎

PR #2713, opened 2024-02-06
fix(SecondaryButton, IconButton): background colors

### LINKED ISSUE #2547: HSL colors with alpha used for `background-color` blend with the container's `background-color`
**Describe the bug**
Certain design tokens are defined as HSL colors with `alpha`. Components using these tokens for styling e.g. `background-color` cannot be placed on darker backgrounds due to color blending. For instance, `SecondaryButton` in `active` or `hover`ed state uses the following tokens:
https://github.com/commercetools/ui-kit/blob/5ce0b2aaffe76e0d9466b068c8ddc636be5ae0e2/design-system/src/design-tokens.ts#L143-L144

This results in:
Jun-30-2023 15-38-22

In particular this can be observed in the Save Toolbar:
Screenshot 2023-06-30 at 15 52 46

kark:
#### Summary

Fixes #2547

Filip outlined the solution [here](https://github.com/commercetools/ui-kit/issues/2547#issuecomment-1923765807).

| Before  | After  | 
|---|---|
|Screenshot 2024-02-06 at 14 03 33<br />Screenshot 2024-02-06 at 14 05 19 | Screenshot 2024-02-06 at 14 04 14<br />Screenshot 2024-02-06 at 14 04 47 |


FilPob (approved):

CarlosCortizasCT:
@kark do you think it would be possible to have new VRTs to secure this behaviour?

kark:
> @kark do you think it would be possible to have new VRTs to secure this behaviour?

@CarlosCortizasCT VRTs updated

CarlosCortizasCT:
> > @kark do you think it would be possible to have new VRTs to secure this behaviour?
> 
> @CarlosCortizasCT VRTs updated

Thanks a lot for giving it a try, however it seems the screenshots that Percy takes can't get the actual UI state of the component we want to validate.
When taking a look, for instance, to the new one in the `SecondaryButton` we can see the background color is white but what we would expect is to see a greyish color because it should be hovered.

Perhaps this is due to the problems we've already experienced with Percy when trying to test dynamic states of components so maybe it's better to not have a test at all as it would provide a false sense of safety.

What do you think?

kark:
> > > @kark do you think it would be possible to have new VRTs to secure this behaviour?
> > 
> > 
> > @CarlosCortizasCT VRTs updated
> 
> Thanks a lot for giving it a try, however it seems the screenshots that Percy takes can't get the actual UI state of the component we want to validate. When taking a look, for instance, to the new one in the `SecondaryButton` we can see the background color is white but what we would expect is to see a greyish color because it should be hovered.
> 
> Perhaps this is due to the problems we've already experienced with Percy when trying to test dynamic states of components so maybe it's better to not have a test at all as it would provide a false sense of safety.
> 
> What do you think?

> we can see the background color is white
Sorry, I failed to notice the difference. I think you're right. 

> maybe it's better to not have a test at all
Agreed

CarlosCortizasCT (approved):
Thank you, Kacper 👏

PR #2714, opened 2024-02-07
💅 Restore theming support

kark:
#### Summary

This pull request reinstates the theming setup in the UIKit. Key points to note:
- any suggestions for a more fitting name than `rebranding` for the new theme name would be appreciated
- I've added VRT setup, despite being aware of Percy limitations.
~~- This pull request does not apply any actual token values, just some test ones~~


design-system/src/theme-provider.tsx
```diff
@@ -110,17 +112,47 @@ type TUseThemeResult = {
...
+    if (nextTheme) {
+      setTheme(nextTheme);
+    }
+  }, []);
+
   return {
     theme: 'default',
     themedValue,
-    isNewTheme: true,
+    isNewTheme: false,
```
kark, line 154: I'm not sure, can this be reused? Or could/should that be removed?

FilPob:
My suggestion would for the name would be "recolouring"

FilPob (approved):

design-system/src/theme-provider.tsx
CarlosCortizasCT: From my point of view, I'm not expecting us to actually need neither `isNewTheme` nor `themedValue` this time.

Perhaps we can double check we're not using them in the MC internal applications (monorepo and satellites) and remove them.
If I remember correctly, removing its usage was part of the MC cleanup work we did after last year's MC redesign.

kark: I think `isNewTheme` / `themedValue` will be needed to apply those changes 🤔
Screenshot 2024-02-09 at 11 01 29

About `isNewTheme`, yes it's been cleaned up, but in here only deprecated, so I wonder if removing it would be any kind of breaking change.

design-system/materials/internals/definition.yaml
```diff
@@ -264,6 +264,43 @@ choiceGroupsByTheme:
...
+        color-warning-25: 'hsl(33, 83%, 25%)'
+        color-warning-40: 'hsl(33, 80%, 34%)'
+        color-warning-60: 'hsl(35, 90%, 55%)'
+        color-warning-85: 'hsl(33, 90%, 80%)'
+        color-warning-95: 'hsl(45, 100%, 92%)'
+        color-error: 'hsl(3, 65%, 58%)'
+        color-error-25: 'hsl(4, 69%, 37%)'
+        color-error-40: 'hsl(3, 60%, 46%)'
+        color-error-85: 'hsl(1, 55%, 74%)'
+        color-error-95: 'hsl(349, 66%, 92%)'
```
kark, line 302: @FilPob As discussed with Carlos, I added the new colors so hope you don't mind me asking for another review 🙂
(especially since there were a few cases when a color inspected in Figma dev mode was different than in the color square label)

design-system/src/theme-provider.tsx
kark: ℹ️ Following the discussion with Carlos, the component-level values should be implemented using designated design tokens. While the `themedValue` helper remains available, its usage is encouraged only where necessary for applying new colors in satellite repositories.

CarlosCortizasCT (approved):
Looks good to me 💯 

Thanks a lot, Kacper 👍

chloe0592 (approved):
Thank you Kacper! 🙏

.changeset/purple-snails-do.md
```diff
@@ -0,0 +1,6 @@
+---
+'visual-testing-app': patch
```
tdeekens, line 2: Nit: minor. Restoring a feature is adding a feature :)

tdeekens (approved):

FilPob (approved):

PR #2717, opened 2024-02-08
Stamp spacings for the label and icon

chloe0592:
The spacing between the icon and label in Condensed `Stamp` should be smaller (4px).
We need to make this change to the Stamps component as we're currently adding it to the App bar.

Screenshot 2024-02-08 at 14 33 41


chloe0592:
> Preview looks correct. Should we also update the VRT?

@FilPob good point, thanks! You can check Percy now.

FilPob (approved):

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks Klaudija

PR #2718, opened 2024-02-09
Remove design legacy plain tokens

CarlosCortizasCT:
#### Summary

Remove design legacy plain tokens

## Description

Since we already removed (#2703) temporary design tokens we introduced for last year's redesign, we think this is a good time to also remove the legacy plain tokens as well.

Consumer applications have already been prepared for this.


emmenko (approved):
❤️

chloe0592 (approved):
Thanks Carlos!

PR #2719, opened 2024-02-09
fix(ui-kit): override default anchor card wrapper display style

jaikamat:
## Summary

This PR fixes a bug encountered in the Merchant Center, when using clickable `Card` functionality.

### To reproduce: 
Without `to` prop defined:
image

With `to` prop defined:
image

## Description

TIL `<a>` tags are `display: inline;` by default. By overriding this style, we can match that of the `div` and ensure expected behavior when anchor tags are parent `Card` elements.

### DIFF
--- a/.changeset/tricky-humans-invite.md
+++ b/.changeset/tricky-humans-invite.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/card': patch
+---
+
+Fixes a display error when using clickable Card functionality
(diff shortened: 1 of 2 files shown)


tylermorrisford (approved):
Surgical! Was fun to watch you two solve that problem in 14-ish seconds :)

kark (approved):
Thanks! 🙌

chloe0592 (approved):
Thank you!

PR #2725, opened 2024-02-19
Cut 965 date time input default time

leventekobor:
#### Summary

Adds the `defaultDaySelectionTime` prop to `DateTimeInput` component, and exposes it in the `DateTimeField` component


packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -186,8 +191,15 @@ class DateTimeInput extends Component<
...
         ? null
         : getDateInMonth(this.props.value, this.props.timeZone) - 1,
-    timeString: '',
+    timeString: this.props.defaultDaySelectionTime
+      ? formatDefaultTime(
+          this.props.defaultDaySelectionTime,
+          this.props.intl.locale,
+          this.props.timeZone
+        )
+      : '',
```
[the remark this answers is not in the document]
leventekobor: I might misunderstood the specifications outlined for me. It reads:
```
Bear in mind that this is an internationalized value, so we need to transform the received value to the appropriate format based on the user’s locale
```
Based on this it should be only localised and timezone should not effect it, so ill be removing the timezone part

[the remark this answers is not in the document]
CarlosCortizasCT: Hi Lucas ✋ , 

What we meant by that is that, when rendering the value, we should bear in mind the locale the user is using: is some locales is preferred to use 24 hours notation while in others in AM/PM.

If we want to set the default time to be 10:00AM, we want to always be that specific time in the day, no matter in what timezone the user is located, so the comment in the specification was about the rendering side.

Does this make sense to you?

CarlosCortizasCT:
I think we're still missing a couple of technical requirements:
* The value must be validated and, if it’s not valid, we should log a warning to the console and use the default value.
* Bear in mind that this (provided default time) is an internationalized value, so we need to transform the received value to the appropriate format based on the user’s locale (24 hours or AM/PM)

packages/components/inputs/date-time-input/src/date-time-input.tsx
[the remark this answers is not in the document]
leventekobor: Thanks for the quick clarification @CarlosCortizasCT! i removed the timezone part and added some test for `formatDefaultTime` function

leventekobor:
hey @CarlosCortizasCT thank you for the feedback! 

> I think we're still missing a couple of technical requirements:
> 
> * The value must be validated and, if it’s not valid, we should log a warning to the console and use the default value.

currently in `calendar-time.ts` (`formatDefaultTime` function) there is the following validation for the time: `if (moment(time, 'HH:mm', true).isValid())` if this is false we have a warning in the console notifying the user. Is this type of validation is insufficient? 

> * Bear in mind that this (provided default time) is an internationalized value, so we need to transform the received value to the appropriate format based on the user’s locale (24 hours or AM/PM)

for this also in `calendar-time.ts` there is the following: `moment(today).locale(locale).format('LT')`. should i add something else here as well?

CarlosCortizasCT:
> hey @CarlosCortizasCT thank you for the feedback!
> 
> > I think we're still missing a couple of technical requirements:
> > 
> > * The value must be validated and, if it’s not valid, we should log a warning to the console and use the default value.
> 
> currently in `calendar-time.ts` (`formatDefaultTime` function) there is the following validation for the time: `if (moment(time, 'HH:mm', true).isValid())` if this is false we have a warning in the console notifying the user. Is this type of validation is insufficient?
> 
> > * Bear in mind that this (provided default time) is an internationalized value, so we need to transform the received value to the appropriate format based on the user’s locale (24 hours or AM/PM)
> 
> for this also in `calendar-time.ts` there is the following: `moment(today).locale(locale).format('LT')`. should i add something else here as well?

I'm sorry Lucas, you're right.

packages/calendar-time-utils/src/calendar-time.ts
```diff
@@ -43,6 +43,27 @@ export const changeTime = (
...
+    return moment(today).locale(locale).format('LT'); // 5:13 PM
+  } else {
+    console.warn(
+      `DataTimeInput: the specified defaultDaySelectionTime '${time}' is not supported. The format should be hh:mm, e.g. 11:10. Using 00:00 as default time.`
+    );
+    today.set({
+      hour: 0,
+      minute: 0,
+    });
+    return moment(today).format('LT'); // 12:00 AM
```
CarlosCortizasCT, line 63: I think we should also use the `locale` here so we don't always return '12:00 AM` as there are locales where we should be returning `00:00`.

.changeset/tricky-cherries-behave.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-uikit/date-time-field': minor
+'@commercetools-uikit/date-time-input': minor
+'@commercetools-uikit/calendar-time-utils': minor
+---
+
+Add defaultDaySelectionTime prop to date time input and date time field component.
```
CarlosCortizasCT, line 7:
```suggestion
We've included a new property in the `DateTimeInput` and `DateTimeField` components for consumers to be able to set the default time the calendar will show upon users opening it.
```

packages/calendar-time-utils/src/calendar-time.ts
leventekobor: definitely! thanks

packages/calendar-time-utils/src/calendar-time.spec.js
```diff
@@ -79,3 +82,18 @@ describe('getLocalizedDateTimeFormatPattern', () => {
     });
   });
 });
+
+describe('formatDefaultTime', () => {
+  it('should format the time', () => {
+    expect(formatDefaultTime('09:00', 'en')).toBe('9:00 AM');
+    expect(formatDefaultTime('13:00', 'en')).toBe('1:00 PM');
```
CarlosCortizasCT, line 89: Can we add at least another locale example where the rendered time would be in 24 hours format?

Example:
```js
expect(formatDefaultTime('13:00', 'es')).toBe('13:00');
```

CarlosCortizasCT (approved):
Looks good to me now 💯 

Thanks a lot for your collaboration, Levente 🙇

leventekobor:
> Looks good to me now 💯
> 
> Thanks a lot for your collaboration, Levente 🙇

Thank you for the feedback and the help! are there any more steps? Should i tag anyone else in this PR?

packages/components/inputs/date-time-input/src/date-time-input.tsx
```diff
@@ -186,8 +191,14 @@ class DateTimeInput extends Component<
       this.props.value === ''
         ? null
         : getDateInMonth(this.props.value, this.props.timeZone) - 1,
-    timeString: '',
+    timeString: this.props.defaultDaySelectionTime
+      ? formatDefaultTime(
+          this.props.defaultDaySelectionTime,
+          this.props.intl.locale
+        )
+      : '',
```
kark, line 199: Just a side note - this computation is executed only once, before the first render. Any changes of the `defaultDaySelectionTime` prop won't make difference. It's unlikely that this prop will change and this computation should be done again, so I'm not sure we should consider that a bug. But in that case I would definitely delete the `defaultDaySelectionTime` knob from both stories, since it won't ever work and gives impression that the component is buggy.

kark (commented):
Many thanks, looks good from my end as well. 
I'm just leaving some final remarks.

packages/components/inputs/date-time-input/src/date-time-input.spec.js
```diff
@@ -178,3 +178,14 @@ describe('date picker keyboard navigation', () => {
     expect(screen.getByText('August')).toBeInTheDocument();
   });
 });
+
+describe('date picker defaultDaySelectionTime prop', () => {
+  it('should set the time prop when it is set', async () => {
```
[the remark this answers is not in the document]
leventekobor: removed it! i dont know what that was doing over there

packages/calendar-time-utils/src/calendar-time.ts
```diff
@@ -43,6 +43,27 @@ export const changeTime = (
...
+    const [hour, minute] = time.split(':');
+    today.set({
+      hour: parseInt(hour, 10),
+      minute: parseInt(minute, 10),
+    });
+    return moment(today).locale(locale).format('LT'); // 5:13 PM
+  } else {
+    console.warn(
+      `DataTimeInput: the specified defaultDaySelectionTime '${time}' is not supported. The format should be hh:mm, e.g. 11:10. Using 00:00 as default time.`
+    );
```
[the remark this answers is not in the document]
CarlosCortizasCT: Yes. I agree with Kacper we should be using the `warning` helper.

packages/components/inputs/date-time-input/src/date-time-input.tsx
CarlosCortizasCT: I also think that's not a use case we need to worry about and removing the knob from the story is a good idea.

leventekobor: amazing, i am removing the knob (from date-time-field as well)

PR #2736, opened 2024-03-05
New components Storybook stories migration

CarlosCortizasCT:
#### Summary

New components Storybook stories migration

## Description

In this PR we work in migrating the Storybook stories for Spacing components.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,6 +1,7 @@
 import { dirname, join, resolve } from 'path';
 import type { StorybookConfig } from '@storybook/react-vite';
 import react from '@vitejs/plugin-react';
+import remarkGfm from 'remark-gfm';
 
 function getAbsolutePath(value: string): any {
   return dirname(require.resolve(join(value, 'package.json')));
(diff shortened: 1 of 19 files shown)


.storybook/main.ts
```diff
@@ -16,6 +17,16 @@ const config: StorybookConfig = {
     getAbsolutePath('@storybook/addon-essentials'),
     getAbsolutePath('@storybook/addon-interactions'),
     getAbsolutePath('@storybook/addon-links'),
+    {
```
CarlosCortizasCT, line 20: This is required so we can use advanced Markdown syntax in the READMEs such as tables.

.storybook/preview.tsx
```diff
@@ -19,12 +19,12 @@ const preview: Preview = {
...
-    onChange: { action: 'onChange' },
-    onFocus: { action: 'onFocus' },
-    onInputChange: { action: 'onInputChange' },
-  },
+  // argTypes: {
+  //   onBlur: { action: 'onBlur' },
+  //   onChange: { action: 'onChange' },
+  //   onFocus: { action: 'onFocus' },
+  //   onInputChange: { action: 'onInputChange' },
+  // },
```
CarlosCortizasCT, line 27: This will be revisited in a follow-up PR.
Currently we were adding these types to all components stories, which is not expected.

PR #2739, opened 2024-03-08
ThemeProvider component SSR support

CarlosCortizasCT:
#### Summary

ThemeProvider component SSR support.

## Description

In this PR we try to include server side rendering support for the `ThemeProvider` component.

We also extend the design-system package assets generation to build a CSS files with all design tokens value per theme.
Until now we were only building one for the default theme.


design-system/src/theme-provider.tsx
```diff
@@ -94,6 +94,16 @@ const ThemeProvider = (props: ThemeProviderProps) => {
...
   }, [props.theme, props.themeOverrides]);
 
+  // The previous hook call is not executed on the server, so we just
+  // apply the theme directly if we are on the server.
+  if (typeof window === 'undefined') {
+    applyTheme({
+      newTheme: props.theme,
+      parentSelector: parentSelectorRef.current,
+      themeOverrides: props.themeOverrides,
+    });
```
kark, line 104: The generated css files work fine when imported in the server side rendered app 🙌
👆 This fragment didn't work.
The other issue I found was this warning: 
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
    at ThemeProvider
```

> **_NOTE:_** I created a [demo app](https://github.com/kark/vite-ssr-project) to validate this solution. It's based on `npm init vite-plugin-ssr@latest`. There is a lot of boilerplate 

Perhaps I'm missing something or should have checked differently but just wanted to share my findings.

design-system/scripts/generate-design-tokens.js
```diff
@@ -290,11 +290,26 @@ fs.writeFileSync(
...
   path.join(__dirname, '../materials/custom-properties.css'),
-  prettier.format(printCss(designTokens), {
+  prettier.format(printCss(designTokens.default), {
     ...prettierConfig,
     parser: 'css',
   })
 );
+Object.keys(designTokens).forEach((theme) => {
+  fs.writeFileSync(
+    path.join(__dirname, `../materials/custom-properties_${theme}.css`),
```
emmenko, line 300: For the `default` I would keep the file name without the suffix, so just `custom-properties.json`

design-system/src/theme-provider.tsx
emmenko: Yeah, I guess we would need to decouple the implementation for client/server side rendering.

I think we don't need to spend time on this, having the css file for the new theme should be good enough for the use case of the docs.

Of course having this SSR compatible would be great but if it's not that straightforward maybe it's ok to leave it as-is for now.

emmenko (commented):
Thanks!

design-system/src/theme-provider.tsx
CarlosCortizasCT: Thanks for giving it a try.

I'll check again on Monday.

design-system/scripts/generate-design-tokens.js
CarlosCortizasCT: I'm keeping the default `custom-properties.json` as it is right now to be backwards compatible.
I'm only adding two new files with theme suffixes because I think it would be easier for docs team to use them.

design-system/src/theme-provider.tsx
CarlosCortizasCT: My initial attempt was way too naive and it doesn't make sense.

As suggested by Nicola, I'll keep it simple with the new CSS files and we'll explore SSR support in the future.

kark (approved):
Thank you 🙌
Just a tiny reminder about the changeset before merging

design-system/src/theme-provider.tsx
emmenko: Thanks for looking into it and giving it a try! ❤️

emmenko (approved):

CarlosCortizasCT:
`foo`

> Thank you 🙌 Just a tiny reminder about the changeset before merging

Thanks, Kacper 👍 

Changeset added here: 5a7b12cd5d0836c0fe12853e4406b1feec6e150e

PR #2742, opened 2024-03-12
Recolouring follow-up

kark:
#### Summary

[Figma file](https://www.figma.com/file/hymZlJ7T3NyHsmo1FrSFoP/New-homepage-UI?node-id=4415%3A53959&mode=dev)


packages/components/text/src/text.styles.ts
```diff
@@ -43,7 +43,7 @@ const getTone = (tone: string) => {
     case 'secondary':
       return `color: ${designTokens.colorNeutral40};`;
     case 'positive':
-      return `color: ${designTokens.colorPrimary25};`;
+      return `color: ${designTokens.fontColorForTextAsPositive};`;
```
kark, line 46: That's an additional change not listed in the Figma file

packages/components/tag/src/tag.tsx
```diff
@@ -93,14 +93,14 @@ const Tag = (props: TTagProps) => {
...
-            : designTokens.colorNeutral95};
+            : designTokens.backgroundColorForTag};
 
           ${props.onClick &&
           `&:hover {
             background-color: ${
               props.type === 'warning'
                 ? designTokens.colorWarning95
-                : designTokens.colorNeutral90
+                : designTokens.backgroundColorForTagWhenHovered
```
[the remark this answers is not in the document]
FilPob: yes, correct

design-system/materials/internals/definition.yaml
```diff
@@ -618,15 +711,43 @@ decisionGroupsByTheme:
...
+        background-color-for-tag-when-hovered:
+          choice: color-primary-90
+        background-color-for-dropdown-when-hovered:
+          choice: color-primary-95
+        background-color-for-dropdown-when-active:
+          choice: color-primary-90
+        background-color-for-dropdown-option-when-active:
+          choice: color-primary-95
+        background-color-for-input-as-quiet-when-hovered:
+          choice: 'hsla(240, 64%, 58%, 4%)'
```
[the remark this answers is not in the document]
FilPob: looks good :)

FilPob (approved):

CarlosCortizasCT:
Is this PR ready for review?
I'm asking because it still has the WIP label

kark:
> Is this PR ready for review? I'm asking because it still has the WIP label

@CarlosCortizasCT Thanks for checking, yes, it is

packages/components/text/src/text.styles.ts
kark: Got a late request from Filip. Hope Filip won't mind me sharing the DM: 
```
In the text components we have the tone prop in `Caption` `Detail` `Body` `Subheadline`. In those components we would need to change the `tone: positive` to use `color-success-40`
```
cc @FilPob

design-system/materials/internals/definition.yaml
```diff
@@ -45,6 +45,7 @@ choiceGroupsByTheme:
         color-primary-85: 'hsl(175, 70%, 85%)'
         color-primary-90: 'hsl(175, 70%, 90%)'
         color-primary-95: 'hsl(175, 90%, 95%)'
+        color-primary-98: 'hsl(244, 100%, 99%)' # design token copied from recolouring theme for export
```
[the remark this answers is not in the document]
kark: Sorry, I don't fully understand this one. Is the idea to create a placeholder token value solely due to limitations in our tooling? I believe I would need clarification from Filip on this as I'm not familiar with the pattern.

Upon further reflection, it might be beneficial to use a placeholder value like `unused` or something similar. What do you think?

kark (commented):
@CarlosCortizasCT Thank you for the review 🙏 I covered token renaming in 2a0f27902a3f0d9deb3f08690b9395391ec2a586. The PR is ready for another review round

packages/components/text/src/text.styles.ts
CarlosCortizasCT: Thanks for sharing.

I think it would be beneficial to, aside of the DM, to also update the Figma file so we can have the requirements updated in case we need to check this again in the future so we know we implemented what we've been asked for.

cc/ @FilPob

design-system/materials/internals/definition.yaml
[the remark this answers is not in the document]
CarlosCortizasCT: Since `default` is the base theme, I think it should have all the available tokens.
I understand it might not needed in this case, but I think it's a good idea from the maintainability point of view.

The way I see it, I don't see the benefit in including a placeholder value but just think about what the name of the token represent and have its value in the default theme. That's why I suggest to use the proposal I shared in my previous comment.

We can have a call to discuss this further if you want to 👍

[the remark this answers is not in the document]
kark: Thanks 🙏 Still, I don't fully get the full picture and what is the proposed value for this choice, so I'd rather schedule a call to discuss it

design-system/materials/custom-properties_recolouring.css
```diff
@@ -15,44 +15,45 @@
...
-  --color-brown-98: hsl(35, 90%, 98%);
-  --color-purple-10: hsl(248, 90%, 10%);
+  --color-accent-10: hsl(195, 100%, 10%);
+  --color-accent-20: hsl(195, 80%, 20%);
+  --color-accent-30: hsl(195, 70%, 30%);
+  --color-accent-40: hsl(195, 70%, 40%);
+  --color-accent-50: hsl(195, 70%, 50%);
+  --color-accent-60: hsl(195, 70%, 60%);
+  --color-accent-85: hsl(195, 85%, 85%);
+  --color-accent-90: hsl(200, 90%, 90%);
```
ddouglasz, line 27: The `hue` value from the figma table here is 195. Perhaps there is a reason why you use 200? 
```suggestion
  --color-accent-90: hsl(195, 90%, 90%);
```

design-system/materials/custom-properties_recolouring.css
```diff
@@ -15,44 +15,45 @@
...
+  --color-accent-95: hsl(200, 95%, 95%);
```
ddouglasz, line 28: Same as above too
```suggestion
  --color-accent-95: hsl(195, 95%, 95%);
```

ddouglasz (commented):
Great work @kark 🙌🏾
Just some minor observation I pointed out.

design-system/materials/custom-properties_recolouring.css
kark: Thanks @ddouglasz, great catches 🙌 Those values have changed over time, missed those few updates. Updated in b432d630f2602f7ace57dad982859b8652507926

CarlosCortizasCT:
> @CarlosCortizasCT Thank you for the review 🙏 I covered token renaming in [2a0f279](https://github.com/commercetools/ui-kit/commit/2a0f27902a3f0d9deb3f08690b9395391ec2a586). The PR is ready for another review round

Regarding [this one](https://github.com/commercetools/ui-kit/pull/2742#discussion_r1524386604), do you prefer to not change it?

kark:
> > @CarlosCortizasCT Thank you for the review 🙏 I covered token renaming in [2a0f279](https://github.com/commercetools/ui-kit/commit/2a0f27902a3f0d9deb3f08690b9395391ec2a586). The PR is ready for another review round
> 
> Regarding [this one](https://github.com/commercetools/ui-kit/pull/2742#discussion_r1524386604), do you prefer to not change it?

Updated in 6131eb2312b592107a68dfe45a2d935b1304d617

ddouglasz (approved):
Thank you 🙏🏾

CarlosCortizasCT (approved):
Thanks a lot, Kacper! 🙇 💯

Issue #2740, opened 2024-03-08
NumberInput changes value when you scroll


jocelyn-dunkley:
**Describe the bug**
This is a common UX issue with the number input. If a number input is focused and you scroll, the value of the input is changed inadvertently. 
We recently got a [support ticket](https://commercetools.atlassian.net/browse/SUPPORT-25569) about it and it's annoying our customers:
'The impact on the day to day business is that they need to spend more time entering new wines (granted: not so much) and a high chance of mistakes. But more importantly in CLIENT's eyes, this makes the MC look "immature, buggy and very frustrating to use" (these are the words of their CEO that actually entered the first products himself into the system).'

**To Reproduce**
This is a universal issue but it's most easily seen in the products app.

1. Go to Product list 
2. Click on a product that has a number attribute
3. Scroll down to the attribute and click on it
4. Scroll down and up - you'll see that the number from before and after is different

**Expected behavior**
Nothing changes when you scroll on NumberInput/NumberField component or anything with an underlying `<input type="number"/>` type

**Additional context**
A common solution is to add something like `onWheel={(e) => e.target.blur()}` to the number input field but you can figure out what would be best.

This issue/solution is documented in other open-source libraries like:
[Material UI](https://github.com/mui/material-ui/issues/7960) 
[Carbon](https://github.com/carbon-design-system/carbon/issues/11947)

CarlosCortizasCT:
Thanks for letting us know about this issue @jocelyn-dunkley 👍 

We'll fix it as soon as we have some time for it.

PR #2750, opened 2024-03-19
feat(number input): prevent input change on mouse wheel

ddouglasz:
#### Summary

Fixes https://github.com/commercetools/ui-kit/issues/2740

## Description

If a number input is focused and you scroll, the value of the input is changed inadvertently as can be seen below. We want to be able to scroll with the mouse without having it affect the number input, whether it's focused or not.


kark (approved):
Thanks Douglas 🙌 Looks good to me!

.changeset/calm-moons-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/number-input': minor
+---
+
+prevent number input change when scrolling with the mouse.
```
CarlosCortizasCT, line 5:
```suggestion
Preventing the value of the input to be updated when it has the focus and the user scrolls the page content using the mouse wheel.
```

CarlosCortizasCT (approved):
Thanks 👍

.changeset/calm-moons-run.md
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2750/commits/5020ad798c4a22d7ab62fb41ec3a68391b43f509`

Thank you.

PR #2752, opened 2024-03-19
Export leading icon types

ddouglasz:
#### Summary

Export leading icon types so it can be properly consumed the right way. 

For more context, see https://github.com/commercetools/merchant-center-frontend/pull/16448/files#r1525033953

### DIFF
--- a/.changeset/mean-crabs-perform.md
+++ b/.changeset/mean-crabs-perform.md
@@ -2,4 +2,4 @@
 '@commercetools-uikit/icons': minor
 ---
 
-export types for leading icon
+We now export props typescript types for both `LeadingIcon` and `InlineSvg` components.
--- a/.changeset/mean-crabs-perform.md
+++ b/.changeset/mean-crabs-perform.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/icons': minor
+---
+
+export types for leading icon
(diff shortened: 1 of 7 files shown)


CarlosCortizasCT:
Please @ddouglasz, take a look at the alternative proposal I'm suggesting in [this PR](https://github.com/commercetools/ui-kit/pull/2753) 🙏

ddouglasz:
> Please @ddouglasz, take a look at the alternative proposal I'm suggesting in [this PR](https://github.com/commercetools/ui-kit/pull/2753) 🙏

Yes the change proposal is clear, it makes sense to export the `TLeadingIconProps` from it's own icon-types file in the root of the `leading-icon` folder so that we can import both the `leading-icon` component and it's type from the same directory.  
Thank you.

.changeset/mean-crabs-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/icons': minor
+---
+
+export types for leading icon
```
CarlosCortizasCT, line 5:
```suggestion
We now exports props typescript types for both `LeadingIcon` and `InlineSvg` components.
```

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2752/commits/e96762b7c0e01ecdbbbe20671cb4d3a1ba5ef25a

PR #2753, opened 2024-03-19
refactor(icons): update ts type exporting for custom icon components

CarlosCortizasCT:
#### Summary

This is al alternative proposal for #2752.

#### Description

Here I'm proposing a change in the way we want to expose custom icon components types so they can be consumed from their own entry point.

With the implementation proposed in the original PR, a user wanting to use the `LeadingIcon` component and also access its type would need two different _imports_:
```
import LeadingIcon from '@commercetools-uikit/icons/leading-icon';
import { type TLeadingIconProps } from '@commercetools-uikit/icons';
```

With the approach I'm proposing here, only one import is required:
```
import LeadingIcon, { type TLeadingIconProps } from '@commercetools-uikit/icons/leading-icon';
```

I'm also doing the same for the ` InlineSvg` component.

### DIFF
--- a/packages/components/icons/src/export-types.ts
+++ b/packages/components/icons/src/export-types.ts
@@ -1,2 +1,1 @@
 export type { Props, SVGProps } from './templates/icon.styles';
-export type { TLeadingIconProps } from './leading-icon';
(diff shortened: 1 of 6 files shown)


emmenko (approved):

ddouglasz (approved):
Thank you.

PR #2758, opened 2024-03-21
Icon button changes

ddouglasz:
#### Summary

Icon button changes

When disabled:

### DIFF
--- a/.changeset/two-colts-build.md
+++ b/.changeset/two-colts-build.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/icon-button': major
+'@commercetools-uikit/design-system': minor
+---
+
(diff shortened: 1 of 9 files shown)


FilPob (approved):

CarlosCortizasCT:
Looks good to me but I wanted to double check with @FilPob about the border being removed in the new theme not only for the `default` button theme but also for the `info` and `primary` ones

FilPob:
@CarlosCortizasCT  didnt even see that the border is there in the old theme :) so yes, no border for `info` and `primary` in the new theme

chloe0592 (approved):
Thank you @ddouglasz!

CarlosCortizasCT (approved):
Thanks! 👍

PR #2759, opened 2024-03-26
PasswordField - adding a new prop to render the "Show/Hide" button

chloe0592:
We had an inquiry for adding a prop to `PasswordField` that determines whether to render the "Show/Hide" button for password.

https://github.com/commercetools/ui-kit/assets/52276952/15231966-ba95-48b3-aff7-24801b2f1d1b


FilPob (approved):

FilPob:
UI wise everything looks correct. One thing I found in Percy though: Even though the screenshot doesn't render the showhide button, that description in the purple box says `renderShowHideButton: true` that's the case in all screenshots where the button is not rendered
image

chloe0592:
>UI wise everything looks correct. One thing I found in Percy though: Even though the screenshot doesn't render the showhide button, that description in the purple box says `renderShowHideButton: true` that's the case in all screenshots where the button is not rendered image

Thanks @FilPob for noticing! I do think that's expected, because if we don't specifically set the prop `renderShowHideButton` to `false` it's always set to `true` by default. 
Furthermore, the button won't be rendered even if it's set to `true` in case if the field is "Read-only" or "Disabled". 
Does it makes sense to you?

FilPob:
yes, if it was expected, it's fine for me :)

packages/components/fields/password-field/src/password-field.tsx
```diff
@@ -176,6 +176,8 @@ export type TPasswordField = {
    * Might be used to display additional information about the content of the field (E.g verified email)
    */
   badge?: ReactNode;
+  // Determines whether to render the "Show/Hide" button for the password field.
```
kark, line 179: nit: we typically use JSDoc notation for comments here

kark (approved):
Thanks 🙌

packages/components/fields/password-field/src/password-field.tsx
chloe0592: Oh great catch! Thanks @kark - [f6691c8](https://github.com/commercetools/ui-kit/pull/2759/commits/f6691c8c8f93f96bfc0bdf1405f99dd90a419317).

ddouglasz (approved):
Thank you!  🙌🏾

PR #2761, opened 2024-04-01
Update json5 transitive dependency version

CarlosCortizasCT:
#### Summary

Update json5 transitive dependency version.

## Description

We have [this security alert](https://github.com/commercetools/ui-kit/security/dependabot/106) about the `json5` dependency having a security issue.

We found we're using an unpatched version of that library because of our direct `rcfile` dependency.
* `rcfile`
  * our version: `1.0.3` 
  * latest version: `1.0.3` (8 years old)
  * `json5` dependency version: `^0.5.1` 

image

Since we can't update the `rcfile` dependency because there are no newer version, I'm proposing to update the resolutions sections of the `package.json` file so we make sure we don't import anything below version `1.0.2`.

I also considered replacing the library for a newer one ([example](https://github.com/develar/read-config-file)) but went for the simplest solution.
Please, let me know your thoughts about it.


ddouglasz (approved):
Looks like we are only using a specific version `resolution: "json5@npm:1.0.2"` . We can consider `"json5@npm:^1.0.2"` mentioned as a minimum like you mentioned.

PR #2764, opened 2024-04-01
Update `node-fetch` transitive dependency version

CarlosCortizasCT:
#### Summary

Update `node-fetch` transitive dependency version

## Description

We have [this security alert](https://github.com/commercetools/ui-kit/security/dependabot/44) about the `node-fetch` dependency having a security issue.

This dependency is only used in Storybook.

Here is the dependency path:
```
@storybook/core@npm:5.3.21 -> isomorphic-fetch@npm:2.2.1 -> node-fetch@npm:1.7.3 (via npm:^1.0.1)
```


ddouglasz (approved):
Thank you

PR #2767, opened 2024-04-08
PANGOLIN- 3683 - <Tag/> styling updates

valoriecarli:
Figma references [here](https://www.figma.com/file/PaImaWorWz1RbV85gqw4MH/Filter-%26-Table-controls?type=design&node-id=1133%3A127328&mode=dev)

This PR contains the following updates (screenshots below):
1.) **border-radius** from `2px` to `20px`    --> all tags, in any state
2.) **typography** - `Text.Body` to `Text.Detail` -->   in `TagBody` component
3.) **border-color** to `warning85` --> when type is warning
4.)  **border-color** to `colorPrimary90` --> when type is default w/ hover (onClick)
5.) **background-color** to `neutral95` && **border-color** to `neutral` --> when type is default & isDisabled

Updates 09/04/24:
6.) **background-color** to `neutral95` && **border-color** to `neutral` --> when type is warning && isDisabled (this now matches the color schema of type=default && isDisabled)
7.)  **background-color** to `colorWarning85` --> when type is default w/ hover (onClick)

1.) Before vs After
Screenshot 2024-04-07 at 19 38 21
Screenshot 2024-04-07 at 19 38 46

2.) After
Screenshot 2024-04-07 at 19 52 29

3.) Before vs After
Screenshot 2024-04-07 at 19 41 38
Screenshot 2024-04-07 at 19 42 09

4.) Before vs After
Screenshot 2024-04-08 at 10 15 19
Screenshot 2024-04-08 at 10 19 08

5.) Before vs After
Screenshot 2024-04-08 at 11 55 32

6.) screenshot in [comments](https://github.com/commercetools/ui-kit/pull/2767#discussion_r1558173559)

7.) Before vs After
Screenshot 2024-04-09 at 14 40 59
Screenshot 2024-04-09 at 14 42 36

### DIFF
--- a/.changeset/smooth-baboons-beg.md
+++ b/.changeset/smooth-baboons-beg.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-uikit/tag': minor
+---
+
+Add additional border-radius and color styling to Tag components.  
+
--- a/.changeset/smooth-baboons-beg.md
+++ b/.changeset/smooth-baboons-beg.md
@@ -2,5 +2,5 @@
 '@commercetools-uikit/tag': minor
 ---
 
-Add additional border-radius and color styling to Tag components.  
+Component UI styling updates.  
 
--- a/packages/components/tag/src/tag-body.tsx
+++ b/packages/components/tag/src/tag-body.tsx
@@ -53,7 +53,7 @@ const getContentWrapperStyles = (props: TTagBodyProps) => {
     display: flex;
     box-sizing: border-box;
     align-items: center;
-    border-radius: ${designTokens.borderRadius2};
+    border-radius: ${designTokens.borderRadius20};
     padding: ${designTokens.spacing05} ${designTokens.spacing25};
     white-space: normal;
     text-align: left;
@@ -96,8 +96,8 @@ const TagBody = (props: TTagBodyProps) => {
         Boolean(props.onRemove) &&
           css`
             border-right: ${!props.isDisabled && '0'};
-            border-top-right-radius: 0;
-            border-bottom-right-radius: 0;
+            border-top-right-radius: ${designTokens.borderRadius20};
+            border-bottom-right-radius: ${designTokens.borderRadius20};
           `,
         !props.isDisabled &&
           Boolean(props.onClick) &&
--- a/packages/components/tag/src/tag-body.tsx
+++ b/packages/components/tag/src/tag-body.tsx
@@ -30,18 +30,6 @@ const defaultProps: Pick<TTagProps, 'type' | 'isDisabled' | 'isDraggable'> = {
 type TBody = Pick<TTagBodyProps, 'to' | 'as'>;
 const Body = styled.div<TBody>``;
 
-const getClickableContentWrapperStyles = (type: TTagBodyProps['type']) => {
-  return type === 'warning'
-    ? []
-    : [
-        css`
-          &:hover {
-            border-color: ${designTokens.colorPrimary90};
-          }
-        `,
-      ];
-};
-
 const getTextDetailColor = (isDisabled: TTagBodyProps['isDisabled']) => {
   if (isDisabled) return designTokens.colorNeutral60;
   return designTokens.colorSolid;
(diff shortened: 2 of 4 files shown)


ddouglasz (commented):
looks good, so far 🙌🏽
Please remember to add a changeset.

FilPob:
Looks good from my side @valoriecarli for president ! :)

FilPob (approved):

FilPob:
Just one thing: the new screenshots in Percy seem to use the old theme. But I guess maybe that's expected? cc @CarlosCortizasCT

CarlosCortizasCT:
> ts in Percy seem to use the old theme. But I guess maybe that's expected? c

Yes. We would be changing the default theme in a later PR but we can already do it here.

@valoriecarli is this PR ready to be reviewed?
I'm asking because it's in draft mode and also with the WIP tag.

.changeset/smooth-baboons-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/tag': minor
+---
+
+Add additional border-radius and color styling to Tag components.
```
CarlosCortizasCT, line 5: In this case, we can be more generic so maybe something like this is good enough:
```
Component UI styling updates.
```

packages/components/tag/src/tag-body.tsx
```diff
@@ -53,7 +53,7 @@ const getContentWrapperStyles = (props: TTagBodyProps) => {
     display: flex;
     box-sizing: border-box;
     align-items: center;
-    border-radius: ${designTokens.borderRadius2};
+    border-radius: ${designTokens.borderRadius20};
     padding: ${designTokens.spacing05} ${designTokens.spacing25};
```
CarlosCortizasCT, line 57: In the Figma design they are using a `5px` vertical padding (which is a non-supported value) but here we're still using `2px`.

@FilPob Could you please confirm what value we should be using? 🙏

packages/components/tag/src/tag-body.tsx
```diff
@@ -63,7 +63,9 @@ const getContentWrapperStyles = (props: TTagBodyProps) => {
     border-style: solid;
     border-width: 1px;
     border-color: ${props.type === 'warning'
-      ? designTokens.colorWarning
+      ? designTokens.colorWarning85
+      : props.isDisabled
```
CarlosCortizasCT, line 67: I think the disabled state should override both `warning` and `normal` types.
image

@FilPob Could you please confirm? 🙏

packages/components/tag/src/tag.tsx
```diff
@@ -91,18 +91,25 @@ const Tag = (props: TTagProps) => {
           cursor: default;
           min-width: 0;
           display: flex;
-          background-color: ${props.type === 'warning'
+          border-radius: ${designTokens.borderRadius20};
+          background-color: ${props.isDisabled
```
CarlosCortizasCT, line 95: I believe the disabled state should override styles equally for `normal` and `warning` types.

@FilPob Could you please confirm? 🙏

packages/components/tag/src/tag-body.tsx
FilPob: Sorry for the confusion. No changes to the vertical padding were intended

packages/components/tag/src/tag-body.tsx
FilPob: Yes, the disabled state should be same for both `Default` and `warning` type. @CarlosCortizasCT  good catch, thanks

CarlosCortizasCT:
@valoriecarli Regarding how border related styles were currently managed in this component, I think we can simplify it so I opened [this PR](https://github.com/commercetools/ui-kit/pull/2770) with a proposal.

Please let me know what do you think about it 🙏

packages/components/tag/src/tag-body.tsx
FilPob: Sorry for the missing values... I added the missing styles **[here](https://www.figma.com/file/PaImaWorWz1RbV85gqw4MH/Filter-%26-Table-controls?type=design&node-id=1133%3A127328&mode=design&t=00bf6LyKQTa9dmS5-1)**

I added:
- Warning tag in the hover state: bg-color: `color-warning-85`

> This is nbd, but just to confirm.. on the warning default, do you want the border color to match the neutral of default disabled - or do you want it to remain orange (with neutral background color)?

@valoriecarli  the warning tag in the default state should have border-color: `color-warning-85`. Is that what you asked?

CarlosCortizasCT:
> Just one thing: the new screenshots in Percy seem to use the old theme. But I guess maybe that's expected? cc @CarlosCortizasCT

@FilPob @valoriecarli We just updated the default theme used in both Storybook and the VRTs (used in Percy) to use the new color schema.

We just need to refresh this branch with the latest from `main`.

FilPob:
@valoriecarli looks good now. Thanks!

CarlosCortizasCT (approved):
Thanks for your work, @valoriecarli 🙇 💯

CarlosCortizasCT:
> @valoriecarli looks good now. Thanks!

@FilPob if it's ok for you, could you please approve the Percy differences?

ddouglasz (approved):
Thank you!

PR #2768, opened 2024-04-09
New `DropdownMenu` component

CarlosCortizasCT:
#### Summary

New `DropdownMenu` component

## Description

In this PR we add a new component which can be used for those situations where we need to display a floating panel positioned beside an element when the latter is triggered. It's like a fine-grained contextual content panel.

This new component is abstract in the sense that any element can be used as the triggering one and also the contents of the floating panel can be whatever is needed for each use case.
However, we think an item selection dropdown will be a common one so we provide helpers to easily achieve this UI.

The component can be manually tested in [these Storybook stories](https://ui-kit-git-cm-dropdown-menu-commercetools.vercel.app/?path=/story/components-dropdowns-dropdownmenu--dropdownmenu-list-menu-content).

Here is an example of a general use case:
``` jsx
<DropdownMenu
  triggerElement={
    <SecondaryButton label="Filters" iconLeft={<FilterIcon />} />
  }
  menuHorizontalConstraint={6}
  menuPosition="right"
>
  <SpacingsStack scale="m">
    <Text.Body>Store</Text.Body>
    <CheckboxInput isChecked value="store" onChange={(event) => {}}>
      Canada (FR)
    </CheckboxInput>
  </SpacingsStack>
</DropdownMenu>
``` 

And here is an example of a item selection use case:
```jsx
<DropdownMenu
      triggerElement={<IconButton icon={<ColumnsIcon />} label="list" />}
      menuHorizontalConstraint={6}
      menuPosition="left"
      menuType="list"
    >
      <DropdownMenu.ListMenuItem onClick={() => {}}>
        Option 1
      </DropdownMenu.ListMenuItem>
      <DropdownMenu.ListMenuItem onClick={() => {}} isDisabled>
        Option 2
      </DropdownMenu.ListMenuItem>
      <DropdownMenu.ListMenuItem onClick={() => {}}>
        Option 3
      </DropdownMenu.ListMenuItem>
    </DropdownMenu>
```

**NOTE**: This new component is not sharing anything with the already existing [PrimaryActionDropdown](https://uikit.commercetools.com/?path=/story/components-dropdowns--primaryactiondropdown) one at the moment due to deadline restrictions but that's something we will be doing as a follow up.


packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
```diff
@@ -0,0 +1,65 @@
+import SecondaryButton from '@commercetools-uikit/secondary-button';
+
+import { screen, render } from '../../../../../test/test-utils';
+import DropdownMenu from './dropdown-menu';
+
+describe('DropdownMenu', () => {
```
CarlosCortizasCT, line 6: There are some error logs about state modifications no wrapped in an `act` statement that I don't understand where they're coming from.

Any help would be much appreciated?

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+    // Dropdown shold be closed by default
+    expect(screen.queryByText('Content')).not.toBeVisible();
+
+    // Open the dropdown
+    screen.getByLabelText('Trigger').click();
+    expect(await screen.findByText('Content')).toBeVisible();
+
+    // Clicking outside the dropdown should close it
+    screen.getByText('Header').click();
+    expect(await screen.findByText('Content')).not.toBeVisible();
```
ddouglasz, line 63:
```suggestion
  await act(async () => {
      // Clicking outside the dropdown should close it
    screen.getByText('Header').click();
    expect(await screen.findByText('Content')).not.toBeVisible();
    });
```

I was able to clear the error by using the suggested act to wait for the state to be completed before asserting, from what I found.

ddouglasz (commented):
Thank you 💯

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
ddouglasz: `import { act } from 'react-dom/test-utils';`

FilPob:
@CarlosCortizasCT thanks! looks awesome! I have just one issue: The hover effect in the case bellow should be bg-color: primary-98. The same as in a select menu, or the primary action dropdown 
image

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
ddouglasz: fair point. 
I saw some js tests using it although act imported from a different package.

Maybe another alternative could to waitFor.

```suggestion
    await waitFor(() => {
      screen.getByText('Header').click();
      // Check if Content is not visible
      const contentElement = screen.queryByText('Content');
      expect(contentElement).not.toBeVisible();
    });
```

CarlosCortizasCT:
> @CarlosCortizasCT thanks! looks awesome! I have just one issue: The hover effect in the case bellow should be bg-color: primary-98. The same as in a select menu, or the primary action dropdown image

Sorry about that 🤦 

It has been fixed here: 9cc149d03392943d5732741341e5eec28645b519

Please give it another try when you have some time.

FilPob:
perfect now. Thanks :)

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
CarlosCortizasCT: I also tried that one but didn't work in my machine.
Did it work for you?

ddouglasz: Yes, it works.

ddouglasz: image

CarlosCortizasCT: But bear in mind the tests won't fail locally. They will fail on CI because over there we raise an error if we use `console.error`.

Maybe you can create another PR our of this one with that change and check if it works over there, because it failed when I tried it.

ddouglasz: Okay, let me try that now.

ddouglasz: Looks like its just the vrt that's failing now. But I cannot see that error on this PR: https://github.com/commercetools/ui-kit/pull/2771

packages/components/dropdowns/dropdown-menu/docs/additional-info.md
```diff
@@ -0,0 +1,15 @@
...
+
+When using the list floating panel, the `DropdownMenu` component exposes an inner sub-component called `DropdownMenu.ListMenuItem` that should be used to render each item in the list.
+
+Clicking on an item will close the panel and call the `onClick` callback with the item's value.
+
+### Properties
+
+| Props        | Type                                                | Required | Default | Description                                       |
+| ------------ | --------------------------------------------------- | :------: | ------- | ------------------------------------------------- |
+| `isDisabled` | `union`<br/>Possible values:<br/>`'left' , 'right'` |          | `false` | Whether the item should be disabled.              |
```
kark, line 13:
```suggestion
| `isDisabled` | `boolean` |          | `false` | Whether the item should be disabled.              |
```

packages/components/dropdowns/dropdown-menu/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+When using the list floating panel, the `DropdownMenu` component exposes an inner sub-component called `DropdownMenu.ListMenuItem` that should be used to render each item in the list.
+
+Clicking on an item will close the panel and call the `onClick` callback with the item's value.
+
+### Properties
+
+| Props        | Type                                                | Required | Default | Description                                       |
+| ------------ | --------------------------------------------------- | :------: | ------- | ------------------------------------------------- |
+| `isDisabled` | `union`<br/>Possible values:<br/>`'left' , 'right'` |          | `false` | Whether the item should be disabled.              |
```
kark, line 110:
```suggestion
| `isDisabled` | `boolean` |          | `false` | Whether the item should be disabled.              |
```

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(screen.queryByText('Content')).not.toBeVisible();
+
+    // Open the dropdown
+    screen.getByLabelText('Trigger').click();
+    expect(await screen.findByText('Content')).toBeVisible();
+
+    // Clicking outside the dropdown should close it
+    act(() => {
+      screen.getByText('Header').click();
+    });
```
kark, line 64: I think `act` is needed here, due to [this effect](https://github.com/commercetools/ui-kit/blob/842892e4a33ece64c64dd741b532ce4126164941/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.tsx#L88-L93) which changes component state after click. 
Just as a side note, I noticed though that changing this line to `screen.getByRole('button', { name: 'Trigger' }).click();` makes all work without `act`

kark (commented):
Thanks Carlos 🙇‍♂️ Massive work!

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
CarlosCortizasCT: Thanks @kark 

However, I still don't understand as that effect does not change the component state but just removes an event listener from the `window` object.

ddouglasz (approved):
Really great work, Carlos 🙌🏽. 
Looks good to me 🚀

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
kark: Sorry for not being precise, the effect doesn't change the state directly, the event handler it adds does. It's hard for me to reason about it in details but I guess reasoning aligns with what was mentioned [here](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#the-dreaded-act-warning:~:text=If%20you%27re%20still-,experiencing,-the%20act%20warning)

packages/components/dropdowns/dropdown-menu/src/trigger/dropdown-menu-trigger.tsx
```diff
@@ -0,0 +1,18 @@
+import { ReactNode, forwardRef } from 'react';
+
+export type TDropdownTriggerProps = {
+  onClick: () => void;
+  children: ReactNode;
+};
+const DropDownTrigger = forwardRef<HTMLDivElement, TDropdownTriggerProps>(
```
kark, line 7: nit: 
```suggestion
const DropdownTrigger = forwardRef<HTMLDivElement, TDropdownTriggerProps>(
```
(for consistency)

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.visualroute.jsx
```diff
@@ -0,0 +1,36 @@
...
+    <Spec label="default dropdown menu">
+      <DropdownMenu triggerElement={<SecondaryButton label="Trigger default dropdown" />}>
+        <Spacings.Stack>
+          <h2>Some headline</h2>
+          <p>Some content</p>
+        </Spacings.Stack>
+      </DropdownMenu>
+    </Spec>
+
+    <Spec label="opened default menu dropdown">
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: a157a52d6437a364437fbed70f6e79a1b2f6d1e2

packages/components/dropdowns/dropdown-menu/docs/additional-info.md
CarlosCortizasCT: Updated here: a157a52d6437a364437fbed70f6e79a1b2f6d1e2

packages/components/dropdowns/dropdown-menu/README.md
CarlosCortizasCT: Updated here: a157a52d6437a364437fbed70f6e79a1b2f6d1e2

kark (approved):
🌟

PR #2774, opened 2024-04-10
feat(select-input): update styling of select input component for opti…

Sarah4VT:
…onal condensed view

#### Summary

Update the styling of the select-input component to allow for a condensed view.

## Description

The [UX design](https://www.figma.com/file/PaImaWorWz1RbV85gqw4MH/Filter-%26-Table-controls?type=design&node-id=1133-128124&mode=design&t=Mwxpkjj2wgYVa21v-4) is as follows:

image

## Results
image

isMulti
image


packages/components/inputs/select-utils/src/types.ts
```diff
@@ -0,0 +1,15 @@
+import type SelectComponents from 'react-select';
+
+type ComponentProps<T> = T extends React.ComponentType<infer P> ? P : never;
```
Sarah4VT, line 3: This is a conditional type that extracts the props type from a React component type.

packages/components/inputs/select-utils/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type TReactSelectCustomComponentsProps = {
```
Sarah4VT, line 5: This is a mapped type that transforms each property in the SelectComponents type to its corresponding prop type. It does this by iterating over each key K in SelectComponents and mapping it to ComponentProps<SelectComponents[K]>. This results in an object type where each property is the prop type of the corresponding component in SelectComponents. The [keyof SelectComponents] part then creates a union of all these prop types. If a new key is added to SelectComponents by the react-select team, TReactSelectCustomComponentsProps will automatically include the prop type for the new key.

packages/components/inputs/select-utils/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  [K in keyof SelectComponents]: ComponentProps<SelectComponents[K]>;
+}[keyof SelectComponents];
+
+export type TSelectInputCustomComponentProps<
```
Sarah4VT, line 9: This is a generic type that extends TReactSelectCustomComponentsProps. It represents the props for a custom component that can be used as an input in a select component. It adds a selectProps property to the props type T, which includes all the properties from T['selectProps'] and an optional isCondensed property. This allows you to pass additional props to the select component through the selectProps property.

jmcreasman (approved):
Pulled down and tested locally, seems to work as expected 🚀

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
```diff
@@ -5,20 +5,22 @@ import { useIntl } from 'react-intl';
...
+import { TSelectInputCustomComponentProps } from '../types';
 
 type TInnerProps = {
   ref: LegacyRef<HTMLButtonElement>;
 } & JSX.IntrinsicElements['button'];
 
 export type TClearIndicatorProps = {
   innerProps: TInnerProps;
-} & ClearIndicatorProps;
+} & TSelectInputCustomComponentProps<ClearIndicatorProps>;
```
kark, line 16: Maybe we could simply use the same pattern as seen here: 
https://github.com/commercetools/ui-kit/blob/c2447dbae4048b2a06017c6ef3db88efdbc955fd/packages/components/inputs/select-utils/src/tag-remove/tag-remove.tsx#L9-L15 for typing the `selectProps` prop

packages/components/inputs/select-input/src/select-input.visualroute.jsx
```diff
@@ -113,6 +122,16 @@ const DefaultRoute = () => (
...
+    <Spec label="with multiple values selected and condensed">
+      <SelectInput
+        value={['one', 'two']}
+        onChange={() => {}}
+        options={options}
+        isMulti={true}
+        isCondensed={true}
+        horizontalConstraint={7}
+      />
+    </Spec>
```
kark, line 134: It appears that the height of the multivalue wrapper is set to `32px`, which prevents us from seeing the condensed padding.
| Current  | Expected  |
|---|---|
| Screenshot 2024-04-15 at 09 23 04  | Screenshot 2024-04-15 at 09 25 59  |

kark (commented):
Thanks, looks good overall from my end

FilPob:
Hey @Sarah4VT thanks for the updates. Some points here:

1. I forgot about the leftIcon prop. This one should be also set to 16px when condensed.
2. Secondly the version when `isCondensed` and `isMulti` is turned on is not quite correct. The overall height of the input should be 32px. If you select multiple values the Input's height grows to from 32 to 38px. 
3. The "x" icon inside the tag should be also 12px when `isCondensed` is turned on 
image

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -165,7 +166,9 @@ const getHoverInputBorderColor = (props: TProps) => {
 const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
   return {
     ...base,
-    fontSize: designTokens.fontSize30,
+    fontSize: props.isCondensed
+      ? designTokens.fontSize20
+      : designTokens.fontSize30,
```
ddouglasz, line 171: I think we should have a token for input when condensed since we are doing this change for multiple input components.
```suggestion
    fontSize: props.isCondensed
      ? designTokens.fontSizeForInputWhenCondensed
      : designTokens.fontSize30,
```
@commercetools/shield-team-design-system what do you think ?

ddouglasz (commented):
Awesome work 🙌🏽
Just a minor feedback.

packages/components/inputs/select-utils/src/clear-indicator/clear-indicator.tsx
CarlosCortizasCT: I think it would be a better approach what we are proposing here as not only known custom components but the ones provided by consumers would need to extend the typing.

Please, reach out to me and we can discuss this in a meeting if you think we're going in the wrong direction.

Sarah4VT:
> Hey @Sarah4VT thanks for the updates. Some points here:
> 
>     1. I forgot about the leftIcon prop. This one should be also set to 16px when condensed.
> 
>     2. Secondly the version when `isCondensed` and `isMulti` is turned on is not quite correct. The overall height of the input should be 32px. If you select multiple values the Input's height grows to from 32 to 38px.
> 
>     3. The "x" icon inside the tag should be also 12px when `isCondensed` is turned on
> 
> 
> image

Fixed in [this commit](https://github.com/commercetools/ui-kit/pull/2774/commits/bf8794eff8a421d3f6c7ac8a4a06fe13827605ef)

packages/components/inputs/select-input/src/select-input.visualroute.jsx
Sarah4VT: Fixed in [this commit](https://github.com/commercetools/ui-kit/pull/2774/commits/bf8794eff8a421d3f6c7ac8a4a06fe13827605ef)

packages/components/inputs/select-utils/src/export-types.ts
```diff
@@ -1,10 +1,7 @@
 export type { TClearIndicatorProps } from './clear-indicator/clear-indicator';
-export type {
-  TTagRemoveSelectProps,
```
Sarah4VT, line 3: Not sure if this constitutes a breaking change?

packages/components/inputs/select-utils/src/export-types.ts
```diff
@@ -1,10 +1,7 @@
...
-  TTagRemoveProps,
-} from './tag-remove/tag-remove';
+export type { TTagRemoveProps } from './tag-remove/tag-remove';
 export type {
   TWrapperWithIconProps,
-  TWrapperWithIconSelectProps,
```
Sarah4VT, line 8: same here

packages/components/inputs/select-input/README.md
```diff
@@ -73,6 +73,7 @@ export default Example;
...
 | `components`               | `ReactSelectProps['components']`                                                                         |          |             | Map of components to overwrite the default ones, see what components you can override&#xA;<br>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                    |
+| `isCondensed`              | `boolean`                                                                                                |          |             | Is the component is displayed in a condensed layout                                                                                                                                                                                                                                                                                                                                      |
```
CarlosCortizasCT, line 76: I think we need to regenerate the Readme in order to get the latest JSDoc of this property

packages/components/inputs/select-utils/src/export-types.ts
CarlosCortizasCT: That's a good point.

This would actually be a breaking change, although this package is more internal than external.

What we can do is to keep those types just for backwards compatibility and mark them as deprecated. We can remove them later when we plan for a major `ui-kit` release.

Since this types are not needed anymore in their components, I would maybe add something like this to the `types.ts` file:

```ts

/**
 * @deprecated Use `TTagRemoveProps['selectProps]` instead
 */
export type TTagRemoveSelectProps = {
  isReadOnly: boolean;
} & MultiValueGenericProps['selectProps'];

/**
 * @deprecated Use `TWrapperWithIconProps['selectProps]` instead
 */
export type TWrapperWithIconSelectProps = {
  iconLeft?: ReactElement;
};
```

FilPob:
UI wise looks correct from my side. 
one question for @ddouglasz though. The tags inside the input when `isMulti` do not (again) have the same styles as in the recent updates in the tag component. I know you worked on this in a separate PR but not sure how that affects this PR

CarlosCortizasCT:
> @Sarah4VT thanks! UI wise looks correct now one question for @ddouglasz though. The tags inside the input when `isMulti` do not (again) have the same styles as in the recent updates in the tag component. I know you worked on this in a separate PR but not sure how that affects this PR

@ddouglasz opened [this PR](https://github.com/commercetools/ui-kit/pull/2780) to address those styles so, whatever gets merged first will be updated in the other PR.

FilPob:
Am just a bit hesitant  to approve the Percy screenshots cause the "when condensed" screenshot slipped into the 3rd position and therefore the rest of the screenshots are marked as they would have changes. Would it be possible to put the "when condensed" screenshot on the end of the list?

CarlosCortizasCT:
> Am just a bit hesitant to approve the Percy screenshots cause the "when condensed" screenshot slipped into the 3rd position and therefore the rest of the screenshots are marked as they would have changes. Would it be possible to put the "when condensed" screenshot on the end of the list?

@FilPob I think Sarah tried to follow the grouping we already had and it seems it makes sense to keep what she did.

I triple checked the changes (it's just a matter of staring stronger at the screenshots 🤣) and changes seem ok.

By the way, you can also see the new remove tags in action.

FilPob:
@CarlosCortizasCT thanks! Unfortunately the new tag styles did somehow overwrite Sarah's implementation. 🙈 Now the `isMulti` when `condensed` looks wrong again. The overall height of the Input should be 32px when condensed. This was addressed already by Sarah in some of the previous commits
image

FilPob (approved):

FilPob:
@CarlosCortizasCT looks good now! Many thanks!

packages/components/inputs/select-utils/src/export-types.ts
CarlosCortizasCT: I included these types here: cb411c1ce2bccbe8808d466e8f87a738c3ab918c

packages/components/inputs/select-input/README.md
CarlosCortizasCT: Readme updated here: cc7989be06c2d337ef6f7408629d9b0909a912a5

kark (approved):
Big thanks for the contribution and for the type refactor 🧹

Sarah4VT:
Thanks @CarlosCortizasCT  for making the fixes! Merge away!

Issue #2773, opened 2024-04-10
SelectableSearchInput state doesn't update when input changes


ChristianMoll:
**Describe the bug**

The SelectableSearchInput component has a state `searchValue`, this is displayed within the text input.
The component also accepts an input object `value` which has a property `text`.
If the `text` property of the  `value` input changes the value that is sent when submitting the input does change but the shown value inside the input doesn't.
One simple way to avoid this would be to add a useEffect to the SelectableSearchInput to update the state when the input changes. i.E.

````
 useEffect(() => {
    setSearchValue(
      props.value.text
    );
  }, [props.value.text]);

````
**To Reproduce**

1. Use the SelectableSearchInput
2. Update the props.value.text through code
3. Press enter or click the SearchIcon button
4. The value being passed to the onSubmit handler is not the same being shown in the input.

**Expected behavior**
The shown value is in sync with the value that is sent on submitting.

ChristianMoll:
Hi @CarlosCortizasCT 

Thanks for the fast response.

In our case it is a problem when we restore the state of the SelectableSearchInput from the myview in the customer search.

Other hypothetical scenarios could be restoring a form draft that uses a SelectableSearchInput when the user wants to continue with it or a SelectableSearchInput which is pre-filled based on the value of another input.

CarlosCortizasCT:
Related to this issue #2573

PR #2776, opened 2024-04-11
Update selectable search input state when the input changes

ddouglasz:
#### Summary

Update selectable search input state when the input changes

Closes https://github.com/commercetools/ui-kit/issues/2773

### DIFF
--- a/.changeset/plenty-snails-draw.md
+++ b/.changeset/plenty-snails-draw.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/selectable-search-input': minor
+---
+
+When the input of the selectable search input changes, it should show the currently updated value on the input field.
--- a/packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
+++ b/packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
@@ -6,6 +6,7 @@ import {
   useState,
   useCallback,
   useRef,
+  useEffect,
 } from 'react';
 import SecondaryIconButton from '@commercetools-uikit/secondary-icon-button';
 import Constraints from '@commercetools-uikit/constraints';
@@ -260,6 +261,11 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
   const legacyDataProps = filterDataAttributes(props);
   const transformedSelectDataProps = transformDataProps(props.selectDataProps);
   const transformedInputDataProps = transformDataProps(props.inputDataProps);
+
+  // Ensure input state is always updated when the input changes
+  useEffect(() => {
+    setSearchValue(props.value.text);
+  }, [props.value.text]);
 
   const optionsWithoutGroups = props.options.flatMap((option) => {
     if (isOptionObject(option)) {


CarlosCortizasCT:
To be honest, I have mixed feelings with this.

The `value` property is currently used more like a default value to be used when initializing the component than the value itself as this component is more an uncontrolled one but it seems the requirement is to be able to control its state from the outside (controlled component).

I think we're mixing both strategies here.

My proposal would be to have both `value` and `defaultValue` properties and make the component be able to work in the two modes.

What do others think about it? @commercetools/shield-team-fe

ddouglasz:
> My proposal would be to have both `value` and `defaultValue` properties and make the component be able to work in the two modes.

updated here: https://github.com/commercetools/ui-kit/pull/2776/commits/be21255d07a7e5d918a22d839099b251a51147aa
Thank you

ddouglasz:
>However, I still would like to read the rest of the team point of view about this.
>cc/ commercetools/shield-team-fe

That makes sense and this may likely be a breaking change.
 @kark @chloe0592 @emmenko Please what would you suggest?

chloe0592:
I agree with the proposed changes. Introducing both `value` and `defaultValue` would offer flexibility for both controlled and uncontrolled states. 👍

kark:
Thanks for checking with the team. 
> My proposal would be to have both `value` and `defaultValue` properties and make the component be able to work in the two modes.

This proposal makes sense to me and this approach is justified in my opinion

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -89,6 +90,10 @@ export type TSelectableSearchInputProps = {
    * Value of the input. Consists of text input and selected option.
    */
   value: TValue;
+  /**
+   * Default value of the input. Consists of text input and selected option.
+   */
+  defaultValue: TValue;
```
emmenko, line 96: Is this supposed to be a required field?

If yes, it should probably be considered a breaking change.
If not, it should be handled as an optional field.

emmenko (commented):
From my side, I'm also good with having the two props. Thanks for the proposal / fix!

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
ddouglasz: Oh true, seems to me like it should be optional, assuming we do not expect users to always provide a default value.
The initial implementation already sets it to an empty string in the absence of a default value.

CarlosCortizasCT: The problem I see is that we currently only have the `value` property but we're actually using it as default value.
If we now change the behaviour of the already exposed `value` property to drive the internal state and add the new `defaultValue`, all current consumers will be affected so hence this would be a breaking change as Nicola mentioned.

We need to double think if we can somehow come up with a solution that can avoid the breaking change.

CarlosCortizasCT: Nicola and I had a conversation about this and we agreed releasing a major `ui-kit` version just for this change currently is not convinient.
Although the long term solution is having both a `defaultValue` and a `value` properties to handle both controlled and uncontrolled states, we think it could benefit us to hold that change until we have a better reason to release a breaking change version.

What we can do in the meantime is to introduce a new temporary property (example: `experimentalValue`) that can be used for the controlled version of the component. Then, we will have the two properties that we need, but with the "wrong" names and at least we will have a escape hatch for those (very few) use cases we know so far that require the controlled version of the component.

Once we get to the point when it will be convenient to release an `ui-kit` major version, we will rename those properties and probably publish a _codemod_ which consumers can run in their codebase, to migrate from one names to the others automatically.

ddouglasz:
All recent feedback updated here: https://github.com/commercetools/ui-kit/pull/2776/commits/3a3bb9dd7d2468339b5b67113fb7a6916c8957bc

.changeset/plenty-snails-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/selectable-search-input': minor
+---
+
+When the input of the selectable search input changes, it should show the currently updated value on the input field.
```
CarlosCortizasCT, line 5: Perhaps we can provide just a little context:

```suggestion
We're introducing a **temporary** property to handle an internal case.
Please do not use it as it will be removed in the near future.
```

packages/components/inputs/selectable-search-input/src/selectable-search-input.spec.tsx
```diff
@@ -134,6 +134,35 @@ describe('SelectableSearchInput', () => {
     ).toHaveTextContent('Bar');
   });
 
+  it('should show the passed experimental value when controlled externally', () => {
```
CarlosCortizasCT, line 137: Thanks for working in these new tests.

I think we should also test that, when the consumer state changes, the internal `SelectableSearchInput` component state also change when using the new property but does not when using the old one.

.changeset/plenty-snails-draw.md
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2776/commits/1e62e9989cd8dd9e9f24cfa0f87955cf7e681ad3
Thanks

packages/components/inputs/selectable-search-input/src/selectable-search-input.spec.tsx
ddouglasz: Sorry, maybe I do not understand this clearly but it seems like that is what the first test checked for.

CarlosCortizasCT: My bad.
Thanks for explaining in a meeting.

ddouglasz:
@CarlosCortizasCT PR updated based on our last discussion here : https://github.com/commercetools/ui-kit/pull/2776/commits/5d1edeafec747dca4d3dc786aa5bba566f58e595

Thank you.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -296,7 +298,7 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
     componentName: 'SelectableSearchInput',
   });
 
-  const { onFocus, onBlur, name } = props;
+  const { onFocus, onBlur, name, onChange } = props;
```
CarlosCortizasCT, line 301: If we want to destructure the `onChange` prop, I think we should use the destructured value everywhere in the file. Currently we have been using `props.onChange` in other parts of the file and we should be consistent.

packages/components/inputs/selectable-search-input/src/selectable-select.tsx
```diff
@@ -50,23 +51,6 @@ const SelectableSelect = (props: TSelectableSelect) => {
     dropdownHasFocus: props.dropdownHasFocus,
   }) as ReactSelectProps['styles'];
 
-  const { onChange, name, id, textInputRef } = props;
```
CarlosCortizasCT, line 53: Do we really need to mode this code to the parent?

Perhaps we can just keep this here and, in the parent, listen to the `onChange` event, update the selected dropdown state and bubble up the event.

ddouglasz: Thank you for the observation, I actually noticed all other methods ( `handleContainerBlur`, `handleDropdownBlur`, `handleDropdownFocus`) are calculated in the `selectable-search-input` and then passed to the `selectable-select` as props, except the `handleDropdownChange`. It felt a bit inconsistent to me. That's why I moved it there. 
I didn't see a practical reason to leave it there. Either options are still fine by me. What do you think?

CarlosCortizasCT: Fair point 👍

CarlosCortizasCT:
I think we still need to update the Storybook story. As discussed last week, currently it's implemented as if it's controlling the state of the component where it is not.

Also, now that we've introduced a new function called `handleDropdownChange`, then one we have to handle the input change has a potentially confusing name (`handleChange`).
I suggest we update its name to use the same pattern we use for other callbacks we pass to the input (`handleTextInputFocus` and `handleTextInputBlur`) so for instance it could be something like `handleTextInputChange`.

CarlosCortizasCT:
I also was wondering if the last change we've introduced with managing the dropdown option could be a breaking change for current consumers.
I think we should check some examples to make sure.

ddouglasz:
> I also was wondering if the last change we've introduced with managing the dropdown option could be a breaking change for current consumers. I think we should check some examples to make sure.

Great point 💯
Let me do some findings on this. If that is the case, we may need to go back to the initial implementation since we are trying to avoid a breaking change for now.

ddouglasz:
>I think we still need to update the Storybook story. As discussed last week, currently it's implemented as if it's controlling the state of the component where it is not.

updated here: https://github.com/commercetools/ui-kit/pull/2776/commits/c0e73a3f0769fd1e2cfcd1cb402de6f3571dc480
Thank you

packages/components/inputs/selectable-search-input/src/selectable-search-input.story.js
```diff
@@ -112,15 +108,6 @@ storiesOf('Components|Inputs', module)
             id={text('id', '')}
             name={name}
             value={value}
-            onChange={(event) => {
```
CarlosCortizasCT, line 115: **nit**: we might want to keep the call to the `action` as we had before so we can check the callback is indeed called from the Storybook interface.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -424,7 +424,7 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
       }
       textInputRef.current?.focus();
     },
-    [onChange, selectablSearchInputId, name]
+    [props, selectablSearchInputId, name]
```
CarlosCortizasCT, line 427: this will make the callback to be regenerated on every props updates. Perhaps we can use the `props.onChange` property in the array so it gets recreated only if that prop in new.

CarlosCortizasCT (approved):

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2776/commits/12a589e2aef12601ea4cd772895d4b79145d2bc7 and here: https://github.com/commercetools/ui-kit/pull/2776/commits/12d690966cbe724414dbd466a44c1191f082e583

Thank you.

PR #2780, opened 2024-04-12
Apply new tag styles in select inputs multivalue

ddouglasz:
#### Summary

Apply new tag styles in select inputs multi value

## Description

Before:

| | isDisabled: false  | isDisabled: true |
|--------------------|--------|-----------------|
|Before |image |image|
|After | image|image|


FilPob (approved):

FilPob:
Thanks :)

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -449,15 +449,22 @@ const menuPortalStyles = (props: TProps) => (base: TBase) => ({
...
-const multiValueStyles = () => (base: TBase) => {
+const multiValueStyles = (props: TProps) => (base: TBase) => {
   return {
     ...base,
     display: 'flex',
     alignItems: 'center',
     height: '32px',
-    backgroundColor: designTokens.colorNeutral95,
-    padding: '0',
+    backgroundColor: (() => {
```
CarlosCortizasCT, line 458: Perhaps we can extract this to a function an place it at the top of the file the same we do for other similar calculations (eg: `renovate/major-all-flopflip-packages`)

ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2780/commits/415007433042f7e6889d00d7384adab592b95ed9

Thank you

CarlosCortizasCT: Thanks for updating the `controlStyles` part, but I think we should also do the same here for the `multiValueStyles`.

ddouglasz: Oh! missed that.
Updated here: https://github.com/commercetools/ui-kit/pull/2780/commits/50df31e2d57a38dba5859cf4ce6117e352201809

Thank you

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks!

kark (approved):
Thanks, looks great!

PR #2781, opened 2024-04-12
UI Kit | FE | SearchTextInput add new isCondensed prop

jmcreasman:
Adding a new boolean isCondensed prop that when set to true condenses the input height, font size and icon size.

isCondensed = false
Screenshot 2024-04-12 at 11 48 40 AM

isCondensed = true
Screenshot 2024-04-12 at 11 48 48 AM


FilPob (approved):

FilPob:
Looks good from my side, thanks! :)

.changeset/good-mayflies-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/search-text-input': minor
+---
+
+"feat(searchtext-input): add new isCondensed prop"
```
ddouglasz, line 5: Add `isCondensed` prop that when set to true, condenses the input height and font size.

packages/components/inputs/search-text-input/src/search-text-input.styles.ts
```diff
@@ -53,6 +54,9 @@ const getSearchTextInputStyles = (props: TInputProps) => [
     &:hover {
       background-color: transparent !important;
     }
+    font-size: ${props.isCondensed
+      ? `${designTokens.fontSize20}`
+      : `${designTokens.fontSize30}`};
```
ddouglasz, line 59: I am not so sure of the full context of this PR on why do we need `isCondensed` for just the SearchTextInput?

Also, Maybe we can generate a token for this state
```suggestion
      ? `${designTokens.fontSizeForInputWhenCondensed}`
      : `${designTokens.fontSize30}`};
```

So at least if this change is going to be made across other input components later on, then we have a variable for it. @FilPob what do you think ?

.changeset/good-mayflies-notice.md
jmcreasman: Added!

packages/components/inputs/search-text-input/README.md
```diff
@@ -60,6 +60,7 @@ export default Example;
 | `onSubmit`             | `Function`<br/>[See signature.](#signature-onSubmit)                                                  |    ✅    |           | Handler when the search button is clicked.                                                                                |
 | `onReset`              | `Function`<br/>[See signature.](#signature-onReset)                                                   |    ✅    |           | Handler when the clear button is clicked.                                                                                 |
 | `isAutofocussed`       | `boolean`                                                                                             |          |           | Focus the input on initial render                                                                                         |
+| `isCondensed`          | `boolean`                                                                                             |          |           | Whether the options appear in a condensed layout                                                                          |
```
CarlosCortizasCT, line 63: You need to generate the Readme again to get the latest JSDoc comment from the propw type.

jmcreasman: Thanks!

CarlosCortizasCT (approved):
Thanks! 🙇

PR #2784, opened 2024-04-15
Update design tokens readme reference

ddouglasz:
## Description
Update design tokens readme reference

The [custom-properties.ts](https://github.com/commercetools/ui-kit/blob/main/design-system/src/custom-properties.ts) link seems to have been updated to [design-tokens.ts](https://github.com/commercetools/ui-kit/blob/main/design-system/src/design-tokens.ts) but we have not updated it on the README documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ postcss([
 You can also access the JavaScript variables like this
 
 ```js
-import { customProperties } from '@commercetools-uikit/design-system';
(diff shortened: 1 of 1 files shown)


kark (approved):
Thanks!

CarlosCortizasCT (approved):
Thanks! 👍

PR #2785, opened 2024-04-17
Remove unused old theme

ddouglasz:
#### Summary
Remove the unused (old) theme in UI-Kit

## Description

- [x] Clean up most tokens, especially in the `definition.yaml`,  tokens that are only used in fewer places. Use the direct value where the styling is referenced in the file
- [x] Move the `recolouring` tokens to the `default` tokens. Remove any token in the initial `default` tokens that is not used.


packages/components/buttons/primary-button/src/primary-button.styles.ts
```diff
@@ -34,7 +34,7 @@ const getButtonStyles = (
     align-items: center;
     color: ${designTokens.colorSurface};
     transition: background-color ${designTokens.transitionLinear80Ms};
-    padding: ${designTokens.paddingForButton};
+    padding: 0 var(--spacing-30);
```
CarlosCortizasCT, line 37:
```suggestion
    padding: 0 ${designTokens.spacing30};
```

packages/components/inputs/input-utils/src/styles.ts
```diff
@@ -93,7 +93,7 @@ const getInputStyles = (props: TInputProps) => {
       : 'unset'}; /* fix for mobile safari */
     outline: none;
     overflow: hidden;
-    padding: ${designTokens.paddingForInput};
+    padding: 0 var(--spacing-30);
```
CarlosCortizasCT, line 96:
```suggestion
    padding: 0 ${designTokens.spacing30};
```

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -188,8 +188,8 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
     })(),
     padding:
       props.appearance === 'quiet'
-        ? designTokens.paddingForInputAsQuiet
-        : designTokens.paddingForInput,
+        ? '0 var(--spacing-20)'
+        : '0 var(--spacing-30)',
```
CarlosCortizasCT, line 192: Please use de designTokens from the `designTokens` object.

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -211,7 +211,7 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
       backgroundColor: (() => {
         if (!props.isDisabled && !props.isReadOnly) {
           if (props.appearance === 'quiet') {
-            return designTokens.backgroundColorForInputAsQuietWhenHovered;
+            return 'hsla(240, 64%, 58%, 4%)';
```
CarlosCortizasCT, line 214: This is the same as the `color-primary` design token plus `alpha` 4%.
I'm wondering if there's any way we can leverage the existing design token.

packages/components/buttons/primary-button/src/primary-button.styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/098974e6596998284f5d665ce8a0a416310509e0
Thank you

packages/components/inputs/input-utils/src/styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/098974e6596998284f5d665ce8a0a416310509e0
Thank you

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/098974e6596998284f5d665ce8a0a416310509e0
Thank you

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: Yes this one is a bit tricky. I was thinking maybe I could pass the alpha value and form a hsla like this `hsla(${designTokens.colorPrimary}, 4%)` but this will output `(hsl(240, 64%, 58%), 4%)`.
Perhaps I could write a function to extract the value. What do you think?

ddouglasz: I tried the color-mix but it seems like we will likely be circling back to the complexity we are trying to avoid.
`background-color: color-mix(in hsl, var(--color-primary), hsla(0deg, 0%, 0%, 4%));`  but it gives a different result from what we want to achieve.
image

design-system/materials/internals/definition.yaml
```diff
@@ -99,28 +51,75 @@ choiceGroupsByTheme:
...
         color-solid-05: 'hsl(0deg 0% 10% / 5%)'
         color-solid-10: 'hsl(0deg 0% 10% / 10%)'
         color-surface: '#fff'
         color-transparent: 'transparent'
-        color-success: 'hsl(152, 77%, 39%)' # Design token copied from recolouring theme
-        color-success-25: 'hsl(155, 84%, 20%)' # Design token copied from recolouring theme
-        color-success-40: 'hsl(155, 90%, 24%)' # Design token copied from recolouring theme
-        color-success-85: 'hsl(144, 69%, 80%)' # Design token copied from recolouring theme
-        color-success-95: 'hsl(141, 76%, 92%)' # Design token copied from recolouring theme
+        color-primary: 'hsl(240, 64%, 58%)'
```
CarlosCortizasCT, line 60: Any reason to move this block of tokens from where they were? 🤔

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -57,8 +57,8 @@ const getSizeStyles = (size: TIconButtonProps['size']) => {
   switch (size) {
     case 'small':
       return css`
-        height: ${designTokens.heightForButtonAsSmall};
-        width: ${designTokens.heightForButtonAsSmall};
+        height: 16px;
```
CarlosCortizasCT, line 60: **nit**:
In this case, since we have tokens for the other button sizes, I would suggest to keep a token for the small size as well.
Even if we are just using it here, perhaps we will need to adjust other buttons and it will also helps us to keep consistency here and maintainability.

packages/components/buttons/primary-button/src/primary-button.styles.ts
```diff
@@ -45,7 +45,7 @@ const getButtonStyles = (
         &,
         &:active,
         &:hover {
-          background-color: ${designTokens.backgroundColorForButtonWhenDisabled};
```
CarlosCortizasCT, line 48: This token was used here and also in the `SecondaryButton` styles so perhaps it would make sense to keep it. 
It seems like a token all buttons may share.

What do you think?

packages/components/buttons/secondary-button/src/secondary-button.styles.ts
```diff
@@ -124,13 +124,13 @@ const getSizeStyles = (size: TSecondaryButtonProps['size']) => {
     case 'medium':
       return css`
         height: ${designTokens.heightForButtonAsMedium};
-        border-radius: ${designTokens.borderRadiusForButtonAsMedium};
+        border-radius: ${designTokens.borderRadius4};
```
CarlosCortizasCT, line 127: I'm wondering if we need this style as both for `medium` and `big` size is the same value and also this is the default value in the `secondary-button.tsx` file below .

packages/components/inputs/select-utils/src/create-select-styles.ts
CarlosCortizasCT: These one haven't been updated.

packages/components/inputs/select-utils/src/create-select-styles.ts
CarlosCortizasCT: Yes, you're right it doesn't seem like an feasible solution.
We can however add a comment to let our future selves about this is the primary color with an `alpha` channel we only need here.

Thanks for giving it a try.

CarlosCortizasCT:
@ddouglasz Please remember to take a look at Percy and try to find out why we have differences 🙏

design-system/materials/internals/definition.yaml
ddouglasz: Oh wait, we are still leaving the design tokens for the recolouring block of the `choiceGroupsByTheme` as it is right  ?

ddouglasz:
> @ddouglasz Please remember to take a look at Percy and try to find out why we have differences 🙏

Yes, I am looking into it too and I spotted some slight changes that I already discussed with @FilPob  about the current border color for input when error that is not correct on our currently deployed version.
Thank you

packages/components/buttons/primary-button/src/primary-button.styles.ts
ddouglasz: That makes sense, I left most of the button tokens as they seemed commonly shared. I guess I missed this one.

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/065e27db2ab90e9c6103bb78679b8c7d9ee5283f
Thank you

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/065e27db2ab90e9c6103bb78679b8c7d9ee5283f
Thank you

packages/components/buttons/icon-button/src/icon-button.styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/065e27db2ab90e9c6103bb78679b8c7d9ee5283f
Thank you

packages/components/inputs/select-utils/src/create-select-styles.ts
```diff
@@ -210,8 +210,8 @@ const controlStyles = (props: TProps) => (base: TBase, state: TState) => {
     })(),
     padding:
       props.appearance === 'quiet'
-        ? designTokens.paddingForInputAsQuiet
-        : designTokens.paddingForInput,
+        ? designTokens.spacing30
+        : designTokens.spacing30,
```
CarlosCortizasCT, line 214: Is this ok? Both conditions use the same value.

design-system/materials/internals/definition.yaml
CarlosCortizasCT: I'm just saying that the colors tokens started with the `color-primary` and now they start with the `color-neutral`.
I was expecting to see just the values of the tokens changed, not the tokens moved around.

If it's not clear, I can push a commit to explain what I mean if that's ok for you.

ddouglasz:
> I was expecting to see just the values of the tokens changed, not the tokens moved around.

It is clear to me now. I will an update as expected. 
Thank you

packages/components/inputs/select-utils/src/create-select-styles.ts
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/bdc818a94588d626b008a5a2a56681bad3e1f882

Thank you

CarlosCortizasCT:
> > I was expecting to see just the values of the tokens changed, not the tokens moved around.
> 
> It is clear to me now. I will an update as expected. Updated here: [bdc818a](https://github.com/commercetools/ui-kit/commit/bdc818a94588d626b008a5a2a56681bad3e1f882)
> 
> Thank you

@ddouglasz bear in mind you correctly updated the default theme tokens, but you need to remove the `recouloring` color tokens as well.
We should not have the `recouloring` theme in the `definitions.yml` anymore.

ddouglasz:
> We should not have the recouloring theme in the definitions.yml anymore.
> Also, we need to review the `states`, `variants` and `componentGruops` to remove the entries that are not used.

updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/a9e6c67b87fa5461b3966c967be1b2b47259177b and https://github.com/commercetools/ui-kit/pull/2785/commits/fab99deb46e23b85c57af976c82d441a38ce6906

ddouglasz:
@CarlosCortizasCT @chloe0592  please what do you think about this last [change](https://github.com/commercetools/ui-kit/pull/2785/commits/d48129bb9b203df5e971259433f6fc55bc1892db) to fix the vrt as we no longer support `recolouring` as a list of theme options from our `ThemeProvider`? 🤔

CarlosCortizasCT:
> @CarlosCortizasCT @chloe0592 please what do you think about this last [change](https://github.com/commercetools/ui-kit/pull/2785/commits/d48129bb9b203df5e971259433f6fc55bc1892db) to fix the vrt as we no longer support `recolouring` as a list of theme options from our `ThemeProvider`? 🤔

I think that's a valid change but I would suggest to comment  the `Interactive` [test group](https://github.com/commercetools/ui-kit/blob/SHIELD-1166-remove-unused-old-theme/design-system/src/theme-provider.visualspec.js#L11) from the `theme-provider.visualspec.js` as we don't have more that one theme anymore.
We can leave that commented just in case we introduce a new theme in the future so we have a reference test ready.

Also, I took a look at the repository and I still found some references to the theme we're trying to remove on files [checkbox-input.tsx](https://github.com/commercetools/ui-kit/blob/SHIELD-1166-remove-unused-old-theme/packages/components/inputs/checkbox-input/src/checkbox-input.tsx) and [primary-action-dropdown.tsx](https://github.com/commercetools/ui-kit/blob/SHIELD-1166-remove-unused-old-theme/packages/components/primary-action-dropdown/src/primary-action-dropdown.tsx).

ddouglasz:
>We can leave that commented just in case we introduce a new theme in the future so we have a reference test ready.


>Also, I took a look at the repository and I still found some references to the theme we're trying to remove on files [checkbox-input.tsx](https://github.com/commercetools/ui-kit/blob/SHIELD-1166-remove-unused-old-theme/packages/components/inputs/checkbox-input/src/checkbox-input.tsx) and [primary-action-dropdown.tsx](https://github.com/commercetools/ui-kit/blob/SHIELD-1166-remove-unused-old-theme/packages/components/primary-action-dropdown/src/primary-action-dropdown.tsx).

Thank you for your response, updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/e2c5fd94698266773cd2a0b996421d8d505c867b

design-system/src/theme-provider.visualspec.js
```diff
@@ -8,30 +8,30 @@ describe('ThemeProvider', () => {
...
+//     const localThemeChangeButton = await queries.findByText(
+//       doc,
+//       'change local theme'
+//     );
+//     await localThemeChangeButton.click();
+//     await page.waitForSelector('[data-theme="default"]');
+//     // TODO: uncomment when issue with Percy is resolved
+//     // await percySnapshot(page, 'ThemeProvider - after local theme change');
+//   });
+// });
```
kark, line 37: Perhaps we could just delete that?

kark (commented):
Thanks Douglas, overall looks good to me👍
I think we could also remove `recolouring` theme passed as prop in:
https://github.com/commercetools/ui-kit/blob/2c9b378019e6253c0bbdec2f937a275327658c96/docs/.storybook/decorators/theme-wrapper/theme-wrapper.js#L7
and 
https://github.com/commercetools/ui-kit/blob/2c9b378019e6253c0bbdec2f937a275327658c96/visual-testing-app/src/App.jsx#L32

ddouglasz:
> Thanks Douglas, overall looks good to me👍 I think we could also remove `recolouring` theme passed as prop in:
> 
> https://github.com/commercetools/ui-kit/blob/2c9b378019e6253c0bbdec2f937a275327658c96/docs/.storybook/decorators/theme-wrapper/theme-wrapper.js#L7
> 
> 
> and
> https://github.com/commercetools/ui-kit/blob/2c9b378019e6253c0bbdec2f937a275327658c96/visual-testing-app/src/App.jsx#L32

updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/f00e0a17d61a7a27cf474a3e4ac6d7cb3e0a66ed
Thank you.

design-system/src/theme-provider.visualspec.js
ddouglasz: We agreed to leave it commented out just so we have a sample of the test block in the event we introduce new themes later on.

design-system/src/theme-provider.tsx
```diff
@@ -130,9 +130,11 @@ const useTheme = (parentSelector = defaultParentSelector): TUseThemeResult => {
     attributeFilter: ['data-theme'],
   });
 
+  // TODO - make sure old and new theme return same value as new defaultThemeValue
+  // At least for the remaining places that we still use this function
   const themedValue: TUseThemeResult['themedValue'] = useCallback(
-    (defaultThemeValue, newThemeValue) =>
-      theme === 'default' ? defaultThemeValue : newThemeValue,
+    (defaultThemeValue, _newThemeValue) =>
+      theme === 'default' ? defaultThemeValue : defaultThemeValue,
```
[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/c7e1be8008f1bc8e11b7e2304158a722a4e256d3
Thank you

kark (approved):
Thanks, looks great 💯

packages/components/stamp/src/stamp.tsx
```diff
@@ -77,7 +71,7 @@ const tonesPropsMap = (
       borderColor: designTokens.colorPrimary85,
       color: designTokens.colorSuccess40,
     },
-    iconColor: themedValue('primary', 'success'),
+    iconColor: 'primary',
```
CarlosCortizasCT, line 74: It seems this value be `success`; that's what we had when we were using the theming hook.

Currently we can see the icon not rendering in the expected color:

Screenshot 2024-04-23 at 18 24 11

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/6ba27608f57b6f0121caaf6d2293725bca63a576

Thank you

CarlosCortizasCT:
Almost there 👍 

Could you please include the changesets?

ddouglasz:
> Could you please include the changesets?

updated here: https://github.com/commercetools/ui-kit/pull/2785/commits/7775f7d563735e066359edee6cdab9ac06d4dcc9

Thank you

CarlosCortizasCT:
After updating this branch with the latest changes from main, we have some Percy differences again.

@ddouglasz could you please take a look? 🙏

CarlosCortizasCT (approved):
Thanks a lot Douglas! 🙇

PR #2787, opened 2024-04-19
UI Kit | FE | TextField add new isCondensed prop

jmcreasman:
Adding a new boolean isCondensed prop that when set to true condenses the input height and font size by passing isCondensed into the nested TextInput component.

isCondensed is set to false:
Screenshot 2024-04-19 at 08 58 30

isCondensed is set to true:
Screenshot 2024-04-19 at 08 58 42


CarlosCortizasCT (approved):
Nice and simple 💯 

Many thanks @jmcreasman 🙇

ddouglasz (approved):

PR #2793, opened 2024-04-23
feat(localized-multiline-text-input): add support for isCondensed pro…

Sarah4VT:
…perty on LocalizedMultilineTextInput

#### Summary

Adds support for isCondensed prop on LocalizedMultilineTextInput component to display a condensed layout.

## Description

[Figma Design](https://www.figma.com/file/PaImaWorWz1RbV85gqw4MH/Filter-%26-Table-controls?type=design&node-id=1305-135823&mode=design&t=E8vYa0ewveJolju1-4)
image

Results with isCondensed toggled on
TODO


Sarah4VT:
@CarlosCortizasCT There is some weirdness with this one that I can't figure out. I'm not sure if it has to do with the 3rd party library that we use for this resizable text area. But this !important style for height is getting set somewhere that I think may be calculated based on the initial rendering. This means if you toggle isCondensed in storybook, it doesn't recalculate. Not sure if this is something I need to handle?
Components Inputs - LocalizedMultilineTextInput ⋅ Storybook 2024-04-23 at 4 33 42 PM jpg

kark:
> @CarlosCortizasCT There is some weirdness with this one that I can't figure out. I'm not sure if it has to do with the 3rd party library that we use for this resizable text area. But this !important style for height is getting set somewhere that I think may be calculated based on the initial rendering. This means if you toggle isCondensed in storybook, it doesn't recalculate. Not sure if this is something I need to handle? I did confirm if I have isCondensed turned on, and refresh storybook, it does render the inputs correctly. 🤷‍♀️
> 
> Components Inputs - LocalizedMultilineTextInput ⋅ Storybook 2024-04-23 at 4 33 42 PM jpg

@Sarah4VT After investigating this case, it appears that the observed behavior results from the underlying `<TextareaAutosize>` component from the external library, as you mentioned. More precisely, we've set `cacheMeasurements={true}` 
https://github.com/commercetools/ui-kit/blob/2c9b378019e6253c0bbdec2f937a275327658c96/packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx#L106
which prevents re-renders. In my view, it's very unlikely for the `isCondensed` prop to change once this component is rendered, so I think this behavior is negligible. 
Another option is switching to `cacheMeasurements={false}` but I'm not sure if this might cause any issues. 

@commercetools/shield-team-design-system what is your take on that?

ddouglasz:
> @commercetools/shield-team-design-system what is your take on that?

Thank you Kacper, I noticed it too and was playing around with it a little bit to be sure if switching it to false actually causes any issues. At least, so far, I have not noticed any issues. 
The documentation defines the prop to `Reuse previously computed measurements when computing height of textarea` which seems to stop the height from updating.

Sarah4VT:
@commercetools/shield-team-design-system I went ahead and removed the `cacheMeasurements` value and tested it in storybook and verified that it does allow the size to update when toggling the isCondensed property now in storybook. Hopefully this does the trick. Thanks for the suggestion!

ddouglasz (approved):
Looks good now  💯
Thank you @Sarah4VT

CarlosCortizasCT:
Thank you all for looking into this.

As you already spotted, the issue is because of the usage of the `cacheMeasurements` props which is used to enhance the `TextareaAutosize` performance by avoiding going to the DOM to perform the measurements calculations needed as it's an "expensive" operation that will happen in every render otherwise (eg: typing anything in the input).
I'm not sure how "expensive" this operation can be; maybe it could be seen in low-end devices.

I can think of different options for use to consider:
* Do nothing: In this case we assume (as @kark mentioned) is very unlikely the value of the `isCondensed` prop will change after the component is rendered, so it would be ok to accept the new behaviour.
  * In this case I would say consumers can be confused when coming to our Storybook and seeing that prop does nothing on the component and this is something we want to avoid.
* Expose a prop to control that behaviour: Let's expose the `cacheMeasurements` as part of the component's API and set it to `true` by default. We can explain in the docs of that prop where it can be useful to use it (We would be using it in our Storybook story so the component reacts to changing the `isCondensed` knob).
* Keep track of the `isCondensed` prop history in the `MultilineInput` component: if we have a ref in that component with the last `isConsensed` valued, we can compare it with the new one and, only if it has changed, we set the `cacheMeasurements` to `false`.

Personally I would try to avoid removing that property altogether from the `MultilineInput` components because it will introduce a performance degradation from our baseline.
I believe exposing the `cacheMeasurements` property and having it to `true` by default is the most convenient solution.

packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx
```diff
@@ -52,6 +59,20 @@ const MultilineInput = (props: TMultiLineInputProps) => {
   const { onHeightChange } = props;
   const ref = useRef<HTMLTextAreaElement | null>(null);
 
+  const [prevIsCondensed, setPrevIsCondensed] = useState(props.isCondensed);
+  const [cacheMeasurements, setCacheMeasurements] = useState(true);
```
ddouglasz, line 63:
```suggestion
  const [isCacheMeasurements, setIsCacheMeasurements] = useState(true);
```

packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx
```diff
@@ -52,6 +59,20 @@ const MultilineInput = (props: TMultiLineInputProps) => {
...
+  useEffect(() => {
+    // If props.isCondensed has changed, the height of the layout will only be automatically recalculated if
+    // cacheMeasurements is set to false. So let's set cacheMeasurements to true by default to get the performance
+    // benefits, unless isCondensed value has changed, in which case we will set it to false for this render.
+    if (props.isCondensed !== prevIsCondensed) {
+      setCacheMeasurements(false);
```
ddouglasz, line 69:
```suggestion
      setIsCacheMeasurements(false);
```

packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx
```diff
@@ -52,6 +59,20 @@ const MultilineInput = (props: TMultiLineInputProps) => {
...
+    } else {
+      setCacheMeasurements(true);
```
ddouglasz, line 71:
```suggestion
      setIsCacheMeasurements(true);
```

packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx
```diff
@@ -103,7 +124,7 @@ const MultilineInput = (props: TMultiLineInputProps) => {
       role="textbox"
       minRows={MIN_ROW_COUNT}
       maxRows={props.isOpen ? undefined : MIN_ROW_COUNT}
-      cacheMeasurements={true}
+      cacheMeasurements={cacheMeasurements}
```
ddouglasz, line 127:
```suggestion
      cacheMeasurements={isCacheMeasurements}
```

ddouglasz (commented):
Some minor feedback it we would be going with this approach.

CarlosCortizasCT:
> In theory if someone else wanted to toggle this value (which I agree is very unlikely), why not handle it for them since we have to handle it for ourselves anyways

Thanks @Sarah4VT for your feedback 👍 

>  In theory if someone else wanted to toggle this value (which I agree is very unlikely), why not handle it for them since we have to handle it for ourselves anyways.

My argument to avoid this is that any lines of code we add to a component increases it's complexity and, thus, it's maintainability.
I believe we should balance that with the features we provide and, in this case, since we (more or less) agree is unlikely consumers would need that, exposing the property would be (in my opinion) a good balance between complexity added to the component in comparison with the requirement.

kterry1 (approved):
Nice work! Tested it on the preview branch and it works as expected.

Sarah4VT:
> Thanks @Sarah4VT for your feedback 👍
> 
> > In theory if someone else wanted to toggle this value (which I agree is very unlikely), why not handle it for them since we have to handle it for ourselves anyways.
> 
> My argument to avoid this is that any lines of code we add to a component increases its complexity and, thus, its maintainability. I believe we should balance that with the features we provide and, in this case, since we (more or less) agree is unlikely consumers would need that, exposing the property would be (in my opinion) a good balance between complexity added to the component in comparison with the requirement.

@CarlosCortizasCT I made the changes to move cacheMeasurements out to a prop on the component to allow ultimate control for consumers and set it to true by default. Let me know if I need to change anything else!

Sarah4VT:
> Some minor feedback it we would be going with this approach.

I ended up undoing the implementation due to Carlos's feedback, so these changes were not applied as the code was removed.

CarlosCortizasCT:
> I ended up undoing the implementation due to Carlos's feedback, so these changes were not applied as the code was removed.

Thanks @Sarah4VT 🙇 

I've just pushed a small commit in order to use the component's default props better.

Also, bear in mind we're currently not respecting the expected left margin of the text input value when the component is in condensed mode.

Design
image

Storybook
image

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.story.js
```diff
@@ -69,7 +69,7 @@ storiesOf('Components|Inputs', module)
               }
               defaultExpandMultilineText={defaultExpandMultilineText}
               isAutofocussed={boolean('isAutofocussed', false)}
-              cacheMeasurements={boolean('cacheMeasurements', true)}
+              cacheMeasurements={boolean('cacheMeasurements', false)}
```
Sarah4VT, line 72: @CarlosCortizasCT  Do you typically set the default value in storybook to false even if the default value in the component is true?

Sarah4VT:
> > I ended up undoing the implementation due to Carlos's feedback, so these changes were not applied as the code was removed.
> 
> Thanks @Sarah4VT 🙇
> 
> I've just pushed a small commit in order to use the component's default props better.
> 
> Also, bear in mind we're currently not respecting the expected left margin of the text input value when the component is in condensed mode.
> 
> Design image
> 
> Storybook image

@CarlosCortizasCT  I fixed the left/right padding to go back to the original value and only reduced the top/bottom padding on the input when isCondensed is true. Good catch!

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.story.js
CarlosCortizasCT: I think it's better to use `false` here because, otherwise, the `isCondensed` check won't work and users won't know why.

FilPob:
Thanks for all the work! Looks good from my side

CarlosCortizasCT:
Something that I wanted to check with @FilPob is the horizontal padding on the text in the input.
In the Figma file, this padding is 16px but we already had 8px in the component. I think this was wrong as the other inputs also use 16px.

I think we should take the opportunity to also adjust that.

Filip, what do you think?

FilPob:
@CarlosCortizasCT true! good catch. should be 16px for all inputs

FilPob:
@Sarah4VT awesome, looks good. thanks :)

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks Sarah!

PR #2801, opened 2024-04-29
feat(search-select-input): add isCondensed prop to SearchSelectInput …

Sarah4VT:
…component

#### Summary
Adds support for isCondensed prop on SearchSelectInput component to display a condensed layout.

## Description

image

Results
image

I did notice that with a long input, the layout is a little off, but it was like this prior to my changes as well.
Existing code:
image

With my changes:
image


Sarah4VT:
@commercetools/shield-team-design-system Not sure if you want me to change anything about the layout for that long input value example. It was working that way before my changes, so I wanted to be sure it wasn't intentional since it was not in my list of changes to implement.

ddouglasz:
> @commercetools/shield-team-design-system Not sure if you want me to change anything about the layout for that long input value example. It was working that way before my changes, so I wanted to be sure it wasn't intentional since it was not in my list of changes to implement.

 If this is not a blocker, Maybe we can create a ticket to see if we can have a follow-up fix from our end. Unless if it is something you think you can add to this PR :)
 cc: @FilPob

ddouglasz (approved):
Looks good 💯

FilPob:
the condensed prop works correctly from my side. I can't reproduce the bug with the weird layout with long value in Storybook. @ddouglasz can we do this in a follow up?

Sarah4VT:
> the condensed prop works correctly from my side. I can't reproduce the bug with the weird layout with long value in Storybook. @ddouglasz we can fix this in a follow up if you think that's better

Here is how I was able to create the bug. This is on the current version of UI kit. It's fine until you add the iconLeft.
Zight Recording 2024-04-30 at 04 35 40 PM

Sarah4VT:
@commercetools/shield-team-design-system I took a stab at trying to figure out the layout issue for a long value. I tried changing the width on the span inside WrapperWithIcon to be
`width: ${`calc(100% - ${props.selectProps.isCondensed ? 24 : 32}px)`} `
(I know the variables would need to be fixed up, but trying to do math with variables that are strings and include px I wasn't sure if it would work) There was still an element causing the span and the svg above it to be pushed up. I am happy to keep looking, but I am not sure what this input is doing so went ahead and opened a separate bug ticket and included a video of how to recreate it on the current code.
Components Inputs SelectInputs - SearchSelectInput ⋅ Storybook 2024-04-30 at 4 43 05 PM jpg

ddouglasz:
Thank you @Sarah4VT for giving it a try 🙏🏾. We will create a ticket and fix this as a follow up.

Sarah4VT:
> Thank you @Sarah4VT for giving it a try 🙏🏾. We will create a ticket and fix this as a follow up.

Yeah, I did look at it more yesterday. That input that is causing the problem is the input that is used when the user types into the search box. The problem is the parent element is a flex layout and because the span that contains the value is so long, that input is wrapping to the next line, then I think getting flex'd to take up the full space possibly. If you shrink the span down to the maximum amount that also allows space for the iconLeft, padding between iconLeft and the span, the clear button, and the search button (if they are all present), it will all go on one line and look correct. It's a bit of math to do in the code because you have a lot of conditions that may or may not be present. I probably could have gotten it working, but just wasn't sure if your team would have a different preferred solution.

CarlosCortizasCT (approved):

PR #2802, opened 2024-04-30
Decoupling datatablemanager: window state poc

ddouglasz:
We want to be able to consume the data table settings in the data table without having so much restrictions to how they are currently structured in the DOM tree. 
The settings dropdown should be flexible to exist anywhere in the component that it is used. 

An alternative approach is to managing the state using a Provider as seen in this [POC](https://github.com/commercetools/ui-kit/compare/decoupling-datatablemanager-using-provider-POC).


ddouglasz:
After comparing both options, I would go with this option since it does not seem to require the user importing a new Provider or any other form of import.
But I am also interested in knowing what @commercetools/shield-team-design-system think about this.

Issue #2779, opened 2024-04-11
`LeadingIcon` - passing `svg` prop as `string` results in unexpected SVG styling


ByronDWall:
**Describe the bug**
The `LeadingIcon` component API currently accepts an `svg` prop, which is currently a `string` that is rendered internally by the `InlineSVG` component.

The intention behind allowing for an `svg` prop as well as an `icon` prop is so that the `LeadingIcon` can display "custom" SVG's.

The `InlineSVG` component sets all `fill` values within the SVG to `fill: inherit` by default [in this style method](https://github.com/commercetools/ui-kit/blob/0ebf4f87866bd6dd0e7cd4c75f97e2adecedb756/packages/components/icons/src/templates/icon.styles.tsx#L105) which means that all elements of the SVG are rendered in a single color.  

This means that custom SVG's displayed by `LeadingIcon` will always have a single color, which leads to unexpected results when passing in SVG's that have multiple fill colors. 

**To Reproduce**

1. Import `LeadingIcon` into a file
2. Implement an svg that has multiple `fill` values as a string
3. Pass string to `LeadingIcon` component via `svg` prop
4. `LeadingIcon` displays an svg with only one `fill` color

**Expected behavior**
SVG's passed to `LeadingIcon` are displayed with all declared fill colors.

**Screenshots**

https://github.com/commercetools/ui-kit/assets/87667330/8dd8f957-e9c6-4017-a388-03a88cdf9320

**Additional context**
Being able to display custom SVG's as-expected in `LeadingIcon` is one of the main use-cases our team needs to be able to achieve with this component, so I would like to suggest a couple of solutions that the first contact team could implement relatively quickly:

1) remove the `*:not([fill='none'])` selector from SVG's built by the `InlineSVG` component for SVG's rendered by `LeadingIcon`

2) accept a `ReactComponent` instead of a `string` for the `LeadingIcon` `svg` prop.

ddouglasz:
Hey Byron,
Thanks for reporting. I can also see this from my end too.
We will take a look at a bit further and let you know how we would like to proceed.

emmenko:
I would suggest to not use the `InlineSVG` implementation for the `LeadingIcon` component if you don't want the default icons behavior.

On another note, if you want to preserve the SVG as-is, why not loading the SVG directly as a React component? The MC tooling already allows that if the file name ends with `.react.svg`.
Then you just need a "container" to render the SVG component within certain constraints.

Thanks

[...]

PR #2803, opened 2024-04-30
FCT-808: Create `CustomIcon` component

ByronDWall:
#### Summary

Based on our discussion about how `LeadingIcon` handles non-ui-kit icons in this [issue](https://github.com/commercetools/ui-kit/issues/2779), this PR introduces a new `CustomIcon` component that handles displaying non-ui-kit SVG's in a consistent manner.

## Description

The `CustomIcon` is similar to the `LeadingIcon`, but is intended to display non-ui-kit icons.  See the linked issue for more details.


stephsprinkle (approved):

packages/components/icons/src/custom-icon/custom-icon.tsx
```diff
@@ -0,0 +1,58 @@
...
+  icon: ReactElement | string;
+};
+
+const defaultProps: Required<Pick<TCustomIconProps, 'size' | 'hasBorder'>> = {
+  size: '20',
+  hasBorder: true,
+};
+
+const CustomIcon = (props: TCustomIconProps) => {
+  if (!props.icon) {
```
CarlosCortizasCT, line 31: I would avoid this check. The `icon` property is the only one required and it's stated in the docs and in the TS types.
For those rare cases where a consumer forgets to pass the icon, I think they will notice right away.

I'm saying because the less code we write, the less code we need to maintain so I believe we should focus on the things that provide the biggest value.

packages/components/icons/README.md
```diff
@@ -105,3 +105,34 @@ const app = () => <LeadingIcon icon={<ExportIcon />} />;
...
+
+const app = () => <Icon icon={<YourCustomIcon />} />;
+```
+
+### Properties
+
+| Props       | Type                                                      | Required | Values                                              | Default | Description                                                                                                     |
+| ----------- | --------------------------------------------------------- | :------: | --------------------------------------------------- | ------- | --------------------------------------------------------------------------------------------------------------- |
+| `size`      | `string`                                                  |          | '10', '20', '30', '40'                              | '20'    | Specifies the icon size                                                                                         |
+| `icon`      | `union`<br/>Possible values:<br/>`, ReactElement, string` |    -     | A `ReactNode` or `string` that display a custom SVG |         | Icon that is displayed within the component, you must supply a child icon with with this prop or the `svg` prop |
```
CarlosCortizasCT, line 133: Is seems the description is out of sync.
I think we need to generate the readme again.

packages/components/icons/src/custom-icon/custom-icon.styles.ts
```diff
@@ -0,0 +1,27 @@
+// use negative margin to make it so that border doesnt shift content?
```
CarlosCortizasCT, line 1: Could you maybe explain this comment? 🙏

packages/components/icons/src/custom-icon/custom-icon.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+import { designTokens } from '@commercetools-uikit/design-system';
+import { css } from '@emotion/react';
+import { sizeMap } from '../leading-icon/leading-icon.styles';
```
CarlosCortizasCT, line 4: I think we should try to avoid coupling components the better we can.

In this case, I would suggest to have the same values here.

Maybe as a follow up we can think about having some component-level design tokens, but we will be adjusting the sizes for the icons in the short term, so I would hold the new design tokens for now.

packages/components/icons/src/fixtures/CustomIconReact.tsx
```diff
@@ -0,0 +1,37 @@
+const SvgCustomIcon = () => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width={35}
+    height={35}
```
CarlosCortizasCT, line 5: I think it would be better for this icon to not have size attributes.
I believe the intention is that the size of the `svg` is controlled by the icon size so it grows or shrinks with it.

packages/components/icons/src/custom-icon/custom-icon.tsx
ByronDWall: good point, it will be removed

packages/components/icons/src/custom-icon/custom-icon.styles.ts
ByronDWall: this was left here accidentally, it was a note to myself that I didnt end up using/following, will delete, thanks for pointing it out!

packages/components/icons/src/fixtures/CustomIconReact.tsx
ByronDWall: This is meant to be an example of a non-standard "custom" svg that a consumer of the library would pass to this component.

To me "custom" means that we do not override the properties that the user specified. If the user wanted the svg to scale, all they would have to do is remove the `width` and `height` from the svg.

ByronDWall: I can remove the width/height here if you'd prefer that for the docs examples though

tylermorrisford (approved):

FilPob:
@ByronDWall thanks for tackling this! I found one issue with the border. If the border is turned on, the underlaying svg seems to cut through the border-radius. 
image

ByronDWall:
> @ByronDWall thanks for tackling this! I found one issue with the border. If the border is turned on, the underlaying svg seems to cut through the border-radius.

@FilPob thanks for pointing that out, the issue should be fixed now

FilPob:
@ByronDWall thanks, border looks good now. One more thing I noticed is that sometimes the custom svg doesn't scale if you change the size. It could be that this is just a storybook issue cause if you toggle the `use stringified svg for icon` prop on and off then suddenly it works fine. 
CPT2405061618-442x210

packages/components/icons/src/fixtures/CustomIconReact.tsx
CarlosCortizasCT: Thanks for sharing your reasoning, Byron 👍 

The way I see it, the main point of this component is to make sure the icons size is consistent across applications, not matter if they are from `ui-kit` or custom. I actually think we should update the description of the component to better explain about this.

If users provide a custom icon with a specific size, we won't get the expected behaviour because, as in the current state with the fixture svg, making the `CustomIcon` component bigger does not make the svg bigger, which I think that's what we want.

So yes, I would suggest to remove the size properties of the fixture SVG and also to update the description of the component (we can ask Filip for help).

packages/components/icons/README.md
```diff
@@ -105,3 +105,34 @@ const app = () => <LeadingIcon icon={<ExportIcon />} />;
 ### Where to use
 
 This component can be used wherever it is necessary to display a themed icon.
+
+## Custom Icon
+
+The custom icon is a an eye-catching visual element that should be used when an additional visual prominence is needed for displaying a custom SVG icon in the UI.
```
CarlosCortizasCT, line 111: I think we would need to be more specific about this component.

@FilPob could you help us coming up with another description which resembles what you think this component should be used for? 🙏 

Here is my attempt, in case it could be useful:
> This component is meant to be used whenever consumers need to render an icon which is not part of the [ui-kit icon set](https://uikit.commercetools.com/?path=/story/components-icons--all-icons). 
In order to keep visual consistency, we want to keep the available sizes of all icons equal. 
Bear in mind we would expect custom SVG icons to not contain size attributes so it can be controlled based on the components size attribute.

packages/components/icons/README.md
```diff
@@ -105,3 +105,34 @@ const app = () => <LeadingIcon icon={<ExportIcon />} />;
...
+
+### Properties
+
+| Props       | Type                                                      | Required | Values                                              | Default | Description                                     |
+| ----------- | --------------------------------------------------------- | :------: | --------------------------------------------------- | ------- | ----------------------------------------------- |
+| `size`      | `string`                                                  |          | '10', '20', '30', '40'                              | '20'    | Specifies the icon size                         |
+| `icon`      | `union`<br/>Possible values:<br/>`, ReactElement, string` |    -     | A `ReactNode` or `string` that display a custom SVG |         | Icon displayed as a child of this component     |
+| `hasBorder` | `boolean`                                                 |          | `true`, `false`                                     | `false` | Specifies whether the element displays a border |
+
+### Where to use
```
CarlosCortizasCT, line 136: Not sure if we need both the description and this section.

@FilPob What do you think?

packages/components/icons/README.md
FilPob: Thanks @CarlosCortizasCT that sounds totally fine for me

packages/components/icons/README.md
FilPob: not sure what is meant here. (too much code for my eyes :D ) Are there two descriptions somewhere? I would probably suggest to follow whatever approach is there in other components

CarlosCortizasCT: We currently have a description of the component and where to use it:

Screenshot 2024-05-07 at 15 32 29

Does it make sense to have both. If so, what do you suggest to explain in each section?

FilPob: ah okay got it... No, the first description is rather targeted for the Leading icon. The second fits better I think

CarlosCortizasCT: You mean it's ok to not include the "Where to use" section and just keep the initial component description but changing it to the one I suggested, don't you?

> This component is meant to be used whenever consumers need to render an icon which is not part of the [ui-kit icon set](https://uikit.commercetools.com/?path=/story/components-icons--all-icons).
In order to keep visual consistency, we want to keep the available sizes of all icons equal.
Bear in mind we would expect custom SVG icons to not contain size attributes so it can be controlled based on the components size attribute.

FilPob: yes exactly :)

ByronDWall: hey @CarlosCortizasCT would you like me to make these changes?

CarlosCortizasCT: Yes, please

ByronDWall: PR is updated to include the requested wording in the description and removes the "where to use" section as specified

ByronDWall:
> @ByronDWall thanks, border looks good now. One more thing I noticed is that sometimes the custom svg doesn't scale if you change the size. It could be that this is just a storybook issue cause if you toggle the `use stringified svg for icon` prop on and off then suddenly it works fine.

@FilPob I believe this is an issue with storybook, the stringified svg is using the `scale` `InlineSvg` sizing prop, which is used elsewhere without issue.  Also, I don't believe this component will be dynamically sized in consuming apps the way it is in the storybook, so this should not be an issue.

CarlosCortizasCT:
> > @ByronDWall thanks, border looks good now. One more thing I noticed is that sometimes the custom svg doesn't scale if you change the size. It could be that this is just a storybook issue cause if you toggle the `use stringified svg for icon` prop on and off then suddenly it works fine.
> 
> @FilPob I believe this is an issue with storybook, the stringified svg is using the `scale` `InlineSvg` sizing prop, which is used elsewhere without issue. Also, I don't believe this component will be dynamically sized in consuming apps the way it is in the storybook, so this should not be an issue.

@ByronDWall I think Filip is not talking about an inline svg issue but an icon one. If you use the icon and change its size, we can see we still have some extra space between the icon and the border, but this works if you change from icon to inline svg and back to icon (to be honest, I don't know why).

I believe the solution is to change the flex styles from the wrapping div and just use a `display: inline-block` style:
image

ByronDWall:
> @ByronDWall I think Filip is not talking about an inline svg issue but an icon one. If you use the icon and change its size, we can see we still have some extra space between the icon and the border, but this works if you change from icon to inline svg and back to icon (to be honest, I don't know why).

This was an interesting issue, because I could not reproduce it in firefox.  Was able to in chrome.  Fixed

packages/components/icons/src/leading-icon/leading-icon.styles.ts
```diff
@@ -8,7 +8,7 @@ type TColorThemeConfiguration = {
   border?: string;
 };
 
-const sizeMap = {
+export const sizeMap = {
```
CarlosCortizasCT, line 11: Do we actually need to export this constant?

CarlosCortizasCT:
@FilPob I think UI wise this is ready.

Could you please take another look when you have the time? 🙏

FilPob:
Looks perfect now! thanks @CarlosCortizasCT @ByronDWall :)

FilPob (approved):

packages/components/icons/src/leading-icon/leading-icon.styles.ts
ByronDWall: no, that must have slipped back in at some point, the export has been removed

CarlosCortizasCT: Thanks.

CarlosCortizasCT (approved):
Thanks for your contribution, Byron 👍

PR #2804, opened 2024-04-30
Fix `DropdownMenu` component issues

### LINKED ISSUE #2786: dropdownmenu list menu content: menu content is not reachable if out of viewport
A quick video describing the issue:

https://github.com/commercetools/ui-kit/assets/106664971/a315b7aa-2194-4022-8392-16233c0f9e9a

**Expected behavior**
All the items should be accessible even if they initially appear out of the viewport

CarlosCortizasCT:
#### Summary

Fix `DropdownMenu` component issues related to floating menu size and scrolling.

Closes #2786

## Description

This PR tries to fix two problems:
1. We want to block the scroll (if any) in the page when the dropdown is opened. This allows to avoid the complexity of updating the placement of the floating menu while scrolling and it seems a fair compromise.
2. We were not controlling the size of the floating menu, so long lists could potentially get out of the viewport with no menu scroll available.

This is how we tackle those issues:
1. Previously we were blocking the document `body` scroll, but this component might be rendered in a scrollable container, so now we look for its first scrollable ancestor to block it while the dropdown is opened. It can be the case of multiple scroll containers in the page which might lead to potential issues, but we want to have a quick fix released for the majority of the use cases.
2. Now the floating menu height will grow until it reaches the bottom of the viewport and applies scrolling to its contents if they need more vertical space. On top of that, we've introduced a new property (`menuMaxHeight`) which consumers can use to limit the height of the floating menu however they want.

Something that we intentionally left out of the scope of this PR is how to deal with dropdown menus opening the floating menu always downwards, which is not convenient when the triggering element is at the bottom of the page.
We will work on that as a follow up so we can release these fixes to make the component usable.


packages/components/dropdowns/dropdown-menu/src/dropdown-menu.story.js
```diff
@@ -24,30 +24,31 @@ storiesOf('Components|Dropdowns|DropdownMenu', module)
...
-          Option 2
-        </DropdownMenu.ListMenuItem>
-        <DropdownMenu.ListMenuItem onClick={action('onClick')}>
-          Option 3
-        </DropdownMenu.ListMenuItem>
-      </DropdownMenu>
-    </Section>
-  ))
+  .add('DropdownMenu - List menu content', () => {
+    const optionsCount = number('Options count', 5);
```
CarlosCortizasCT, line 28: This is to allow having long lists in the dropdown menu so the height and scrollability can be tested.

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.tsx
```diff
@@ -40,6 +47,48 @@ export type TDropdownMenuProps = {
   children: ReactNode;
 };
 
+function getScrollableParent(element: HTMLElement | null): HTMLElement | null {
```
CarlosCortizasCT, line 50: We want to block the scroll of the nearest ancestor with scrolling enabled, if any.

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.tsx
```diff
@@ -40,6 +47,48 @@ export type TDropdownMenuProps = {
...
+  const overflowY = window.getComputedStyle(element).overflowY;
+  const isScrollable = overflowY !== 'visible' && overflowY !== 'hidden';
+  if (isScrollable && element.scrollHeight >= element.clientHeight) {
+    return element;
+  }
+
+  return getScrollableParent(element.parentElement);
+}
+
+function useScrollBlock(isOpen: boolean, triggerRef: RefObject<HTMLElement>) {
```
CarlosCortizasCT, line 63: Just to extract a bit of logic from the component itself.

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.story.js
ddouglasz: 💯

ddouglasz (approved):
Works as expected 🙇🏽‍♂️
Thank you

chloe0592 (approved):
Thanks for fixing it! 🙏

PR #2807, opened 2024-05-02
Decoupling data table manager components

ddouglasz:
We want to be able to consume the data table settings in the data table without having so much restrictions to how they are currently structured in the DOM tree. 
We want to decouple the DataTableManager such that the settings dropdown component can live anywhere in the DOM tree from the DataTable
The current way it works is one in which the `Datatable` is always  directly after the Datatable settings dropdown.
The settings dropdown should be flexible to exist anywhere in the component that it is used. 

This PR implements this feature using the Provider method, where the provider allows both the data table settings dropdown and the datatable to share state from a common source, allowing us the flexibility we want. 

An alternative approach is to managing the state using a Window method as seen in this [POC](https://github.com/commercetools/ui-kit/pull/2807/commits/1d3236139e306f2d9d084b057c759c224c65ff49).  But this posed to have some complexities regarding how to manage the initial states and also the security implications of storing the state in a global window object

An example of where this can be useful:
image


ddouglasz:
After comparing both options, I would go with this option since it does not seem to require the user importing a new Provider or any other form of import.
But I am also interested in knowing what @commercetools/shield-team-design-system think about this.

packages/components/data-table/src/data-table.tsx
```diff
@@ -239,8 +239,15 @@ export type TDataTableProps<Row extends TRow = TRow> = {
 };
 
 const DataTable = <Row extends TRow = TRow>(props: TDataTableProps<Row>) => {
+  const columns =
```
ddouglasz, line 242: if we use this alternative, we would miss the initial state. and it seems to me like we simply just need to pass the columns object just the same way it was used in the data-table-manager. 
```suggestion
const [columns, setColumns] = useState<[]>();
  const handleCustomEvent = (event) => {
      const { columns } = event.detail;
      setColumns(columns);
    };
    useEffect(() => {
      window.addEventListener(‘DataTaBleManagerEvent’, handleCustomEvent);
      return () => {
        window.removeEventListener(‘DataTaBleManagerEvent’, handleCustomEvent);
      };
    }, []);

File: packages/components/data-table-manager/src/data-table-manager.tsx [hunk 2]
```diff
@@ -166,7 +166,10 @@ const DataTableManager = <Row extends TRow = TRow>(
     [areDisplaySettingsEnabled, props.columns, isWrappingText]
   );
 
-  return (
+  //  @ts-ignore
+  window.dataTableColumns = { columns };
```
ddouglasz, line 170: I would use the window.dispatch but it seems like we would be missing the initial state using this approach
```suggestion
  useEffect(()=> {
    const customEvent = new CustomEvent(‘DataTaBleManagerEvent’, {
      detail: {columns}
    })
    window.dispatchEvent(customEvent)
  },[columns])
```

ddouglasz: Thank you for the comprehensive feedback. I will push a change with the Provider approach since it is now clear that we would have some security/complexity concerns using the window approach 🙏🏾

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/39a97a46ad4d52bc58293c40b7f07325c8ad5a55

Thank you.

chloe0592 (approved):
Thanks @ddouglasz for working on it, this was a tricky one.

packages/components/data-table-manager/src/data-table-manager.spec.js
```diff
@@ -153,3 +186,97 @@ describe('rendering', () => {
     ).toBeInTheDocument();
   });
 });
+
+describe('rendering with decoupled data table', () => {
```
ddouglasz, line 190: We also want to be sure that when the datatable manager is decoupled, all it's behaviours are still intact.

packages/components/data-table/src/data-table.tsx
```diff
@@ -20,6 +21,7 @@ import HeaderCell from './header-cell';
 import DataRow from './data-row';
 import useManualColumnResizing from './use-manual-column-resizing-reducer';
 import ColumnResizingContext from './column-resizing-context';
+import { DataTableManagerContext } from '@commercetools-uikit/data-table-manager/src/data-table-manager';
```
ddouglasz, line 24: This is meant to just be a root import from data-table-manager but I am yet to find out why it would not return column data when I import that way.
```suggestion
import { DataTableManagerContext } from '@commercetools-uikit/data-table-manager';
```

packages/components/data-table-manager/src/data-table-manager.spec.js
```diff
@@ -41,8 +45,37 @@ const TestComponent = (props) => {
     </DataTableManager>
   );
 };
-/* eslint-enable react/prop-types */
 
+const DecoupledDatatableTestComponent = (props) => {
```
CarlosCortizasCT, line 49: I don't see we're actually testing the decoupling here as it seems the provider state is not used in the `TestTable`.

`DataTableManager` and `TestTable` components should be in different leafs of the component's tree.

packages/components/data-table-manager/README.md
```diff
@@ -12,6 +12,7 @@ This component enhances the `<DataTable>` component and additionally provides a
 
 - The `disableDisplaySettings` enables / disables the layout settings panel, allowing the user to select wrapping text and density display options.
 - The `disableColumnManager` enables / disables the column manager panel, allowing the user to select which columns are visible.
+- To Detach the `DatatableManager` settings dropdown from the Datatable and position it anywhere else, you would need to import a `DataTableManagerProvider` and wrap both components with the provider.
```
CarlosCortizasCT, line 15: I believe we would need a dedicated section to explain this behaviour.

packages/components/data-table-manager/README.md
```diff
@@ -53,18 +56,19 @@ const columns = [
 ];
 
 const Example = () => (
-  <DataTableManager columns={columns}>
+  <DataTableManagerProvider>
```
CarlosCortizasCT, line 59: I think we need to provide two different examples. The current one and another one with the detached behaviour where the `DataTable` and `DataTableManager` should be in different component's tree leaves.

packages/components/data-table/package.json
```diff
@@ -22,6 +22,7 @@
     "@babel/runtime": "^7.20.13",
     "@babel/runtime-corejs3": "^7.20.13",
     "@commercetools-uikit/accessible-button": "19.2.0",
+    "@commercetools-uikit/data-table-manager": "19.2.0",
```
CarlosCortizasCT, line 25: I'm concerned about this. Now whoever uses the `DataTable` component, will include the `DataTableManager` in their application JS bundles even if they don't use the latter component.

An idea would be to export the provider from the `DataTableManager` package as an independent entry point (we do something similar in the icons package) so consumers can import only the provider code.

packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -146,6 +162,33 @@ type TDataTableManagerProps<Row extends TRow = TRow> = {
   managerTheme?: 'light' | 'dark';
 };
 
+type TDataTableManagerContext = {
+  columns: TDataColumns;
+  updateColumns: (columns: TDataColumns) => void;
+};
+
+export const DataTableManagerContext = createContext<TDataTableManagerContext>({
```
CarlosCortizasCT, line 170: The context should live in a new file and also export a custom hook to access its information.

This will also help with having a better export from the package.

packages/components/data-table/src/data-table.tsx
CarlosCortizasCT, line 24: We should not import from internals of a package but just from what they export.
Importing from internals is fragile as we would be depending on its current implementation and will fail if it is refactored.

packages/components/data-table/src/data-table.tsx
```diff
@@ -273,6 +280,11 @@ const DataTable = <Row extends TRow = TRow>(props: TDataTableProps<Row>) => {
         (tableRef.current.offsetWidth - tableRef.current.clientWidth)
       : undefined;
 
+  const updatedProps = {
```
CarlosCortizasCT, line 283: We can also do this inline when using the `DataRow` and we don't need an intermediate constant:
```tsx
<DataRow<Row>
  {...props}
  columns={columnsData}
  row={row}
  key={row.id}
  rowIndex={rowIndex}
```

packages/components/data-table-manager/src/data-table-manager.tsx
```diff
@@ -181,9 +233,19 @@ const DataTableManager = <Row extends TRow = TRow>(
...
       })}
     </Spacings.Stack>
+  ) : (
+    <DataTableSettings
+      topBar={props.topBar}
+      onSettingsChange={props.onSettingsChange}
+      columnManager={props.columnManager}
+      displaySettings={props.displaySettings}
+      managerTheme="light"
+    />
```
CarlosCortizasCT, line 243: We can just keep the previous code and conditionally render the `children`.

packages/components/data-table-manager/src/data-table-manager.spec.js
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/bd8447871f0d99ac366b5adc9d865055c9a4506e

Thank you

packages/components/data-table-manager/README.md
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/bd8447871f0d99ac366b5adc9d865055c9a4506e

Thank you

packages/components/data-table-manager/README.md
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/bd8447871f0d99ac366b5adc9d865055c9a4506e

Thank you

packages/components/data-table-manager/src/data-table-manager.tsx
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/bd8447871f0d99ac366b5adc9d865055c9a4506e

Thank you

packages/components/data-table/package.json
ddouglasz: On it. Thank you

packages/components/data-table/src/data-table.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/e02436f523fc54b93c131ade1cb3b5b26bb4a20ahttps://github.com/commercetools/ui-kit/pull/2807/commits/e02436f523fc54b93c131ade1cb3b5b26bb4a20a

packages/components/data-table/package.json
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/e02436f523fc54b93c131ade1cb3b5b26bb4a20a

packages/components/data-table/src/data-table.tsx
ddouglasz: I missed this initially. 
Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/95ca76b86725e211f2fc69748866c26db1019ca6 
Thank you.

packages/components/data-table-manager/src/data-table-manager.story.js
```diff
@@ -321,4 +324,157 @@ storiesOf('Components|DataTable', module)
...
+        }}
+        columnManager={columnManager}
+      >
+        <Spacings.Stack>
+          <DataTableManager />
+          <SearchTextInput
+            placeholder={text('placeholder', 'Dummy search component')}
+            isReadOnly
+          />
+          <DataTable
```
CarlosCortizasCT, line 469: I think it would be great to have the components in a more deep component tree to really showcase consumers can use them wherever they want.

Just an example:
```jsx
return (
      <DataTableManagerProvider
        columns={withRowSelection ? columnsWithSelect : visibleColumns}
        displaySettings={displaySettings}
        onSettingsChange={(action, nextValue) => {
          tableSettingsChangeHandler[action](nextValue);
        }}
        columnManager={columnManager}
      >
        <header>
          <Spacings.Inline justifyContent="flex-end">
            <DataTableManager />
          </Spacings.Inline>
          <hr style={{ marginBottom: '24px' }} />
        </header>
        <main>
          <DataTable
            rows={withRowSelection ? rowsWithSelection : rows}
            sortedBy={sortedBy}
            onSortChange={onSortChange}
            sortDirection={sortDirection}
          />
        </main>
        <br />
        <hr />
      </DataTableManagerProvider>
```

packages/components/data-table-manager/src/types.tsx
```diff
@@ -0,0 +1,139 @@
+import { type ReactNode, type MouseEventHandler, ReactElement } from 'react';
+
+import type { TDataTableSettingsProps } from './data-table-settings';
```
CarlosCortizasCT, line 3: I think we need to do this the other way around: define here the type and use it in the `DataTableSettings` component.
The way is currently implemented means the `DataTableManagerProvider` build package will include the `DataTableSettings` component because it's a transitive dependency.

packages/components/data-table/src/data-table.tsx
```diff
@@ -239,8 +239,16 @@ export type TDataTableProps<Row extends TRow = TRow> = {
 };
 
 const DataTable = <Row extends TRow = TRow>(props: TDataTableProps<Row>) => {
+  const { columns, isCondensed } = useDataTableManagerContext();
+  const isValueFromProvider = Boolean(columns && columns.length !== 0);
+  const columnsData = isValueFromProvider ? columns : props.columns;
+  const condensedValue =
+    isValueFromProvider && isCondensed !== undefined
+      ? isCondensed
+      : props.isCondensed;
```
CarlosCortizasCT, line 248: Would it be easier and also work if check for the props first and then for the context values?
```suggestion
  const dataTableManagerProvider = useDataTableManagerContext();
  const columnsData = props.columns.length > 0 ? props.columns : dataTableManagerProvider.columns;
  const condensedValue = props.isCondensed ?? dataTableManagerProvider.isCondensed;
```

packages/components/data-table-manager/src/types.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/d1c167920c31da23e74275d1a3aa0912d539dc2d

Thank you

packages/components/data-table-manager/src/data-table-manager.story.js
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/d1c167920c31da23e74275d1a3aa0912d539dc2d

Thank you

packages/components/data-table-manager/src/data-table-manager-provider/data-table-manager-provider.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { columns, topBar, displaySettings, onSettingsChange, columnManager } =
+    useContext(DataTableManagerContext);
+
+  const areDisplaySettingsEnabled = Boolean(
+    displaySettings && !displaySettings.disableDisplaySettings
+  );
+  const isWrappingText =
+    areDisplaySettingsEnabled && displaySettings!.isWrappingText;
+
+  const contextValue = useMemo(
```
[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/d1c167920c31da23e74275d1a3aa0912d539dc2d

Thank you

packages/components/data-table-manager/README.md
```diff
@@ -61,10 +63,46 @@ const Example = () => (
...
+const SomeOtherComponent = () => {
+  return <div>Some other component</div>;
+};
+
+const Example = () => (
+  <DataTableManagerProvider columns={columns}>
+    <DataTableManager />
+    <SomeOtherComponent />
+    <DataTable rows={rows} />
+  </DataTableManagerProvider>
```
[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/d1c167920c31da23e74275d1a3aa0912d539dc2d

Thank you

packages/components/data-table/src/data-table.tsx
ddouglasz: Yes I tried this and the other way too, it looks simpler but where it becomes tricky is we are trying to check for the existence of one boolean value or the other. In which case, the absence of it is also a boolean value. That's why I tied it to the column from the provider `isValueFromProvider` so when we see that value or not, we can do a much better check for the `condensedValue`

packages/components/data-table-manager/src/data-table-manager-provider/data-table-manager-provider.tsx
```diff
@@ -17,43 +14,8 @@ const DataTableManagerContext = createContext<TDataTableManagerContext>({
...
-      columnManager,
-      areDisplaySettingsEnabled,
-      isWrappingText,
-    ]
-  );
-
-  return contextValue;
-};
+export const useDataTableManagerContext = () =>
+  useContext(DataTableManagerContext);
```
CarlosCortizasCT, line 18: It would be great to check for the context: if the context does not exist, we can throw an error explaining what's happening ([some reasoning about it](https://kentcdodds.com/blog/how-to-use-react-context-effectively#the-custom-consumer-hook)).

packages/components/data-table-manager/README.md
```diff
@@ -15,6 +15,8 @@ This component enhances the `<DataTable>` component and additionally provides a
 
 Both panels delegate the handling of the settings change on the parent through function properties, allowing the settings to be persisted or just used as state props.
 
+- The `children` prop can only be used when the datatable manager is used in it's coupled or undetached form, and not when it is detached.
```
CarlosCortizasCT, line 18: This is my proposal for updating the README:
```
# DataTableManager

## Description

This component enhances the `<DataTable>` component and additionally provides a UI and state management to handle configuration of the table such as column manager.

- The `disableDisplaySettings` enables / disables the layout settings panel, allowing the user to select wrapping text and density display options.
- The `disableColumnManager` enables / disables the column manager panel, allowing the user to select which columns are visible.

Both panels delegate the handling of the settings change on the parent through function properties, allowing the settings to be persisted or just used as state props.

This component will render a triggering element (icon button) above the `<DataTable>` (top-right corner) which users can click to select the panel to open.

## Installation

```
yarn add @commercetools-uikit/data-table-manager
```

```
npm --save install @commercetools-uikit/data-table-manager
```

Additionally install the peer dependencies (if not present)

```
yarn add react react-dom react-intl
```

```
npm --save install react react-dom react-intl
```

## Usage

There are two ways this component can be used:

### Attached

The basic usage is when it just wraps the DataTable component.

```jsx
import DataTableManager from '@commercetools-uikit/data-table-manager';
import DataTable from '@commercetools-uikit/data-table';

const rows = [
  { id: 'parasite', title: 'Parasite', country: 'South Korea' },
  { id: 'portrait', title: 'Portrait of a Lady on Fire', country: 'France' },
  { id: 'wat', title: 'Woman at War', country: 'Iceland' },
];

const columns = [
  { key: 'title', label: 'Title' },
  { key: 'country', label: 'Country' },
];

const Example = () => (
  <DataTableManager columns={columns}>
    <DataTable rows={rows} />
  </DataTableManager>
);

export default Example;
```

### Dettached

As mentioned earlier, the default behaviour is for the triggering element to be placed above the `<DataTable>` component (top-right corner), but there can be use cases where it would be needed to place that triggering element in a different place in the page. This is also possible but requires the usage of one more component (`DataTableManagerProvider`) in order to share the manager state between the manager panels and the `DataTable` component.

```jsx
import DataTableManager, {
  DataTableManagerProvider,
} from '@commercetools-uikit/data-table-manager';
import DataTable from '@commercetools-uikit/data-table';
import SearchTextInput from '@commercetools-uikit/search-text-input';

const rows = [
  { id: 'parasite', title: 'Parasite', country: 'South Korea' },
  { id: 'portrait', title: 'Portrait of a Lady on Fire', country: 'France' },
  { id: 'wat', title: 'Woman at War', country: 'Iceland' },
];

const columns = [
  { key: 'title', label: 'Title' },
  { key: 'country', label: 'Country' },
];

const SomeOtherComponent = () => {
  return <div>Some other component</div>;
};

const Example = () => (
  <DataTableManagerProvider columns={columns}>
    <header>
      <DataTableManager />
      <SearchTextInput />
    </header>
    <main>
      <SomeOtherComponent />
      <DataTable rows={rows} />
    </main>
  </DataTableManagerProvider>
);

export default Example;
```

CarlosCortizasCT: I would also add this note below the properties table:
```
> `*`: `DataTableManagerProvider` component accepts the same properties as the `DataTableManager`
```

ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/e04a5226ed7a7bde3743ec5f6a9f1d607e92b562

Thank you

packages/components/data-table-manager/src/data-table-manager-provider/data-table-manager-provider.tsx
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/e04a5226ed7a7bde3743ec5f6a9f1d607e92b562

Thank you

packages/components/data-table-manager/src/data-table-manager-provider/data-table-manager-provider.tsx
```diff
@@ -62,32 +24,43 @@ export const DataTableManagerProvider = ({
...
   onSettingsChange: () => void;
   columnManager: TColumnManagerProps;
-  isCondensed: boolean;
 }) => {
+  const areDisplaySettingsEnabled = Boolean(
+    displaySettings && !displaySettings.disableDisplaySettings
+  );
+
+  const isWrappingText =
+    areDisplaySettingsEnabled && displaySettings!.isWrappingText;
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/e04a5226ed7a7bde3743ec5f6a9f1d607e92b562

Thank you

packages/components/data-table-manager/README.md
```diff
@@ -61,10 +66,52 @@ const Example = () => (
 export default Example;
 ```
 
+### Dettached
+
+As mentioned earlier, the default behaviour is for the triggering element to be placed above the `<DataTable>` component (top-right corner), but there can be use cases where it would be needed to place that triggering element in a different place in the page. This is also possible but requires the usage of one more component (`DataTableManagerProvider`) in order to share the manager state between the manager panels and the `DataTable` component.
+
```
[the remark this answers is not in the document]
ddouglasz: This makes sense. 
updated: https://github.com/commercetools/ui-kit/pull/2807/commits/e4aacf488d6e932f17cecedde608c092ac2e3c78

packages/components/data-table-manager/src/data-table-manager.spec.js
```diff
@@ -41,8 +49,49 @@ const TestComponent = (props) => {
...
+    [UPDATE_ACTIONS.IS_TABLE_WRAPPING_TEXT_UPDATE]: setIsWrappingText,
+  };
+  return (
+    <DataTableManagerProvider
+      columns={props.columns}
+      isCondensed={isCondensed}
+      displaySettings={props.displaySettings}
+    >
+      <div>
+        <DataTableManager
```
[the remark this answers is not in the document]
ddouglasz: updated: https://github.com/commercetools/ui-kit/pull/2807/commits/e4aacf488d6e932f17cecedde608c092ac2e3c78

packages/components/data-table-manager/README.md
```diff
@@ -37,6 +36,12 @@ npm --save install react react-dom react-intl
 
 ## Usage
 
+There are two ways this component can be used:
+
+### Attached
```
CarlosCortizasCT, line 41:
```suggestion
### Atached
```

packages/components/data-table-manager/README.md
```diff
@@ -61,10 +66,53 @@ const Example = () => (
 export default Example;
 ```
 
+### Detached
+
+As mentioned earlier, the default behaviour is for the triggering element to be placed above the `<DataTable>` component (top-right corner), but there can be use cases where it would be needed to place that triggering element in a different place in the page. This is also possible but requires the usage of one more component (`DataTableManagerProvider`) in order to share the manager state between the manager panels and the `DataTable` component.
+In this mode, you should pass the manager props to the `DataTableManagerProvider` component and the `DataTableManager` does not need to receive any prop; it can be placed anywhere in the component's tree without requiring any prop.
```
CarlosCortizasCT, line 72: It would be great to include a breaking line between the two paragraphs.

CarlosCortizasCT (approved):
Thanks a lot for all your work in this PR, Douglas! 🙇

packages/components/data-table-manager/README.md
ddouglasz: This is the right spelling. Thanks! 🙃

packages/components/data-table-manager/_docs/usage-example.js
```diff
@@ -12,10 +14,24 @@ const columns = [
...
+};
+
+// With the data table settings decoupled.
+//  It can also be exported as default.
+export const ExampleWithDecoupledDataTableManager = () => (
+  <DataTableManagerProvider>
+    <DataTableManager columns={columns} />
+    <SomeOtherComponent />
+    <DataTable rows={rows} />
+  </DataTableManagerProvider>
```
kark, line 36: I wonder if this will work? Shouldn't columns be passed to the provider?

packages/components/data-table-manager/README.md
```diff
@@ -61,10 +66,54 @@ const Example = () => (
 export default Example;
 ```
 
+### Detached
+
+As mentioned earlier, the default behaviour is for the triggering element to be placed above the `<DataTable>` component (top-right corner), but there can be use cases where it would be needed to place that triggering element in a different place in the page. This is also possible but requires the usage of one more component (`DataTableManagerProvider`) in order to share the manager state between the manager panels and the `DataTable` component.
```
kark, line 71: nit: maybe this reads better:
```suggestion
As mentioned earlier, the default behaviour places the triggering element above the `<DataTable>` component (top-right corner), but there may be use cases where the triggering element needs to be positioned differently on the page. This is also possible but requires the usage of one more component (`DataTableManagerProvider`) in order to share the manager state between the manager panels and the `DataTable` component.
```

kark (commented):
Thank you Doulgas! Looks good! Great work overall! 🚀

packages/components/data-table-manager/src/data-table-manager-provider/data-table-manager-provider.tsx
```diff
@@ -0,0 +1,72 @@
+import { createContext, useContext, useMemo } from 'react';
+import type { TDataTableSettingsProps, TColumnManagerProps } from '../types';
+import { TDataTableManagerColumnProps, TRow } from './types';
```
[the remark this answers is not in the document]
ddouglasz: updated: https://github.com/commercetools/ui-kit/pull/2807/commits/37d6c7c20176cd563a6ad743408f2df3a22a752f

Thank you

packages/components/data-table-manager/src/types.tsx
```diff
@@ -0,0 +1,233 @@
+import { type ReactNode, type MouseEventHandler, ReactElement } from 'react';
```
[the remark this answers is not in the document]
ddouglasz: updated: https://github.com/commercetools/ui-kit/pull/2807/commits/37d6c7c20176cd563a6ad743408f2df3a22a752f

Thank you

packages/components/data-table-manager/_docs/usage-example.js
ddouglasz: Yes, that is correct.
updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/37d6c7c20176cd563a6ad743408f2df3a22a752f

Thank you

packages/components/data-table-manager/README.md
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2807/commits/37d6c7c20176cd563a6ad743408f2df3a22a752f

Thanks!

kark (approved):

packages/components/data-table-manager/README.md
CarlosCortizasCT: Sorry

PR #2809, opened 2024-05-02
feat(selectable search): add isCondensed prop to `SelectableSearchInput` component

jmcreasman:
Adding a new boolean isCondensed prop that when set to true condenses the input height, font size and icon size for the following component:

SelectableSearchInput
isCondensed is false
Screenshot 2024-05-03 at 10 06 42

isCondensed is true
Screenshot 2024-05-03 at 10 06 47


jmcreasman:
@ddouglasz 
I'm running into a problem with this component getting the left dropdown height to shrink (see picture). Everything else is done but that. It sure seems like the style are coming from components/inputs/selectable-search-input/src/selectable-search-input.styles.ts with createSelectableSelectStyles yet I can’t seem to figure out where to set height and get it to work. Played around with inspect for a good deal of time too and still no luck. 

Any thoughts? Could be I've just looked at the wrong part too long and am missing the obvious...  

Screenshot 2024-05-02 at 14 29 40

jmcreasman:
> @ddouglasz I'm running into a problem with this component getting the left dropdown height to shrink (see picture). Everything else is done but that. It sure seems like the styling is coming from components/inputs/selectable-search-input/src/selectable-search-input.styles.ts with createSelectableSelectStyles yet I can’t seem to figure out where to set height and get it to work. Played around with inspect for a good deal of time too and still no luck.
> 
> Any thoughts? Could be I've just looked at the wrong part too long and am missing the obvious...
> 
> Screenshot 2024-05-02 at 14 29 40

This has been resolved now but about to leave a couple comments on some curiosities we had.

packages/components/inputs/selectable-search-input/src/selectable-select.tsx
jmcreasman, line 1: Is there a reason select (from react) is being used/imported here instead of the select component already set up in UI Kit? Then all that would be needed is just to pass in the isCondensed instead of re-setting up the conditional styling again.

ddouglasz: Hey @jmcreasman 👋
Sorry I totally missed your previous mentions. Regarding your observation above, thats a good question. I think that is a viable option as it seems they are similar components. However, I guess it is either these use cases are not ideal for the outcome we want using a SelectInput or they were just implemented in this isolated form. I can confirm this when I have some time to investigate further.

ddouglasz (approved):
Thank you 🙏🏾

packages/components/inputs/selectable-search-input/src/selectable-select.tsx
CarlosCortizasCT: Thanks @jmcreasman for your feedback 👍 

I remember we run into som issues we couldn't overcome when we first tried to implement the `SelectableSearchInput` using the `SelectInput` component, although I don't remember the exact reason.

I agree with you that it's weird not all selectable components share more code and that's something that we would like to address in one of the initiatives we already discussed within the team for the future on `ui-kit`, however, we need to find the time to kickstart that initiative and I don't think that's going to happen this or next quarter.

Anyway, we really appreciate the feedback so it helps us to confirm what we already had in mind for improving this area of `ui-kit`.

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks!

FilPob:
@jmcreasman  thanks, looks good so far. One thing that is missing:

One of the requirements was to also decreae the font-size inside the dropdown menu when `isCondensed=true` This was done already in the `SelecInput` component as well

image

jmcreasman:
> @jmcreasman thanks, looks good so far. One thing that is missing:
> 
> One of the requirements was to also decreae the font-size inside the dropdown menu when `isCondensed=true` This was done already in the `SelecInput` component as well
> 
> image

I see what you mean, I had thought they were condensed too but after a second glance realize they are not. I'll take a look today and ping you when it's ready for your review. Thanks for the catch!

CarlosCortizasCT:
> @jmcreasman that looks correct. I dontt see it in the preview link though

I don't see it neither, even when pulling the branch and running it locally.

@jmcreasman could it be you have some locally uncommitted/unpushed changes?

jmcreasman:
@FilPob / @CarlosCortizasCT 
I apologize, I committed the code but never pushed it. Sorry about that it should [be there now](https://github.com/commercetools/ui-kit/pull/2809/commits/e60401e18bfb72f4aa58b3f70b04d13214ee1a36).

FilPob:
@jmcreasman Nice! the condensed one looks great now. However somehow there is also an unexpected change in the non-condensed version. The padding for each select item seems to be decreased. The non-condensed version should stay as it is currently. 
image

jmcreasman:
> @jmcreasman Nice! the condensed one looks great now. However somehow there is also an unexpected change in the non-condensed version. The padding for each select item seems to be decreased. The non-condensed version should stay as it is currently. image

@CarlosCortizasCT / @ddouglasz 
Any idea what's going on here? So in order to get the drop down menu font to resize when isCondensed is true I had to add 
`option` to `createSelectableSelectStyles` (which already had control/singleValue/dropdownIndicator). When I added `option` initially, all the styling was stripped from the drop down menu, but when I passed in `base: TBase` it looked like all the originally styling came back, only not all of it as @FilPob pointed out the padding now seems off. Is there something else that should be passed in too? Thanks!

CarlosCortizasCT:
> > @jmcreasman Nice! the condensed one looks great now. However somehow there is also an unexpected change in the non-condensed version. The padding for each select item seems to be decreased. The non-condensed version should stay as it is currently. image
> 
> @CarlosCortizasCT / @ddouglasz Any idea what's going on here? So in order to get the drop down menu font to resize when isCondensed is true I had to add `option` to `createSelectableSelectStyles` (which already had control/singleValue/dropdownIndicator). When I added `option` initially, all the styling was stripped from the drop down menu, but when I passed in `base: TBase` it looked like all the originally styling came back, only not all of it as @FilPob pointed out the padding now seems off. Is there something else that should be passed in too? Thanks!

Hi @jmcreasman 

So the issue here is that we are already creating some custom styles for the select component [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/selectable-search-input/src/selectable-search-input.styles.ts#L243):
```js
const selectStyles = createSelectStyles({
    hasWarning,
    hasError,
    menuPortalZIndex,
    isDisabled,
    isReadOnly,
    horizontalConstraint,
  });
```

`selectStyles` contains an object with the base **customized** styles we want to apply to all select inputs (we add them to the result of this function [here](https://github.com/commercetools/ui-kit/blob/main/packages/components/inputs/selectable-search-input/src/selectable-search-input.styles.ts#L253)).

That shared function (`createSelectStyles`) already contains the styles we need to apply to the `option` property and it already handles the condensed version but we're not passing the `isCondensed` value to that function so it can return the adjusted styles.

I just pushed a commit to adjust that and it seems to work fine.

@FilPob will you take another look when you have the time? 🙏

jmcreasman:
Thanks @CarlosCortizasCT that makes sense

FilPob:
Looks good now. Thank you all!

FilPob (approved):

CarlosCortizasCT (approved):

PR #2811, opened 2024-05-08
Fix `useTheme` hook helper

CarlosCortizasCT:
#### Summary

Fix `useTheme` hook helper.

## Description

The `themedValue` helper is now always returning the left-side provided parameter (meaning "old" value) where it should be always returning the right-side one ("new" value) as we already migrated to the new theme.


tdeekens (approved):

chloe0592 (approved):
Thanks @CarlosCortizasCT!

PR #2822, opened 2024-05-20
Refactor `useMutationObserver` hook to manage subtree modifications

CarlosCortizasCT:
#### Summary

Refactor `useMutationObserver` hook to manage subtree modifications.

## Description

If we currently pass the subtree option to the `useMutationObserver` hook, we're not going to have our mutation callback called. This is because we can have a same `MutationObserver` observing different elements and, when we get the internal mutation callback in our hook, we check to see wether the mutated element is one of the ones we have previously registered.
In the case of adding a direct children to the observed element, that's not a problem; however, if we add a grandchildren to the observer element, consumers won't know about it because the mutated element will be an observer element child and we didn't registered those in our hook.

With this change, we introduce some changes for consumers to be allowed to be notified about nested elements mutated within the one registered in the hook.

In this first attempt, a selector is needed by the hook to check whether a mutated element is contained in the observed element.

### DIFF
--- a/packages/hooks/src/use-mutation-observer/use-mutation-observer.spec.jsx
+++ b/packages/hooks/src/use-mutation-observer/use-mutation-observer.spec.jsx
@@ -1,0 +1,61 @@
+/* eslint-disable react/prop-types */
+import { useRef, useState } from "react";
+import { render, screen, waitFor } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


CarlosCortizasCT:
Closing as this PoC was abandoned.

PR #2824, opened 2024-05-21
fix(export-icon): replaced the export-icon with the fixed version

misama-ct:
Replaced the export-icon with a new version that makes the intent more clear.


CarlosCortizasCT (approved):
Thanks Michael 🙇

kark (approved):

ddouglasz (approved):
Thank you, Michael 🙌🏽

PR #2827, opened 2024-05-24
refactor(search-text-input): swap icon and input order (visually)

misama-ct:
I used the css `order` property to change the order of appearance on screen while keeping the logical tabbing order for keyboard-users.


FilPob (approved):

kark (approved):
Awesome!

PR #2830, opened 2024-05-28
Add new sizes for icon buttons

ddouglasz:
#### Summary

We want to be able to introduce some intermediate sizes to take care of some design needs we have. In order to do this, we are introducing a new naming pattern for the icon  `icon` `button secondary` `icon button`

- [ ] Include a new size for `icon` `icon button` `secondary icon button`
- [ ] Introduce a new numeric sizing scale for all buttons that is backwards compatible


CarlosCortizasCT:
Hello @ddouglasz 

I'd like to double check this requirement:
> Change icon size in icon button when size: medium

I think we should avoid changing the current size parameters as that would affect current consumers.

The way I remember the discussion about it, we would keep the medium value size as it is. As we are introducing the numeric range size values, only the size value `30` would actually use the `20px` size for the icon. This way, current consumers won't be affected and those who want to use the `20px` icon size, they can update the value from `medium` to `30`.

@FilPob can you please also share your point of view? 🙏

ddouglasz:
> I'd like to double check this requirement:
> Change icon size in icon button when size: medium

Thank you @CarlosCortizasCT 🙏🏾
I just did not want to paste the ticket here so I copied it's action items. Maybe I can take that off the ticket and here too. I am not sure we are making any of those changes, from the code, I only just added the new field to pass in icon-size:30 and so on, but no change to the initial sizes was made.

misama-ct:
I just pushed the rest of the changes:

- I introduced the new sizing-scale to all buttons mentioned in the figma-file attached to the story and made the new sizes the default in storybook.

- the previous tokennames will continue to work, but won’t show up in sb anymore.

- I’ve been quite granular with the TS annotations in the hope that users will not just see that the previous tokennames are deprecated but also get the new alternative tokenname suggested directly in their editor

Let me know what you think.

ddouglasz:
Thank you @misama-ct 🙏🏾
Is this ready for review so we can move the PR from Draft to ready for review ?

misama-ct:
> Is this ready for review [ ... ] ?

I was wondering about the changeset file. I though I've seen one, but couldn't find any now when I went through the file-changes. Can you have a short look and let me know.

packages/components/buttons/primary-button/src/primary-button.styles.ts
```diff
@@ -3,15 +3,17 @@ import { css } from '@emotion/react';
 import { designTokens } from '@commercetools-uikit/design-system';
 import type { TPrimaryButtonProps } from './primary-button';
 
+// TODO
```
ddouglasz, line 6: Please help me remove this //TODO I left here. Thanks.

ddouglasz (commented):
I just checked, no changeset yet, now that you have finished making all the file changes, I think this is a good time to add a changeset.

.changeset/empty-coins-lick.md
```diff
@@ -0,0 +1,12 @@
...
+'@commercetools-uikit/secondary-button': patch
+'@commercetools-uikit/primary-button': patch
+'@commercetools-uikit/rich-text-utils': patch
+'@commercetools-uikit/checkbox-input': patch
+'@commercetools-uikit/icon-button': patch
+'@commercetools-uikit/icons': patch
+'@commercetools-uikit/design-system': patch
+---
+
+New size and size-names introduced for icons, buttons & IconButtons
```
CarlosCortizasCT, line 12: It would be great to also include here that we're deprecating some value names and also some design tokens.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -70,6 +74,26 @@ const getSizeStyles = (size: TIconButtonProps['size']) => {
         height: ${designTokens.heightForButtonAsBig};
         width: ${designTokens.heightForButtonAsBig};
       `;
+    case '10':
```
CarlosCortizasCT, line 77: Since the previous design tokens are now deprecated, I think we should not use them anymore in our code.
I'm referring to the ones above (`heightForButtonAsSmall`, `heightForButtonAsMedium` and `heightForButtonAsBig`).

packages/components/buttons/icon-button/src/icon-button.tsx
```diff
@@ -18,6 +18,31 @@ import {
...
+
+/**
+ * @deprecated Use '30' from `TSizes` instead.
+ */
+type TMedium = 'medium';
+
+/**
+ * @deprecated Use '40' from `TSizes` instead.
+ */
+type TBig = 'big';
```
CarlosCortizasCT, line 34: I was wondering if we actually need these types since this is a pattern we haven't used before.

Could you maybe explain what's the value of these against using those values in the `TLegacySizes` type?

```ts
type TLegacySizes = 'small` | `medium` | `big`;
``` 

I'm asking because they are not exported nor used within this file.

Thanks.

packages/components/buttons/primary-button/src/primary-button.styles.ts
```diff
@@ -6,12 +6,13 @@ import type { TPrimaryButtonProps } from './primary-button';
...
     case 'small':
     case 'medium':
       return css`
         height: ${designTokens.heightForButtonAsMedium};
       `;
-
+    case '20':
     case 'big':
       return css`
         height: ${designTokens.heightForButtonAsBig};
```
CarlosCortizasCT, line 18: I believe we should not use the deprecated tokens in our code anymore and use the new ones instead.

packages/components/icons/src/templates/icon.styles.tsx
```diff
@@ -15,14 +15,18 @@ export type Props = {
     | 'warning'
     | 'error'
     | 'success';
-  size?: 'small' | 'medium' | 'big' | 'scale';
+  size?: 'small' | 'medium' | 'big' | 'scale' | '10' | '20' | '30' | '40';
```
CarlosCortizasCT, line 18: Perhaps we can add JSDocs here and explain the t-shirt sizes are deprecated.

packages/components/buttons/icon-button/src/icon-button.tsx
```diff
@@ -67,9 +92,9 @@ export type TIconButtonProps<
    */
   theme?: 'default' | 'primary' | 'info';
   /**
-   * Indicates the size of the icon.
+   * Indicates the size of the icon. Available sizes are '10', '20', '30', '40'.
    */
-  size?: 'small' | 'medium' | 'big';
+  size?: TLegacySizes | TSizes;
```
CarlosCortizasCT, line 97: Perhaps we can also use a warning within the updated components to let users know about the deprecated size values. 

[Here](https://github.com/commercetools/ui-kit/blob/main/packages/components/buttons/primary-button/src/primary-button.tsx#L101) is an example on how to use the `useWarning` hook

packages/components/buttons/icon-button/src/icon-button.tsx
misama-ct: Not completely aware of how everything is wired together yet, I had the slight hope that - from a consumer perspective - their editor won’t just indicate that “medium” is a deprecated value, but rather show the deprecation-message for the individual token. This way users could fix it right away in their editor instead of finding the docs online and a possible alternative value.

CarlosCortizasCT: Thanks for the explanation, Michael 🙇 

I think that's a great idea but I'm not sure how we can make it work.

I've just done a quick test and I don't seem to have the message you're looking for in the editor.
Do you know if that's actually possible?

image

misama-ct: I’m tempted to say it is, but It might be just wishful thinking. I’ve seen stroke-lines when using react components, but am not sure anymore if it was the property or the value that got it :D

packages/components/buttons/icon-button/src/icon-button.styles.ts
misama-ct: Right. Does anything speak against removing those from the [token definitions](https://github.com/commercetools/ui-kit/blob/fcc083bba5e0c2d3d4e5130172d9bc2bb61423d1/design-system/materials/internals/definition.yaml#L396) as well or could that have side-effects?

packages/components/buttons/icon-button/src/icon-button.tsx
CarlosCortizasCT: My point of view is that we should introduce the changes for a reason that adds more value for developer experience or project maintainability.
In this case, I would love to see what you are proposing but if we can't achieve that goal, I would vote for going for the usual pattern so we don't introduce anything new that does not provide an actual value.

Let's see what others think about it.

ddouglasz: To be fair, I agree that it would have been be very nice to have it display a stroke for deprecated  properties on that granular level. But if that is not the case at the moment, it would be great to keep the initial workflow we already have since that doesn't seem to show up.

misama-ct:
I managed to incorporate all suggestions & requests and fixed related tests. Please have another look once you have time. Thanks!

.changeset/empty-coins-lick.md
misama-ct: addressed [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-6e938d9c08323975ef0e88bd5e6cd151540b1a0344c267a23d37c428bfb28a22R12) @CarlosCortizasCT

packages/components/buttons/icon-button/src/icon-button.styles.ts
misama-ct: Addressed [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-92bf6a7b4ce14dccc55ad3c0ee5db75a599f7d62e83a1667d1fbebb0117653b0L60-L71), [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-c10084319127c43b8f3365a94b946414de422e1b5df5272a1c28e43375b35e74L12-L18), [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-c10084319127c43b8f3365a94b946414de422e1b5df5272a1c28e43375b35e74L12-L18) and [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-6fc4fa62a60abfb40503b4213ef3a4acf4799c1f0dd1ab890ad10b1f31970ef5L220-L222). (since no test broke, I removed the now unused tokens from the design-system repo as well). @CarlosCortizasCT

packages/components/buttons/icon-button/src/icon-button.tsx
misama-ct: addressed [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-864898bbe411df91fd117fb7208de606c7d979c399fabb1cf5dbe90933fcff94R24), [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-3a56ecdd8c0ec240bd13af3c79568baf5b6c619a5b6e6443dacdae064f24a7a4R21), [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-c7dc6615dcde1e6e98272ffe5ee052955de20e77af64f177c5f0c78b9792acc0R26) and [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-ab4d692c622bc07c621c7eb40977e7f7ab012f23e7d0674d4944849b880f0319L15). @CarlosCortizasCT

packages/components/buttons/primary-button/src/primary-button.styles.ts
misama-ct: addressed [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-c10084319127c43b8f3365a94b946414de422e1b5df5272a1c28e43375b35e74L11-L18) @CarlosCortizasCT

packages/components/icons/src/templates/icon.styles.tsx
misama-ct: addressed [here](https://github.com/commercetools/ui-kit/pull/2830/files#diff-e2b96990d679733055c76ed55fb98bd14d8d66deddd6f44eb8511c64d7ec9636R19)

packages/components/buttons/icon-button/src/icon-button.tsx
misama-ct: I added it to all buttons. [IconButton](https://github.com/commercetools/ui-kit/pull/2830/files#diff-864898bbe411df91fd117fb7208de606c7d979c399fabb1cf5dbe90933fcff94R120), [PrimaryButton](https://github.com/commercetools/ui-kit/pull/2830/files#diff-3a56ecdd8c0ec240bd13af3c79568baf5b6c619a5b6e6443dacdae064f24a7a4R120), [SecondaryButton](https://github.com/commercetools/ui-kit/pull/2830/files#diff-c7dc6615dcde1e6e98272ffe5ee052955de20e77af64f177c5f0c78b9792acc0R153) and [SecondaryIconButton](https://github.com/commercetools/ui-kit/pull/2830/files#diff-ab4d692c622bc07c621c7eb40977e7f7ab012f23e7d0674d4944849b880f0319R93).

packages/components/buttons/icon-button/src/icon-button.styles.ts
CarlosCortizasCT: Bear in mind this is a breaking change as we export the design tokens and consumers might be using them.
To be honest, I don't expect them to be actually using them, but we can't just remove them without notice. That's a package change that could break consumer's code.

misama-ct: @CarlosCortizasCT I just saw that I was smart enough 🤡  to change it the generated design-tokens.ts file and not in the source yaml file. So basically no harm done. Should I still revert it or will it be overwritten anyway?

CarlosCortizasCT: You need to build the packages again so the generated file will be updated and you can commit it with the right data.

CarlosCortizasCT:
Taking a look at the VRTs in Percy, I found some things we might want to fix:
* IconButton: The test named "sizes - when '20'" is using the size 10 instead of 20
* SearchTextInput: the cross icon on the right is now bigger than it should
* SelectableSearchInput: the cross icon on the right is now bigger than it should
* FieldLabel: the info icon is bigger than it should

misama-ct:
> * IconButton: The test named "sizes - when '20'" is using the size 10 instead of 20

Done.

> * SearchTextInput: the cross icon on the right is now bigger than it should
> * SelectableSearchInput: the cross icon on the right is now bigger than it should
> * FieldLabel: the info icon is bigger than it should

Good catch, I replaced the former “medium” token with the token name-equivalent ’30’ instead of it’s former value-equivalent ’20’. Done.

packages/components/buttons/icon-button/src/icon-button.styles.ts
misama-ct: [Done](https://github.com/commercetools/ui-kit/pull/2830/commits/6ac8da4fda401999d2960ac4d717cf9e2c81ec3a)

packages/components/buttons/primary-button/src/primary-button.tsx
```diff
@@ -129,7 +152,10 @@ const PrimaryButton = <TStringOrComponent extends ElementType = 'button'>(
             {props.iconLeft &&
               cloneElement(props.iconLeft, {
                 color: props.isDisabled ? 'neutral60' : 'surface',
-                size: props.size === 'big' ? 'big' : 'medium',
+                size:
+                  props.size === 'big' || props.size === '20'
+                    ? 'big'
+                    : 'medium',
```
kark, line 158: Considering I didn't participate in the discussions about the task, I hope I got everything right. 
Are we only deprecating T-shirt sizes for icon buttons, or for icons as well?
If we do that for icons, I wonder, shouldn't the ternary return the new sizes? And then we could also update https://github.com/commercetools/ui-kit/blob/6ca49442df321fe6e1142c0f36fed073177b019a/packages/components/icons/src/icons.visualroute.jsx#L47 and https://github.com/commercetools/ui-kit/blob/6ca49442df321fe6e1142c0f36fed073177b019a/packages/components/icons/src/icon.story.js#L81.
Otherwise please disregard

packages/components/buttons/secondary-button/src/secondary-button.tsx
```diff
@@ -174,7 +208,10 @@ export const SecondaryButton = <
             {props.iconLeft &&
               cloneElement(props.iconLeft, {
                 color: getIconColor(props),
-                size: props.size === 'big' ? 'big' : 'medium',
+                size:
+                  props.size === 'big' || props.size === '20'
+                    ? 'big'
+                    : 'medium',
```
kark, line 214: Same here

packages/components/buttons/secondary-button/src/secondary-button.visualroute.jsx
```diff
@@ -93,6 +93,14 @@ export const component = () => (
       <SecondaryButton label="A label text" onClick={() => {}} size="medium" />
     </Spec>
 
+    <Spec label='size - when "20"'>
+      <SecondaryButton label="A label text" onClick={() => {}} size="big" />
```
kark, line 97:
```suggestion
      <SecondaryButton label="A label text" onClick={() => {}} size="20" />
```

packages/components/buttons/secondary-button/src/secondary-button.visualroute.jsx
```diff
@@ -93,6 +93,14 @@ export const component = () => (
...
+    </Spec>
+
+    <Spec label='size - when "10"'>
+      <SecondaryButton label="A label text" onClick={() => {}} size="medium" />
```
kark, line 101:
```suggestion
      <SecondaryButton label="A label text" onClick={() => {}} size="10" />
```

kark (commented):
Thanks Michael 🙌 looks good overall!
I'm only leaving minor feedback.

packages/components/buttons/primary-button/src/primary-button.tsx
misama-ct: Indeed, it is a little ambiguous. The ticket mentions introducing the new sizes to the buttons, figma makes it appear as if Icons should get them as well. 

@FilPob Should Icons get the numeric sizing scale as well?

packages/components/buttons/secondary-button/src/secondary-button.visualroute.jsx
misama-ct: I left this intentionally, so that we can visually verify that the mapping (big->20) is leading to the same result as before. I added the new sizes below.

packages/components/buttons/secondary-button/src/secondary-button.visualroute.jsx
misama-ct: Same here.

packages/components/buttons/primary-button/src/primary-button.tsx
kark: Thanks! Just for the record, regardless of the ticket's title, this PR already [covers T-shirt size deprecation in all icons](https://github.com/commercetools/ui-kit/blob/1f5223bc86a985d4fa531f8356f31f27ba76f736/packages/components/icons/src/templates/icon.styles.tsx#L18-L22), so I just wanted to make sure that's also expected here.

CarlosCortizasCT: You're right, Kacper 👍 

Then I agree with you we should use the new sizes values.

CarlosCortizasCT:
> > * IconButton: The test named "sizes - when '20'" is using the size 10 instead of 20
> 
> Done.
> 
> > * SearchTextInput: the cross icon on the right is now bigger than it should
> > * SelectableSearchInput: the cross icon on the right is now bigger than it should
> > * FieldLabel: the info icon is bigger than it should
> 
> Good catch, I replaced the former “medium” token with the token name-equivalent ’30’ instead of it’s former value-equivalent ’20’. Done.
> 
> [Changes are here](https://github.com/commercetools/ui-kit/pull/2830/commits/7c9a6bd30231163e8af659c06f9a924f97b6d286)

@misama-ct 

After taking a second look, there seems to be some things to review yet:
* IconButton: I know we already had VRTs for the legacy sizes but we are including four new values (10, 20, 30 and 40) but we have only included VRTs for 10 and 20. Could we please add VRTs for the other two?
* SearchTextInput: The last VRT ("with is condensed") has a difference and it seems to be related to the size of the lens icon
* SelectableSearchInput: The last VRT ("is condensed") has a difference and it seems to be related to the size of the lens icon

packages/components/buttons/primary-button/src/primary-button.tsx
misama-ct: It also makes sense, considering that LeadingIcon and CustomIcon are already using the new scale-system. However, one problem that I see is, the cleanup afterwards. Cause ideally I'd want to make sure that no icon in ui-kit is used anywhere with one of the old tokens. But since I am merely editing a template, and then generate 147 standalone react-components from it, this will be extremely time-consuming.

misama-ct: I just did a quick implementation, comparable to the button solution (sizeMapping old:new & console warning-helper), resulting in 583 broken tests due to the console.warn.

Any suggestions?

CarlosCortizasCT: I would suggest to change in this PR only the two places Kacper pointed out and let the cleanup for a follow-up PR so we keep the scope of this one contained.

Since I'm FLD this week, I can help with the cleanup (I don't think it would take me much time).

misama-ct:
@CarlosCortizasCT 

> * IconButton: I know we already had VRTs for the legacy sizes but we are including four new values (10, 20, 30 and 40) but we have only included VRTs for 10 and 20. Could we please add VRTs for the other two?

[Done](https://github.com/commercetools/ui-kit/pull/2830/commits/2d46f915caa90bb1f15eb3b6ff4efb9f8fbe2f5d#diff-9ff736d1b759945c5a7dcacc7816445ba450feeb8b3091dfda75d28157bee73cR166-R185)

> * SearchTextInput: The last VRT ("with is condensed") has a difference and it seems to be related to the size of the lens icon
> * SelectableSearchInput: The last VRT ("is condensed") has a difference and it seems to be related to the size of the lens icon

[Done](https://github.com/commercetools/ui-kit/pull/2830/commits/2d46f915caa90bb1f15eb3b6ff4efb9f8fbe2f5d#diff-ddc097b6f908955b04890144b4f5d5d55f408ff34901c3574f8a40d732e63e80L219-R219) & [Done](https://github.com/commercetools/ui-kit/pull/2830/commits/2d46f915caa90bb1f15eb3b6ff4efb9f8fbe2f5d#diff-b60926e4731eb887873b06d24a403a667eddb8bb5b68b714ba4b97bf2e2da175L516-R516).

packages/components/buttons/secondary-button/src/secondary-button.tsx
misama-ct: @kark addressed [here](https://github.com/commercetools/ui-kit/pull/2830/commits/9e54d44555323ba6be36e775edaeee863d5cf338#diff-c7dc6615dcde1e6e98272ffe5ee052955de20e77af64f177c5f0c78b9792acc0L211-R211)

CarlosCortizasCT (approved):
Looks good to me now.

Thanks 🙇 

Please remember to wait for @FilPob's approval

kark (approved):

FilPob:
Thanks guys! looks all good from my side :)

FilPob (approved):

PR #2832, opened 2024-05-31
Shield 794 explore a better way to indicate focused state in checkbox and radiofield components

misama-ct:
The focus-styles for the labels now only kick in if the users' browser detects keyboard-navigation (:focus-visible). Since the actual focusable element is a parent of the label-element, I had to combine it with the :has selector.


FilPob:
works great! thanks :)

FilPob (approved):

CarlosCortizasCT (approved):
Awesome 💯

ddouglasz (approved):
🙌🏽

Issue #2805, opened 2024-04-30
Adding iconLeft to SearchSelectInput with a long value selected shows incorrect layout of input


Sarah4VT:
To recreate the layout issue, look at the SelectSearchInput component with a long value selected. Add an iconLeft value to add an icon to the left side of the input. The result is the icon is on a top row above the input value and icons on the right.

![Zight Recording 2024-04-30 at 04 35 40 PM](https://github.com/commercetools/ui-kit/assets/4260326/bcc66299-d6ac-4150-b11a-c78746ac994a)

_Originally posted by @Sarah4VT in https://github.com/commercetools/ui-kit/issues/2801#issuecomment-2087167748_

ddouglasz:
Thank you for catching this . We will take care of it.

PR #2833, opened 2024-06-03
fix(create-select-styles): preserve default behavior for multi select

dogayuksel:
#### Summary

We should preserve default wrap behaviour for multi selection.
See reference for an example: https://github.com/commercetools/merchant-center-frontend/pull/16911#issuecomment-2144777982

nowrap should be added only when select input is not multi to keep the left icon wrapping fix.

### DIFF
--- a/.changeset/forty-tips-deny.md
+++ b/.changeset/forty-tips-deny.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/select-utils': patch
+---
+
+Allow select inputs with multiple items to wrap.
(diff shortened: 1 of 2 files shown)


CarlosCortizasCT (approved):
Thanks a lot, @dogayuksel 🙇

dogayuksel:
| isMulti false with icon | isMulti true |
| --- | --- | 
| Screenshot 2024-06-03 at 13 03 47 | Screenshot 2024-06-03 at 13 04 03 |

tdeekens (approved):

ddouglasz (approved):
🙌🏾

CarlosCortizasCT:
@commercetools/shield-team-design-system as a follow-up, it would be great to have new VRTs to make sure we don't break this again (it all started [here](https://github.com/commercetools/ui-kit/issues/2805))

Issue #2777, opened 2024-04-11
`LeadingIcon` - component not available via `@commercetools-frontend/ui-kit` package


ByronDWall:
The `LeadingIcon` ui-kit component is not available from the `@commercetools-frontend/ui-kit` package, and instead must be imported by installing `@commercetools-uikit/icons`.

**To Reproduce**

1. install `@commercetools-frontend/ui-kit`
2. try to import `LeadingIcon` into a file, e.g. `import { LeadingIcon } from @commercetools-frontend/ui-kit`
3. Get an error: `Module '"@commercetools-frontend/ui-kit"' has no exported member 'LeadingIcon'.`

**Expected behavior**

`LeadingIcon` component should be bundled with / made available by ``@commercetools-frontend/ui-kit`.  

Having to install `@commercetools-uikit/icons` separately, and import the component using this syntax: `import LeadingIcon from '@commercetools-uikit/icons/leading-icon';` is unexpected and not documented

**Screenshots**

Error when importing from `@commercetools-frontend/ui-kit`:
![Screenshot 2024-04-11 at 10 37 23 AM](https://github.com/commercetools/ui-kit/assets/87667330/0fbedf24-6ada-482c-b484-bda2b58d3737)

Successful import from `@commercetools-uikit/icons/leading-icon`:
![Screenshot 2024-04-11 at 10 52 47 AM](https://github.com/commercetools/ui-kit/assets/87667330/c01b29a1-2a3e-46af-a9a9-66cc8f265144)

PR #2834, opened 2024-06-04
Export LeadingIcon, CustomIcon, and InlineSvg fro…

ddouglasz:
#### Summary
Export `LeadingIcon`, `CustomIcon`, and `InlineSvg` from the global preset.

closes https://github.com/commercetools/ui-kit/issues/2777

## Description
The `LeadingIcon`, `CustomIcon`, and `InlineSvg` ui-kit components are not available from the @commercetools-frontend/ui-kit package.
To fix this, we need to export `LeadingIcon`, `CustomIcon`, and `InlineSvg` from the global preset, allowing users to access them from `@commercetools-frontend/ui-kit`.

With this, we can now use, for example;
`import { LeadingIcon, CustomIcon, InlineSvg } from @commercetools-frontend/ui-kit`


CarlosCortizasCT (approved):
Thanks! 👍

PR #2836, opened 2024-06-04
Set `glob-parent` dependency minimum version

CarlosCortizasCT:
#### Summary

Set `glob-parent` dependency minimum version

## Description

This is due to this security alert about that dependency having a [vulnerability](https://github.com/commercetools/ui-kit/security/dependabot/16).

This is a transitive dependency coming from the `@preconstruct/cli` which we already have in its latest version so that's why I added a new resolution to make sure it also uses the minimum version.
Building seems to work as well, so all good.


ddouglasz (approved):
Thank you

PR #2837, opened 2024-06-07
improve dropdown menu component placement #SHIELD-1235

misama-ct:
#### Summary

The DropdownMenu component has some flaws when it comes to displaying & positioning the dropdown-content. This pull-request attempts to resolve known issues, introducing the following logic changes:

#### Changes

- [x] If there is no space below the trigger to display the dropdown content, display the content above the trigger
- [x] if there is not enough space to display the menu on the preferred position (left|right), display it on the opposite one
- [x] if the menu width is set to ```scale```, the maximum menu-width is limited to ensure it fits into the viewport and preserves the dropdown appearance
- [x] part of the positioning logic is shifted into a`setTimeout` function, doing the positioning after an initial render, since ```horizontalConstraint``` options like 'auto' or 'scale' don't result in height- or width-values that can be used for computing positions.


packages/components/dropdowns/dropdown-menu/src/dropdown-menu.story.js
```diff
@@ -111,4 +115,72 @@ storiesOf('Components|Dropdowns|DropdownMenu', module)
         </SpacingsStack>
       </DropdownMenu>
     </Section>
-  ));
+  ))
```
misama-ct, line 118: I don't think this story is needed for devs (or is it?), but I'm leaving it here for now, in case someone wants to check the positioning-logic in storybook during review.

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.story.js
```diff
@@ -111,4 +115,72 @@ storiesOf('Components|Dropdowns|DropdownMenu', module)
...
+  .add('DropdownMenu - Reposition menu if necessary', () => {
```
misama-ct, line 119: I don't think this story is needed for devs (or is it?), but I'm leaving it here for now, in case someone wants to check the positioning-logic in storybook during review.

CarlosCortizasCT:
Thanks a lot for working on this, Michael 🙇 

To be honest, it's difficult for me to follow along the changes without going too deep since it's a lot of different styles calculation due to the nature of the task.

The thing that is more surprising to me in the `setTimeout` to wait for some styles to be applied as I wasn't expecting it.
I think I understand your explanation but it seems it might be introducing some flickering in some situations (right aligned menu).
It does not always happen, but here are some examples of why I found:


https://github.com/commercetools/ui-kit/assets/97907068/90375db9-20ca-4e92-a8c3-259ae5e98cda


https://github.com/commercetools/ui-kit/assets/97907068/ffb5ec64-74ab-44e3-aba0-a42fb836a165

misama-ct:
> The thing that is more surprising to me in the `setTimeout` to wait for some styles to be applied as I wasn't expecting it.
> I think I understand your explanation but it seems it might be introducing some flickering in some situations (right aligned menu).
> It does not always happen, but here are some examples of why I found:

If there wasn't an 'auto' or 'scale' option, this function wouldn't be needed, because then we'd have full-control over the dropdowns dimensions. But those 2 options outsource the decision of how wide the dropdown will be to the browser (based on dropdown content) and only after the browser has rendered the dropdown, I can query for the actual width and do the proper positioning.

What I could do is, hide the dropdown from the eye with `visibility: hidden` initially and restore the visibility within the `setTimeout` function. This way the dimension calculation would work without flickering, but this only works in a browser-environment, it's basically not mockable in a test.

CarlosCortizasCT:
> > The thing that is more surprising to me in the `setTimeout` to wait for some styles to be applied as I wasn't expecting it.
> > I think I understand your explanation but it seems it might be introducing some flickering in some situations (right aligned menu).
> > It does not always happen, but here are some examples of why I found:
> 
> If there wasn't an 'auto' or 'scale' option, this function wouldn't be needed, because then we'd have full-control over the dropdowns dimensions. But those 2 options outsource the decision of how wide the dropdown will be to the browser (based on dropdown content) and only after the browser has rendered the dropdown, I can query for the actual width and do the proper positioning.
> 
> What I could do to avoid the flickering is, hide the dropdown from the eye with `visibility: hidden` initially and restore the visibility within the `setTimeout` function. This way the dimension calculation would work without flickering, but this only works in a browser-environment, it's basically not mockable in a test.

I think using the `visibility` style is a good option 👍 

Another thing I'd like @FilPob to confirm is the size of the menu when we use the `scale` value.
In the current production version, this makes the menu to grow until the horizontal edge based on its direction but in this PR we're changing that behaviour for it to always use the whole horizontal space available.
Is this something we actually want to change?

misama-ct:
> Another thing I'd like @FilPob to confirm is the size of the menu when we use the `scale` value. In the current production version, this makes the menu to grow until the horizontal edge based on its direction [...]

The `scale` option basically translates to - css-wise - `width: 100%` which in this case means 100% of the available width of the closest parent which enforces a certain width. It actually is a terrible option, because you never know how wide it will be.

Screenshot 2024-06-18 at 11 08 05

But in any case: What I do is limit the maximum width. If the parents-width is not exceeding the viewport this behavior won't be applied.

FilPob:
Uff, I am a bit lost with the `scale` thingy :D Could we rather have a call?

misama-ct:
> Uff, I am a bit lost with the `scale` thingy :D Could we rather have a call?

Sure. Just ping me once you have time or schedule a remeet.

misama-ct:
@CarlosCortizasCT FYI @FilPob and I were checking out the previous behavior of the `scale` option yesterday.

It was consistently creating a dropdown which was as wide as the viewport, aligned with the trigger, appearing to end on the screen-edge but actually overflowing it. 

The solution I proposed will - while not ideal - at least force-fit it on the screen, so it's at step forward. However, we failed to see a use-case for a window-wide dropdown. Is anything speaking against deprecating the `scale` value?

CarlosCortizasCT:
> @CarlosCortizasCT FYI @FilPob and I were checking out the previous behavior of the `scale` option yesterday.
> 
> It was consistently creating a dropdown which was as wide as the viewport, aligned with the trigger, appearing to end on the screen-edge but actually overflowing it.
> 
> The solution I proposed will - while not ideal - at least force-fit it on the screen, so it's at step forward. However, we failed to see a use-case for a window-wide dropdown. Is anything speaking against deprecating the `scale` value?

Hi Michael ✋ 

> However, we failed to see a use-case for a window-wide dropdown. Is anything speaking against deprecating the `scale` value?

I think this is a question for @FilPob to answer.
If we don't see a use case for it, I'm all for deprecating it,

FilPob:
Am also all in for removing the `scale` option for now

misama-ct:
> I think using the `visibility` style is a good option 👍

@CarlosCortizasCT I added the `visibility: hidden` solution and it should™ work without flickering now. I was also able to make sure the tests continue to work (which I failed to do in my initial attempt).

packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
```diff
@@ -3,6 +3,10 @@ import SecondaryButton from '@commercetools-uikit/secondary-button';
 import { act, screen, render } from '../../../../../test/test-utils';
 import DropdownMenu from './dropdown-menu';
 
+// Pause test execution to wait for the DOM-operations to finish
+const waitForDOMOperations = async () =>
```
CarlosCortizasCT, line 7: Perhaps we can use [the utilities](https://jestjs.io/docs/timer-mocks) Jest already providers for dealing with timers in tests. What do you think?

misama-ct: As far as I understood, those utilities are there to avoid actual waiting while running the tests (basically mocking timer-functions). But the `setTimeout` function here is indeed needed to defer the execution of the code following after. The value of `0` milliseconds is making sure, that execution happens immediately after the recent call stack (= calculating the final DOM in this case) is cleared, so there is no actual waiting involved and test-execution won't take any longer than before.

CarlosCortizasCT: Thanks for the explanation, Michael.

I understand the use case and why we need somehow managing the timeout in these tests, what I'm suggesting is that, instead of using custom functions, we can use what the testing library we use already provides.

In this case, using `jest.useFakeTimers()` and `await jest.runAllTimersAsync();` will work and we wouldn't need to introduce custom code in the test suite.

Would that be reasonable for you?

misama-ct: I see, thanks for the hint, i would've never guessed that `runAllTimersAsync` is the key. I went through the jest docs but couldn't find anything fitting, and the crucial part that would eventually have made me realize it is, only shows up in my editor as inline-help after putting the function in use:

Screenshot 2024-06-20 at 09 27 56

I pushed the changes, please have another look. Thanky you very much!

CarlosCortizasCT (approved):
Thanks Michael 👍

ddouglasz (approved):
Quite an interesting change 💯

PR #2842, opened 2024-06-17
feat(selectable-search-input): multiple adjustments

misama-ct:
This pull-request changes the visual order of submit-button (magnifier icon) + input and adds the possibility to add an additional action triggered by an `<IconButton/>` to the `<SelectableSearchInput/>`.

One thing to note is, that the `<IconButton/>` currently has a bug, causing the `theme="primary"` version of it to not look like expected and shown in figma. I double checked though, the IconButton integration is technically correct (the best kind of correct) and will look as expected once the`<IconButton/>` bug is tackled in [another ticket](https://commercetools.atlassian.net/browse/SHIELD-1273).


FilPob (approved):

FilPob:
@misama-ct thanks, looks all good from my side. Could you maybe just change the default right icon to the `codeView` icon in the Storybook story?

misama-ct:
> @misama-ct thanks, looks all good from my side. Could you maybe just change the default right icon to the `codeView` icon in the Storybook story?

Was even thinking about that, but other stories always start with the first icon when it's selectable as input, so I thought I'd stick to that pattern as well.

But I agree, I find the currently displayed dropdown arrow in combination with the unfixed primary icon-button problematic as well.

The only problem that I see is, that we'd tie that particular Icon to the story and we have a bit of unnecessary logic, all in all - in my opinion - neglectable, but let's hear what others have to say.

ddouglasz:
>The only problem that I see is, that we'd tie that particular Icon to the story and we have a bit of unnecessary logic
Yes I agree, I think I have the same feeling about this too, it would slightly go against the patterns other components already use since in the end, the user can choose their icon of choice. 
If it doesn't hurt or bring any complications to have the `code-icon` as default, lets's go for it.

.changeset/late-bees-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/selectable-search-input': minor
+---
+
+support for an additional action added + minor visual changes
```
ddouglasz, line 5: Support for an optional custom action icon to be displayed on the right side of the selectable search input.

ddouglasz (approved):
Thank you, Michael 🙌🏽
I left a tiny feedback

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
```diff
@@ -520,6 +543,23 @@ const SelectableSearchInput = (props: TSelectableSearchInputProps) => {
...
+
+          {props.rightActionIcon && props.rightActionProps && (
+            <div
+              css={css`
+                order: 4;
+                margin-left: ${designTokens.spacing30};
+              `}
+            >
+              <IconButton
+                theme="primary"
```
CarlosCortizasCT, line 555: @FilPob do you think this should be static?
I know we're now thinking on the filters, but from the component point of view, I'm not sure if other use cases would require a different theme.
Perhaps this can also be a property which will use the "primary" value by default.

What do you think?

CarlosCortizasCT:
I was expecting the new VRTs to have a purple icon button but they don't look like that.
Am I missing something?

image

.changeset/late-bees-kiss.md
ddouglasz, line 5:
```suggestion
Support for an optional custom action icon to be displayed on the right side of the selectable search input.```

**Comment by misama-ct** [comment]
> I was expecting the new VRTs to have a purple icon button but they don't look like that. Am I missing something?

@CarlosCortizasCT I wrote in the [description](https://github.com/commercetools/ui-kit/pull/2842#issue-2357783197) that this is expected since the `IconButton` currently does actually not look like design expects it to look. I'll fix that issue soon as well, but in a separate ticket.

File: .changeset/late-bees-kiss.md (hunk 1 above)
Comment (misama-ct) [reply .changeset/late-bees-kiss.md:5]: Reads indeed more sophisticated 🧐 Thank you!

(I adjusted your suggestions slightly, to prevent the backticks from the suggestions to be committed as well)

**Comment by CarlosCortizasCT** [comment]
> > I was expecting the new VRTs to have a purple icon button but they don't look like that. Am I missing something?
> 
> @CarlosCortizasCT I wrote in the [description](https://github.com/commercetools/ui-kit/pull/2842#issue-2357783197) that this is expected since the `IconButton` currently does actually not look like design expects it to look. I'll fix that issue soon as well, but in a separate ticket.

Oh! sorry. I didn't understand that part.

File: packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx (hunk 2 above)
Comment (FilPob) [reply packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx:555]: Yes that was the idea right? @misama-ct . To have the `primary` theme as default, but still changeable by the user

File: packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx [hunk 3]
```diff
@@ -227,6 +231,19 @@ export type TSelectableSearchInputProps = {
    * [Props from React select was used](https://react-select.com/props)
    */
   selectCustomComponents?: ReactSelectProps['components'];
+
+  /**
+   * Custom action icon to be displayed on the right side of the input.
+   * [Available Icons](/?path=/story/components-icons--all-icons)
```
kark, line 237: Including this URL is an interesting concept. However, please note that this JS Doc comment will only work in the deployed Storybook context. So unless we use a FQDN this won't work for Github readme readers, TypeScript users, and in [the npm package readme](https://www.npmjs.com/package/@commercetools-uikit/selectable-search-input)

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
misama-ct: @FilPob Yes, it can be overridden via the `rightActionProps` properties, it accepts all configs that the IconButton does and would look like this:

```
rightActionProps={{ label: 'Click me', onClick: alert('clicked'), theme: 'whatever' }}
```

But I think @CarlosCortizasCT meant - correct me if I'm wrong Carlos - making even the component configurable, e.g `<SecondaryIconButton/>` instead of `<IconButton/>`.

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
misama-ct: Oh no, didn't think about that case :(

I'd rather remove the links then. Thanks!

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
CarlosCortizasCT: No. Sorry, that was my mistake.
I was confused because how you use the `theme` property with the `primary` value in the `IconButton` and didn't realize you're using it as the default value that can be overwritten by the `{...props.rightActionProps}` line below.

A little confusing to me (in terms of readability) but if others engineers find it easy to reason about, it's ok for me.

FilPob:
> > @misama-ct thanks, looks all good from my side. Could you maybe just change the default right icon to the `codeView` icon in the Storybook story?
> 
> Was even thinking about that, but other stories always start with the first icon when it's selectable as input, so I thought I'd stick to that pattern as well.
> 
> But I agree, I find the currently displayed dropdown arrow in combination with the unfixed primary icon-button problematic as well.
> 
> The only problem that I see is, that we'd tie that particular Icon to the story and we have a bit of unnecessary logic, all in all - in my opinion - neglectable, but let's hear what others have to say.

No then I guess it's fine to stick to the existing approach and not add some unnecessary logic that is relevant just for one use case in the MC

packages/components/inputs/selectable-search-input/src/selectable-search-input.tsx
misama-ct: Done.

packages/components/inputs/selectable-search-input/README.md
```diff
@@ -94,6 +94,8 @@ export default Example;
...
 | `selectCustomComponents`   | `ReactSelectProps['components']`                                                       |          |           | Map of components to overwrite the default ones, see what components you can override&#xA;<br/>&#xA;[Props from React select was used](https://react-select.com/props)                                                                                                                                                                                                                   |
+| `rightActionIcon`          | `ReactElement`                                                                         |          |           | Custom action icon to be displayed on the right side of the input.&#xA;[Available Icons](/?path=/story/components-icons--all-icons)                                                                                                                                                                                                                                                      |
```
kark, line 97: Kindly run the generate-readmes script 🙏

packages/components/inputs/selectable-search-input/src/selectable-search-input.styles.ts
```diff
@@ -91,6 +91,10 @@ const getSelectableSearchInputStyles = (props: TInputProps) => [
     }
     width: 100%;
     color: ${getInputFontColor(props)};
+    order: 2;
+    margin-left: 4px;
```
kark, line 95: nit:
```suggestion
    margin-left: ${designTokens.spacing10};
```

kark (approved):

CarlosCortizasCT (approved):
Thanks Michael 👍

PR #2843, opened 2024-06-18
fix(icon-button): Shield 1273 icon button fix styles

misama-ct:
IconButton styles have been restored according to the designs in Figma.

[Bug-Ticket](https://commercetools.atlassian.net/browse/SHIELD-1273)
[Figma](https://www.figma.com/design/UoqmtHSHwxvEZRcbsM4A3Z/CT-product-design-system?node-id=4882-50372&t=ujTbz4AjKJkRERZD-0)


packages/components/buttons/icon-button/src/icon-button.story.js
```diff
@@ -34,8 +34,8 @@ storiesOf('Components|Buttons', module)
             icon={createElement(icons[select('icon', iconNames, iconNames[0])])}
             onClick={action('onClick')}
             label={text('label', 'Accessibility text')}
-            isToggleButton={boolean('isToggleButton', true)}
-            isToggled={boolean('isToggled', true)}
+            isToggleButton={boolean('isToggleButton', false)}
+            isToggled={boolean('isToggled', false)}
```
misama-ct, line 38: I felt like this 2 options shouldn't be active by default, since a toggle-button seems to be less frequently needed than a regular button.

packages/components/buttons/icon-button/src/icon-button.styles.ts
```diff
@@ -155,13 +158,15 @@ const getBaseStyles = (
...
+
         &:hover {
-          background-color: ${designTokens.colorInfo};
+          background-color: ${designTokens.colorInfo40};
           box-shadow: ${designTokens.shadow0};
         }
         ${isActive ? '&,' : ''}
         &:active {
-          background-color: #057fcc;
+          background-color: ${designTokens.colorInfo};
```
misama-ct, line 169: @FilPob  I implemented it as shown in figma, but since it uses the same color for the default and active state, a  `IconButton` with `info` theme can not be used as a toggle button as it wouldn't give any visual feedback.

packages/components/buttons/icon-button/src/icon-button.story.js
FilPob: +1 to that.

packages/components/buttons/icon-button/src/icon-button.styles.ts
FilPob: Oh I see... the toggled state for the info IconButton should have the background-color: `color-info-40` will correct that also in Figma.

misama-ct: `color-info-40` is currently also used in the hover-state. I'm happy to change it, but checking the updated figma screen, the hover-color user there is an undocumented color (`#089BF7`) and not referring to a token.

FilPob: Yes exactly. `color-info-50` for the hover state. Thanks

FilPob (approved):

.changeset/red-queens-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/icon-button': patch
+---
+
+fix the appearance of misc theme-options
```
CarlosCortizasCT, line 5: Since we're introducing a new design token, we should have another changeset for that package telling about it.

CarlosCortizasCT (approved):
Looks good to me. 💯 

Thanks!

ddouglasz (approved):
Thank you 🙌🏽

PR #2845, opened 2024-06-20
feat(multiline text input): multiple adjustments

ddouglasz:
#### Summary

multiline text input multiple adjustments

## Description

Multiple adjustments introduced to the multiline text input which now allows users to be able to add an action icon button of choice to the top right position of the text area, add the condensed feature to accomodate more text and finally, limit text rows as a way of controlling text area height.

- [x] Add a new `<IconButton/>`
- [x] Introduce maxRows property
- [x] isCondensed property


packages/components/inputs/multiline-text-input/src/multiline-text-input.tsx
```diff
@@ -99,6 +108,27 @@ export type TMultilineTextInputProps = {
...
+   * Props for the right-action icon-button. Required when rightActionIcon is provided.
+   * At least a `label` and an `onClick` prop/function need to be provided.
+   */
+  rightActionProps?: TSecondaryButtonIconProps;
+  /**
+   * Set this to `true` to reduce the paddings of the input allowing the input to display
+   * more data in less space.
+   *
+   */
+  isCondensed?: boolean;
```
misama-ct, line 126: I don't know if this is a thing, but if the `isCondensed` property changes from`false` to `true` during runtime , it does not seem to re-trigger some height-calculations and the input is ending up in this state:

image

packages/components/inputs/multiline-text-input/src/multiline-text-input.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+import styled from '@emotion/styled';
+import { designTokens } from '@commercetools-uikit/design-system';
+import { type TMultilineTextInputProps } from './multiline-text-input';
+
+const getMultilineTextInputActionIconStyles = (
+  props: TMultilineTextInputProps
+) => css`
+  position: absolute;
+  right: ${designTokens.spacing30};
+  top: ${props.isCondensed ? '7px' : designTokens.spacing20};
```
misama-ct, line 11: If you want to vertically center the icon within the boundaries of the first row, the following works as well and without a magic number:

```suggestion
top: 0;
  height: ${props.isCondensed
    ? `${designTokens.heightForInputAsSmall}`
    : `${designTokens.heightForInput}`};
  padding: 0;
  display: flex;
  justify-content: center;
  align-items: center;
```

misama-ct (changes requested):
Thanks!

packages/components/inputs/multiline-text-input/src/multiline-text-input.tsx
ddouglasz: Thank you @misama-ct 
Exactly! a question I would also like to ask, especially from your suggestion with the isCondensed height change. Since this is a multiline text input, should the condensed value just work for when there is 1 row of text, what should happen when we have multiple rows? 🤔

My initial thought was basically to just reduce the text size and the padding but I guess we would need @FilPob 's help here 🙏🏾

misama-ct: > should the condensed value just work for when there is 1 row of text, what should happen when we have multiple rows?

I don't think so. And if it starts in the `isCondensed` state everything looks okay (try setting it to true in the sb story and reload). 

So basically this only happens if `isCondensed` changes to true *after* the component has already been rendered once in non-condensed state. The `TextareaAutosize` sets the fixed height, so this component has to be made aware of the size-change somehow.

The only solution I have right now is adding a key-prop to the`TextareaAutosize` usage [here](https://github.com/commercetools/ui-kit/blob/56a2cfeb3a44315a51578e15a17facdc51035569/packages/components/inputs/input-utils/src/multiline-input/multiline-input.tsx#L81), essentially forcing a complete re-render of the `TextareaAutosize` component whenever the prop changes.

```
<TextareaAutosize ...  key={props.isCondensed}/>
```

packages/components/inputs/multiline-text-input/src/multiline-text-input.visualroute.jsx
```diff
@@ -73,6 +74,47 @@ export const component = () => (
         hasWarning={true}
       />
     </Spec>
+    <Spec label="with right action">
```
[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2845/commits/022f3b1b0ff8a0368ef2cad6b3cd9f2867eb1491
thank you

packages/components/inputs/multiline-text-input/src/multiline-text-input.styles.ts
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2845/commits/022f3b1b0ff8a0368ef2cad6b3cd9f2867eb1491
Thank you

packages/components/inputs/multiline-text-input/src/multiline-text-input.tsx
ddouglasz: I see your point. Thanks for the pointer 🙏🏾 

What I also noticed is that the `TextareaAutosize` uses a  `cacheMeasurements` prop to compare before rendering. I think. a sort of optimisation feature. If this value does not "change", then it uses the initial value without any further computing. So I am not so sure how deep it goes with this comparisons but it currently seems like it does not notice the prop change. 
setting `cacheMeasurements={false}` from the `MultilineTextInput` gives us the behaviour we want. 
The question is what significance is what impact is it, turning off `cacheMeasurements` ?

ddouglasz: From what I am reading, this feature would be useful if we make freequent updates that change the measurements of properties like scroll and text area height rapidly like in a live chat or real-time collaborations. 
I don't think it would hurt to turn it off, in my opinion.

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2845/commits/fce428fecc73f6056c9da3533c93a920ad20c49e

@commercetools/shield-team-design-system what do you think of this change please?

misama-ct: I agree. By default caching is not turned on either.

misama-ct (approved):

FilPob:
@ddouglasz thanks! Looks and works beautifully. I would just ask to increase max. possible number of rows in the `max.rows` property to let's say `20`

ddouglasz:
> @ddouglasz thanks! Looks and works beautifully. I would just ask to increase max. possible number of rows in the `max.rows` property to let's say `20`

@FilPob Updated here: https://github.com/commercetools/ui-kit/pull/2845/commits/6bf4fe982fbd60fb2142743c48bbe3245dd308d6
Thank you.

PR #2847, opened 2024-06-25
Users should be able to lock a given currency to the money input/field

ddouglasz:
#### Summary

Ability to have a fixed currency that is locked to the money input/field without a select option to change it once this value is turned.

image


.changeset/shy-poems-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-uikit/money-field': minor
+'@commercetools-uikit/money-input': minor
+---
+
+This change provides the option to be able to set a fixed currency that cannot be modified from the currency select dropdown.
```
CarlosCortizasCT, line 6:
```suggestion
Introduced a new property (`isCurrencyInputDisabled`) which can be used to lock the currency selector element in the component so users can't change its value.
```

CarlosCortizasCT (approved):
Looks good to me 👌 

Thanks

packages/components/inputs/money-input/src/money-input.tsx
```diff
@@ -805,7 +809,11 @@ const MoneyInput = (props: TMoneyInputProps) => {
             inputId={MoneyInput.getCurrencyDropdownId(moneyInputId)}
             name={getCurrencyDropdownName(props.name)}
             value={option}
-            isDisabled={props.isDisabled}
+            isDisabled={
+              props.isCurrencyInputDisabled
+                ? props.isCurrencyInputDisabled
+                : props.isDisabled
+            }
```
misama-ct, line 816: Just so that I have something to comment, here is a little nitpicking 😄 :

```suggestion
isDisabled={props.isCurrencyInputDisabled || props.isDisabled}
```

misama-ct (approved):

packages/components/inputs/money-input/src/money-input.tsx
ddouglasz: Yes, we can do this too. 
Updated here: https://github.com/commercetools/ui-kit/pull/2847/commits/588183887d960bb9caa00de97655d0e6ae6bee68
Thank you

PR #2852, opened 2024-07-10
feat(data table manager): fix data table manager scroll issue

ddouglasz:
When there is a horizontally long table that requires scrolling,
The user should be able to horizontally scroll only the table container  without any other element of view scrolling along.

image

### DIFF
--- a/packages/components/data-table/src/data-table.styles.tsx
+++ b/packages/components/data-table/src/data-table.styles.tsx
@@ -55,7 +55,7 @@ const TableContainer = styled.div<TTableContainer>`
 
   ${(props) =>
     // this is needed in order to have a sticky header
-    props.maxHeight ? `overflow-x: auto;` : ''}
+    props.maxHeight ? `overflow-x: auto;` : 'overflow-x: scroll;'}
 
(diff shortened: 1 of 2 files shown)


packages/components/data-table/src/data-table.styles.tsx
```diff
@@ -55,7 +55,7 @@ const TableContainer = styled.div<TTableContainer>`
 
   ${(props) =>
     // this is needed in order to have a sticky header
-    props.maxHeight ? `overflow-x: auto;` : ''}
+    props.maxHeight ? `overflow-x: auto;` : 'overflow-x: scroll;'}
```
misama-ct, line 58:
```suggestion
overflow-x: auto;
```

In theory, `auto` is solving the same problem as `scroll` does, but in a more elegant way. Introducing scrollbars only when needed instead of having them visible at all times. This should be sufficient, if it is a scrolling-issue.

ddouglasz: Oh yes, I realised and moved it back to wip. Thank you 🙏🏾

PR #2853, opened 2024-07-12
feat(NumberField): add `belowMin` and `aboveMax` error keys to `NumberField`

ragafus:
#### Summary

Adds two new error keys, `belowMin` and `aboveMax`, to improve `NumberField` DX.

## Description

`NumberField` provides a `min` and `max` props which are used as property on input field. Unfortunately, users can introduce numbers above the `max` prop and below the `min` props. ATM, the developer using `NumberField` has to handle all the logic regarding `min` and `max`  errors.
With this PR we provide default error messages for `belowMin` and `aboveMax` error keys, so, the developer using `NumberField` has to just generate these new keys when validating their forms.


packages/components/fields/number-field/src/number-field.tsx
```diff
@@ -278,7 +280,36 @@ class NumberField extends Component<TNumberFieldProps, TNumberFieldState> {
             id={sequentialErrorsId}
             errors={this.props.errors}
             isVisible={hasError}
-            renderError={this.props.renderError}
+            renderError={(key: string, error?: boolean) => {
+              const element = this.props.renderError?.(key, error);
```
misama-ct, line 284: @ragafus Do you have a bit of time for us (remeet) to elaborate as to why this change is needed?

ragafus: > @ragafus Do you have a bit of time for us (remeet) to elaborate as to why this change is needed?

Of course! I can explain my point of view as a "consumer" of `NumberField` component.

ddouglasz: Great! I will set up a meeting for us. 
Thanks again

ragafus:
Hey @misama-ct @ddouglasz, I've refactored the code as we agreed in the meeting. Please, let me know your thoughts.

ddouglasz (approved):
Thank you @ragafus 
Looks good to me now 🙌🏾

ddouglasz:
Please note that we would also need a sign off from our designer @FilPob too.

misama-ct:
Thanks for taking the time to do this change.

@FilPob you can provoke the error-message by setting a min and/or max value [here](https://ui-kit-git-rg-number-field-min-max-errors-commercetools.vercel.app/?path=/story/components-fields--numberfield), entering a number out of the min/max range in the input and then checking the "touched" checkbox.

misama-ct (approved):

FilPob (approved):

FilPob:
Thanks! Looks good from my side

PR #2854, opened 2024-07-12
feat(TextField): forward `maxLength` prop to `TextField` and `TextInput` input element

ragafus:
Closes https://github.com/commercetools/ui-kit/issues/2860

#### Summary
The new prop will allow developers to restrict the max. number of characters of a text input.

## Description
In Priceless we need to render a field that will only accept 1 character.


ddouglasz (approved):
Thank you! 🙏🏾. 
Looks good to me. I left some comments after checking in with our designer. 
We at least have to remind the developers that they should provide a form of visual notification when they use this property. 
I added requested changes in line with this.

image

FilPob (approved):

misama-ct (approved):
Thank you!

PR #2856, opened 2024-07-12
fix(RadioField): widen value prop type to `string | boolean`

ragafus:
#### Summary

The `value` prop of `RadioField` to support boolean values. Boolean values can be used when there are just two `RadioInput.Option`.

## Description

In priceless we want two render two options for a `RadioField` which value is `true` and `false`

Screenshot 2024-07-15 at 11 33 57


ddouglasz (approved):
Thank you!

misama-ct (approved):
Thank you as well!

PR #2858, opened 2024-07-18
SHIELD-1347-storybook-ga-removal-cookie-banner

misama-ct:
Goal was to get rid of unneeded code, unnecessary third-party domain cookies and achieve GDPR compliance by adding the ct cookie-banner:

- [x] Removed Google Analytics from storybook
- [x] Self-Host the “Inter” font in Storybook, get rid of Google Fonts dependency
- [x] Remove the broken circle-ci badge from ui-kit frontpage (it’s setting a cookie)
- [x] Integrate the ct Cookie Banner when sb is served via a commercetools.com subdomain


docs/.storybook/manager-head.html
```diff
@@ -1,3 +1,26 @@
...
+    Object.keys(attributes).forEach((key) => {
+      tag.setAttribute(key, attributes[key]);
+    });
+
+    if (onLoad) tag.onload = onLoad;
+      document.head.appendChild(tag);
+    };
+
+    // Inject the cookie consent scripts only if the page is served on a `*.commercetools.com` domain.
+  if (window && window.location.host.includes('.commercetools.com')) {
```
ddouglasz, line 16: Maybe this is an obvious question but I'll ask anyway 😅
We are injecting two scripts here right? 
One is the consent while the other is the template correct?

ddouglasz (approved):
Thank you for the work. 
I left a small question just out of curiosity.

docs/.storybook/manager-head.html
misama-ct: I basically just copied the code that @emmenko linked in the slack-conversation leading to this request. As far as I managed to figure out, the first script attempts to makes sure that cookies are blocked automatically and the second one is the actual app that displays the UI and allows consenting.

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks, Michael 🙇

PR #2859, opened 2024-07-22
chore: get rid of excess ';' character

misama-ct:
Screenshot 2024-07-22 at 14 18 44

Somehow an excess `;` character was overseen by all of us and is (was) messing with the layout.


emmenko (approved):
😅

PR #2864, opened 2024-07-24
Add empty Second Storybook

misama-ct:
This pull-request is part of the storybook-migration effort (version 5 to 8).
It sets up and configures a second storybook in its newest version, which can be run in parallel together with the legacy storybook.

I separated this storybook-setup from the actual migrated stories, since the review/organisation of the stories will probably drag for a while and meanwhile we would either have no vercel deployments for this second storybook or we would be unable to merge changes to main (since the build commands simply wouldn't be available in the main branch and lead to an automatic 'failed' check).

This is how I imagine things will work:

1. This pull request is merged
2. Since storybook v8 is in the main-branch now, the vercel integration can be set up
3. Once set up, `main` will build an empty second storybook and not throw errors
4. Story migration work can continue on a branch containing the migrated stories only


tsconfig.json
```diff
@@ -39,7 +39,10 @@
...
       // Original: docs/node_modules/@storybook/api/node_modules/@emotion/core/types/index.d.ts
       // Since `@emotion/core` does not exist anymore, let's resolve it to `@emotion/react`.
-      "@emotion/core": ["./node_modules/@emotion/react"]
+      "@emotion/core": ["./node_modules/@emotion/react"],
+      "@/storybook-helpers": ["./storybook/src/helpers/index"]
     }
-  }
+  },
+  // @todo: remove, once storybook v8 replaced the v5 version
+  "exclude": ["./storybook"]
```
misama-ct, line 47: Currently, the 'storybook' folder needs to be excluded from typechecking, since the typechecker [complains](https://github.com/commercetools/ui-kit/actions/runs/10076964662/job/27858442821?pr=2864) about missing dependencies. As long as the v5 storybook version is still present and in use, all the tooling relies on the v5 storybook packages being present, means the failing can not be prevented either.

Once we're confident all v8 stories (separate branch) provide feature parity to the v5 ones, the v5 storybook packages & stories can be removed (as well as this line) and storybook can become a regular workspace.

storybook/serve.json
```diff
@@ -0,0 +1,6 @@
+{
+  "public": "./storybook-static",
+  "cleanUrls": false,
+  "directoryListing": false,
+  "redirects": [{ "source": "/", "destination": "/index.html" }]
+}
```
misama-ct, line 6: Maybe not obvious either, this is the config file for the `serve` command in the package json. It needed a bit of configuration to be able to serve the production-build of storybook in a comparable fashion to what is currently live.

PR #2865, opened 2024-07-25
Add additional Storybook v8

misama-ct:
This pull-request is part of the storybook-migration effort (version 5 to 8). It sets up and configures a second storybook in its newest version, which can be run in parallel together with the legacy storybook.

I separated this storybook-setup from the actual migrated stories, since the review/organisation of the stories will probably drag for a while and meanwhile we would either have no vercel deployments for this second storybook or we would be unable to merge changes to main (since the build commands simply wouldn't be available in the main branch and lead to an automatic 'failed' check).

This is how I imagine things will work:

1.  This pull request is merged
2. Since storybook v8 is in the main-branch now, the vercel integration can be set up
3. Once set up, main will build an empty second storybook and not throw errors
4. Story migration work can continue on a branch containing the migrated stories only (with a working preview accessible to all of us)

### Changes
The changes look scarier than they are. While there are a lot of files changed or added, most of them is unrelated to ui-kit 
 or its build-process, but rather necessary to later demo the design-tokens.

As outcome of the review I just need to be sure that the new storybook works locally for you and the config changes in the linter files and the tsconfig.json won’t impact the current processes (the pipeline already says so).

Trying out the storybook locally:

The storybook folder is technically not part of the mono-repo (yet), so you have to `cd` into the folder and install the dependencies manually (`yarn`). After that you can run storybook by running `yarn start`.


tsconfig.json
```diff
@@ -39,7 +39,10 @@
...
       // Original: docs/node_modules/@storybook/api/node_modules/@emotion/core/types/index.d.ts
       // Since `@emotion/core` does not exist anymore, let's resolve it to `@emotion/react`.
-      "@emotion/core": ["./node_modules/@emotion/react"]
+      "@emotion/core": ["./node_modules/@emotion/react"],
+      "@/storybook-helpers": ["./storybook/src/helpers/index"]
     }
-  }
+  },
+  // @todo: remove, once storybook v8 replaced the v5 version
+  "exclude": ["./storybook"]
```
misama-ct, line 47: Currently, the `storybook` folder needs to be excluded from typechecking, since the typechecker [complains](https://github.com/commercetools/ui-kit/actions/runs/10076964662/job/27858442821?pr=2864) about missing dependencies. As long as the v5 storybook version is still present and in use, all the tooling relies on the v5 storybook packages being present, means the failing can not be prevented either.
Once we're confident all v8 stories (separate branch) provide feature parity to the v5 ones, the v5 storybook packages & stories can be removed (as well as this line) and storybook can become a regular workspace.

storybook/serve.json
```diff
@@ -0,0 +1,6 @@
+{
+  "public": "./storybook-static",
+  "cleanUrls": false,
+  "directoryListing": false,
+  "redirects": [{ "source": "/", "destination": "/index.html" }]
+}
```
misama-ct, line 6: This file is used by the `serve` command of the package.json. It configures the server to be able to serve the production-build of storybook locally in a comparable fashion to what is currently live.

storybook/scripts/postinstall.sh
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env bash
```
misama-ct, line 1: This empty file was necessary, to prevent the `yarn install` command from failing. It is basically run instead of the regular postinstall script. This script does nothing, and thus does not fail at doing so.

kark (commented):
Thank you for working on this and making it work 🙌
> As outcome of the review I just need to be sure that the new storybook works locally for you and the config changes in the linter files and the tsconfig.json won’t impact the current processes (the pipeline already says so).

It looks good in that regard, but I have a question. Do we really need to merge this PR now? It seems like we are cutting corners here and there. Why can't we keep it as a long-running feature branch (since everything is located in the `storybook` folder, there won't be many conflicts), use Vercel branch deployments, and merge it once it can fully replace Storybook v5? At least, that was the previous assumption for the migration, I think.

misama-ct:
> [...] I have a question. Do we really need to merge this PR now? It seems like we are cutting corners here and there. Why can't we keep it as a long-running feature branch (since everything is located in the `storybook` folder, there won't be many conflicts. 

Thank you for having a look.

I don't want to cut corners, the plan was actually exactly as you described, to only deploy the working branch. Unfortunately, after setting up the vercel-project for V8 Storybook and connecting it to the repo, vercels integration assumed that the code for the build-tooling is already in main, and caused all other completely unrelated branches' CI pipelines to fail. I spend a good afternoon going through the docs of vercel but couldn't find a solution to set up a project to only run on a specific branch or only take a specific branch into account, so I tried to find another solution (putting the necessary tooling in main upfront and move on).

> Also, implicitly, we are breaking with the practice of collocating stories with components, right?)

No. The stories for components will be co-located with the components as usual. In my previous [working branch](https://github.com/commercetools/ui-kit/compare/main...storybook-migration-v2#files_bucket) they are basically just *.stories.tsx files side-by-side to the original *.story.js file.

(I added the mdx-files demoing the design-tokens inside of the storybook folder to keep things contained for now, this can change later of course.)

ddouglasz:
@kark If I remember correctly, I think that is the strategy here. To let this run as a long-term feature branch in parallel to the already existing storybook.

kark:
> > [...] I have a question. Do we really need to merge this PR now? It seems like we are cutting corners here and there. Why can't we keep it as a long-running feature branch (since everything is located in the `storybook` folder, there won't be many conflicts.
> 
> Thank you for having a look.
> 
> I don't want to cut corners, the plan was actually exactly as you described, to only deploy the working branch. Unfortunately, after setting up the vercel-project for V8 Storybook and connecting it to the repo, vercels integration assumed that the code for the build-tooling is already in main, and caused all other completely unrelated branches' CI pipelines to fail. I spend a good afternoon going through the docs of vercel but couldn't find a solution to set up a project to only run on a specific branch or only take a specific branch into account, so I tried to find another solution (putting the necessary tooling in main upfront and move on).
> 
> > Also, implicitly, we are breaking with the practice of collocating stories with components, right?)
> 
> No. The stories for components will be co-located with the components as usual. In my previous [working branch](https://github.com/commercetools/ui-kit/compare/main...storybook-migration-v2#files_bucket) they are basically just *.stories.tsx files side-by-side to the original *.story.js file.
> 
> (I added the mdx-files demoing the design-tokens inside of the storybook folder to keep things contained for now, this can change later of course.)

I see. Thanks for the explanations 🙇‍♂️. I don't want to block progress, so I'm personally okay with giving it a go.

kark (approved):

ddouglasz (approved):
Thank you for working on this 🙌🏽. 
Exciting to see where it's heading 🙏🏾

storybook/.storybook/theme.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Typography
+  fontBase: '"Inter", serif',
+  fontCode: 'monospace',
+
+  //
+  // colorPrimary: '#3A10E5',
+  // colorSecondary: '#585C6D',
+
+  colorPrimary: 'hsl(240, 64%, 58%)',
+  colorSecondary: 'hsl(195, 70%, 50%)',
```
ByronDWall, line 23: At some point it might be worth using token/theme values from the `/design-system` package for the storybook theme for consistency/ease of change management.

ByronDWall (approved):
As a means of setting up the tooling necessary to move to storybook v8 on `main`, I think this PR works really well.

Both the v5 and v8 storybooks build for me locally.

I also really like how the tokens are separated out here, it makes it more straightforward to understand what's available.

PR #2870, opened 2024-08-01
fix: update font dependencies in visual-testing-app

misama-ct:
**Problem**
Google updating the font assets for the 'Inter' font in google font caused our visual-testing-app to create diffs for every testcase that has text in it. 

**Solution**
Remove the dependency on google-fonts from the visual-testing-app code.

**Caveats**
While I was able to remove the dependency, I was not able do determine the exact version of the font that google was using before, so I can not avoid having to go through all percy diffs and have them signed off (once).


visual-testing-app/index.html
```diff
@@ -6,18 +6,10 @@
       name="viewport"
       content="width=device-width, initial-scale=1, viewport-fit=cover"
     />
-    <link rel="preconnect" href="https://fonts.googleapis.com">
-    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=Open+Sans:ital,wght@0,300;0,400;0,700;1,400;1,700&display=swap" rel="stylesheet">
     <title>UI-Kit | Visual testing app</title>
```
misama-ct, line 9: There were 2 fonts loaded. One used by ui-kit, one used by the testing app. I felt it's unnecessary to load a custom font, just to use for the title of the violet test-case-header so I removed it as well.

FilPob (approved):

ddouglasz (approved):
Thank you!

PR #2873, opened 2024-08-05
fix: add new grey tone to fix accessibility issues

misama-ct:
The `Text` components' `tertiary` tone was using a grey-tone that did not meet our `AA` contrast goal on default background colors. This PR fixes the insufficient contrast-ratio by adding and additional color to the neutral palette and putting this color to use in the `Text` component.


ddouglasz (approved):
🙌🏾

PR #2874, opened 2024-08-07
refactor: keep js comments when building bundles

emmenko:
We shouldn't remove JS comments from the type declarations, as otherwise we lose important type hints such as `@deprecated`, etc.


misama-ct (approved):
✅  Fully agree, thanks.

PR #2876, opened 2024-08-08
Storybook v8 config + tooling changes

misama-ct:
This pull request includes necessary changes to the to Storybook v8 config / repo-tooling, namely:

- storybook stories written in typescript are temporarily excluded from linters and typecheking (fixing ts-errors may require touching the components themselves, which is out of scope)
- the story sort order in the storybook preview.tsx file was changed to better reflect the hierarchy
- `@storybook/addon-storysource` was added to allow checking the generated source-code of a story from within each stories sidebar


FilPob (approved):

storybook/package.json
```diff
@@ -5,27 +5,29 @@
   "devDependencies": {
     "@chromatic-com/storybook": "^1.6.1",
     "@emotion/css": "^11.10.5",
-    "@emotion/react": "^11.10.5",
```
ddouglasz, line 8: We don't really need this in the root package.json right?

ddouglasz (approved):
LGTM. Thanks!

PR #2877, opened 2024-08-08
Storybook v8 stories batch 1

misama-ct:
## Background

## Goal

## Review instructions

### DIFF
--- a/packages/components/avatar/src/avatar.stories.tsx
+++ b/packages/components/avatar/src/avatar.stories.tsx
@@ -1,0 +1,22 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import Avatar from './avatar';
+import { iconArgType } from '@/storybook-helpers';
+
+const meta: Meta<typeof Avatar> = {
+  title: 'components/Avatar',
+  component: Avatar,
+  argTypes: {
+    icon: iconArgType,
+  },
+};
+export default meta;
+
+type Story = StoryObj<typeof Avatar>;
+
+export const BasicExample: Story = {
+  args: {
+    firstName: 'John',
+    lastName: 'Doe',
+    size: 'm',
+  },
+};
--- a/packages/components/avatar/src/avatar.readme.mdx
+++ b/packages/components/avatar/src/avatar.readme.mdx
@@ -1,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw';
+
+<Meta title="components/Avatar/Readme" />
+
+<Markdown>{ReadMe}</Markdown>
--- a/packages/components/buttons/flat-button/src/flat-button.readme.mdx
+++ b/packages/components/buttons/flat-button/src/flat-button.readme.mdx
@@ -1,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw';
+
+<Meta title="components/Buttons/FlatButton/Readme" />
+
+<Markdown>{ReadMe}</Markdown>
--- a/packages/components/buttons/flat-button/src/flat-button.stories.tsx
+++ b/packages/components/buttons/flat-button/src/flat-button.stories.tsx
@@ -1,0 +1,26 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import FlatButton from './flat-button';
+import { iconArgType } from '@/storybook-helpers';
+
+const meta: Meta<typeof FlatButton> = {
+  title: 'components/Buttons/FlatButton',
+  component: FlatButton,
+  argTypes: {
+    icon: iconArgType,
+    as: {
+      control: 'text',
+    },
+  },
+};
+export default meta;
+
+type Story = StoryObj<typeof FlatButton>;
+
+export const BasicExample: Story = {
+  args: {
+    tone: 'primary',
+    label: 'A label text',
+    onClick: () => alert('Button clicked'),
+    isDisabled: false,
+  },
+};
--- a/packages/components/buttons/icon-button/src/icon-button.readme.mdx
+++ b/packages/components/buttons/icon-button/src/icon-button.readme.mdx
@@ -1,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw';
+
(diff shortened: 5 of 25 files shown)


FilPob:
Awesome! I like the structure of the stories much better! I found one thing that differs from the old storybook. In the Dropdown component, the dropdown menu displays with unnecessary padding. The menu is supposed to have this padding in the case of the "complex menu content" (e.g Filters) But when it's just a simple list of options, there should be no padding. This works correctly in the current storybook.

image

misama-ct:
> I found one thing that differs from the old storybook. In the Dropdown component, the dropdown menu displays with unnecessary padding.

Interesting. I'll have a look why this is the case.

ddouglasz:
Landing page README: 
I don't know if this is something we can change, I noticed the quotes markdown formatting in the new storybook is a little different. I am not so sure we can do anything about that. 
My other observation on the landing page readme is on v5, the link formatting is clear enough to indicate a link. in the v8 it does not give a link clue to the user.
V5
image

V8 
image

misama-ct:
> [...] on the landing page readme is on v5, the link formatting is clear enough to indicate a link. in the v8 it does not give a link clue to the user. 

Great catch. Its a code tag in a link tag and the code styling probably took precedence. This is for sure something I can pick up and improve.

ddouglasz:
README (FOUNDATION):
I like the new folder structure 🙌🏽
I just wonder if the bold formatting on this notice message of the bottom of the `Example` section was taken off deliberately.
V5
image

V8
image

ddouglasz:
Decisions:
we are missing color previews for the decisions:
V5
image

V8
image

misama-ct:
> [...] I just wonder if the bold formatting on this notice message of the bottom of the `Example` section was taken off deliberately.

Not deliberately. However, the markup looks like this:

image

I'd say the styling of old storybook is off. An _em_ tag should be italic because it emphasizes text _subtly_, while **bold** is better suited for the **strong** tag. If both are bold we have a semantic difference but no visual one.

(We could fix the markdown document though later and use a strong tag, if someone insists in having it strong.)

ddouglasz:
> > [...] I just wonder if the bold formatting on this notice message of the bottom of the `Example` section was taken off deliberately.
> 
> Not deliberately. However, the markup looks like this:
> 
> image
> I'd say the styling of old storybook is off. An _em_ tag should be italic because it emphasizes text _subtly_, while **bold** is better suited for the **strong** tag. If both are bold we have a semantic difference but no visual one.
> 
> (We could fix the markdown document though later and use a strong tag, if someone insists in having it strong.)

Yeah, thanks for clarifying 🙏🏾. A simple italic is fine in my opinion. I just felt it's worth drawing your attention to it.

misama-ct:
> Decisions: we are missing color previews for the decisions

@FilPob what do you think? What's the state of "decisions" category? I felt like it's outdated, fragmented and hardly in use anymore so I took a shortcut. Is this important or is the reference to the underlying "choice" good enough for now?

packages/components/avatar/src/avatar.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+  component: Avatar,
+  argTypes: {
+    icon: iconArgType,
+  },
+};
+export default meta;
+
+type Story = StoryObj<typeof Avatar>;
+
+export const BasicExample: Story = {
```
ByronDWall, line 16: This syntax is _much easier to implement_, what an improvement over v5!

ByronDWall (approved):
Props and functionality match for storied between v5 and v8

packages/components/avatar/src/avatar.stories.tsx
misama-ct: What I also noticed is, that if you use `StoryObj` instead of `StoryFn`, you can use the Storybook-UI to create new stories after changing the controls (lower right corner: Save, New, Reset):

image

misama-ct:
> [...] I noticed the quotes markdown formatting in the new storybook is a little different. I am not so sure we can do anything about that. My other observation on the landing page readme is on v5, the link formatting is clear enough to indicate a link. in the v8 it does not give a link clue to the user. 

@ddouglasz  I'll leave the quote style untouched for now (it's still recognizable as quote) but I did an underline to all links, so that they are clearly recognizable as such. [^1]

[^1]: Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.(https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html).

ddouglasz (approved):
🙌🏽🙌🏽

FilPob:
> > Decisions: we are missing color previews for the decisions
> 
> @FilPob what do you think? What's the state of the "decisions" category? I felt like it's outdated, fragmented and hardly in use anymore so I took a shortcut. Is this important or is the reference to the underlying "choice" good enough for now? I wouldn't want to invest more time in it if it's clear that we are gonna remove it.

Yes I think it's fine. No need to put in extra effort

FilPob:
> > Awesome! I like the structure of the stories much better! I found one thing that differs from the old storybook. In the Dropdown component, the dropdown menu displays with unnecessary padding. The menu is supposed to have this padding in the case of the "complex menu content" (e.g Filters) But when it's just a simple list of options, there should be no padding. This works correctly in the current storybook.
> > image
> 
> This is - in my opinion - a design flaw of the component. When `menuType` is not set to `list`, it simply adds `padding` to whatever is inside regardless of the`children`. The `children` are not changed though if you toggle this feature in sb and since the children are currently `Dropdown.ListMenuItem`'s you essentially get double-padding. The exact opposite happens if you open the "Complex Menu Content" story and set the `menuType` from `default` to `list`.
> 
> We may want to create 2 separate components for this 2 different behaviors in the future.
> 
> But in any case, I think the explanation on the prop-page for `menuType` is sufficient for now:
> 
> > **menuType**
> > The type of the menu. The 'default' type just renders a dropdown container but the 'list' type is intended to be used with the DropdownMenu.ListMenuItem component.

Sure if you think that the technical approach should be improved on the component level, than it doesn't need a fix in context of this PR. The dropdown component will be very much needed for the upcoming filter pattern, so we can definitely revisit it as a follow up and see if it fits all requirements for the filters

FilPob (approved):

PR #2878, opened 2024-08-08
Storybook v8 stories batch 2

misama-ct:
## Background

## Goal

## Review instructions

### DIFF
--- a/packages/components/data-table-manager/src/data-table-manager.readme.mdx
+++ b/packages/components/data-table-manager/src/data-table-manager.readme.mdx
@@ -1,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw';
+
+<Meta title="components/DataTable/DataTableManager/Readme" />
+
+<Markdown>{ReadMe}</Markdown>
--- a/packages/components/data-table-manager/src/data-table-manager.stories.tsx
+++ b/packages/components/data-table-manager/src/data-table-manager.stories.tsx
@@ -1,0 +1,412 @@
+// @ts-expect-error
+import type { Meta, StoryFn } from '@storybook/react';
+import DataTable from './../../data-table';
+import DataTableManager from './index';
+import { useState } from 'react';
+import times from 'lodash/times';
+import PrimaryButton from '@commercetools-uikit/primary-button';
+import SecondaryButton from '@commercetools-uikit/secondary-button';
+import { useSorting } from '@commercetools-uikit/hooks';
+import { UPDATE_ACTIONS } from './constants';
+import { DataTableManagerProvider } from '@commercetools-uikit/data-table-manager/data-table-manager-provider';
+import Spacings from '@commercetools-uikit/spacings';
+import SearchTextInput from '@commercetools-uikit/search-text-input';
+
+const meta: Meta<typeof DataTableManager> = {
+  title: 'components/DataTable/DataTableManager',
+  component: DataTableManager,
+  argTypes: {
+    topBar: {
+      control: 'text',
+    },
+  },
+};
+export default meta;
+
+type Story = StoryFn<typeof DataTableManager>;
+
+const items = [
+  {
+    id: '5e188c29791747d9c54250e2',
+    name: 'Morgan Bean',
+    customRenderer: 'CYCLONICA',
+    phone: '+1 (895) 529-3300',
+    age: 23,
+    about:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Arcu dictum varius duis at consectetur lorem donec.',
+  },
+  {
+    id: '5e188c295ae0bb19afbb115f',
+    name: 'Franklin Cochran',
+    customRenderer: 'TINGLES',
+    phone: '+1 (835) 571-3268',
+    age: 36,
+    about:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Arcu dictum varius duis at consectetur lorem donec.',
+  },
+  {
+    id: '5e188c298f0ea901553c517f',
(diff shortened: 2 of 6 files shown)


FilPob:
Only thing missing here are the available props for each column that we currently have. Is this by accident, or is it bringing some complexity? We can think about displaying them in a different way, but having them is quite useful. 
image

misama-ct:
> Only thing missing here are the available props for each column that we currently have. Is this by accident, or is it bringing some complexity? 

This is an example of a rather radical change. While I felt it's necessary to demonstrate all those settings (width, align etc.) I did not replicate the existing UI to set those values from within the story. 

Whenever possible, I want devs to be able to copy & pasta the code that is shown when they hit the "Show code" button under an example on a [props-page](https://ui-kit-storybook-v8-git-storybook-v8-stori-00c4a3-commercetools.vercel.app/?path=%2Fdocs%2Fcomponents-datatable--props):

image

Right now they can use that code with just tiny changes. If I'd add the previous ui, this ui-code would show up there as well and it would be tedious to remove.

You can still set the column properties from the props-table:

image

However, I do see a point. I think there needs to be at least a link to [the signature a column-item](https://ui-kit-storybook-v8-git-storybook-v8-stori-00c4a3-commercetools.vercel.app/?path=%2Fdocs%2Fcomponents-datatable-readme--props#signature-columns) can have, since the `TColumn<Row>[]` type is not much of a hint.

FilPob:
Aah nice! I missed the option in the prop table... thanks!

FilPob (approved):

ddouglasz:
I was going to come here to comment on the missing columns props then I saw your explanation. Brilliant! 👏
First review looks great so far. Maybe something to consider is if we need to mention these properties or possibilities somewhere so the user would know they exist(Or what the table is capable of), even without inspecting the code or seeing a UI for it. 
I am referring to the column settings. 
image

ByronDWall:
> However, I do see a point. I think there needs to be at least a link to [the signature a column-item](https://ui-kit-storybook-v8-git-storybook-v8-stori-00c4a3-commercetools.vercel.app/?path=%2Fdocs%2Fcomponents-datatable-readme--props#signature-columns) can have, since the TColumn<Row>[] type is not much of a hint.

@misama-ct would it be worth implementing a separate story for `column-item`?

misama-ct:
> @misama-ct would it be worth implementing a separate story for `column-item`?

Personally I am just missing the link to the signature, I'm gonna fix that first. But if you think it adds value to have a separate story, I can add one as well. @ByronDWall @ddouglasz Should it then be equivalent to what was visible in the legacy sb or are there any other ideas floating around?

PS: If a column was some sort of subcomponent, we could have used [this feature](https://storybook.js.org/docs/writing-stories/stories-for-multiple-components) of storybook.

ddouglasz:
_The table for each documented subcomponent does not include [controls](https://storybook.js.org/docs/essentials/controls) to change the value of the props, because controls always apply to the main component's args._

I am not sure we want to treat the column as a separate story on it's own.
Also, In the [link](https://storybook.js.org/docs/writing-stories/stories-for-multiple-components) you shared, it looks like we may not really be able to include controls for the columns component. If that is the case, then the main aim is defeated. I think fixing the signature would suffice for now in my opinion.

ByronDWall:
Not adding it as a story is fine with me, it would definitely be a weird usecase.

One thing I noticed, the `column` array is editable on the `props` page, but if you edit the array in the `Basic Example` page, you get an error in both the `Basic Example` and `props` pages
Screenshot 2024-08-13 at 9 08 44 AM
Screenshot 2024-08-13 at 9 08 58 AM

misama-ct:
> [...] One thing I noticed, the `column` array is editable on the `props` page, but if you edit the array in the `Basic Example` page, you get an error in both the `Basic Example` and `props` pages 

@ByronDWall  I was able to figure out the cause. Essentially, it seems to be some sort of (de)serialization issue. Looking at the initial `columns` payload, the second column item contains a `renderItem` and a `headerIcon` property, both claiming to accept `ReactNode`'s.

Both keys are showing up as editable json-structure for `columns` on the props page, but only the `headerIcon` shows up in the _Controls_ section of the _Basic Example_.

Now, if you remove `headerIcon` property from the json-structure on that page (by hitting the ⊖ button next to it), the `renderItem` property seems to get silently ditched (the phone number is not linked anymore) and the json-payload is freely editable again.

Storybook does not seem to be able to deal with JSX-elements directly, but needs a function instead which returns a JSX-element:

works:
```
renderIcon: ()=> (
  <IconButton icon={<InformationIcon />
)
```

doesn't work:
```
renderIcon: (
  <IconButton icon={<InformationIcon />
)
```

I made the Basic Example work by adjusting the `HeaderCell` component of the DataTable, [introducing a`typeof` check for the `iconComponent` property](https://github.com/commercetools/ui-kit/pull/2878/commits/e70db570cfd4461d5463d423d14c63933a1f7538) and change rendering accordingly, @ddouglasz (fyi, since you're working on the DataTable)

ddouglasz (approved):
Looks good from my end 🙌🏽. 
I think we can do a follow up to add the latest changes of the DataTableManager (Nested/Custom settings).
Maybe a separate PR. I Will be happy to make a ticket for this and follow up on this later. 
What do you think about this strategy?

misama-ct:
@ddouglasz Is there anything open on your side?

> Looks good from my end 🙌🏽. I think we can do a follow up to add the latest changes of the DataTableManager (Nested/Custom settings). Maybe a separate PR. I Will be happy to make a ticket for this and follow up on this later. What do you think about this strategy?

Thank you for the offer, I take it!

misama-ct:
@ByronDWall Do you mind having a look at this [comment](https://github.com/commercetools/ui-kit/pull/2878#issuecomment-2288390513) and let me know if there is anything left that needs to be addressed? Thank you.

ByronDWall (approved):
Thanks for all the work on this, it looks really good.  I completely agree that any further datatable-related storybook updates can wait.

PR #2880, opened 2024-08-08
Storybook v8 stories batch 4

misama-ct:
description follows

### DIFF
--- a/packages/components/fields/multiline-text-field/src/multiline-text-field.stories.tsx
+++ b/packages/components/fields/multiline-text-field/src/multiline-text-field.stories.tsx
@@ -1,0 +1,78 @@
+import type { Meta, StoryFn } from '@storybook/react';
+import { iconArgType } from '@/storybook-helpers';
+import MultilineTextField from './multiline-text-field';
+import { useState } from 'react';
+
+const meta: Meta<typeof MultilineTextField> = {
+  title: 'Form/Fields/MultilineTextField',
+  // @ts-expect-error, fix component and/or types
+  component: MultilineTextField,
+  argTypes: {
+    // @ts-expect-error
+    title: { control: 'text' },
+    hint: { control: 'text' },
+    description: { control: 'text' },
+    hintIcon: iconArgType,
+  },
+};
+export default meta;
+
+type Story = StoryFn<typeof MultilineTextField>;
+
+export const BasicExample: Story = (args) => {
+  // @ts-expect-error
+  const { defaultExpandMultilineText } = args;
+  const [value, onChange] = useState('');
+
+  return (
+    <MultilineTextField
+      {...args}
+      key={
+        defaultExpandMultilineText ? 'default-expanded' : 'not-default-expanded'
+      }
+      value={value}
+      onChange={(event) => {
+        onChange(event.target.value);
+      }}
+    />
+  );
+};
+
+BasicExample.args = {
+  // @ts-expect-error
+  id: 'multiline-text-field-id',
+  name: 'multiline-text-field-name',
+  horizontalConstraint: 7,
+  warnings: {
+    customWarning: true,
+  },
+  errors: { missing: true, customError: true },
+  renderError: (key: string) => {
+    switch (key) {
+      case 'customError':
+        return 'A custom error.';
+      default:
+        return null;
+    }
+  },
+  renderWarning: (key: string) => {
+    switch (key) {
+      case 'customWarning':
+        return 'A custom warning.';
+      default:
+        return null;
+    }
+  },
+  isRequired: false,
+  touched: false,
+  isAutofocussed: false,
+  isDisabled: false,
+  defaultExpandMultilineText: false,
+  isReadOnly: false,
+  placeholder: 'Placeholder text',
+  title: 'Description',
+  hint: 'Enter a description',
+  description: '',
+  onInfoButtonClick: () => alert('info button clicked'),
+  badge: '',
+};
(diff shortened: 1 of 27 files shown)


FilPob (approved):

ddouglasz:
Some nice upgrades in the icons. I particularly like that we can copy import snippet to clipboard 😍
Thanks for this! 🙌🏽

Every other view looks great! just some tiny observations here: 

**AsyncCreatableSelectInput**:
1.  `isMulti` is `false` by default but we are using it as `true` by default in V8.
2. In the basic example, isValidNewOption is treated like an object, it seems to me like a boolean is supposed to be returned from its [type definition](https://react-select.com/props)
image
 And maybe this is why it breaks here:

misama-ct:
> **AsyncCreatableSelectInput**:
> 
> 1. `isMulti` is `false` by default but we are using it as `true` by default in V8.
> 2. In the basic example, isValidNewOption is treated like an object, it seems to me like a boolean is supposed to be returned from its [type definition](https://react-select.com/props)
> 

@ddouglasz I set `isMulti` to false and removed the input for the `isValidNewOption` option entirely, as it requires a function as value.

ddouglasz:
> > **AsyncCreatableSelectInput**:
> > 
> > 1. `isMulti` is `false` by default but we are using it as `true` by default in V8.
> > 2. In the basic example, isValidNewOption is treated like an object, it seems to me like a boolean is supposed to be returned from its [type definition](https://react-select.com/props)
> 
> @ddouglasz I set `isMulti` to false and removed the input for the `isValidNewOption` option entirely, as it requires a function as value.

Thank you 🙌🏽. Just caught another one. I think we can do the same thing to the `inputValue ` property too. Clicking it breaks the Story. Maybe we can take that control out too. What do you think?

packages/components/grid/src/grid.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+  align-items: center;
+  justify-content: center;
+  background-color: pink;
+  padding: 16px;
+`;
+
+Placeholder.displayName = 'Placeholder';
+
+/** Use `<Grid/>` & `<Grid.Item/>` to create layouts as simple or complex as you wish: */
+export const BasicExample: Story = (args) => {
```
ByronDWall, line 24: I like these examples a lot, thanks for adding them!

packages/components/fields/multiline-text-field/src/multiline-text-field.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+  },
+};
+export default meta;
+
+type Story = StoryFn<typeof MultilineTextField>;
+
+export const BasicExample: Story = (args) => {
+  // @ts-expect-error
+  const { defaultExpandMultilineText } = args;
+  const [value, onChange] = useState('');
```
ByronDWall, line 25: I understand why, but I should point out that the `value` control doesn't do anything on this story
Screenshot 2024-08-15 at 4 18 21 PM

misama-ct: Good catch. [I made it work](https://github.com/commercetools/ui-kit/pull/2880/commits/0cea1d8300db73a6641bedcb1f1fbe7a176a047d).

ddouglasz: Almost there! 👏
@misama-ct thanks for the fix! I was doing another quick round of review and I found that if `isMulti` is true and the `value` object is clicked, an empty object is rendered in the Input.

misama-ct: @ddouglasz What exactly would be your expectation? Checking the component, to me it behaves as expected, as the initial payload is simply not in the shape that the field expects. As soon as you define a json structure there (Array with object containing a label + value prop) it displays properly.

ddouglasz: Yes, it works as expected @misama-ct . It's just the initialisation is a bit weird on the UI before a value is set. 
Maybe this is a stretch, but I would expect that nothing happens on the interface until the key-value pair have been set so we can prevent the weird empty tag when no value has been set yet.

misama-ct: We'd probably have to touch storybook internals for that, something we do want to avoid. Only thing I can offer for now is to deactivate the `value` prop, which may even make sense, cause even if you manage to configure the value-prop. The Component is then in some sort of _"controlled"_ mode, and the initially given functionality of selecting an item from the suggestion-list after entering some character is not working anymore. What do you think, is deactivating the `value` control an option?

ddouglasz: I think deactivating the value prop from the control should be fine. Since the user can already set values by interacting with the UI, see the json output and also code example. In my opinion, all these are already sufficient to provide the user with enough context for usage example.

misama-ct: I [agree](https://github.com/commercetools/ui-kit/pull/2880/commits/2bf86e4233033e511441127c02718bd60d9d2c71).

ByronDWall (approved):
Affter a lot of back and forth, this is looking good to me

ddouglasz (approved):
Looks good now. Thank you for all the work 🙇🏽‍♂️

PR #2881, opened 2024-08-08
Storybook v8 stories batch 5

misama-ct:
## Background

## Goal

## Review instructions

### DIFF
--- a/packages/components/inputs/creatable-select-input/src/creatable-select-input.stories.tsx
+++ b/packages/components/inputs/creatable-select-input/src/creatable-select-input.stories.tsx
@@ -1,0 +1,163 @@
+import type { Meta, StoryFn } from '@storybook/react';
+import CreatableSelectInput from './creatable-select-input';
+import Spacings from '@commercetools-uikit/spacings';
+import { iconArgType } from '@/storybook-helpers';
+import { useEffect, useState } from 'react';
+
+const meta: Meta<typeof CreatableSelectInput> = {
+  title: 'Form/Inputs/CreatableSelectInput',
+  component: CreatableSelectInput,
+  argTypes: {
+    createOptionPosition: {
+      control: 'select',
+      options: ['first', 'last'],
+    },
+    iconLeft: iconArgType,
+    isMulti: { control: 'boolean' },
+    backspaceRemovesValue: { control: 'boolean' },
+    isClearable: { control: 'boolean' },
+    isSearchable: { control: 'boolean' },
+    closeMenuOnSelect: { control: 'boolean' },
+    allowCreateWhileLoading: { control: 'boolean' },
+    placeholder: { control: 'text' },
+    'aria-label': { control: 'text' },
+    'aria-labelledby': { control: 'text' },
+    'aria-invalid': { control: 'boolean' },
+    'aria-errormessage': { control: 'text' },
+    id: { control: 'text' },
+    containerId: { control: 'text' },
+    isDisabled: { control: 'boolean' },
+    isOptionDisabled: { control: 'boolean' },
+    maxMenuHeight: { control: 'number' },
+    menuShouldBlockScroll: { control: 'boolean' },
+    name: { control: 'text' },
+    tabIndex: { control: 'text' },
+    tabSelectsValue: { control: 'boolean' },
+  },
+};
+export default meta;
+
+type Story = StoryFn<typeof CreatableSelectInput>;
+
+const options = [
+  {
+    options: [
+      { value: 'platypus', label: 'Platypus' },
+      { value: 'goat', label: 'Goat' },
+      { value: 'giraffe', label: 'Giraffe' },
+      { value: 'whale', label: 'Whale' },
+      { value: 'killer-whale', label: 'Killer Whale', isDisabled: true },
+      { value: 'otter', label: 'Otter' },
+      { value: 'elephant', label: 'Elephant' },
+      { value: 'rat', label: 'Rat' },
+      { value: 'anteater', label: 'Anteater' },
+      { value: 'alligator', label: 'Alligator' },
+      { value: 'dog', label: 'Dog' },
+      { value: 'pig', label: 'Pig' },
+      { value: 'hippopotamus', label: 'Hippopotamus' },
+      { value: 'lion', label: 'Lion' },
+      { value: 'monkey', label: 'Monkey' },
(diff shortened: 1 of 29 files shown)


FilPob:
In the DateInput Story the calendar dropdown is cut. Otherwise looks all good. 
image

ddouglasz:
Thanks! 🙌🏽
At the moment, we can only interact with the `CheckboxInput` using the `controls`(formerly `knobs`). Do you think we should at least allow the user to click on the checkbox in the basic example to toggle the state from checked to unchecked too?

misama-ct:
> Thanks! 🙌🏽 At the moment, we can only interact with the `CheckboxInput` using the `controls`(formerly `knobs`). Do you think we should at least allow the user to click on the checkbox in the basic example to toggle the state from checked to unchecked too?

I think so, it's certainly irritating if it does not toggle when clicked. I'll adjust the story or create an additional one.

ddouglasz:
**CreatableSelectInput**

isMulti is false by default in v5 unless the user switches it to true.

isValidNewOption is likely missing the expected value so it breaks the input too.

ddouglasz:
**LocalizedMultilineTextInput**
could you please check the formatting of this again in the README? Looks like it is missing the javascript formatting, unlike the others.

image

ddouglasz:
LocalizedRichTextInput: Basic example
I think we should make the `selectedLanguage` control a select dropdown with already known options, an text input allows the user to put any value, and that seems to break the story.

ddouglasz:
LocalizedTextInput:
We also need to make the `selectedLanguage` a dropdown here too

ddouglasz:
**MultilineTextInput**

1. The `rightActionIcon` is meant to display when `hasRightAction` is set to true. Mainly for storybook usage. It seems like that is what the `rightActionProps` is doing now. But it also behaves like we need to set an object as its value when in reality, it acts as a boolean because just merely toggling it gives us the behaviour we want. 
Pardon me if I may be missing some context, otherwise maybe want to use the `hasRightAction` boolean as it was in SB v5.

V5 


V8

misama-ct:
@FilPob I fixed the DateInput issue, please have another look

@ddouglasz 

**CheckboxInput** 
✅  Fixed, the checkbox can be toggled now (in the basic example).


**CreatableSelectInput**
✅ `isMulti` is set to false by default now
✅  `isValidNewOption` can not be set anymore (since it requires a function)

**LocalizedMultilineTextInput**
> could you please check the formatting of this again in the README? Looks like it is missing the javascript formatting, unlike > the others.

✅ Fixed. In the docs/readme file the code-block was annotated as `js` code while actually being `jsx`code 

**LocalizedRichTextInput + LocalizedTextInput**
> [...] we should make the selectedLanguage control a select dropdown [...]

✅ `selectedLanguage` is now a dropdown including the 3 available languages for this example

image


**MultilineTextInput**
> 
> 1. The `rightActionIcon` is meant to display when `hasRightAction` is set to true. Mainly for storybook usage. It seems like that is what the `rightActionProps` is doing now. But it also behaves like we need to set an object as its value when in reality, it acts as a boolean because just merely toggling it gives us the behavior we want.
>    Pardon me if I may be missing some context, otherwise maybe want to use the `hasRightAction` boolean as it was in SB v5.

I didn't want to introduce the `hasRightAction` toggle cause so far the prop-tables of all components only show actual props that a component has to offer (and it should stay that way). 

I updated the TypeScript annotation for `rightActionIcon` so that it only appears if `rightActionProps` is also set.

image

ddouglasz:
> I updated the TypeScript annotation for rightActionIcon to clarify that it only shows up if rightActionProps is also set.
Yes, this alternative makes sense to me. Thank you.

ddouglasz (approved):
Thank you 🙌🏽

FilPob (approved):

packages/components/inputs/creatable-select-input/src/creatable-select-input.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+BasicExample.args = {
+  horizontalConstraint: 'scale',
+  options,
+  isMulti: false,
+  hasError: false,
+  hasWarning: false,
+  backspaceRemovesValue: true,
+  isClearable: false,
+  isSearchable: true,
```
ByronDWall, line 158: not in scope for this PR, but why even make `isSearchable` a prop if there's no way to create a new option if it's `false`?

packages/components/inputs/date-time-input/src/date-time-input.readme.mdx
```diff
@@ -0,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
```
ByronDWall, line 1: The `props` page for `DateTimeInput` is missing most of the props
Screenshot 2024-08-19 at 10 54 52 AM

packages/components/inputs/localized-multiline-text-input/src/localized-multiline-text-input.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const BasicExample: Story = ({
+  defaultExpandMultilineText,
+  defaultExpandLanguages,
+  ...args
+}: TLocalizedMultilineTextInputProps) => {
+  const [value, setValue] = useState({
+    en: 'Horse\nCow\nDuck',
+    de: 'Pferd\nKuh\nEnte',
+    'nan-Hant-TW': '马\n奶牛\n鸭子',
```
ByronDWall, line 24: nit, Hant-TW is traditional, these translations are in simplified
```suggestion
    'nan-Hant-TW': '馬\n奶牛\n鴨子',
```

packages/components/inputs/date-time-input/src/date-time-input.readme.mdx
misama-ct: This happened before, sb has issues parsing the types from class-based components. I [added a wrapper-component](https://github.com/commercetools/ui-kit/pull/2881/commits/e943fef6605e26e0738a9136e9c60105a759ae61) for storybook. Alternatively one could have manually added each prop-control via the `argTypes` config of the story, but then you'd have to remember keeping those in sync when updating the component.

ByronDWall (approved):
Overall looks good to me, once again definitely an improvement

PR #2897, opened 2024-09-09
chore: bring back the cookie banner in storybook

misama-ct:
#### Summary

The cookie banner got lost since it was added only recently during the migration process. This PR restores it.

## Description
When talking with @jaikamat I was made a ware that we have [a package](https://www.npmjs.com/package/@commercetools-frontend/cookie-consent) for dealing with the cookiebanner, I tried to use it, but unfortunately I could not find a non-hacky way to integrate React components into the storybook app itself, so this initial implementation (which we used in the former storybook) will have to do for now.

I'll do some more investigation once I am working on the UserGuidance survey implementation (as the UserGuidance snippet should only be loaded if the user gave consent via the banner).

> [!NOTE]  
> If you want to see the cookie banner in action, [here](https://github.com/commercetools/ui-kit/pull/2897#discussion_r1756601973) is the comment explaining how


storybook/.storybook/manager-head.html
```diff
@@ -1,3 +1,25 @@
...
+    };
+
+    const bootCookieBanner = ()=>{
+      injectScript('https://cdn.cookielaw.org/consent/b104027d-4d10-4b75-9675-9ffef11562a8/OtAutoBlock.js');
+      injectScript('https://cdn.cookielaw.org/scripttemplates/otSDKStub.js', {
+        'data-domain-script': 'b104027d-4d10-4b75-9675-9ffef11562a8',
+      });
+    }
+
+    if(window && window.location.host.includes('.commercetools.com')) {
```
misama-ct, line 19: If you want to test the banner:

1. Check out the branch
2. Add a `true ||` in front of this  ⬆️  condition
3. Start storybook (`yarn start`)
4. Open storybook in a private tab

ddouglasz (approved):

PR #2902, opened 2024-09-11
FCT-1112 | display story only when running storybook locally

ByronDWall:
feat(local only storybook): add ability to only display a story in the sidebar when running storybook locally by adding a 'local-dev' tag to the story meta

this is a POC - feedback welcome

### DIFF
--- a/packages/components/filters/filter-menu/src/filter-menu.readme.mdx
+++ b/packages/components/filters/filter-menu/src/filter-menu.readme.mdx
@@ -1,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw'
+
+<Meta tags={['local-dev']} title="components/Filters/FilterMenu/Readme" />
+
+<Markdown>{ReadMe}</Markdown>
--- a/storybook/package.json
+++ b/storybook/package.json
@@ -34,6 +34,6 @@
     "start": "storybook dev -p 6006",
     "build-and-serve": "yarn build & yarn serve",
     "build": "NODE_ENV=production storybook build",
-    "serve": "serve"
+    "serve": "serve ./storybook-static"
   }
 }
--- a/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.readme.mdx
+++ b/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.readme.mdx
@@ -1,6 +1,6 @@
 import { Meta, Markdown } from '@storybook/blocks';
 import ReadMe from './../README.md?raw';
 
-<Meta title="components/Dropdowns/DropdownMenu/Readme" />
+<Meta tags={['local-dev']} title="components/Dropdowns/DropdownMenu/Readme" />
 
 <Markdown>{ReadMe}</Markdown>
--- a/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.readme.mdx
+++ b/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.readme.mdx
@@ -1,6 +1,6 @@
 import { Meta, Markdown } from '@storybook/blocks';
 import ReadMe from './../README.md?raw';
 
-<Meta tags={['local-dev']} title="components/Dropdowns/DropdownMenu/Readme" />
+<Meta title="components/Dropdowns/DropdownMenu/Readme" />
 
 <Markdown>{ReadMe}</Markdown>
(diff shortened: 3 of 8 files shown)


storybook/package.json
```diff
@@ -34,6 +34,6 @@
     "start": "storybook dev -p 6006",
     "build-and-serve": "yarn build & yarn serve",
     "build": "NODE_ENV=production storybook build",
-    "serve": "serve"
+    "serve": "serve ./storybook-static"
```
misama-ct, line 37: Was there an issue with this command? The `serve` tool config is in the `serve.json` file and refers to this folder already.

misama-ct:
I like the tag + filter option. This would allow us to have one story-file with multiple stories and the capability to hide only specific ones. 

One feature I'd love to have as well are `mdx`-files that only show up in development and I think you could do that easily by adding a new (or adjusting the existing) glob-pattern in the `main.ts` file.

e.g. `*.dev.mdx` files are only showing up in the dev-storybook.

storybook/package.json
ByronDWall: exactly, `yarn serve` would crash bc it wasn't finding an `index.html` unless I specified the filepath.

ByronDWall:
> One feature I'd love to have as well are `mdx`-files that only show up in development and I think you could do that easily by adding a new (or adjusting the existing) glob-pattern in the `main.ts` file.
> 
> e.g. `*.dev.mdx` files are only showing up in the dev-storybook.

I agree - my first approach was actually an attempt to update the globs in the `main.ts` `stories.config` to conditionally build `*.dev.stories.ts`, and let's just say that part of it was probably my lack of familiarity with glob patterns because it turned out to be more frustrating then expected.

I'll take another go at it though, because I like the pattern, especially if we can use it for `mdx` files as well.

storybook/package.json
ByronDWall: fwiw, here is the output I get
Screenshot 2024-09-12 at 8 59 42 AM

packages/components/filters/filter-menu/src/filter-menu.readme.mdx
```diff
@@ -0,0 +1,6 @@
+import { Meta, Markdown } from '@storybook/blocks';
+import ReadMe from './../README.md?raw'
+
+<Meta tags={['local-dev']} title="components/Filters/FilterMenu/Readme" />
```
ByronDWall, line 4: @misama-ct we can use `tags` in the functional-component flavor of `Meta` for `*.mdx`.  I think I prefer this over a separate glob-based approach in `main.ts`.

misama-ct: Very nice. Perfect.

storybook/package.json
misama-ct: The only way I can reproduce this is by running the `build-and-serve` command and I quickly noticed why it happened: I forgot the second **&** character and the tasks are started in parallel instead of consecutive order 🤦 

Can you fix this command and run it again, or run the build + serve commands manually one after another.
I just ran both commands from the project-root and from within the storybook folder and had no issues.

misama-ct: Another reason I could imagine is that your system is picking up a globally installed version of `serve` which does not support the `serve.json` file yet. Is `serve -v` showing the version specified in the `package.json`?

ByronDWall: @misama-ct I spent _entirely_ too long trying to understand what was causing different behavior on my machine, and I've found that what's causing the issue is that `serve` has a dependency `serve-handler`, which has a dependency on `minimatch:3.1.2`.

In our root package.json, we have a [resolution](https://github.com/commercetools/ui-kit/blob/fba85949e7e95ff84a9f4a874ff8c4a3931c4aab/package.json#L173) to set `minimatch` to always resolve to `^9.0.0`.

The syntax used to call `minimatch` in `serve-handler` is valid in `3.1.2`, but not in `9.x`, thus the error I am seeing.

We resolve the `minimatch` version to greater than `9.x` because older versions have security vulnerabilities, and from what I can tell there is no way to tell `yarn v3` to resolve `3.1.2` only for `serve-handler`.

I verified that this issue also occurs for @tylermorrisford when he installs `main` and attempts to run `yarn serve`.


So out of curiosity, how are you building packages for the monorepo? because I cannot get this to work without modifying something.

ByronDWall: @misama-ct I'm going to revert the change for the purposes of this PR - it can be handled separately if necessary.

PR #2905, opened 2024-09-13
Filters - `FilterMenu` component POC 

ByronDWall:
Implements a working API for `FilterMenu` and shows effective functionality.

In order to see the menu, run `pnpm start` and navigate to http://localhost:6006/?path=/story/components-filters--filter-menu

Proposed initial `FiltersMenu` component API based on this POC:
```
export type TFilterMenuProps = {
  key: string;
  label: string;
  filter: ReactElement;
  operatorsInput?: ReactElement;
  appliedFilterValues:
    | TAppliedFilterValue
    | TAppliedFilterValue[]
    | undefined
    | null;
  isPersistent?: boolean;
  isDisabled?: boolean;
  onRemoveFilter?: Function;
  onApplyFilter?: Function;
  onClearFilter?: Function;
  onFilterOptionsSortClick?: Function;
};
```

[JIRA TICKET FOR THIS POC](https://commercetools.atlassian.net/browse/FCT-1160)


packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -0,0 +1,109 @@
+import { type ReactElement } from 'react';
+import DropdownMenu from '@commercetools-uikit/dropdown-menu';
+import { Footer } from './footer';
+import { Header } from './header';
+import { TriggerButton } from './trigger-button';
+
+export type TAppliedFilterValue = {
+  key?: string;
+  label: string | ReactElement;
```
misama-ct, line 9: If you use `ReactNode` you cover everything that react can render and can ditch the `string |`. Additionally, if a consumer needs to render a number, there is no need to transform it into a string first to do so.

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * the input in which the user selects values for the filter
+   *
+   * NOTE / OPINIONS WELCOME IN PR COMMENTS:
+   * can either pass a react element and let the consuming app choose what input to use,
+   * or start defining different menu bodies based on props and only make predefined input
+   * types available, which is restrictive and much higher maintenance
+   */
+  filter: ReactElement;
```
misama-ct, line 29: 1. Having a `Field`-version of almost every input already takes a mental toll due to all the duplication, personally I don't need a `Filter`-version of every input on top. 
2. We're not enabling product-development if we'd have to whitelist every possible input first, we'd become a bottleneck. 

I'd let the consumer decide what input to use.

And what about using the `children` property for it?

```jsx
<FilterMenu we={take} care={of} repetitive={stuff} only>
   {`<-- complex, individual inputs -->`}
</FilterMenu>
```

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -0,0 +1,109 @@
...
+   * whether or not the filter is disabled
+   */
+  isDisabled?: boolean;
+  /**
+   * controls whether `x` in Trigger Button is displayed - required if `isPersistent` is `false`
+   *
+   * NOTES / OPINIONS WELCOME IN PR COMMENTS:
+   * is the presence/absence of the function sufficient for display logic, or do we need a separate `show-` prop?
+   */
+  onRemoveFilter?: Function;
```
misama-ct, line 64: I'd love to generalize the naming of this sort of actions. What about `onRemoveRequest`, `onApplyRequest`, `onClearRequest` etc. for components that do *NOT* maintain their own state?

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -0,0 +1,109 @@
...
+   */
+  filter: ReactElement;
+  /**
+   * the input in which the user can select which operator should be used for this filter
+   *
+   * NOTE / OPINIONS WELCOME IN PR COMMENTS:
+   * the alternative to a rect element is passing in an array of operators and a handler function,
+   * which is more achievable here than for the input in the body, but could still be restrictive
+   */
+  operatorsInput?: ReactElement;
```
misama-ct, line 37: The inputs never really change, only their payload, right? If so I'd go with the array config.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,83 @@
...
+   */
+  label: string | ReactElement;
+  /**
+   * the values applied to this filter by the user
+   *
+   * NOTE / OPINIONS WELCOME IN PR COMMENTS:
+   * this will almost certainly be some sort of generic w/validation - all the `Chip` really
+   * needs is the selected option's `label`, but these options need to contain a lot more data
+   */
+  appliedFilterValues:
```
misama-ct, line 20: Even if we deliberately only want to deal with a single value, the value can still be fed into the component as an array, this would reduce complexity.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * indicates whether FilterMenu can be removed from the filtersList
+   */
+  isPersistent?: boolean;
+  /**
+   * controls whether `x` in Trigger Button is displayed - required if `isPersistent` is `false`
+   *
+   * NOTES / OPINIONS WELCOME IN PR COMMENTS:
+   * is the presence/absence of the function sufficient for display logic, or do we need a separate `show-` prop?
+   */
```
misama-ct, line 38: > is the presence/absence of the function sufficient for display logic, or do we need a separate `show-` prop?

I am not sure it is. It might work but it might explode in our faces as well if things are getting more complex over time.

Alternatively, we could try if it's possible to have those buttons as separate components and use css grid to render them where they are supposed to be in the layout (if they are present in the DOM-structure).

```jsx
<FilterMenu>
  <FilterMenu.Filter>{...}</FilterMenu.Filter>
  <FilterMenu.ClearButton onClick={bleh()}/>
  <FilterMenu.ApplyButton onClick={bleh()}/>
</FilterMenu>
```

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onRemoveFilter?: Function;
+};
+
+const TriggerButton = (props: TFilterMenuTriggerButtonProps) => {
+  return (
+    <button
```
misama-ct, line 44: I just noticed that this will generate invalid markup since the `<IconButton>` below can not be inside another `<button>`-element.

packages/components/filters/filter-menu/src/filter-menu.tsx
ByronDWall: What's been asked for by product is a `Filters` "pattern". So the approach that I'm setting up for is that consuming apps pass an array of filter configuration objects to the parent `Filters`/`FiltersList` component, and that component then handles creating each `FilterMenu` based on the config.

So, passing it as `children` could work, but I went with `filter` because the component needs to be called something other than `children` when defined in the config.  If that makes sense?

packages/components/filters/filter-menu/src/filter-menu.tsx
ByronDWall: @misama-ct I like the idea. So the pattern would look like:
`onApplyRequest`
`onClearRequest`
`onClearAllRequest`
`onRemoveRequest`
`onRemoveAllRequest`
`onSortRequest`
?

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
ByronDWall: youre not wrong - this is one of those places where I go tired of fighting the types in our `select` inputs in favor of getting a working POC.  But I think that enforcing `appliedFilterValues` as an array makes a lot of sense.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
ByronDWall: This is a really good point. Updated with a dirty hack.  I'm not sure how to achieve this design otherwise

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
Sarah4VT: I would also highly encourage an array after dealing with the filter types in AL. There is a lot of extra code required to handle single value vs. array if you allow both.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
Sarah4VT: A couple of random thoughts:
- no matter which direction we go on the presence of a function being sufficient for display logic, we should try to be consistent across components. I personally don't want to set 2 props to get a specific functionality when one will do.
- I've built compound components like Michael's suggesting and I have really liked that pattern. If we are not reusing these components outside of the filter menu, that makes it super easy to understand the code and compartmentalize it.
- I have the same questions about the overlap of "isPersistent" and the presence of the onRemoveFilter function to show the "x" button as I did on the overall filter menu component

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -461,6 +461,7 @@ const SelectInput = (props: TSelectInputProps) => {
             } as ReactSelectProps['components']
           }
           menuIsOpen={props.isReadOnly ? false : props.menuIsOpen}
+          hideSelectedOptions={false}
```
Sarah4VT, line 464: Why was this added?

Sarah4VT (commented):
This looks really great Byron! An amazing start. This will make it easy for people to come in and fill out the components. I had very few comments, but mostly they are based on my PTSD from working on AL filters.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
misama-ct: > I personally don't want to set 2 props to get a specific functionality when one will do.

I just remember that storybook was (and still is) injecting `log` functions in every property that accepts functions but have no value specified in the story-config.

I noticed this when migration a story and it essentially means:  to *not* display one of those buttons in a sb-story we'd have to manually (remember to) set the callback-property to `undefined`.
 
I'm saying this because I could imagine some testing-tools behave similar and this implicit behavior might cause issues further down the line.

> I've built compound components like Michael's suggesting and I have really liked that pattern. If we are not reusing these components outside of the filter menu, that makes it super easy to understand the code and compartmentalize it.

Indeed. We could avoid bloat in the parent components' props and consumers could just render the button if needed and use standard button event-handlers.

packages/components/inputs/select-input/src/select-input.tsx
ByronDWall: to not displaythe selected value(s) in the search box of the input, they should only display in the `TriggerButton`

packages/components/filters/filter-menu/src/filter-menu.tsx
ByronDWall: `ReactNode` works for me

packages/components/filters/filter-menu/src/filter-menu.tsx
ByronDWall: The rough idea is that wherever the overall `Filters`/`FiltersList` component is defined, the consuming app would pass down an array of objects that configure each `FilterMenu`, as pseudocoded below:

```jsx
const filterDefinitions = [
 {
  key: 'filterName',
  filter: <ComponentPassedFromConsumingApp props={...data/handlersFromConsumingApp}
 ...otherConfig
 }
]
<FiltersList
  filters={filterDefinitions}
  {...otherProps}
/>

```

`FiltersList` would then pass down `filterDefinitions.filter` to `FilterMenu`.  

So the approach is an "option 1" approach, in that the consuming app defines the component that is rendered in the `FilterMenu` body.  Does that help to explain what I meant in my previous comment?

As far as the naming of the prop, I'm fine with passing it as `children`.

I'm also open to suggestions for the key used in the config object passed to the parent `FiltersList` component.  We are expecting some sort of input, so `filterInput` works for me.  `menuBody` or `filterMenuBody` would also make sense.

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -1,19 +1,107 @@
...
+  isPersistent?: boolean;
+  /**
+   * whether or not the filter is disabled
+   */
+  isDisabled?: boolean;
+  /**
+   * controls whether `x` in Trigger Button is displayed - required if `isPersistent` is `false`
+   *
+   * NOTES / OPINIONS WELCOME IN PR COMMENTS:
+   * is the presence/absence of the function sufficient for display logic, or do we need a separate `show-` prop?
```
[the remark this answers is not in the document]
ByronDWall: I also don't think we need a separate "show-" prop.

We don't want to display a "remove" button without a handler function, so it sounds like we agree that displaying the remove button when `!isPersistent && onRemoveRequest` is sufficient.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
ByronDWall: The approach is fine in abstract, and makes sense in a lot of cases.  I'm not sure this is one of them.

The idea behind having a more restricted `header` and `footer` is to enforce consistent appearance and behavior, as specified in the UX specs.  

Making the buttons available in a compound component relies on consumers to import and declare the buttons on their side, which I think makes consumers more likely to create custom headers and footers because they are declaring the markup.

packages/components/inputs/select-input/src/select-input.tsx
ByronDWall: updating inputs is a whole other side to this, there will need to be a lot done in all the `select` inputs in particular.  But that is a separate scope of work.

packages/components/filters/filter-menu/src/filter-menu.tsx
ByronDWall: fixed in [59a151d](https://github.com/commercetools/ui-kit/pull/2905/commits/59a151d2939c54b37b9a36acc3c70f09fd6fdfde)

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
ByronDWall: fixed in [90c15d8](https://github.com/commercetools/ui-kit/pull/2905/commits/90c15d85c7db4ffd21c45e1407294b7049a0c7f2)

packages/components/filters/filter-menu/src/filter-menu.tsx
misama-ct: > this also allows null and undefined on what we are considering to be a required prop. How does TS resolve that? (I still feel like a TS newbie sometimes)

@Sarah4VT I was curious and tried it out:
image

Apparently you'd still be required to explicitly set the label property to `undefined` etc.

Sarah4VT: Ah that makes sense.

packages/components/filters/filter-menu/src/filter-menu.tsx
Sarah4VT: I see what you mean about not calling it children if you are setting it up in the filterDefinitions. If this is where you set it up, I like having a more descriptive name for it. I actually like incorporating `menuBody` somehow as that tells me where in the filter it's going to be used/rendered. Throwing out other options: `filterMenuInput` or `filterBodyInput` or `menuBodyInput`... 🤷‍♀️  sometimes I hate naming...

packages/components/inputs/select-input/src/select-input.tsx
Sarah4VT: Yeah, this was just a flag for me because it's at the input level, which means it affects all input's, and by setting it to false, it's not configurable at all. Just wasn't sure if this had side effects outside filters that would be undesired, but sounds like there is more to figure out here.

jmcreasman (approved):
Read over and agree with the decisions made and the conversations around them.

Sarah4VT (approved):
Approve the overall direction of the POC.

ByronDWall:

PR #2911, opened 2024-09-17
feat: pretty filter labels

misama-ct:
creating a draft to have a storybook which displays the current filter status


packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -0,0 +1,124 @@
+import { type ReactElement } from 'react';
+
+import { Footer } from './footer';
+import { Header } from './header';
+import { TriggerButton } from './trigger-button';
+import * as Popover from '@radix-ui/react-popover';
```
ByronDWall, line 6: I'm really happy with how radix handles the `popover` compared to the `DropdownMenu`.  I think we should use it.

I also think we should consider replacing the `DropdownMenu` internals with the radix `popover` instead of trying to implement complicated bugfixes and a11y updates to it ourselves.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,216 @@
...
+  RefObject,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
+const { colorNeutral85, colorNeutral40, colorPrimary, fontSize20 } =
+  designTokens;
+
+const useScrollObserver = (ref: RefObject<HTMLElement>, totalCount: number) => {
```
ByronDWall, line 20: This is a really elegant solution to hiding `Chip`s and displaying a count of how many `Chip`s are hidden in the `Badge`.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,216 @@
...
+  paddingLeft: 0,
+});
+
+const BadgeContainer = styled.div({
+  position: 'absolute',
+  right: 0,
+  height: '2rem',
+  display: 'flex',
+  alignItems: 'center',
+  background: 'linear-gradient(to right, transparent, white 25%)',
```
ByronDWall, line 113: the gradient is a nice touch

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,216 @@
...
+              e.stopPropagation();
+              props.onRemoveFilter!();
+            }}
+            size="10"
+          />
+        )}
+      </ClearButtonContainer>
+
+      <CaretDownIcon size="small" color="neutral60" />
+      <OverlayButton ref={ref} aria-label={label} {...rest} />
```
ByronDWall, line 211: This is a good solution to not requiring nested buttons

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  return (
+    <Container>
+      <Label>{props.label}</Label>
+      {filtersApplied && (
+        <ValueContainer ref={chipListRef}>
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'row-reverse',
```
ByronDWall, line 193: Reversing the chips so that the most recently selected value is displayed first leads to some unexpected behavior in which the count becomes unreliable, as shown in the screen grab below:

https://github.com/user-attachments/assets/470be51f-6af4-4dd9-8334-7fe0142c53b2


This is the one update in this POC that I don't think we should use in the `FilterMenu`

misama-ct: I reverted the reversing :trollface:

jmcreasman (approved):
I really love how clean the design looks with this approach, the fade in, gradient, it all works well together. I looked over the Decisions Made section of the ticket and agree with it all.

Sarah4VT (approved):
Love the slick UI of this.

ByronDWall:

PR #2916, opened 2024-09-18
chore(uikit canary version): update gh actions

ddouglasz:
#### Summary

ensure canary version does not use already existing version.


misama-ct (approved):

.github/workflows/preview-release.yml
```diff
@@ -74,28 +74,13 @@ jobs:
...
-          githubTagName: v${{ steps.release_version.outputs.VALUE }}
-        env:
-          GITHUB_TOKEN: ${{ steps.generate_github_token.outputs.token }}
-          SKIP_POSTINSTALL_DEV_SETUP: true
-
-      # Publish canary releases only if the packages weren't published already
+      # Publish preview releases only if the branch starts with "preview/" and packages weren't published already
       - name: Publishing preview releases to npm registry
-        if: steps.changesets.outputs.published != 'true' && startsWith(github.ref, 'refs/heads/preview/')
+        if: startsWith(github.ref, 'refs/heads/preview/')
```
ByronDWall, line 79: what if we make publishing opt-in instead of publishing every commit to branches with open PRs?

That would drastically reduce the amount of artifacts we are publishing to npm, reduce the amount of compute used per action, and still allow for preview publishing whenever needed.
```suggestion
        if: startsWith(github.ref, 'refs/heads/preview/')
        if: contains(github.event.head_commit.message, '[publish_preview]'
```

ddouglasz: Thats a good idea. So if we want to trigger a version then we add a trigger string to the commit message too. Makes it more granular.
Thank you.

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2916/commits/ac083e1f7f7bf4dd1996adc851748c31a7eaa3bc

.github/workflows/preview-release.yml
```diff
@@ -74,28 +74,13 @@ jobs:
...
+      # Publish preview releases only if the branch starts with "preview/" and packages weren't published already
       - name: Publishing preview releases to npm registry
-        if: steps.changesets.outputs.published != 'true' && startsWith(github.ref, 'refs/heads/preview/')
+        if: startsWith(github.ref, 'refs/heads/preview/') && contains(github.event.head_commit.message, '[publish_preview]')
         run: |
           BRANCH_NAME=$(echo "${GITHUB_REF#refs/heads/}" | tr '/' '-')
-          DATE=$(date +"%Y%m%d")
           git checkout ${{ github.head_ref }}
-          yarn changeset version --snapshot "0.0.0-${BRANCH_NAME}-${DATE}"
+          yarn changeset version --snapshot ${BRANCH_NAME}"
```
ByronDWall, line 83: @ddouglasz I think this `"` is what's causing [this workflow](https://github.com/commercetools/ui-kit/actions/runs/10926553069/job/30333018704?pr=2918) to fail

PR #2922, opened 2024-09-19
FCT-1150-ui-kit-make-icon-components-pass-a-11-y-tests

misama-ct:
The goal of this PR is to eliminate false positive accessibility (a11y) alerts in Storybook. Since we use icons to complement existing content or functionality, applying `role="img"` to every svg-element is not helpful. Instead, I replaced the `role="img"` attribute with `aria-hidden="true"`,  ensuring that icons are hidden from screen readers, as they do not convey a consistent, generalizable meaning across different applications/features.

> [!NOTE]
> There are only 2 file-changes ([one](https://github.com/commercetools/ui-kit/pull/2922/files#diff-1ce2266b986e8dcb3357d2c1bd9fdbefdfd1157e985ab4c940d39bf24adf79bb), [two](https://github.com/commercetools/ui-kit/pull/2922/files#diff-952a56be4574bc44b985ba2a7001236ef95c59a49e01633d7a2b13b498cb1e6d)) that need to be reviewed, the other changes are caused by `yarn generate-icons`.


packages/components/stamp/src/stamp.spec.js
```diff
@@ -8,11 +8,22 @@ jest.mock('@commercetools-uikit/utils', () => ({
...
-  render(<Stamp tone="positive" label="Hello" icon={<AngleDownIcon />} />);
+  render(
+    <Stamp
+      tone="positive"
+      label="Hello"
+      icon={<AngleDownIcon data-testid="tested-icon" />}
+    />
+  );
   expect(screen.getByText('Hello')).toBeInTheDocument();
-  expect(screen.getByRole('img')).toBeInTheDocument();
```
misama-ct, line 13: Do I have to be worried, could this happen in product-land as well?

ddouglasz: You mean how this would affect the MC ?

ddouglasz (approved):
Thank you.

packages/components/stamp/src/stamp.spec.js
ByronDWall: from what I can tell, all the tests that use `screen.getByRole('img')` in the MC are testing for actual images, not icons

svgr.config.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
       'prefixIds',
     ],
   },
-  svgProps: { role: 'img' },
+  svgProps: { 'aria-hidden': 'true' },
```
ByronDWall, line 95: Since we require a `label` in the `IconButton` (which is the only place Icons are really displayed by themselves) I think this makes sense

ByronDWall (approved):

jaikamat (approved):

PR #2923, opened 2024-09-19
FCT-1153 - `FiltersList` POC - configuring each filter via config object

ByronDWall:
Initial POC for a data structure for the overall `FiltersList` component.


packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -1,11 +1,135 @@
...
+  <div ref={ref} {...restInnerProps}>
+    {children}
+  </div>
+);
+
 function FiltersList(props: TFiltersListProps) {
-  return <div>{props.label}</div>;
+  const [selectedFilters, setSelectedFilters] = useState<string[]>([]);
+
+  const persistedFilters = props.filters
```
ByronDWall, line 61: I'm going to be the first to admit that there is work to be done on the internals of how we define persistent/selected/active filters (that show up in the filter bar)

jmcreasman:
Will this end up having similar behavior to the `FilterMenu` POC where there's a counter? In this case for the number of Filters selected from the list.

packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -1,11 +1,135 @@
...
+import { MenuProps, MenuListProps } from 'react-select';
+import DropdownMenu from '@commercetools-uikit/dropdown-menu';
+import FilterMenu from '@commercetools-uikit/filter-menu';
+import SecondaryButton from '@commercetools-uikit/secondary-button';
+import SelectInput from '@commercetools-uikit/select-input';
+
+export type TAppliedFilterValue = {
+  filterKey: string;
+  /** does ReactNode | ReactNode[] make sense for single/multiple labels (shown in FilterMenu Chips) */
+  label: ReactNode[];
```
jmcreasman, line 12: I'm guessing there wouldn't be much value in a singleLabel / multiLabel spilt?

packages/components/filters/filters-list/src/filters-list.stories.tsx
```diff
@@ -1,22 +1,137 @@
...
+              Array.prototype.concat(e.target.value ? e.target.value : [])
+            )
+          }
+          components={{ Menu: CustomSelectMenu, MenuList: CustomMenuList }}
+          menuIsOpen={true}
+          controlShouldRenderValue={false}
+        />
+      ),
+      onClearRequest: clearPrimaryColorFilter,
+    },
```
misama-ct, line 97: - This is essentially the config/settings for a single `FilterMenu`, right? What about a `getTags` function as some sort of reducer, which, when executed, returns an array of Tags to be displayed in the underlying FilterMenu. This way any kind of state could be brought into a shape that the FilterMenu understands.

- Should props that are forwarded to the FilterMenu be grouped under a parent key (e.g. `filterMenuConfig`)

- `filter` is so ambiguous/generic, maybe smth. like `filterComponent` or `filterBody` instead? And for performance reasons `filter` (or whatever the new name will be) should be a function returning a React component (otherwise the code is evaluated on every render, regardless of whether this filter is currently displayed or not).

packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -1,11 +1,135 @@
...
+        )!;
+        return (
+          <FilterMenu
+            key={activeFilterConfig.key}
+            filterKey={activeFilterConfig.key}
+            label={activeFilterConfig.label}
+            filter={activeFilterConfig.filter}
+            appliedFilterValues={props.appliedFilters.filter(
+              (appliedFilter) => activeFilter === appliedFilter.filterKey
+            )}
```
misama-ct, line 106: If every item in the`filter` config object would have a `getTags` function, it would be as easy as calling it here and there would be no need to have a separate `appliedFilters` object.

packages/components/filters/filters-list/src/filters-list.tsx
misama-ct: Out of curiosity: What's the definition of a `persistent` filter? And does the component need to know what a persistent filter is? (Or is it enough if it knows that it needs to be rendered or not (e.g. via show/hide prop on a filter-config item).

packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -1,11 +1,135 @@
...
+          .filter((filter) => filter.groupKey === filterGroup.key)
+          .map((filter) => ({ value: filter.key, label: filter.label })),
+      }));
+    } else {
+      return props.filters.map((filter) => ({
+        value: filter.key,
+        label: filter.label,
+      }));
+    }
+  };
```
misama-ct, line 79: Couldn't this also just be part of an initial config if me make a distinction between possible filters and filters that are in use?

misama-ct:
One issue that is not handled yet and where I am wondering how we would solve it is the `Apply` button feature. How would we manage state externally, but at the same time reflect the current state visually, without knowing anything about it's original shape or already applying it?

packages/components/filters/filters-list/src/filters-list.tsx
ddouglasz: I'd go with an array of ReactNodes, as it already is for now, at least this is the simplest form that captures both use cases. If there is a need to change, I am not sure a refactor would be too complex.

packages/components/filters/filters-list/src/filters-list.stories.tsx
ddouglasz: > filter is so ambiguous/generic, maybe smth. like filterComponent or filterBody instead? And for performance reasons filter (or whatever the new name will be) should be a function returning a React component (otherwise the code is evaluated on every render, regardless of whether this filter is currently displayed or not).

Going through the document Sarah put together, I noticed the "filter" word appears a lot in the naming, in some cases, almost duplicate naming in different contexts though, is it worth the time to discuss naming patterns regarding filters ?

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: Requiring the label to always be an array allows the component to handle single/multi values gracefully.  It also means that developers in consuming apps will always pass labels into the component in the same shape, which should lead to less divergent data manipulation to build labels on that side.

packages/components/filters/filters-list/src/filters-list.stories.tsx
ByronDWall: > What about a getTags function as some sort of reducer, which, when executed, returns an array of Tags to be displayed in the underlying FilterMenu.

@misama-ct The only drawback that I can see to this approach is that it might not be straightforward to implement if the filter has an `onApply` action

> Should props that are forwarded to the FilterMenu be grouped under a parent key (e.g. filterMenuConfig)

something like this would work for me (pseudocoded): 
```
{
  key
  label
  filterMenuConfig: {
    menuBody
    onCloseRequest
    ...rest
  }
  ...rest
 }
```

> And for performance reasons filter (or whatever the new name will be) should be a function returning a React component (otherwise the code is evaluated on every render, regardless of whether this filter is currently displayed or not).

This is a good point, in that case should the prop for the input/body start with `render-` e.g. `renderMenuBody`

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: Are you saying that we would pass in the `options` for the "add filters" menu as a separate config array?

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: According to [the figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-6256&m=dev),

> Persistant filter-tag is sort of a quickfilter. It is displayed per default and can’t be removed by the “x” button

I don't know if a show/hide is enough, because the point of a "persistent" filter is that it's not removable from the list, which means that the `FiltersList` needs to know that in order to always display it, and the `FilterMenu` needs to know it to not display an `x`/remove button.

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: This makes sense to me, I'll update the POC.

ByronDWall:
> One problem we haven't solved yet is how the "Apply" button should work. I'm trying to figure out how we can manage the state outside of the component but still show the updated state visually after a UI change (ideally without knowing the original structure of the state or actually applying the new state yet).

this is a really good question, and I can see where things could get very complicated very quickly - loading states, disabled states, debouncing the click event, etc.

It also partially depends on whether selected values should be displayed in the trigger button when they are initially selected, or when they are 'applied'.  I've asked for clarification on the desired behavior [in figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL?node-id=2040-8188&m=dev#961166490).

In the end, I think the best feedback that state has updated in the UI is 'the data being filtered updates to reflect the applied filter'.  But that probably isn't enough.  

Maybe we need to give the ability to pass the 'apply' button as a `() => ReactNode` instead of only passing an `onApplyRequest` function?

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: Updated [here](https://github.com/commercetools/ui-kit/pull/2923/commits/b547b207a8b109b2aa0398063699eb6a02572c0f)

One thing that I hadn't considered is 'how do we persist `FiltersList` state in `localstorage` or an API so that we can recreate it between sessions'?  

I think that an `appliedFilters` value might offer a better way to recreate which filters should be displayed in the `FiltersList` on mount than relying on each filter's `getTags` callback to generate the applied values to be displayed for each filter.

misama-ct: > ... 'how do we persist FiltersList state in localstorage or an API so that we can recreate it between sessions'?

We (ui-kit) can't. If we would, we might be able to restore the visual state, but we wouldn't know to what actual state the ui-state would translate to in the consumer app (e.g. you can't transform a label-string back into a date-range object). The consumer app has to persist app-state which always needs to be reduced to ui-state.

misama-ct: > I think that an appliedFilters value might offer a better way to recreate which filters should be displayed in the FiltersList on mount than relying on each filter's getTags callback to generate the applied values to be displayed for each filter.

The sole purpose of this function is to display exactly what needs to be displayed. What speaks against adjusting the function? Why is an `appliedFilter` object better?

ByronDWall: The `filters` config object allows us to bypass prop-drilling the selected values in each filter through `FiltersList`, which is good.  But what that means is that `FiltersList` has no way to evaluate which filters have data selected.  

This is fine for the purposes of a single session, but if we want to show all active filters in the `FiltersList` when the user resumes their session, the `FiltersList` needs to have some way to know which filters should be visible on mount.

Passing a `getTags` function in each `filters` config object means that there is no way for `FiltersList` to determine which filters should be visible based on the UI state passed to it.

Passing an `appliedValues` array, in which each object specifies the `key` of the filter along with the values/labels for its selected values, allows the `FiltersList` to render filters with applied values on initial render. 

In other words, `appliedValues` gives us the data necessary to recreate UI state for `FiltersList` based on data that was persisted outside of the component.

ByronDWall: @misama-ct [see this PR](https://github.com/commercetools/ui-kit/pull/2928) for a better illustration of what I think `appliedValues` achieves

Sarah4VT:
> > One problem we haven't solved yet is how the "Apply" button should work. I'm trying to figure out how we can manage the state outside of the component but still show the updated state visually after a UI change (ideally without knowing the original structure of the state or actually applying the new state yet).
> 
> this is a really good question, and I can see where things could get very complicated very quickly - loading states, disabled states, debouncing the click event, etc.
> 
> It also partially depends on whether selected values should be displayed in the trigger button when they are initially selected, or when they are 'applied'. I've asked for clarification on the desired behavior [in figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL?node-id=2040-8188&m=dev#961166490).
> 
> In the end, I think the best feedback that state has updated in the UI is 'the data being filtered updates to reflect the applied filter'. But that probably isn't enough.
> 
> Maybe we need to give the ability to pass the 'apply' button as a `() => ReactNode` instead of only passing an `onApplyRequest` function?

I'm not sure if I'm completely following this, but it reminded me of a scenario I worked on in the past where I used inversion of control with a state reducer. Not sure if that's what you are going for here but before I switch over to the chapter, I'd thought I'd just mention that to see if it's the pattern you may be looking for.

packages/components/filters/filters-list/src/filters-list.tsx
Sarah4VT: The date filter in Audit Log is an example of a potential persistent filter, just in case you are curious.

Sarah4VT (approved):

packages/components/filters/filters-list/src/filters-list.stories.tsx
```diff
@@ -1,22 +1,156 @@
-import type { Meta, StoryObj } from '@storybook/react';
+import type { Meta, StoryFn } from '@storybook/react';
+import { useState, useCallback } from 'react';
+import { MenuProps, MenuListProps } from 'react-select';
 import FiltersList from './filters-list';
+import SelectInput from '../../../inputs/select-input';
```
ddouglasz, line 5: Maybe I am missing a change made to the select input to improve something locally. Otherwise I am a bit lost why we are importing select-input locally and not from the package ?

ddouglasz (approved):

packages/components/filters/filters-list/src/filters-list.stories.tsx
ByronDWall: vscode autocompleted the import, and it was fine for PoC purposes - we'll use the workspace namespace for code intended to commit to `main`

packages/components/filters/filters-list/package.json
```diff
@@ -21,6 +21,12 @@
   "dependencies": {
     "@babel/runtime": "^7.20.13",
     "@babel/runtime-corejs3": "^7.20.13",
+    "@commercetools-uikit/design-system": "workspace:^",
+    "@commercetools-uikit/dropdown-menu": "workspace:^",
+    "@commercetools-uikit/filter-menu": "workspace:^",
+    "@commercetools-uikit/flat-button": "workspace:^",
+    "@commercetools-uikit/icons": "workspace:^",
+    "@commercetools-uikit/select-input": "workspace:^",
```
misama-ct, line 29: pnpm uses this syntax, yarn should support it, but I did not encounter it yet here in ui-kit. (Some task seems to update the version numbers after a release.) If you're running into issues, use the latest published version number.

ByronDWall:

PR #2924, opened 2024-09-23
chore(canary preview doc): document canary branch preview

ddouglasz:
## Description

After testing the canary version and concluded that it works as expected, we need to follow up by documenting this new workflow and how to trigger a canary version to preview a specific branch.


misama-ct (approved):
🕺

PR #2928, opened 2024-09-25
FCT-1173 - describe visible filters in `FiltersList` via props

ByronDWall:
POC that implements defining `FiltersList` UI state based on props.

It renders filters on mount based on whether a filter either has a value of `isPersisted=true`, or it has a value defined in `appliedFilters`.

Also updates the shape of `TAppliedFilterValues` to better accommodate discrete chips for multi-selected values.

### Refactoring and Data Structure Improvements:
* [`packages/components/filters/filter-menu/src/filter-menu.tsx`](diffhunk://#diff-80cc90aa6484306f7893b25aa53f02081fabf45c0a9811f16c1ba4de67da5fccL8-R8): Changed the `TAppliedFilterValue` type to include a `value` property instead of `key` and updated related logic to use this new structure. [[1]](diffhunk://#diff-80cc90aa6484306f7893b25aa53f02081fabf45c0a9811f16c1ba4de67da5fccL8-R8) [[2]](diffhunk://#diff-5afab4e1a74d160eadcb5906640b7b6069554466a54a15e9a5d9201bd778bcb5L67-R67)
* [`packages/components/filters/filters-list/src/filters-list.tsx`](diffhunk://#diff-54193cb3dbd32d96375497f45e1a448a439d596bd933f0d469b8da87bf5253bcL67-R93): Refactored the filter list component to use `localVisibleFilters` for managing visible filters, ensuring that persisted and applied filters are correctly handled. [[1]](diffhunk://#diff-54193cb3dbd32d96375497f45e1a448a439d596bd933f0d469b8da87bf5253bcL67-R93) [[2]](diffhunk://#diff-54193cb3dbd32d96375497f45e1a448a439d596bd933f0d469b8da87bf5253bcL95-R120)

### Enhanced Functionality:
* [`packages/components/filters/filter-menu/src/filter-menu.stories.tsx`](diffhunk://#diff-eab71442f43af79f2094c9e0012769b95515e2fc7a3fe5bea105904a11ffac3dL89-R96): Added `hideSelectedOptions` prop to the `SelectInput` component and updated the example to demonstrate its usage. [[1]](diffhunk://#diff-eab71442f43af79f2094c9e0012769b95515e2fc7a3fe5bea105904a11ffac3dL89-R96) [[2]](diffhunk://#diff-eab71442f43af79f2094c9e0012769b95515e2fc7a3fe5bea105904a11ffac3dL104-R117)
* [`packages/components/filters/filters-list/src/filters-list.stories.tsx`](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R114): Updated the example to include new props like `isPersistent`, `backspaceRemovesValue`, and `isClearable` for better demonstration of the filter list component's capabilities. [[1]](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R114) [[2]](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R140-R142)

### Prop Updates and Consistency:
* [`packages/components/filters/filter-menu/src/header/header.tsx`](diffhunk://#diff-09795b1f9a4ad0b001f1de2df750bbd9459f1e0ab38b38a55425fbe0bdb00a69L7-R7): Replaced `operatorsInput` prop with `renderOperatorsInput` to allow for more flexible rendering of operator inputs. [[1]](diffhunk://#diff-09795b1f9a4ad0b001f1de2df750bbd9459f1e0ab38b38a55425fbe0bdb00a69L7-R7) [[2]](diffhunk://#diff-09795b1f9a4ad0b001f1de2df750bbd9459f1e0ab38b38a55425fbe0bdb00a69L23-R24)
* [`packages/components/inputs/select-input/src/select-input.tsx`](diffhunk://#diff-b34243b56690128adca032aef36cb75a7cd769f5fb64368ded23533b4114297dL164-R164): Added `hideSelectedOptions` prop to the `SelectInput` component to provide more control over option visibility. [[1]](diffhunk://#diff-b34243b56690128adca032aef36cb75a7cd769f5fb64368ded23533b4114297dL164-R164) [[2]](diffhunk://#diff-b34243b56690128adca032aef36cb75a7cd769f5fb64368ded23533b4114297dL464-R464)

### Story Updates:
* [`packages/components/filters/filters-list/src/filters-list.stories.tsx`](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R62-R86): Cleaned up and updated the example to better reflect the new data structure and added props for enhanced functionality. [[1]](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R62-R86) [[2]](diffhunk://#diff-21d8d876191cdcfde7d2439398031dfd47abc6a7d4a4f836dca12406109da241R124)

These changes collectively improve the flexibility and usability of the filter components, making them more robust and easier to integrate into various use cases.

### DIFF
--- a/packages/components/filters/filter-menu/src/filter-menu.tsx
+++ b/packages/components/filters/filter-menu/src/filter-menu.tsx
@@ -1,5 +1,7 @@
-import { type ReactNode } from 'react';
+import { type ReactNode, type ReactElement } from 'react';
 import DropdownMenu from '@commercetools-uikit/dropdown-menu';
+import { type TSelectInputProps } from '@commercetools-uikit/select-input';
+import { type TPrimaryButtonProps } from '@commercetools-uikit/primary-button';
 import { Footer } from './footer';
 import { Header } from './header';
 import { TriggerButton } from './trigger-button';
@@ -34,7 +36,7 @@ export type TFilterMenuProps = {
    * the alternative to a rect element is passing in an array of operators and a handler function,
    * which is more achievable here than for the input in the body, but could still be restrictive
    */
-  renderOperatorsInput?: () => ReactNode;
+  renderOperatorsInput?: () => ReactElement<TSelectInputProps>;
   /**
    * the values applied to this filter by the user
    *
@@ -57,19 +59,19 @@ export type TFilterMenuProps = {
    * NOTES / OPINIONS WELCOME IN PR COMMENTS:
    * is the presence/absence of the function sufficient for display logic, or do we need a separate `show-` prop?
    */
-  onRemoveFilter?: Function;
+  onRemoveRequest?: Function;
   /**
    * controls whether `apply` button in Menu Body Footer is displayed
    */
-  onApplyFilter?: Function;
+  renderApplyButton?: () => ReactElement<TPrimaryButtonProps>;
   /**
    * controls whether `clear` button in Menu Body Footer is displayed
    */
-  onClearFilter?: Function;
+  onClearRequest?: Function;
   /**
    * controls whether `sort` button in Menu Body Header is displayed
    */
-  onFilterOptionsSortClick?: Function;
+  onSortRequest?: Function;
 };
 
 function FilterMenu(props: TFilterMenuProps) {
@@ -81,7 +83,7 @@ function FilterMenu(props: TFilterMenuProps) {
           appliedFilterValues={props.appliedFilterValues}
           isDisabled={props.isDisabled}
           isPersistent={props.isPersistent}
-          onRemoveFilter={props.onRemoveFilter}
+          onRemoveRequest={props.onRemoveRequest}
         />
       }
       menuHorizontalConstraint={7}
@@ -90,12 +92,12 @@ function FilterMenu(props: TFilterMenuProps) {
       <Header
         label={props.label}
         renderOperatorsInput={props.renderOperatorsInput}
-        onFilterOptionsSortClick={props.onFilterOptionsSortClick}
+        onSortRequest={props.onSortRequest}
       />
       {props.renderMenuBody()}
       <Footer
-        onApplyFilter={props.onApplyFilter}
-        onClearFilter={props.onClearFilter}
+        renderApplyButton={props.renderApplyButton}
+        onClearRequest={props.onClearRequest}
       />
     </DropdownMenu>
   );
--- a/packages/components/filters/filter-menu/src/filter-menu.tsx
+++ b/packages/components/filters/filter-menu/src/filter-menu.tsx
@@ -5,7 +5,7 @@ import { Header } from './header';
 import { TriggerButton } from './trigger-button';
 
 export type TAppliedFilterValue = {
-  key?: string;
+  value: string;
   label: ReactNode;
 };
 
@@ -34,7 +34,7 @@ export type TFilterMenuProps = {
    * the alternative to a rect element is passing in an array of operators and a handler function,
    * which is more achievable here than for the input in the body, but could still be restrictive
    */
-  operatorsInput?: ReactNode;
+  renderOperatorsInput?: () => ReactNode;
   /**
    * the values applied to this filter by the user
    *
@@ -74,7 +74,6 @@ export type TFilterMenuProps = {
 
 function FilterMenu(props: TFilterMenuProps) {
   return (
-    // dropdown menu needs way to programatically clo
     <DropdownMenu
       triggerElement={
         <TriggerButton
@@ -90,7 +89,7 @@ function FilterMenu(props: TFilterMenuProps) {
     >
       <Header
         label={props.label}
-        operatorsInput={props.operatorsInput}
+        renderOperatorsInput={props.renderOperatorsInput}
         onFilterOptionsSortClick={props.onFilterOptionsSortClick}
       />
       {props.renderMenuBody()}
--- a/packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
+++ b/packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
@@ -64,7 +64,7 @@ const TriggerButton = (props: TFilterMenuTriggerButtonProps) => {
       {props.label}
       {props.appliedFilterValues &&
         props.appliedFilterValues.map((value) => (
-          <Chip key={value.key} label={value.label} />
+          <Chip key={value.value} label={value.label} />
         ))}
       {props.appliedFilterValues && !!props.appliedFilterValues.length && (
         <Badge label={`+${props.appliedFilterValues.length}`} />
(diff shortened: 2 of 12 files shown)


packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -7,17 +7,20 @@ import FlatButton from '@commercetools-uikit/flat-button';
...
+export type TAppliedFilterValue = { label: ReactNode; value: string };
+
+export type TAppliedFilter = {
   filterKey: string;
-  label: ReactNode[];
+  values: TAppliedFilterValue[];
 };
 
-export type TFiltersConfiguration = {
+export type TFilterConfiguration = {
```
ByronDWall, line 17: I think this configuration API for each filter is working well enough to implement it.

packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -64,11 +61,36 @@ const CustomMenuList = ({
...
-  const persistedFilters = props.filters
-    .filter((filter) => filter.isPersistent)
+  /**
+   * persisted filters: always visible
+   * applied filters: filters for which values have been selected
+   * visible filters = persisted + applied
+   *
+   * visibleFiltersFromProps = filters visible based on props
+   * localVisibleFilters = filters actually visible in component currently
+   */
```
ByronDWall, line 71: @misama-ct I personally think that this approach to allowing `FiltersList` to recreate its' UI state based on props works well.

I would also be more than happy to consider an alternate approach if you have any in mind.

packages/components/inputs/select-input/src/select-input.tsx
```diff
@@ -161,7 +161,7 @@ export type TSelectInputProps = {
   // formatOptionLabel: PropTypes.func,
   // getOptionLabel: PropTypes.func,
   // getOptionValue: PropTypes.func,
-  // hideSelectedOptions: PropTypes.bool,
+  hideSelectedOptions?: ReactSelectProps['hideSelectedOptions'];
```
ByronDWall, line 164: Inputs are a separate set of work, but there are some basic props, like `hideSelectedOptions`, that we should make available for every `react-select`-based input

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -34,7 +34,7 @@ export type TFilterMenuProps = {
    * the alternative to a rect element is passing in an array of operators and a handler function,
    * which is more achievable here than for the input in the body, but could still be restrictive
    */
-  operatorsInput?: ReactNode;
+  renderOperatorsInput?: () => ReactNode;
```
ByronDWall, line 37: I think that passing the operators input as a `ReactNode` makes sense here, in large part to minimize the amount of props necessary to achieve the necessary UI.  It forces the consuming application to handle state for the operators completely outside the `Filters` component, which I think is a good thing.

ByronDWall: Also made it a function that returns a `ReactNode`, which I think we should consider making standard practice when passing props that render child components.

packages/components/filters/filters-list/src/filters-list.stories.tsx
```diff
@@ -59,19 +59,31 @@ export const BasicExample: Story = () => {
     clearSecondaryColorFilter();
   };
 
+  // const appliedFilters = [
+  //   {
+  //     filterKey: 'primaryColors',
+  //     values: [{ value: 'blue', label: 'BLUE' }],
+  //   },
+  // ];
```
ByronDWall, line 67: if you set this value, you can see that the `primaryColors` `FilterMenu` is displayed on initial mount.

packages/components/filters/filter-menu/src/filter-menu.tsx
```diff
@@ -5,7 +5,7 @@ import { Header } from './header';
 import { TriggerButton } from './trigger-button';
 
 export type TAppliedFilterValue = {
-  key?: string;
+  value: string;
```
ByronDWall, line 8: the idea being that the `value` is a string that can be used as a key, and the `label` is a `ReactNode` that renders the applied value with the desired format inside the `Chip` in the `TriggerButton`

packages/components/filters/filters-list/src/filters-list.tsx
```diff
@@ -26,21 +29,15 @@ export type TFiltersConfiguration = {
...
-  isDisabled: boolean;
-};
-
 export type TFiltersListProps = {
-  filters: TFiltersConfiguration[];
-  filterGroups?: TFilterGroup[];
-  appliedFilters: TAppliedFilterValue[];
+  filters: TFilterConfiguration[];
+  filterGroups?: TFilterGroupConfiguration[];
+  appliedFilters: TAppliedFilter[];
```
Sarah4VT, line 40: Nit: the prop is called "appliedFilters" but the type is "AppliedFilter". Can we either add an s or not to both? Doesn't have to be updated here, but just going forward.

packages/components/filters/filter-menu/src/trigger-button/trigger-button.tsx
```diff
@@ -64,7 +64,7 @@ const TriggerButton = (props: TFilterMenuTriggerButtonProps) => {
       {props.label}
       {props.appliedFilterValues &&
         props.appliedFilterValues.map((value) => (
-          <Chip key={value.key} label={value.label} />
+          <Chip key={value.value} label={value.label} />
```
Sarah4VT, line 67: I am not crazy about `value.value`. We have this in audit log and it drives me a little crazy.

Sarah4VT: Although I realize here it's just because we call the mapped items "value". But still, if this is being passed to the prop "key", I kinda liked it as "key" better I think.

Sarah4VT (approved):

packages/components/filters/filters-list/src/filters-list.tsx
misama-ct: No, let's go with this approach. I'm confident this implementation will work and once it does, we'll just have a look at the code to see if we see potential and/or need for improvement.

misama-ct (approved):

ddouglasz (approved):

packages/components/filters/filters-list/src/filters-list.tsx
ByronDWall: I figured that `appliedFilters` is the plural of an array of singular `appliedFilter` objects.  But I have a pretty low care amount here so I can change it if necessary

ByronDWall:

PR #2930, opened 2024-09-27
Define available filters via explicitly declared children - POC

ddouglasz:
#### Summary

Define available filters via explicitly declared children -POC

### DIFF
--- a/packages/components/filters/filters-list/src/filters-list.stories.tsx
+++ b/packages/components/filters/filters-list/src/filters-list.stories.tsx
@@ -1,22 +1,162 @@
-import type { Meta, StoryObj } from '@storybook/react';
+import type { Meta, StoryFn } from '@storybook/react';
+import { MenuProps, MenuListProps } from 'react-select';
+import { useState, useCallback } from 'react';
 import FiltersList from './filters-list';
+import SelectInput from '../../../inputs/select-input';
 
 const meta: Meta<typeof FiltersList> = {
   title: 'components/Filters/FiltersList',
   component: FiltersList,
   tags: ['local-dev'],
-  argTypes: {
-    label: {
-      control: 'text',
-    },
-  },
+  argTypes: {},
 };
 export default meta;
 
-type Story = StoryObj<typeof FiltersList>;
+type Story = StoryFn<typeof FiltersList>;
 
-export const BasicExample: Story = {
-  args: {
-    label: 'A list of filters',
-  },
+const CustomSelectMenu = ({
+  children,
+  innerProps: { ref, ...restInnerProps },
+}: MenuProps) => (
+  <div ref={ref} {...restInnerProps}>
+    {children}
+  </div>
+);
+
+const CustomMenuList = ({
+  children,
+  innerProps: { ref, ...restInnerProps },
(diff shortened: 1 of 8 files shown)


packages/components/filters/filters-list/src/filters-list.stories.tsx
```diff
@@ -1,22 +1,162 @@
...
+        }))
+      : [];
+  }, [secondaryColorValue]);
+
+  const clearAllFilters = () => {
+    clearPrimaryColorFilter();
+    clearSecondaryColorFilter();
+  };
+
+  const filters = [
```
ddouglasz, line 80: Where I may need some clarity is if we are only just passing the configs as children or the whole filters array too?

ddouglasz: I can tell you I spent more time figuring out what should go in as children, thereby creating the subset of data. Syntax-wise, I agree that passing the configuration as props seems more straightforward to follow. It just seems  easier to shape and centralise the data consumed from the stories(users) that way.
I will also document this in the ticket. 
Thank you

ByronDWall: That's also my takeaway from this, fwiw, thanks @ddouglasz

packages/components/filters/filters-list/src/filters-list.stories.tsx
```diff
@@ -1,22 +1,162 @@
...
+        justifyContent: 'center',
+      }}
+    >
+      <FiltersList
+        filters={filters}
+        onClearAllRequest={clearAllFilters}
+        filterGroups={filterGroups}
+      >
+        {filterConfigs.map(
+          ({ filterKey, name, options, value, setValue, isMulti = false }) => (
```
misama-ct, line 137: I was actually hoping for some sort of component ensemble which would reduce the need for custom props. 
e.g. a  `<FilterList.ClearFiltersButton/>` component would allow us to ditch the `onClearRequest` on the `FiltersList`, cause the consuming app could just use the regular `onClick` handler of the `<FilterList.ClearFiltersButton/>`.

ByronDWall: `onClearAllRequest` is a required prop because the design spec requires a "clear all" button to be displayed in the `FiltersList`.  Making a `FilterList.ClearFiltersButton` component available and expecting it to be passed as a child of the `FiltersList` opens up the possibility of the consuming application not implementing the "clear all" button.

I think it makes a lot of sense to allow the consuming app to pass components a children/props when what they are defining is optional, but when something is required I think there is a lot to be said for making it as straightforward as possible to implement.

ddouglasz: > I think it makes a lot of sense to allow the consuming app to pass components a children/props when what they are defining is optional, but when something is required I think there is a lot to be said for making it as straightforward as possible to implement.

This is part of what I was saying to you earlier today @misama-ct about the risk of passing required props as children and having the consumer decide on it as though it was optional.

ByronDWall:

ByronDWall:

PR #2932, opened 2024-09-30
FCT-1197 - update file-system for `Filters` and `QuickFilters`

ByronDWall:
#### Summary

Now that we are moving past the PoC phase and into the implementation phase of the `Filters` pattern, it is necessary to update the file structure to facilitate further development.

## Description

There will be 2 new components available in UI Kit as a result of implementing the "filters pattern": a `Filters` component, and a `QuickFilters` component.

The `Filters` component is referred to as `FiltersList` in the PoCs and in JIRA, but since it is the parent component for most of the "filters pattern", it makes sense to call it `Filters`.

The `QuickFilters` are sufficiently functionally and visually distinct from the `Filters` to warrant their own component.


packages/components/filters/src/badge/badge.tsx
ByronDWall, line 1: The `Badge` will be used in both the `FilterMenu` to [display the count of hidden selected values](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2615-13850&m=dev) in the `TriggerButton`, and in the `Filters` [to display the count of active filters](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2001-6880&m=dev) when the filters-list is toggled to be inactive

ddouglasz (approved):
💯

PR #2934, opened 2024-10-01
FCT 1189 - add `hideSelectedOptions` prop to `SelectInput` and `AsyncSelectInput`

ByronDWall:
#### Summary

Adds `hideSelectedOptions` prop to `SelectInput` and `AsyncSelectInput`

## Description

In order to display selected filter options within select inputs as specified, it is necessary to implement the `hideSelectedOptions` [prop from react-select](https://react-select.com/props) in the `SelectInput` and `AsyncSelectInput`.

Options displayed in the `SearchSelectInput` are defined via passing a `loadOptions` function, and are therefore already controlled by the consuming application.  Therefore there is no need to update the `SearchSelectInput` props.

This PR adds `hideSelectedOptions`, updates the unit tests to insure the expected behavior, adds the controls for the prop to storybook, and updates the `README` for both `SelectInput` and `AsyncSelectInput`.


tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "exclude": ["**/node_modules/**", "**/dist/**"],
```
ByronDWall, line 2: The typescript server in vscode kept dying on me in the UI Kit repo - this is an attempt to make TS more reliable.

tdeekens (approved):

ddouglasz (approved):
Thanks!

PR #2935, opened 2024-10-02
fix: codeowners

emmenko:
Not sure if you want to use the `first-contact-team-fe` team or create a different one with a subset of people.


ByronDWall (approved):
I'm fine with `@first-contact-team-fe` for now, we can change it in the future if necessary, thanks for making the change

ddouglasz (approved):

PR #2936, opened 2024-10-02
FCT-1072 - implement filters `Badge` component

ByronDWall:
This PR implements the `Badge` component in the `filters` package.  

It implements the styles and states [specified in figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-6256&m=dev), along with the proper accessibility attributes for the `status` role, [as specified here](https://w3c.github.io/aria/#status).

### Badge Component Implementation:
* [`packages/components/filters/src/badge/badge.tsx`](diffhunk://#diff-3a3cc1bc670245d29b7b51e03ec6378bdee6231c14135a350bea94098e7edaddL1-R53): Fully implemented the `Badge` component with properties for `id`, `isDisabled`, `label`, and `aria-label`. Added styles using `@emotion/react` and `designTokens` from `@commercetools-uikit/design-system`.

### Test Updates:
* [`packages/components/filters/src/badge/badge.spec.tsx`](diffhunk://#diff-d7ea2b19bec24e727a07922300c734f05a6f72d8e74552aa3c5f7832fb5840d0L4-R11): Updated the test for the `Badge` component to use `defaultProps` and check for the `status` role.

### Documentation and Dependency Updates:
* [`packages/components/filters/README.md`](diffhunk://#diff-018b1db07883ff17d596ba7a4ce8a57c65b0a44894b7c1476408706435d8e6daL46-R47): Fixed the table formatting in the properties section.
* [`packages/components/filters/package.json`](diffhunk://#diff-a9b33f9c36305865b265a930a2124d31f24065710b96c192e7ea15a789867dfeR24-R26): Added dependencies for `@commercetools-uikit/design-system` and `@commercetools-uikit/utils`.…ibutes, implement unit test, show states in storybook story

**PLEASE NOTE: The `Filters` storybook must be updated to have a `local-dev` tag before merging this PR to `main` so that the `Filters` docs do not get published to the prod site, these docs are published for this PR as a courtesy to the reviewers.**
[
The badge can be seen here](https://ui-kit-git-fct-1072-badge-commercetools.vercel.app/?path=/docs/components-filters--props)

### DIFF
--- a/packages/components/filters/src/badge/badge.tsx
+++ b/packages/components/filters/src/badge/badge.tsx
@@ -44,7 +44,7 @@ const disabledBageStyles = css`
 function Badge(props: TBadgeProps) {
   return (
     <span
-      aria-label={props['aria-label'] ? props['aria-label'] : props.label}
+      aria-label={props['aria-label']}
       css={[badgeStyles, props.isDisabled && disabledBageStyles]}
       id={props.id}
       role="status"
--- a/packages/components/filters/src/badge/badge.tsx
+++ b/packages/components/filters/src/badge/badge.tsx
@@ -34,6 +34,7 @@ const badgeStyles = css`
   flex-direction: column;
   justify-content: center;
   align-items: center;
+  height: ${designTokens.spacing40};
 `;
 
 const disabledBageStyles = css`
--- a/packages/components/filters/src/badge/badge.tsx
+++ b/packages/components/filters/src/badge/badge.tsx
@@ -1,5 +1,56 @@
-function Badge() {
-  return <div>badge</div>;
+import { css } from '@emotion/react';
+import { designTokens } from '@commercetools-uikit/design-system';
+
+export type TBadgeProps = {
+  /**
+   * CSS ID for badge, used to specify relationship with parent in parent's `aria-controls` property
+   * see https://www.w3.org/TR/wai-aria-1.1/#status
+   */
+  id: string;
+  /**
+   * If `true`, indicates that the element is in a disabled state.
+   */
+  isDisabled?: boolean;
+  /**
+   * String to be displayed in badge, generally a count of some kind
+   */
+  label: string;
+
+  /**
+   * Optional descriptive explanation of label (e.g. "+4 additional filters applied")
+   */
+  ['aria-label']?: string;
+};
+
+const badgeStyles = css`
+  font-size: ${designTokens.fontSize20};
+  font-weight: ${designTokens.fontWeight500};
+  line-height: ${designTokens.lineHeight20};
+  color: ${designTokens.colorSurface};
+  background-color: ${designTokens.colorInfo};
+  padding: 0 calc(${designTokens.spacing05} + ${designTokens.spacing10});
+  border-radius: ${designTokens.borderRadius20};
+  display: inline-flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+`;
+
+const disabledBageStyles = css`
+  background-color: ${designTokens.colorNeutral};
+`;
+
+function Badge(props: TBadgeProps) {
+  return (
+    <span
+      aria-label={props['aria-label'] ? props['aria-label'] : props.label}
+      css={[badgeStyles, props.isDisabled && disabledBageStyles]}
+      id={props.id}
+      role="status"
+    >
+      {props.label}
+    </span>
+  );
 }
 
 export default Badge;
--- a/packages/components/filters/src/filters.tsx
+++ b/packages/components/filters/src/filters.tsx
@@ -1,4 +1,4 @@
-import { Badge } from './badge';
+import { designTokens } from '@commercetools-uikit/design-system';
 
 export type TFiltersProps = {
   /**
@@ -9,12 +9,8 @@ export type TFiltersProps = {
 
 function Filters(props: TFiltersProps) {
   return (
-    <div style={{ display: 'flex', gap: '8px' }}>
+    <div style={{ display: 'flex', gap: designTokens.spacing20 }}>
       <div>{props.label}</div>
-      active:
-      <Badge label="+2" id={`${props.label}-total-applied-filters`} />
-      disabled:
-      <Badge label="+2" id={`${props.label}-applied-filter-count`} isDisabled />
     </div>
   );
 }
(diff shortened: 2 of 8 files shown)


packages/components/filters/src/filters.tsx
```diff
@@ -6,6 +8,14 @@ export type TFiltersProps = {
 };
 
 function Filters(props: TFiltersProps) {
-  return <div>{props.label}</div>;
+  return (
+    <div style={{ display: 'flex', gap: '8px' }}>
```
tdeekens, line 12: It's 8px a spacing constant?

FilPob:
@ByronDWall looks good from my side. Just the same thing about the overall height... should be 24px instead of 22

ByronDWall:
@FilPob height updated [here](https://github.com/commercetools/ui-kit/pull/2936/commits/48a8971d73b72dddfdd894b90e7808e3d6a5d431), thanks!

FilPob (approved):

packages/components/filters/src/badge/badge.tsx
```diff
@@ -1,5 +1,57 @@
...
+   */
+  label: string;
+
+  /**
+   * Optional descriptive explanation of label (e.g. "+4 additional filters applied")
+   */
+  ['aria-label']?: string;
+};
+
+const badgeStyles = css`
```
misama-ct, line 25: May I propose to use  emotion's `styled` function ([click](https://emotion.sh/docs/styled)) instead of `css` and also use [object styles](https://emotion.sh/docs/object-styles) instead of string literals.

With `styled` you get access to the component props and don't need any additional logic for conditional styles outside of your style-definitions. All styles would be in one `StyledBadge` component.

Additional benefits of the object syntax is, that it is a regular javascript object already understood by your editor and other tools (code quality, refactoring etc).

packages/components/filters/src/badge/badge.tsx
```diff
@@ -1,5 +1,57 @@
...
+`;
+
+const disabledBageStyles = css`
+  background-color: ${designTokens.colorNeutral};
+`;
+
+function Badge(props: TBadgeProps) {
+  return (
+    <span
+      aria-label={props['aria-label'] ? props['aria-label'] : props.label}
```
misama-ct, line 47: If there is only a `label` and no `aria-label`, there is no benefit in adding an `aria-label`. I think this prop should only be rendered in the DOM if the consuming component specifies `aria-label` on `Badge`.

packages/components/filters/src/badge/badge.tsx
misama-ct:
```
const StyledBadge = styled.span({
    fontSize: designTokens.fontSize20,
    fontWeight: designTokens.fontWeight500,
    ... etc...
    // dynamic property based on props
    backgroundColor: (props) =>
        props.isDisabled ? designTokens.colorNeutral : designTokens.colorInfo,
})
```

packages/components/filters/src/badge/badge.tsx
```diff
@@ -1,5 +1,57 @@
...
+  height: ${designTokens.spacing40};
+`;
+
+const disabledBageStyles = css`
+  background-color: ${designTokens.colorNeutral};
+`;
+
+function Badge(props: TBadgeProps) {
+  return (
+    <span
```
misama-ct, line 46: Here you would use the `StyledBadge` instead of `<span>`. 

Nice to have: Typescript is aware that `StyledBadge` uses a span tag to render to DOM and only suggests attributes that you can set on a span-tag. Granted, not super useful in this case, but very much more so on `button`'s and `input`'s.

packages/components/filters/src/badge/badge.tsx
ByronDWall: Wow.  I was under the impression that UI Kit heavily favored `css`, but it looks like it uses `styled` just as much.  [Which isnt great](https://emotion.sh/docs/best-practices#use-the-css-prop-or-emotionstyled-but-not-both), but is where we're at.

The one thing that is happening with both `css` and `styled` is that we are _overwhelmingly_ always using the template syntax, and we have linting set up for it.  I personally like the 1-1 template<->css attribute values, but I understand the arguments for object syntax as well.

Also, as far as dynamic styles, the emotion docs [suggest](https://emotion.sh/docs/best-practices#use-the-css-prop-or-emotionstyled-but-not-both) that it is best to use css vars declared in a `style` prop to handle them.

I specifically went with `css` and declaring `css` functions in the array syntax so that we have explicit classes defined for styling certain states, instead of having mostly-similar classes repeated in the DOM with only 1-2 different attrs between each.

All of this is to say that 'wtf do we do about styling' is one of the big maintainability headaches that we need to come to some decisions on, but nothing has been decided yet.

I definitely think that we should at least keep styling consistent _within_ a component, and think I've made a pretty decent case for why we should use `css`, but if we really need to re-evaluate that we can.

packages/components/filters/src/badge/badge.tsx
ByronDWall: yeah, I thought I might have been over-engineering this one, good point.

packages/components/filters/src/badge/badge.tsx
ByronDWall: oh, this is an argument for `styled` I hadn't considered.  personally I think it's weird to couple styling so heavily to markup, but css-in-js pretty much always results in coupling them anyway.

My suggestion here is lets use `css` for this component, and `styled` for the next new component, and then reassess?

misama-ct: As mentioned it's just a proposal. I am aware of the current state in ui-kit. It's just the first time we actually add something instead of just fixing something, so I at least wanted to mention it. As for the markup: `styled` also adds the `as` property to the component, so you can change the underlying semantics on a case by case basis.

misama-ct (approved):

packages/components/filters/src/badge/badge.tsx
ByronDWall: I really more meant that with `css` we could at least theoretically use the `disabled` `css` function in more than one place, kind of like, yknow, css.  But in reality we don't do that enough to optimize for that behavior (although I'm going to argue that maybe we should when it comes to actually standardizing styling practices)

PR #2937, opened 2024-10-02
FCT 1079 - filters `Chip`

ByronDWall:
This PR implements the `Chip` component with necessary styles and state as [specified in figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-6256&m=dev).  

The `Chip` is implemented as an `li`, and should be used as a child of a `ul` component, as it is meant to list the selected filter values in a `FilterMenu`.

### Component Enhancements:

* **Chip Component**: Refactored to accept `label` and `isDisabled` props, and styled using `@commercetools-uikit/design-system` and `@emotion/react`.
* **FilterMenu Chip Test**: Updated the test for the `Chip` component to check for the `label` prop and role attribute.

**PLEASE NOTE: The Filters storybook must be updated to have a local-dev tag before merging this PR to main so that the Filters docs do not get published to the prod site, these docs are published for this PR as a courtesy to the reviewers.**
[
Chip can be seen here](https://ui-kit-git-fct-1078-chip-commercetools.vercel.app/?path=/docs/components-filters--props)

### DIFF
--- a/packages/components/filters/src/filter-menu/chip/chip.tsx
+++ b/packages/components/filters/src/filter-menu/chip/chip.tsx
@@ -1,5 +1,39 @@
-function Chip() {
-  return <div>chip</div>;
+import { type ReactNode } from 'react';
+import { designTokens } from '@commercetools-uikit/design-system';
+import { css } from '@emotion/react';
+
+export type TFilterMenuChipProps = {
+  isDisabled?: boolean;
+  label: ReactNode;
+};
+
+const chipStyles = css`
+  font-size: ${designTokens.fontSize20};
+  font-weight: ${designTokens.fontWeight400};
+  line-height: ${designTokens.lineHeight20};
+  color: ${designTokens.colorPrimary20};
+  background-color: ${designTokens.colorPrimary95};
+  padding: 0 ${designTokens.spacing20};
+  border-radius: calc(
+    ${designTokens.borderRadius20} - ${designTokens.borderRadius4}
+  );
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  list-style: none;
+`;
+
+const disabledChipStyles = css`
+  color: ${designTokens.colorNeutral60};
+  background-color: ${designTokens.colorNeutral90};
+`;
+
+function Chip(props: TFilterMenuChipProps) {
+  return (
+    <li css={[chipStyles, props.isDisabled && disabledChipStyles]}>
+      {props.label}
+    </li>
+  );
 }
 
 export default Chip;
--- a/packages/components/filters/src/filter-menu/chip/chip.tsx
+++ b/packages/components/filters/src/filter-menu/chip/chip.tsx
@@ -25,7 +25,7 @@ const chipStyles = css`
 `;
 
 const disabledChipStyles = css`
-  color: ${designTokens.colorNeutral50};
+  color: ${designTokens.colorNeutral40};
   background-color: ${designTokens.colorNeutral90};
 `;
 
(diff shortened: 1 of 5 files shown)


jaikamat (approved):
Looks like it does the thing!

FilPob:
@ByronDWall thanks! I just realized we introduced `color-neutral-50` recently. Could we please use it for the font color when disabled? That makes it a bit more readable also in the disabled state.

FilPob:
and one more tiny thing. In the designs the overall height is 24px, in the PR it's 22

packages/components/filters/src/filters.tsx
```diff
@@ -6,6 +8,14 @@ export type TFiltersProps = {
 };
 
 function Filters(props: TFiltersProps) {
-  return <div>{props.label}</div>;
+  return (
+    <>
+      <div>{props.label}</div>
+      <ul style={{ display: 'flex', gap: '8px' }}>
```
ddouglasz, line 14: do you think we should use spacing20 here? 

```suggestion
      <ul style={{ display: 'flex', gap: designTokens.spacing20 }}>
```

packages/components/filters/src/filter-menu/chip/chip.tsx
```diff
@@ -1,5 +1,39 @@
...
+};
+
+const chipStyles = css`
+  font-size: ${designTokens.fontSize20};
+  font-weight: ${designTokens.fontWeight400};
+  line-height: ${designTokens.lineHeight20};
+  color: ${designTokens.colorPrimary20};
+  background-color: ${designTokens.colorPrimary95};
+  padding: 0 ${designTokens.spacing20};
+  border-radius: calc(
```
ddouglasz, line 17: smooth!

ddouglasz (approved):
Left a really tiny nit. Otherwise 👌🏽

ByronDWall:
@FilPob how do you feel about using `colorNeutral40` instead of `colorNeutral50`?

`colorNeutral40` meets AA accessibility contrast ratio guidelines:
Screenshot 2024-10-04 at 9 15 44 AM

while `colorNeutral50` does not:
Screenshot 2024-10-04 at 9 16 14 AM

FilPob:
@ByronDWall yeah lets do that! thanks

FilPob (approved):

tdeekens (approved):

packages/components/filters/src/filter-menu/chip/chip.tsx
```diff
@@ -1,5 +1,40 @@
...
+`;
+
+const disabledChipStyles = css`
+  color: ${designTokens.colorNeutral40};
+  background-color: ${designTokens.colorNeutral90};
+`;
+
+function Chip(props: TFilterMenuChipProps) {
+  return (
+    <li css={[chipStyles, props.isDisabled && disabledChipStyles]}>
```
misama-ct, line 34: Personally, to keep this component portable, I'd go with a `span` tag  and wrap the `Badge` with an `li`-tag  when it's put to use. This way you can also make tweaks on the list separate from the `Badge` implementation (e.g. :first-child & :last-child adjustments)

misama-ct (approved):

packages/components/filters/src/filter-menu/chip/chip.tsx
ByronDWall: this component is specifically not meant to be portable, and only for use in this specific place, which is why I went with the `li`. Partly to reinforce that it is meant to be used specifically in a list.

If we need to make it its' own component at some point, we can refactor it then

PR #2938, opened 2024-10-04
Data table manager custom configurations story

ddouglasz:
#### Summary

Add data table manager custom configurations story. This pull request includes a new story for custom data table manager configurations in the `@commercetools-uikit/data-table-manager` package.


packages/components/data-table-manager/src/data-table-manager.stories.tsx
```diff
@@ -255,25 +667,65 @@ export const BasicExample: Story = (args) => {
     ...columnManagerButtons,
   };
 
+  // const NestedComponent = (props: { id: string }) => {
```
ddouglasz, line 670: These were deliberately commented out for the second phase of the work that involves the nested rows

packages/components/data-table-manager/src/types.tsx
```diff
@@ -231,14 +231,15 @@ export type TColumnProps<Row extends TRow = TRow> = {
...
+  withRowSelection?: boolean;
   /**
    * Each object requires a unique `key` which should correspond to property key of
    * the items of `rows` that you want to render under this column, and a `label`
    * which defines the name shown on the header.
    * The list of columns to be rendered.
    * Each column can be customized (see properties below).
    */
-  columns: TColumnProps<Row>[];
+  columns?: TColumnProps<Row>[];
```
ddouglasz, line 242: I made this slight change because now we can also consume the data table without passing the props through the data table component but rather, through it's provider, migrating storybook to typescript actually helped me catch that we need to make the columns prop optional.

packages/components/data-table-manager/src/types.tsx
```diff
@@ -231,14 +231,15 @@ export type TColumnProps<Row extends TRow = TRow> = {
 };
 
 export type TDataTableManagerProps<Row extends TRow = TRow> = {
+  withRowSelection?: boolean;
```
misama-ct, line 234: image

could you add a short description for our future selves :)

packages/components/data-table-manager/src/data-table-manager.stories.tsx
```diff
@@ -177,8 +215,66 @@ export const BasicExample: Story = (args) => {
...
+  const updateTextColor = (
+    newState: string | ((prevState: string) => string)
+  ) => {
+    setTextColor(newState);
+  };
+  const updateColumnResize = (
+    newState: boolean | ((prevState: boolean) => boolean)
+  ) => {
+    setDisableResize(newState);
+  };
```
misama-ct, line 277: Anything speaking against using the `setState` functions directly instead of proxying them? 

Oh, and you can define the expected type of accepted values right at the hook (if a payload is in the wrong shape your editor will let you know).

```
const [rowClick, setRowClick] = useState<boolean>(true);
```

ddouglasz: > Oh, and you can define the expected type of accepted values right at the hook (if a payload is in the wrong shape your editor will let you know).

Thank you, updated here: https://github.com/commercetools/ui-kit/pull/2938/commits/d4ad05447ab56bad68d7fac6104c085a9943877e

> Anything speaking against using the setState functions directly instead of proxying them?

This is like so because we are actually managing the state from a parent component(in this context, where it is global enough for other components to reference the same state for the data table settings) and then the state it triggered from a child component or any custom component that is created by the user. For example: 

https://github.com/commercetools/ui-kit/blob/ccf69ac02cc7aa74e484a07d5413a942d74e8975/packages/components/data-table-manager/src/data-table-manager.stories.tsx#L1022

packages/components/data-table-manager/src/types.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2938/commits/d4ad05447ab56bad68d7fac6104c085a9943877e

Thank you.

packages/components/data-table-manager/README.md
```diff
@@ -110,6 +110,31 @@ const Example = () => (
 export default Example;
 ```
 
+### With custom configuration settings
+
+The `DataTableManager` allows provides a feature for users to add custom configurations to the data table settings.
```
misama-ct, line 115:
```suggestion
The `DataTableManager` provides a feature for users to add custom configurations to the data table settings.
```

ddouglasz: Good catch! updated here: https://github.com/commercetools/ui-kit/pull/2938/commits/22d3a2150007f17b8dde4306beb671c91875e1f8
Thanks!

misama-ct (approved):
🙇‍♂️ Thank you!

ByronDWall (approved):
I'm approving this, because it adds documentation correctly.

Adding the ability to implement custom config settings makes a lot of sense in and of itself.

I'm less thrilled with the whole 'nested tables' thing, but that's another topic for another PR.

PR #2944, opened 2024-10-07
FCT-1196 - DateInput & DateTimeInput bug

misama-ct:
#### Summary
This PR fixes an issue where manually entered, valid dates were not being emitted when the `onBlur` event is triggered by the calendar-ui closing. While I do not see it as a bug (as confirmation of the entered value is possible by hitting the `return` key), I also don't see any reason that speaks against having this functionality.

> [!NOTE]
>  In the original ticket, `DateRangeInput` component was mentioned as well. After investigating the component I decided to create a separate ticket for it, as the mechanism for parsing manually entered dates seems to be quasi broken and needs to be fixed first.


packages/components/inputs/date-input/src/date-input.spec.js
```diff
@@ -221,3 +226,32 @@ describe('date picker keyboard navigation', () => {
...
+  // no change is expected
+  await waitFor(() =>
+    expect(htmlInputElement).toHaveDisplayValue('03/28/2024')
+  );
+
+  // enter an invalid date
+  await fireEvent.change(htmlInputElement, { target: { value: '33/28/2024' } });
+  await fireEvent.blur(htmlInputElement);
+
+  // should reset to the most recent valid date
```
ByronDWall, line 253: nit: This is the only part I'm unclear about - is it better to return the most recent valid date, or for validation to fail and the input to go into an error state?

If this is the behavior for typed values elsewhere, then it's outside of the scope of this PR and this comment should be closed.

misama-ct: Right now, there’s no error handling in place. Before, users couldn’t select an invalid date through the calendar UI. If they entered an invalid date and pressed 'return,' the calendar wouldn't close, and no value would be passed along.

Now, with the possibility that an invalid date could be entered, I wanted to make sure that this incorrect value doesn’t get passed through, as some consumers might assume this input always provides a valid date. The only other option I see would be to clear the input entirely, but that wouldn’t allow users to easily correct small mistakes, like typos.

ByronDWall: that works for me

ByronDWall (approved):

ddouglasz (approved):
Thank you! 🙌🏽

tylermorrisford (approved):

PR #2947, opened 2024-10-07
FCT 1084- Filter Menu - Footer

valoriecarli:
This PR introduces a new `Footer` component for the `FilterMenu`.

[Ticket Details](https://commercetools.atlassian.net/browse/FCT-1084)
[Figma Specs](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?m=auto&node-id=2041-11654)

Requirements Summary:
- The Footer can contain an `Apply` and/or a `Clear all` button. 
  - Each button should be rendered conditionally based on whether an onClick handler is passed for that button.  
  - The `Apply` button will be passed as a react component by the consumer & justified on left side of the footer.
  - The `Clear all` button is a UI Kit `FlatButton`, and should be justified to the right side of the footer.  
  - There should be at least a 24px gap between the 2 buttons if both are displayed.

- There should be a border/divider 8px above the button
- Tests added

Note that as this will only be used in the FilterMenu, it does not need its' own storybook or percy/visual-regression test.
Screenshot 2024-10-07 at 13 23 50
Screenshot 2024-10-07 at 13 26 43
Screenshot 2024-10-07 at 13 27 26

### DIFF
--- a/packages/components/filters/src/filter-menu/footer/footer.tsx
+++ b/packages/components/filters/src/filter-menu/footer/footer.tsx
@@ -1,5 +1,73 @@
-function Footer() {
-  return <div>footer</div>;
-}
+import FlatButton from '@commercetools-uikit/flat-button';
+import messages from './messages';
+import { CloseBoldIcon } from '@commercetools-uikit/icons';
+import { ReactNode } from 'react';
+import { css } from '@emotion/react';
+import { designTokens } from '@commercetools-uikit/design-system';
+import { useIntl } from 'react-intl';
 
+type TFooterProps = {
+  /**
+   * controls whether `apply` button in Menu Body Footer is displayed
+   */
+  renderApplyButton?: () => ReactNode;
+  /**
+   * controls whether a `clear all` button in Menu Body Footer is displayed
+   */
+  onClearAllRequest?: () => void;
+};
+
+const mainContainerStyles = css`
+  height: auto;
+  width: 252px;
+   };
+`;
+
+const footerContainerStyles = css`
+  /* horizontal divider */
+  height: 1px;
+  background-color: ${designTokens.colorWarning};
+  margin-top: ${designTokens.spacingS};
+
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  align-items: center;
+
+  & > div:first-of-type {
+    margin-top: ${designTokens.spacingS};
+    padding-left: ${designTokens.spacingM};
+  }
+
+  & > div:last-of-type {
+    margin-top: ${designTokens.spacingS};
+    padding-right: ${designTokens.spacingM};
+    display: flex;
+    justify-content: flex-end;
+  }
+`;
+
+const Footer = ({ renderApplyButton, onClearAllRequest }: TFooterProps) => {
+  const intl = useIntl();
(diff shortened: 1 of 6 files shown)


packages/i18n/data/core.json
```diff
@@ -19,6 +19,10 @@
     "developer_comment": "Tooltip title when user hovers the today button",
     "string": "Today"
   },
+  "UIKit.ClearAllButton.clearAllButtonLabel": {
+    "developer_comment": "Label for the clear all button",
+    "string": "Clear all"
+  },
```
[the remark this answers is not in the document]
FilPob: yes there is a official [process](https://commercetools.atlassian.net/wiki/spaces/MC/pages/82003203/4.+How+to+Request+Translations+Step-by-Step+Guide+for+UXDs)

TLDR: You need to send me the new keys that should be translated (via slack is enough) and I will create the tralsation request ticket :)

packages/components/filters/src/filter-menu/footer/footer.tsx
```diff
@@ -1,5 +1,70 @@
...
+
+  & > div:first-of-type {
+    margin-top: ${designTokens.spacingS};
+  }
+
+  & > div:last-of-type {
+    margin-top: ${designTokens.spacingS};
+    display: flex;
+    justify-content: flex-end;
+  }
```
misama-ct, line 43: - the t-shirt size named spacing-tokens were deprecated at some point in favor of numeral ones (`spacing20` is the equivalent of `spacingS`) 
- you can use the `padding-top` property on the parent `div` to apply the safespace instead of setting the `margin-top` property two times on the 2 child-div's.
- elements in ui-kit which need styling are usually targeted directly and not via `&` selectors (by avoiding those selectors all css-rules generated have the same specificity).

misama-ct: @FilPob  one more thing I noticed is that the Clear-Button has a different computed height than the Apply button. In case there is no apply button, this will cause the footer to have a different (lower) height than with two buttons. It's probably worth to apply a fixed height to the footer and position the buttons within.

misama-ct (changes requested):

packages/components/filters/src/filter-menu/footer/footer.tsx
FilPob: @misama-ct no that's okay if the footer is smaller, if there is only the clear all button

ddouglasz (approved):
🙌🏽

FilPob:
@valoriecarli  thanks looks good from my side... Just one thing: the primary buttons should have `size: 10` :)

ByronDWall (approved):
This looks good to me, thanks for putting in the PR and addressing the comments!

misama-ct (approved):
💐 Thank you!

ByronDWall:
@FilPob since there is a lot of behavior that is specified in the specs, such as disabled states, loading states, etc that are reflections of the state of the consuming application, we've decided it's better to have the consumer pass the button in, so any sizing will need to be handled on that side of things

FilPob:
@ByronDWall @valoriecarli  got it, thanks :)

FilPob (approved):

jaikamat (approved):

PR #2948, opened 2024-10-07
FCT 1207 - guarantee unique HTML ID's in `FieldError` component fields

ByronDWall:
#### Summary
Guarantees unique HTML ID's in `FieldError` component fields. Is part of the fix for this bug [reported in slack](https://commercetools.slack.com/archives/C03B6NKTA83/p1727955170128789).

## Description
The `FieldErrors` component is meant to display error fields.  If multiple errors are passed to the component in the `errors` object, the component is meant to display multiple fields.  Previously, if `props.id` was passed to the component, it would be used as the `id` attribute for all error fields.  If multiple errors were passed, each field would have the same `id`.

This PR updates the `FieldErrors` component so that each `FieldError` on the page will have a unique id attribute in the DOM.  It does this by generating a sequentialId for each component whose `props.id` is `undefined`, thus ensuring there is always a valid `id`.  It then appends the `index` of each mapped `error` to the "baseId", insuring that each field has a valid `id` attribute in the DOM.

It also updates tests and the `FieldErrors` README to reflect these changes.


tylermorrisford (approved):

ByronDWall:
@tdeekens I suppose I focused too much on the need for unique id's in the error/warning fields, and less on what was happening in the parent fields. like you, I also need to better understand the interplay of fields here. This definitely adds another wrinkle to the fix that I hadn't considered.

> My worry is that this behavior breaks as there is no match any longer between the field and at least the first ErrorMessage using the aria-errormessage and id as a link between the two.

This is a legitimate concern, and an interesting issue since there can be multiple errors, and the id for each error/warning field should be unique.  Making it so that the first id can be accurately referenced in `aria-errormessage` should be relatively straightforward.  Handling multiple errors is less straightforward but now that I understand the problem I have a few ideas.

misama-ct:
> > My worry is that this behavior breaks as there is no match any longer between the field and at least the first ErrorMessage using the aria-errormessage and id as a link between the two.
> 
> This is a legitimate concern, and an interesting issue since there can be multiple errors, and the id for each error/warning field should be unique. Making it so that the first id can be accurately referenced in `aria-errormessage` should be relatively straightforward. Handling multiple errors is less straightforward but now that I understand the problem I have a few ideas.

If it's just about scrolling to the first invalid element and not a specific one, could [this](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) maybe be sufficient?

tdeekens:
> accurately referenced in aria-errormessage should be relatively straightforward

Yes, we could do `error-1-0` as the first and link it up to the field. The others then `error-1-n` incrementing from it.

> If it's just about scrolling to the first invalid element and not a specific one, could [this](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) maybe be sufficient?

That's a good point. Scrolling was what I remembered as the main concern. There is [this](https://github.com/commercetools/merchant-center-frontend/blob/6812a191b224bcd2f4e5a4a56f9493cb6d28da65/packages-shared/experimental-components/src/components/scroll-to-field-error/scroll-to-field-error.js#L5) implementation of the scrolling for which better solutions might exist. As it stands it heavily relies on the aria-errormessage. I am not sure how widly spread the adoption is through UX.

ByronDWall:
@misama-ct that's a good thought, but unfortunately I think it's about scrolling to the first message for a specific invalid element when there are multiple invalid elements on the page

ByronDWall:
> Yes, we could do error-1-0 as the first and link it up to the field. The others then error-1-n incrementing from it.

yeah, I was thinking something like `aria-errormessage={Object.keys(errors).map((_, idx) => ${id}-{idx}}`

That or pass the `id` directly to the first error, then add `-n` to the rest in `FieldError`.

> There is [this](https://github.com/commercetools/merchant-center-frontend/blob/6812a191b224bcd2f4e5a4a56f9493cb6d28da65/packages-shared/experimental-components/src/components/scroll-to-field-error/scroll-to-field-error.js#L5) implementation of the scrolling for which better solutions might exist.

What's interesting about this code is that it is looking to scroll to the `input` and not the warning message.  The only time it references `aria-warningmessage` seems to be for custom fields?

ByronDWall:
OK, it's obvious I was focusing on the wrong issue.  Instead of giving each error an `id` generated by the `index`, I am going to wrap `FieldErrors` in a div, and give that div the `id`.  Which solves the problem of 'how to deal with multiple errors'.  The wrapper `div` shouldn't impact styling, since the `TextField` used as a child only adds font color/size/weight values.

The other issue, 'fields of the same type pass the same id to error/warning fields', will be handled in each field.  But now that I have a better idea of what's going on, I believe the issue can be handled alongside `FieldErrors` and `FieldWarnings` all in a single PR.

PR #2949, opened 2024-10-07
FCT 1208 - guarantee unique HTML ID's in `FieldWarnings` component fields

ByronDWall:
#### Summary
Guarantees unique HTML ID's in `FieldWarnings` component fields. Is part of the fix for this bug [reported in slack](https://commercetools.slack.com/archives/C03B6NKTA83/p1727955170128789).

## Description
The `FieldWarnings` component is meant to display warning fields.  If multiple warnings are passed to the component in the `warnings` object, the component is meant to display multiple fields.  Previously, if `props.id` was passed to the component, it would be used as the `id` attribute for all warning fields.  If multiple warnings were passed, each field would have the same `id`.

This PR updates the `FieldWarnings` component so that each `FieldWarning` on the page will have a unique id attribute in the DOM.  It does this by generating a sequentialId for each component whose `props.id` is `undefined`, thus ensuring there is always a valid `id`.  It then appends the `index` of each mapped `warning` to the "baseId", insuring that each field has a valid `id` attribute in the DOM.

It also updates tests and the `FieldWarnings` README to reflect these changes.


tylermorrisford (approved):

packages/components/field-warnings/src/field-warnings.tsx
```diff
@@ -50,7 +57,7 @@ const FieldWarnings = (props: TFieldWarningsProps) => {
           // Custom warnings take precedence over the default warnings
           if (warningElement)
             return (
-              <WarningMessage key={key} id={props.id}>
+              <WarningMessage key={key} id={fieldId}>
```
ddouglasz, line 60: 💯

ddouglasz (approved):

ByronDWall:
closed in favor of [this PR](https://github.com/commercetools/ui-kit/pull/2950)

PR #2950, opened 2024-10-08
FCT 1207-1224: unique id's for error and warning containers in all `-Field` components

ByronDWall:
## Summary

Guarantees unique HTML ID's in `FieldErrors` and `FieldWarnings` components. Guarantee unique ID's for error/warning field containers for each `-Field` component. Is the fix for this bug [reported in slack](https://commercetools.slack.com/archives/C03B6NKTA83/p1727955170128789).

## Description

The `FieldErrors` and `FieldWarnings` components are meant to display error and warning fields. If multiple errors/warnings are passed to the component in the errors/warnings object, the component is meant to display multiple fields. Previously, if props.id was passed to the component, it would be used as the id attribute for each fields. If multiple errors/warnings were passed, each field would have the same id.

This PR updates the `FieldErrors` and `FieldWarnings` component so that all fields have a parent wrapper on which the `id` is set. This insures that the `id` passed to the component is always unique and refers to all errors/warnings.

Each `-Field` component is updated so that it builds the `id` for the `errors` and `warnings` containers based on the unique `id` used for the `-Field` input.  This insures that the `id` specified in the `aria-errormessage` attribute for each `-Field` is unique.  Tests added to each `-Field` component to ensure the id generation works as expected.

There are a lot of components changed, but changes for all `-Field` components are very similar and so should be fairly straightforward to review.


packages/components/fields/async-select-field/src/async-select-field.spec.js
```diff
@@ -191,6 +191,27 @@ describe('when showing an info button', () => {
...
 describe('when field is touched and has errors', () => {
+  it('should render an id for the error container that is based on the component id', async () => {
+    const { container, findByText } = renderAsyncSelectField({
+      touched: true,
+      errors: { custom: true },
+      renderError: () => 'Custom error',
+    });
+    await findByText(/custom error/i);
+    expect(
+      container.querySelector('#async-select-field-errors')
```
tdeekens, line 202: Nit: wondering if we should `queryByRole` on the status role over the id of the div.

packages/components/fields/date-time-field/src/date-time-field.spec.js
```diff
@@ -151,6 +151,22 @@ describe('when showing an info button', () => {
...
+      errors: { missing: true },
+    });
+    expect(container.querySelector('#date-field-errors')).toBeInTheDocument();
+  });
+  it('should set the aria-errormessage value to the id of the error container', () => {
+    const { getByRole } = renderDateTimeField({
+      touched: true,
+      errors: { missing: true },
+    });
+    expect(getByRole('textbox')).toHaveAccessibleErrorMessage(
```
tdeekens, line 166: This assertion!

tdeekens (approved):
Looks really better than the first approach! Thanks for working through the monoton changes on all components!

.changeset/big-beers-reflect.md
```diff
@@ -0,0 +1,23 @@
+---
+'@commercetools-uikit/localized-multiline-text-field': minor
```
tdeekens, line 2: I think before merging (noticed after submitting review) we should make this a patch? It's not really a new feature we're rolling out unless we calling it: finally we add correct markup for error messages.

packages/components/fields/async-select-field/src/async-select-field.spec.js
ByronDWall: the idea is that this is specifically testing that the desired id is present and formed correctly.  The next test does the more 'rtl approved' testing with roles, etc

.changeset/big-beers-reflect.md
ByronDWall: [done](https://github.com/commercetools/ui-kit/pull/2950/commits/e70c1c470a263e16f9bbeb1c7c76ed1330d4cb4a)

tylermorrisford (approved):

tdeekens (approved):
This nicely fixes the broken setup we have. I liked @misama-ct suggestion around using `:invalid`. Maybe an investigation can be a followup if we can use it.

PR #2951, opened 2024-10-08
Filters - Filter Menu - menu header

ddouglasz:
## Description

Add Filters - Filter Menu - menu header component.

This pull request introduces a new `FilterMenu` header component and updates related files to support its implementation. The changes include adding new props to the `FilterMenu` component, updating tests, and modifying the storybook example.

### Implementation of `FilterMenu` header component:

* [`.changeset/dirty-coins-begin.md`](diffhunk://#diff-c75d29d5c9e495acab20cba8b97ac0ca099c49ff793305c459d9d29ddfc31c79R1-R5): Added a minor change entry for the new `FilterMenu` header component.
* [`packages/components/filters/src/filter-menu/header/header.tsx`](diffhunk://#diff-d9b7690b533a80c08714a93baf2a43d7c7b2a6f4c0a08e863cac62698e6739a5L1-R79): Implemented the `Header` component with new props and styles, including `label`, `operandOptions`, `renderOperandsInput`, `onSelectOperand`, `onSort`, and `menuHeaderWidth`.

### Updates to `FilterMenu` component:

* [`packages/components/filters/src/filter-menu/filter-menu.tsx`](diffhunk://#diff-3354452f244af81a93ace12c41a2e7231f2f06731aea479cdc79c3a9e0180983L15-R31): Updated the `FilterMenu` component to include the new `Header` component with its props.

### Test updates:

* [`packages/components/filters/src/filter-menu/header/header.spec.tsx`](diffhunk://#diff-d7724c18584d41e2bc283455de42d8753daae3c371a1460e6769d09c94b44923L1-R79): Enhanced tests for the `Header` component to cover new functionalities like conditional rendering of `SelectInput` and `IconButton`.

### Storybook update:
This task ideally should have no story so after testing, the story modification will be taken off.


packages/components/filters/src/filters.stories.tsx
```diff
@@ -16,5 +17,6 @@ export default meta;
 type Story = StoryFn<typeof Filters>;
 
 export const BasicExample: Story = () => {
-  return <Filters label={'test'} />;
+  // return <Filters label={'test'} />;
```
ddouglasz, line 20: only for testing purposes, I will return back to its state before merging

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,81 @@
...
+                props.onSelectOperand(event.target.value as string);
+              }}
+            />
+          </div>
+        )}
+
+        {props.onSort && (
+          <IconButton
+            size="20"
+            theme={isActive ? 'info' : 'default'}
```
ddouglasz, line 66: You might notice a slight change between the IconButton when Active. This is because "round" prop is deprecated and I have checked in with Filip to confirm we are using the new prop design(square).

.changeset/dirty-coins-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/filters': minor
+---
+
+Implement FilterMenu header component.
```
ByronDWall, line 5: I don't think we need a changeset here, since we aren't publishing `filters` for public consumption until the entire feature is done

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/270993db7c82ab5975460c5515973f05899a1e3c
Thank you

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,81 @@
...
+
+type TOption = {
+  value: string;
+  label: string;
+};
+
+type THeaderProps = {
+  headerLabel: string;
+  operatorOptions?: Array<TOption>;
+  onSelectOperand: (value: string) => void;
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/7a2182f59627197c95bdf053e692d5559368320d
Thank you.

tylermorrisford (approved):
I'm not _entirely_ contextualized with filters yet, but this looks really nice to me.

packages/components/filters/src/filter-menu/filter-menu.tsx
```diff
@@ -81,6 +82,30 @@ const menuStyles = css`
...
+          // Default to the first option if no value is passed
+          headerSelectOptions ? headerSelectOptions : operatorOptions[0].value
+        }
+        options={operatorOptions}
+        onChange={(event) => {
+          setHeaderSelectOptions(event.target.value as string);
+        }}
+      />
+    );
+  };
```
misama-ct, line 107: I saw the comment below, but what is speaking against moving this part into the story?

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,73 @@
...
+import { designTokens } from '@commercetools-uikit/design-system';
+import { css } from '@emotion/react';
+
+type TOption = {
+  value: string;
+  label: string;
+};
+
+type THeaderProps = {
+  headerLabel: ReactNode;
```
misama-ct, line 14: what about just `label`? If more labels are added later, they can get the qualifier in their name.

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,73 @@
...
+  operandOptions?: Array<TOption>;
+  renderOperandsInput?: boolean;
```
misama-ct, line 16: I couldn't find any usage of a `operandOptions` or `renderOperandsInput` prop. Are those leftovers maybe?

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,73 @@
...
+const getSelectInputStyles = (props: THeaderProps) => css`
+  width: ${props.menuHeaderWidth};
+  margin: ${designTokens.spacing20};
+`;
+
+const Header = (props: THeaderProps) => {
+  const [isActive, setIsActive] = useState(false);
+
+  return (
+    <Spacings.Inline>
```
misama-ct, line 47: `Spacings.Inline` has no effect, it can be safely ditched.  

image

I feel like the `headerContainer` can consume the available space  and make sure the sort-button is always on the right. But the label- and input-div probably need some tinkering.

@ByronDWall can you give a bit of direction?

Can labels wrap, or do they have a fixed width/maxWidth? What happens if this width is reached? wrap lines, ellipsis? Does the select need its size manually set, or can it just occupy the leftover space?

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -1,5 +1,73 @@
...
+  operandOptions?: Array<TOption>;
+  renderOperandsInput?: boolean;
+  /**
+   * the function to render the operator input
+   */
+  renderOperatorsInput?: () => ReactNode;
+  /**
+   * the function to sort the data
+   */
+  onSort?: () => void;
```
misama-ct, line 24: @ByronDWall In the ticket it's written `onSort`, but we used the `on...Request` notation previously when buttons are actually doing nothing except triggering a callback fn.  

`onSortRequest`?

packages/components/filters/src/filter-menu/filter-menu.tsx
ddouglasz: Totally agree! updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/87f00d071d260cc8ba05dc38380993e55887aeca

packages/components/filters/src/filter-menu/header/header.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/87f00d071d260cc8ba05dc38380993e55887aeca

Thank you

packages/components/filters/src/filter-menu/header/header.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/87f00d071d260cc8ba05dc38380993e55887aeca

Lets discuss this and I will follow up

Thank you

packages/components/filters/src/filter-menu/header/header.tsx
ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/87f00d071d260cc8ba05dc38380993e55887aeca

Lets discuss this too and I will follow up

Thank you

ByronDWall: yes, sorry, it should be `onSortRequest`

ddouglasz: Updated: https://github.com/commercetools/ui-kit/pull/2951/commits/ec5a8a9aed833d452099407e0cb097c85b871fcd

Thanks

packages/components/filters/src/filter-menu/header/header.tsx
ByronDWall: the header should always take up 100% of the width of the container, and as long as there is an 8px gap between the label and the input, and a 24px gap between the input and the sort button, the operators input should be able to take up the remaining space. ([as specified in the figma](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2040-8188&m=dev))

As far as min/max widths for the label, and whether it should truncate or wrap, I'd like to get clarification from @FilPob on that.  Same for min/max width for the operators input

FilPob: I put some more examples [here](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?m=auto&node-id=5581-34207&t=KR8QhepMF2i7dnoh-1). Hope that makes it clearer

- yes the header should always take 100% of the parent
- the sort icon should be always on the right edge of the parent
- the operator selector' width should be set to wrap it's content... it should not take the available space
- The filter name label should break into multiple lines if it's too long

ddouglasz: @FilPob  Thank you for clarifying, just want to clarify before I push the change. It should basically look like these right? 
 




image

image

image

ddouglasz: updated here: https://github.com/commercetools/ui-kit/pull/2951/commits/f0a8d312959a49036887fad0bb611df89e3fac1e

FilPob: yes that looks good :)

ByronDWall (approved):
Looking good to me, thanks for going through all the revisions!

packages/components/filters/src/filters.stories.tsx
```diff
@@ -1,10 +1,13 @@
...
 import Filters from './filters';
+import FilterMenu from './filter-menu';
+import { ReactNode, useState } from 'react';
+import SelectInput from '@commercetools-uikit/select-input';
 
 const meta: Meta<typeof Filters> = {
   title: 'components/Filters',
   component: Filters,
-  tags: ['local-dev'],
+  // tags: ['local-dev'],
```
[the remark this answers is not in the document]
ddouglasz: Will do! Thank you

PR #2954, opened 2024-10-09
FCT-1188: filters: quick filters: create quick filters component

misama-ct:
## Summary
This PR adds the QuickFilter component and necessary changes in the Tag component to make the QuickFilters look like expected.

### Description
 **Tag** component
-  the `type` property was deprecated in favor of a new `tone` property. 
- the existing types were mapped against the new tones to ensure backward compatibility
- a new tone (type) 'surface' was added
- visual tests were added for the new 'tone' prop
- layout adjustments were made to the Tag component (~~font-size~~, spacings, gaps)
- text-selection was disabled in the Tag component

**QuickFilters** component
- displays a list of tags
- triggers an action onClick of a tag
- uncontrolled, state (`isActive`) is managed by consumer


packages/components/quick-filters/src/quick-filters.tsx
```diff
@@ -1,12 +1,57 @@
...
+};
+
 export type TQuickFiltersProps = {
   /**
-   * This is a stub prop!
+   * collection of quick filter items
+   *
+   * @param item.id unique identifier for the item.
+   * @param item.label label to display
+   * @param item.isActive the current active state of the item
```
misama-ct, line 20: Not ideal, since the descriptions are duplicated, but this way the expected item-shape shows up directly in the components' props-table.

packages/components/tag/src/tag.stories.tsx
```diff
@@ -10,19 +10,46 @@ export default meta;
...
 };
 
-BasicExample.args = {
-  type: 'normal',
-  isDisabled: false,
-  to: '/project-key/products/icecream',
-  onRemove: () => alert('Remove tag request'),
-  children: 'Ice Cream',
+/** displays the tag as a react-router link, (no hover effects) */
+export const LinkedTag: Story = {
```
misama-ct, line 31: I added more stories, primarily to make sure that the existing behavior stays the same (e.g. unlike tags with `onClick` handler, links don't have a hover-effect).

packages/components/tag/src/tag.tsx
```diff
@@ -65,60 +67,68 @@ export type TTagProps = {
...
-            : designTokens.colorPrimary90};
-          ${props.onClick &&
-          `&:hover {
-            background-color: ${
-              props.type === 'warning'
-                ? designTokens.colorWarning85
-                : designTokens.colorPrimary90
-            };
-          }`}
+          user-select: none;
```
misama-ct, line 120: I added this, cause I frequently marked the label-text when clicking the tag.

packages/components/tag/src/tag.tsx
```diff
@@ -65,60 +67,68 @@ export type TTagProps = {
...
     <Constraints.Horizontal max={props.horizontalConstraint}>
       <div
         css={css`
           a {
-            cursor: pointer;
             text-decoration: none;
           }
           box-sizing: border-box;
-          cursor: default;
+          cursor: ${isInteractive ? 'pointer' : 'default'};
```
ByronDWall, line 114: should we also set `cursor: not-allowed` when `isDisabled` is true?

packages/components/tag/src/tag.tsx
```diff
@@ -65,60 +67,68 @@ export type TTagProps = {
...
+    case Boolean(props.to):
+      tagBodyProps = { as: Link, to: props.to };
+      break;
+    case Boolean(props.onClick):
+      tagBodyProps = { as: 'button' as ElementType };
+      break;
+    default:
+      tagBodyProps = {};
+  }
+
```
ByronDWall, line 102: There should also be a warning if the user uses the deprecated `type` prop, similar to the [warning for size in PrimaryButton](https://github.com/commercetools/ui-kit/blob/c41a012c352e7ddb1d3a0ab4d4a9eb6af2d25583/packages/components/buttons/primary-button/src/primary-button.tsx#L120).

packages/components/tag/src/tag.visualroute.jsx
```diff
@@ -199,9 +199,7 @@ export const component = () => (
         {longText}
       </Tag>
     </Spec>
-    <Spec
-      label="Tag list - with remove"
-    >
+    <Spec label="Tag list - with remove">
       <TagList>
         {examplesLong.map((tag, index) => (
           <Tag type="normal" to="foo/bar" onRemove={() => {}} key={index}>
```
ByronDWall, line 205: Does it make sense to update these tests to use `tone` instead of `type`?

packages/components/quick-filters/src/quick-filters.readme.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta, Markdown } from '@storybook/blocks';
 import ReadMe from './../README.md?raw'
 
-<Meta tags={['local-dev']} title="components/QuickFilters/Readme" />
+<Meta tags={[/*'local-dev'*/]} title="components/QuickFilters/Readme" />
```
ByronDWall, line 4: This is the entire `QuickFilters` component, so there isn't a need to put this back into `local-dev` once the PR is merged

packages/components/quick-filters/src/quick-filters.tsx
ByronDWall: not ideal, but definitely helpful and something we should do in `Filters` as well

packages/components/quick-filters/src/quick-filters.stories.tsx
```diff
@@ -1,22 +1,57 @@
...
-      control: 'text',
-    },
-  },
+  //tags: ['local-dev'],
 };
 export default meta;
 
 type Story = StoryObj<typeof QuickFilters>;
 
+/** display a list of filter tags which can be enabled or disabled. */
```
ByronDWall, line 17:
```suggestion
/** The `QuickFilters` component displays a selection of `Tag` components that represent the available filter actions. */
```

ByronDWall (commented):
A few small nits, but overall this is looking really good!

packages/components/tag/src/tag.tsx
misama-ct: Do we really need this? Implementation is easy, but honestly last time was a mess (since we broke peoples tests and forced them to act, although there was no breaking change).

I feel like once we have a plan how to remove deprecated stuff, we can still add this console.warns to all deprecated components/props and communicate a time- & deadline to teams. What do you think?

packages/components/tag/src/tag.visualroute.jsx
misama-ct: I didn't change existing tests, so that I can make sure the existing `type` property continues to work.

packages/components/tag/src/tag.tsx
misama-ct: There won't be anything in the markup that would indicate a possible action which was disabled (no `<a>` or `<button>` tag) but as a visual clue, why not?

packages/components/tag/src/tag.tsx
ByronDWall: that works for me

packages/components/tag/src/tag.visualroute.jsx
ByronDWall: fair enough, im fine with updating when we remove the prop

ByronDWall (approved):

ddouglasz (approved):
A few nits. Otherwise 🙌🏽
Thank you.

misama-ct:
@FilPob I changed the font-size back to 14px, height is still 32px, gap between drag-icon and label was restored. Please have another look.

packages/components/tag/src/tag.tsx
misama-ct: Let's just [try it out](https://github.com/commercetools/ui-kit/pull/2954/commits/8538bbb53e6d024801727046cab4d1d18455a599).

packages/components/quick-filters/src/quick-filters.spec.tsx
```diff
@@ -1,12 +1,26 @@
...
-  it('should render the quick-filters', async () => {
-    await render(<QuickFilters label="quickfilters" />);
-    await screen.findByText('quickfilters');
+  it('should render items and propagate item-clicks', async () => {
+    const items = [
+      { id: '1', label: 'Foo', isActive: true },
+      { id: '2', label: 'Bar', isActive: false },
+    ];
+
+    const cb = jest.fn();
```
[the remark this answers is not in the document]
misama-ct: [Done](https://github.com/commercetools/ui-kit/pull/2954/commits/a7951091694955396dddd49f6c06b9f77fc60fa7)

FilPob (approved):

PR #2964, opened 2024-10-14
FCT 1079- Filter Menu - Trigger Button - Display selected values in `Chips`

jmcreasman:
This PR introduces a Trigger Button that shows the values selected by the user in discrete Chip components. When the length of those exceeds the max, subsequent Chips are hidden behind a badge displaying a count of how many Chips are hidden or partial hidden.

[Ticket](https://commercetools.atlassian.net/browse/FCT-1079)
[Design](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-6256&node-type=frame&m=dev)

You can manually test different filter scenarios by adding/subtracting labels in the `appliedFilterValues` object in the `filters.tsx file`. Just a reminder, partially hidden labels will count towards the overflow. 

Screenshot 2024-10-10 at 14 52 22

AC Summary:
- the FilterMenu shows each selected value in a Chip component
- when the length of the trigger button exceeds 642px, subsequent Chips should be hidden behind a gradient, and a badge should appear that displays a count of how many Chips are hidden
- tests added


packages/components/filters/src/filter-menu/trigger-button/trigger-button.tsx
```diff
@@ -1,8 +1,63 @@
...
+          const hasOverflow = scrollWidth > clientWidth;
+          setIsOverflowing(hasOverflow);
+
+          let widthOfAllChildren = 0;
+          let visibleChipCount = 0;
+
+          if (hasOverflow) {
+            Array.from(children).forEach((child, idx) => {
+              const { width } = (child as HTMLElement).getBoundingClientRect();
+              widthOfAllChildren += width + 4;
```
jmcreasman, line 38: Adding 4 to each child `width` since the width doesn't take the 4px gap into account

packages/components/filters/src/filter-menu/trigger-button/trigger-button.styles.ts
```diff
@@ -15,7 +16,7 @@ export const triggerWrapper = css`
   box-shadow: 0 0 0 ${designTokens.borderWidth1} ${designTokens.colorNeutral85};
   cursor: pointer;
   height: ${designTokens.spacing50};
-  max-width: 100%;
+  max-width: ${Constraints.getMaxPropTokenValue(13)};
```
misama-ct, line 19: What was the reason to change the value? If it's not 100% (="consume not more than the available horizontal space") it may ignore potential boundaries set by a parent component.

e.g. I limit the width available to the FilterMenu:

```
<div style={{ width: '256px'}}>
    <FilterMenu {...} />
</div>
```

This is what happens:

image

packages/components/filters/src/filter-menu/trigger-button/trigger-button.tsx
misama-ct: Can you try calculating if the boundaries of the current child are within the boundaries of the container instead? Something along this:

```
const childRect = child.getBoundingClientRect();
const containerRect = containerRef.current.getBoundingClientRect();
if (
        childRect.left >= containerRect.left &&
        childRect.right <= containerRect.right
) {
        visible += 1; 
}
```

This way the *current* styling is not tied to the counter implementation.

packages/components/filters/src/filter-menu/trigger-button/trigger-button.styles.ts
ByronDWall: its because the [figma specifies](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-7258&m=dev) that the badge should display if the button is wider than `642px`.

This is a good point though. I think adding a width of 100% satisfies both conditions?

```suggestion
  width: 100%;
  max-width: ${Constraints.getMaxPropTokenValue(13)};
```
which results in:
Screenshot 2024-10-15 at 11 24 40 AM

ByronDWall: this is because of the requirement that the badge container be displayed when the component gets to be [larger than 642px](https://www.figma.com/design/b2S3GGpScVDuf15VAGEqeL/Filter-pattern?node-id=2339-7258&m=dev).

I think this works to satisfy both requirements:
```suggestion
  max-width: min(100%, ${Constraints.getMaxPropTokenValue(13)});
```
Screenshot 2024-10-15 at 12 52 54 PM

packages/components/filters/src/filter-menu/trigger-button/trigger-button.tsx
ByronDWall: done in [this commit](https://github.com/commercetools/ui-kit/pull/2964/commits/11f41ac64e8c912ffe2ddfba53a0e4fe52686b46)

misama-ct (approved):
💐 Thanks!

packages/components/filters/src/filters.stories.tsx
```diff
@@ -1,13 +1,10 @@
...
 import Filters from './filters';
-import FilterMenu from './filter-menu';
-import { ReactNode, useState } from 'react';
-import SelectInput from '@commercetools-uikit/select-input';
 
 const meta: Meta<typeof Filters> = {
   title: 'components/Filters',
   component: Filters,
-  tags: ['local-dev'],
+  // tags: ['local-dev'],
```
ddouglasz, line 7: Please remember to uncomment before merging.

ddouglasz (approved):
This looks good to me now too. 
Thank you! 🚀

packages/components/filters/src/filter-menu/trigger-button/trigger-button.styles.ts
```diff
@@ -81,8 +80,12 @@ export const removeButton = css`
 `;
 
 export const label = css`
-  flex-grow: 0;
+  flex: 0 0 auto;
+  overflow: hidden;
+  text-overflow: ellipsis;
   cursor: inherit;
+  white-space: nowrap;
+  max-width: ${designTokens.constraint6};
```
ByronDWall, line 88: @FilPob this is my attempt to handle cases when the label for the filter is very long, it truncates the label after 284px so that there is always some room left for showing selected values in the button
Screenshot 2024-10-17 at 8 56 45 AM

packages/components/filters/src/filter-menu/header/header.tsx
```diff
@@ -16,26 +21,39 @@ type THeaderProps = {
...
 };
 
 const headerContainerStyles = css`
-  display: flex;
+  color: ${designTokens.colorNeutral40};
+  font-weight: ${designTokens.fontWeight500};
+  font-size: ${designTokens.fontSize20};
+  line-height: ${designTokens.lineHeight20};
+  display: grid;
+  grid-template-columns: 1fr min-content;
```
ByronDWall, line 38: @ddouglasz I updated the header styles so that the menu and the operators input never overflow their available width in the menu

packages/components/filters/src/filter-menu/trigger-button/trigger-button.styles.ts
FilPob: Awesome, that looks good to me

packages/components/filters/src/filter-menu/filter-menu.tsx
```diff
@@ -80,7 +107,29 @@ const menuStyles = css`
...
 function FilterMenu(props: TFilterMenuProps) {
+  const menuBodyRef = useRef<HTMLDivElement>(null);
+
+  const focusMenuBody = useCallback(
+    (e) => {
+      if (menuBodyRef.current) {
+        const firstFocusableElementInMenuBody = findFirstFocusable(
+          menuBodyRef.current
+        );
+        if (firstFocusableElementInMenuBody) {
```
ByronDWall, line 124: If there's a focusable element in the body, then `preventDefault` to prevent radix from putting focus on a focusable item in the header, and focus on the element in the body.

If there is no focusable element, then do nothing and let radix handle focus management

FilPob:
@ByronDWall  thanks! looks great overall. Just one question: Why are the filter values in the chips capitalized? 
image

packages/components/filters/src/filter-menu/filter-menu.tsx
```diff
@@ -1,12 +1,33 @@
...
 import * as Popover from '@radix-ui/react-popover';
 import { Footer } from './footer';
 import { Header } from './header';
 import { TriggerButton } from './trigger-button';
 
+/**
+ * CSS selector to find focusable elements.
+ * @see https://github.com/microsoft/tabster/blob/6bfd54a45f5b20eccd17b8a05f6c86c241b992c3/src/Focusable.ts#L17-L25
+ */
+const FOCUSABLE_CSS_SELECTOR = `a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), *[tabindex], *[contenteditable]`;
```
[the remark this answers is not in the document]
misama-ct: Looks terrible 😄 but certainly an interesting solution. Can you add a TODO comment, so that we can find it easier and abstract away if it works reliably (this could certainly be reused for all kind of modals)

ByronDWall:
@FilPob it's to show that the consuming app can modify the labels as-necessary after theyve been selected in the filter.  The chips don't do that themselves

PR #2967, opened 2024-10-17
Fct 1250 UI kit failing percy workflow

ddouglasz:
Fix failing percy workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,13 +55,7 @@ jobs:
         run: yarn visual-testing-app:build
 
       - name: Running Visual Regression Tests for UI components
-      # aa-exec:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   build_lint_and_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ByronDWall, line 15: do you think it's better to downgrade or upgrade here?  if we upgrade, after oct 31 we should be able to stay with `ubuntu-latest`

ddouglasz: ubuntu-24.04 seems to fail, which is strange because it should have been our latest at least before now.

ddouglasz: Updated. Thank you.

jaikamat (approved):

PR #2979, opened 2024-10-29
FCT-1187 - Filter Selection tests

valoriecarli:
added a few more unit tests and continued to expand on Filter int tests.

### DIFF
--- a/packages/components/filters/src/filters.spec.tsx
+++ b/packages/components/filters/src/filters.spec.tsx
@@ -1,8 +1,18 @@
 import { useState } from 'react';
-import { fireEvent, render, screen, within } from '../../../../test/test-utils';
 import Filters, { TFiltersProps } from './filters';
-import { PrimaryColorsInput } from './fixtures/inputs';
 import { FILTER_GROUP_KEYS } from './fixtures/constants';
+import { PrimaryColorsInput, ColorNameTextInput } from './fixtures/inputs';
+import { fireEvent, render, screen } from '../../../../test/test-utils';
+import {
+  getAddFilterButton,
+  getBadgeStatus,
+  getClearAllFiltersButton,
+  openAddFilterDialog,
+  displayedChips,
+  selectFilter,
+  selectFilterValues,
+  toggleFilterList,
+} from './filters.spec.utils';
 
 const mockRenderSearchComponent = (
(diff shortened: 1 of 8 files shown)


packages/components/filters/src/filters.spec.utils.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+//Click `Add filter` button & return dialog for interaction
+export const openAddFilterDialog = async (): Promise<HTMLElement> => {
+  const addFilterButton = await getAddFilterButton();
+  fireEvent.click(addFilterButton);
+  return await screen.findByRole('dialog');
+};
+
+//Select filter to apply
+export const selectFilter = async (dialog: HTMLElement, optionText: string) => {
```
ByronDWall, line 41:
```suggestion
export const selectFilterOption = async (dialog: HTMLElement, optionText: string) => {
```

packages/components/filters/src/filters.spec.utils.tsx
```diff
@@ -0,0 +1,69 @@
...
+  });
+};
+
+export const getClearAllFiltersButton = (): HTMLElement | null => {
+  return screen.queryByRole('button', {
+    name: /clear all/i,
+  });
+};
+
+export const getBadgeStatus = (): HTMLElement | null => {
```
ByronDWall, line 22:
```suggestion
export const getBadgeElement = (): HTMLElement | null => {
```

packages/components/filters/src/filters.spec.tsx
```diff
@@ -57,73 +82,207 @@ const FilterTestComponent = (
...
+    const chips = displayedChips('primaryColors');
+    expect(chips).toEqual(['blue', 'green', 'pink', 'lavender', 'azure']);
+
+    //Collapse filter list & expect no badge to display
+    await toggleFilterList();
+    const filterTotalBadge = getBadgeStatus();
+    expect(filterTotalBadge).toBeNull();
+  });
+
+  //!is this needed?
```
ByronDWall, line 190: yes, it's good to have

tylermorrisford (approved):
I don't have the level of context that others have in the realm of filters, but this looks real nice 🪄

ddouglasz (approved):
Looks good! 👏

ByronDWall (approved):

PR #2988, opened 2024-11-11
feat: remove support for locale zh-CN

mustafaasif2:
## Summary

Removed support for the Chinese locale (zh-CN) across UI Kit.

## Description

Localization and translation support for the zh-CN locale are no longer needed in MC applications. This change removes all instances of zh-CN loading and support in UI Kit.


emmenko (approved):

tdeekens (approved):

misama-ct (approved):
Thank you!

tylermorrisford (approved):

packages/i18n/data/zh-CN.json
```diff
@@ -1,310 +1 @@
...
-  "UIKit.SelectInput.placeholder": {
-    "developer_comment": "Default input placeholder text",
-    "string": "选择..."
-  },
-  "UIKit.TimeInput.placeholder": {
-    "developer_comment": "Placeholder for time input",
-    "string": "小时：分钟"
-  }
-}
+{}
```
emmenko, line 1: @mustafaasif2 @tdeekens I decided to keep this file for now with empty data.

The reason being is that in App Kit's `i18n` package the file is being referenced.

image

If we release UI Kit with the file removed, and App Kit is not updated to the latest version that removes the reference, there will be an error as the file is not found.

This way, we can release both UI Kit and App Kit independently without causing regressions.

emmenko: https://github.com/commercetools/ui-kit/pull/3025

PR #3002, opened 2024-12-04
FCT-1293: Basic omlet integration

misama-ct:
#### Summary

I added the omlet config file, configured it to the best of my knowledge and added an `omlet-scan` script to the root `package.json`. 

(No changeset cause the changes have no effect on the packages)


package.json
```diff
@@ -38,7 +38,8 @@
     "visual-testing-app:build": "yarn workspace visual-testing-app build",
     "visual-testing-app:preview": "yarn workspace visual-testing-app preview",
     "labels:sync": "github-labels sync",
-    "typecheck": "tsc --noEmit --skipLibCheck"
+    "typecheck": "tsc --noEmit --skipLibCheck",
+    "omlet-scan": "yarn omlet analyze --hook-script ./scripts/omlet-hook-script.js"
```
ByronDWall, line 42:
```suggestion
    "omlet:scan": "yarn omlet analyze --hook-script ./scripts/omlet-hook-script.js"
```
separating the noun from the verb with a colon seems to be the norm for these scripts, my care amount is relatively low but consistency for things like script names really makes a difference as we add more scripts.

misama-ct: hahaha, yeah, why not

ByronDWall (approved):
🪄✨

tylermorrisford (approved):

PR #3019, opened 2024-12-12
FCT-1306 Code Connect POC

tylermorrisford:
#### Summary

This PR is a POC of both Figma's [code connect](https://github.com/figma/code-connect/tree/main) CLI tool and a homegrown script meant to assist with the creation and auditing of Figma connect files. The script is only a starting point meant to show a POC of how we would maintain the proper coverage of code connect files after they are generated.

## Description

In testing the Figma's code connect CLI we learned that it is largely a manual process; the user must supply the figma design component url(as in manually copy it out of the Figma UI) and then manually map the props(the largest part of this effort) before making a few other updates to the file before it is ready to be 'published' to Figma, which results in the 'connection' we want. 

The script added in 0f3b31a attempts to crawl the `./packages/components/` directory and locate the root component file, then if a `.figma.tsx` file is not found, it generates one based on a template. If the `.figma.tsx` file is found, the content is read and checked for strings like 'TODO' and '<placeholder>'. The intent here is to give us an easy way to know what figma connect files we have that are incomplete.

## Challenges

-The Ui-Kit file structure is organized in a way that makes it difficult to explore programmatically.
-There is a lack of understanding on my part of what the root files actually are in UI-KIT, ie. which components are designed in Figma and which are not.
-The script doesn't help us with the largest maintenance effort, which is how component props are mapped to Figma properties.
-There is an API that can tell us what each component's node_id is, but to this point I haven't found a way to programmatically find and map these ids without manual intervention, but more research could be handy.


packages/components/buttons/icon-button/src/icon-button.figma.tsx
```diff
@@ -0,0 +1,41 @@
...
+      // }),
+      // "theme": figma.enum('Theme', {
+      //   "Default": "default",
+      //   "Primary": "primary",
+      //   "Inforimary": "inforimary"
+      // })
+    },
+    example: (props) => <IconButton />,
+  }
+);
```
tylermorrisford, line 41: This file was created by the code connect CLI tool but has already been manually updated: 
The `@ts-nocheck` directive was added
The component import statement was fixed
The url on line 15 was manually supplied by us (pasted into the CLI prompt)
The prop mapping is somewhat usable but also mostly incorrect (and what is 'inforimary'?)

packages/components/buttons/link-button/src/link-button.figma.tsx
```diff
@@ -0,0 +1,17 @@
...
+  'https://www.figma.com/design/UoqmtHSHwxvEZRcbsM4A3Z/CT-product-design-system?node-id=<placeholder>',
+  {
+    props: {
+      // *This file was generated from a script*
+      // TODO: manually map props here, see https://www.figma.com/code-connect-docs/react/#figmaconnect
+      children: figma.children('*'),
+    },
+    example: (props) => <LinkButton>{props.children}</LinkButton>,
+  }
+);
```
tylermorrisford, line 17: This file was created by our script.

- The figma component node-id needs to be updated manually (as with the code connect CLI)
- The prop mapping needs to be updated manually

packages/components/buttons/icon-button/src/icon-button.figma.tsx
ByronDWall: the more that I think about it, the more that I think that code connect without props is 
a) still pretty helpful
b) a lot more maintainable, since we don't have to update props if/when they change

I wonder if there's a decent way to get the figma asset urls in a batch, and map them to components w/o props?

tylermorrisford: I agree, even knowing which components we're dealing with when doing a handoff would be helpful. 
There *IS* a way, but with what I know right now, it's very convoluted (from the figma community forum: `Select the Variant in the Component Set, open the console in Figma, enter figma.currentPage.selection[0].key.` From there you can request that key from an endpoint and get the node-id). Going to spend a bit more time on that and see what I find.

tylermorrisford: [6644d36](https://github.com/commercetools/ui-kit/pull/3019/commits/6644d36a38cd1503d94c7c6ee020aecf23e4b171) adds a "way" to get the figma asset ids and map them to their matching components. The link.figma.tsx file (below) was created from the updated script, which was able to map ids for 22 components. Of those 22, I think 20 or so are probably necessary (ie a component that we actually want to map and not just a small piece of something that is internal to it's parent component and never used elsewhere).

tylermorrisford:
Putting this in draft mode until there is a decision on expanding access to Figma Dev Mode.

PR #3027, opened 2024-12-17
Migrate components default props usage

CarlosCortizasCT:
### Summary

This PRs migrated the way we use `defaultProps` in React components as a preparation step to the upcoming overall migration from React v17 to React v19.

We're using a codemod published from `app-kit` ([reference](https://github.com/commercetools/merchant-center-application-kit/pull/3681)) in order to automate the process as much as possible.

### Description

The way we are using React components `defaultProps` in our codebase is no longer supported in the new React 19 version ([reference](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops)).

This is the main concept:
```tsx
// BEFORE
type TMyComponentProps = {
  prop1: string;
  prop2: string;
  prop3: string;
};

function MyComponent(props: TMyComponentProps) {
  return (
    <ul>
    <li>Prop 1: {props.prop1}</li>
    <li>Prop 2: {props.prop2}</li>
    <li>Prop 3: {props.prop3}</li>
  </ul>
  );
}
MyComponent.defaultProps = {
  prop1: 'My default value',
};


// AFTER
type TMyComponentProps = {
  prop1?: string; // <--- Make sure this is marked as optional
  prop2: string;
  prop3: string;
};

function MyComponent({ prop1: 'My default value', ...props }: TMyComponentProps) { // <--- default values here
  return (
    <ul>
    <li>Prop 1: {prop1}</li> {* <--- Direct access to destructured property *}
    <li>Prop 2: {props.prop2}</li>
    <li>Prop 3: {props.prop3}</li>
  </ul>
  );
}
// <--- No "defaultProps" component property
```

## NOTES
Something to bear in mind about the new pattern for using default props is that those props are not computed until the component is renderer whereas the legacy pattern make them available even before.
This has been discovered while migrating the `RadioGroup` and `PrimaryActionDropdown` components where we are inspecting the received children before rendering.

Here is a code example:
```tsx
const optionElements = Children.map(props.children, (child, index) => {
    // NOTE: Allowing to intersperse other elements than `Option`.
    if (
      child &&
      isValidElement(child) &&
      (child as TReactChild).type.displayName === Option.displayName
    ) {
      const clonedChild = cloneElement(child as TReactChild, { ... });
      const { wrapper } = child.props.components; // <---- HERE WE ARE ACCESSING A CHILD COMPONENT PROPERTY
      return wrapper ? wrapper(clonedChild) : clonedChild;
    }
    return child;
  });
```
When using the legacy `defaultProps` property, the props of the child included the `components` prop as it was a default value (empty object).
With the new pattern (the default value is in the component function declaration), the props of the child at this point does not include that property.


dbgmd-00001.sign
```diff
@@ -0,0 +1,6 @@
+tree 0003675ccdfb583bbe7bf1a3380aefaad9ce7c12
```
ByronDWall, line 1: Is this file something that should be ignored/removed?

CarlosCortizasCT: Sorry. That was a mistake.

I removed those files.

ddouglasz (approved):
builds fine, and just the class components still have defaultProps. Looks good to me 💯
Thank you!

ByronDWall (approved):
Overall looks good.  just out of curiosity, what percentage of this was completed by the codemod?

CarlosCortizasCT:
> Overall looks good. just out of curiosity, what percentage of this was completed by the codemod?

@ByronDWall That's a good question but does not have a direct answer.

The codemod refactored around 60 files.
After the first iteration, I would say I had to adjust between 8 and 10 files. Those were adjustments on top of what the codemod already had done.
Some of the manual adjustments have been incorporated to the codemod as a follow-up but there were some with very specific use cases I don't see trying to implement them in it.

The goal is for the codemod to cover the majority of a codebase, but not getting to a 100% coverage as that would take too much effort.

PR #3035, opened 2025-01-09
Proposal for keeping events simulation in tests clean

CarlosCortizasCT:
#### Summary

Proposal for keeping events simulation in tests clean

## Description

This PR is just to demonstrate a proposal I would like to make regarding the changes required in our tests because of React 19 being stricter with some events simulation (`blur`, `focus`).

This is the reasoning for the stricter rules:
* React 18 introduced automatic batching and stricter rules around state updates. Testing tools must ensure that any React state updates are wrapped in act() to simulate real-world behavior properly.
* When you call focus() on an element, it might trigger event listeners or React updates (e.g., onFocus) that could internally cause state updates. React now expects such state updates to be wrapped in act() during tests.

So now we need to use `act()` for handling those events but we haven't used that function until now in our tests and I found it usually confuses developers about its nature.
My proposal here is to abstract that usage from the tests and add it to the tests helpers so we just use the `fireEvent` API for all events simulation (although awaiting for those events is something we will need anyway).

### DIFF
--- a/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
+++ b/packages/components/dropdowns/dropdown-menu/src/dropdown-menu.spec.tsx
@@ -1,6 +1,6 @@
+import { act } from 'react';
 import SecondaryButton from '@commercetools-uikit/secondary-button';
-
-import { act, screen, render, fireEvent } from '../../../../../test/test-utils';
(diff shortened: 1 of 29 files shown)


ByronDWall (approved):
thanks for the pr and the type solution!

PR #3044, opened 2025-01-17
Update types for former defaultProps to be optional

ByronDWall:
In preparation for upgrading ui kit to react 19, it [was necessary](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops) to remove all `defaultProps` in favor of passing a default param to the prop.

This PR updates the types of any prop that has a default param to be optional so that it is not mandatory to pass the prop in consuming applications.  This reduces the amount of refactoring that consumers will have to do to satisfy typechecking.

### DIFF
--- a/packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
+++ b/packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
@@ -43,7 +43,7 @@ export type TColumnData = {
 
 export type TColumnSettingsManagerProps = {
   title?: string;
-  availableColumns: TColumnData[];
+  availableColumns?: TColumnData[];
   selectedColumns: TColumnData[];
   onUpdateColumns: (updatedColums: TColumnData[]) => void;
   areHiddenColumnsSearchable?: boolean;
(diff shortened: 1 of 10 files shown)


tylermorrisford (approved):
lgtm

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
```diff
@@ -101,7 +101,7 @@ export const handleColumnsUpdate = (
       : selectedColumns;
 
     const columns = isSwap ? selectedColumns : availableColumns;
-    const draggedColumn = columns.find(
+    const draggedColumn = columns!.find(
```
tdeekens, line 104: Sorry, but if this is optionally don't we conditionally chain here?

tdeekens (commented):
Just one question which I assume to lack context on.

packages/components/data-table-manager/src/column-settings-manager/column-settings-manager.tsx
ByronDWall: we know that `availableColumns` will always be `[]` because of the default param on line 134 - optional chaining is probably the better choice though

PR #3050, opened 2025-01-21
Update `TAvatarProps.size` for backwards compatibility

ByronDWall:
Makes `TAvatarProps.size` prop optional to match previous api surface with defaultProps in `TAvatarProps`, updates avatar styles to assert that the size prop will always be assigned due to default param assignment.


packages/components/data-table/README.md
```diff
@@ -162,7 +178,7 @@ export default Example;
    */
   shouldIgnoreRowClick?: boolean;
 }
-```
+````
```
CarlosCortizasCT, line 181: **nit** we might want to remove this extra tick

CarlosCortizasCT (commented):
Many thanks! 👍

kark (approved):

packages/components/data-table/README.md
ByronDWall: fixed in [fbdd863](https://github.com/commercetools/ui-kit/pull/3050/commits/fbdd863f4410f5fa490bc78c4a0432024467ee18)

PR #3056, opened 2025-02-05
Only update react-select child component types where necessary when using `moduleResolution: bundler` tsconfig setting

ByronDWall:
#### Summary
Reduces the amount of typecasting and code in the `react-select` based components to what is necessary when using the `moduleResolution: bundler` TypeScript configuration setting.

### DIFF
--- a/.changeset/beige-wolves-tickle.md
+++ b/.changeset/beige-wolves-tickle.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-uikit/async-creatable-select-input': patch
+'@commercetools-uikit/creatable-select-input': patch
+'@commercetools-uikit/async-select-input': patch
+'@commercetools-uikit/select-input': patch
+---
+
+Keep only patched types for needed components
--- a/.changeset/beige-wolves-tickle.md
+++ b/.changeset/beige-wolves-tickle.md
@@ -1,8 +1,0 @@
----
-'@commercetools-uikit/async-creatable-select-input': patch
-'@commercetools-uikit/creatable-select-input': patch
-'@commercetools-uikit/async-select-input': patch
-'@commercetools-uikit/select-input': patch
----
-
-Keep only patched types for needed components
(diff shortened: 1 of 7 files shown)


emmenko:
Hey @ByronDWall , thanks for looking into that.

Can you clarify how did you determine which components to type and which not? What speaks against having them all typed besides being very verbose?

emmenko (approved):
Let's get this releases as well, as there are still some issues from #3054 

https://github.com/commercetools/merchant-center-application-kit/actions/runs/13199662118/job/36848607466?pr=3722

ByronDWall:
@emmenko the ts error output for these components was specific to types from two specific files in `react-select`.  

I took a look at the types packaged in the `dist` of the uikit components when using `moduleResolution: node` and saw that these three components require types that are imported from direcotries in the `react-select` package instead of `react-select` itself, eg: 
```ts
CrossIcon: (props: import("react-select/dist/declarations/src/components/indicators").CrossIconProps) => import("@emotion/react").jsx.JSX.Element;
DownChevron: (props: import("react-select/dist/declarations/src/components/indicators").DownChevronProps) => import("@emotion/react").jsx.JSX.Element;
MenuPortal: <Option_11, IsMulti_11 extends boolean, Group_11 extends GroupBase<Option_11>>(props: import("react-select/dist/declarations/src/components/Menu").MenuPortalProps<Option_11, IsMulti_11, Group_11>) => import("@emotion/react").jsx.JSX.Element | null;
```

As opposed to working types, which import types from `react-select` directly, eg:
```ts
Control: <Option_1, IsMulti_1 extends boolean, Group_1 extends GroupBase<Option_1>>(props: import("react-select").ControlProps<Option_1, IsMulti_1, Group_1>) => import("@emotion/react").jsx.JSX.Element;
```

My main reason for wanting to reduce the type annotations besides the verbosity is that the annotations are essentially copying and pasting the type declarations from `react-select`, so they are not adding any value when declared for types that are not the three mentioned above. 

Also the type annotations handle the generics a little differently than the types exported from `react-select`.  I don't think this difference should have an impact, but I think keeping the type api's consistent with the underlying package is valuable if we are able to do so.

ddouglasz (approved):

emmenko:
@ByronDWall yeah, makes sense. Thanks again.

Can you please merge it and release it when you have a chance?

misama-ct (approved):

PR #3064, opened 2025-02-18
fix(DataTableManagerProvider): wrong `onSettingsChange` prop type

ragafus:
#### Summary

In the @commercetools/fe-chapter-rotation, we want to migrate `@commercetools-frontend-extensions/change-history` to React 19.

While updating the `ui-kit` to its latest RC version in https://github.com/commercetools/audit-log/pull/4268 I realized (an error appeared) that the params of `onSettingsChang` are missing.


CarlosCortizasCT (approved):
Look good to me 👍 

Let's see what the First Contact team thinks about it.

PS: remember we will need to take this change to the migration branch, publish a new RC and update it in the consumers.

ddouglasz (approved):
Yes, this change makes sense, it is already existing in the legacy use case but missed out on in the provider method. 
Thank you for catching this @ragafus 🙏🏾

misama-ct (approved):
👍

PR #3090, opened 2025-05-01
feat(filters): support filter tag displaying if filter is applied out…

Sarah4VT:
#### Summary

Allow filter tags to render immediately when new applied filter value passed to Filters component.

## Description

We have a feature where we allow users to [set a “quick filter”](https://commercetools.atlassian.net/browse/PANGOLIN-4556) outside of the Filters component. I am assuming this is a scenario that was not designed for with the new Filters component as when I set the filter, everything works fine, but the filter tag does not appear until something forces the Filter component to fully re-render. I checked the internals of the Filter component and there is some internal state that is what is triggering the tags to get rendered and that internal state isn’t getting updated when we pass in a new set of applied filters. I went ahead and updated this and did a little test in storybook to make sure it worked as I expected and it did. I did a local build and tested in Audit Log and confirmed it fixes it for us.

### DIFF
--- a/.changeset/mighty-doors-sit.md
+++ b/.changeset/mighty-doors-sit.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/filters': minor
+---
+
+Support filter tag appearing if filter is set outside Filter component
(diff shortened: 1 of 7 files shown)


Sarah4VT:
Found an issue with this fix. Investigating now. Put the PR back in draft mode.

Sarah4VT:
Fix has been pushed and this is ready for review now.

ByronDWall (approved):
Thanks for the contribution, looks good to me!

Sarah4VT:
@CarlosCortizasCT I will need this merged into the react-19 branch for ui kit so AL can consume it. I assume your team can do this or would you like me to handle it?

PR #3093, opened 2025-05-07
feat(filters): update total badge to display count when applied visib…

Sarah4VT:
…le count > 0

#### Summary

Show the Total badge when any number of filters are applied, not just more than 1.

## Screenshot

image


misama-ct (approved):
LGTM 🙇‍♂️

kterry1 (approved):
Nice! It does look like percy needs to be updated. @misama-ct Is that something you can do?

ByronDWall (approved):

PR #3110, opened 2025-05-15
chore(dependabot): adjust config for automated sec updates only

valoriecarli:
oh, this is happening.


tylermorrisford (approved):
This appears to be not insane

stephsprinkle (approved):

PR #3115, opened 2025-05-16
chore(dependabot): remove config

valoriecarli:
Due to _reasons_ I'm more than happy to discuss, let's just remove this for now. 
Getting it to work the way we want to with our current infrastructure just isn't happening in a reasonable amount of time. I do think this can be achieved, I am just not sure it's worth the resources.


tylermorrisford (approved):
you go to sleep now lil' robot

PR #3119, opened 2025-05-21
[PANGOLIN-3857] update z-index for filter menu dropdown

kterry1:
#### Summary
The `portal-container` element for a drawer has a `z-index of 10000` and when the `Filter` component is inside a drawer, the menu dropdown is covered as it has a `z-index of 5`. I updated the z-index of the menu styles for the `Filter` component but if I missed another spot that needs updated here please let me know. 

## Description

https://github.com/user-attachments/assets/fbc467a7-8ab1-482a-be09-4af859091199

### DIFF
--- a/packages/components/filters/src/filter-menu/filter-menu.tsx
+++ b/packages/components/filters/src/filter-menu/filter-menu.tsx
@@ -111,7 +111,7 @@ export const menuStyles = css`
   will-change: 'transform, opacity';
   margin-top: ${designTokens.spacing10};
   position: relative;
-  z-index: 5;
+  z-index: 10010;
 `;
 
 export const menuBodyStyle = css`


Sarah4VT:
> #### Summary
> 
> The `portals-container` element for a drawer has a `z-index of 10000` and when the `Filter` component is inside a drawer, the menu dropdown is covered as it has a `z-index of 5`. I updated the z-index of the menu styles for the `Filter` component but if I missed another spot that needs updated here please let me know.
> ## Description
> Screen.Recording.2025-05-21.at.4.04.53.PM.mov

One thing to note is the extra space between the Filter row and the grid in @kterry1 's recording was manually added during debugging to figure out what was causing the menu to not appear.

packages/components/filters/src/filter-menu/filter-menu.tsx
```diff
@@ -111,7 +111,7 @@ export const menuStyles = css`
   will-change: 'transform, opacity';
   margin-top: ${designTokens.spacing10};
   position: relative;
-  z-index: 5;
```
Sarah4VT, line 114: Does this change cover all the popups that appear? For example once you select a filter and if you open a dropdown for the logical operator?

kterry1: Yep! Confirmed in preview branch's storybook that it sets the correct z-index of the root dropdown along with subsequent dropdowns.

https://github.com/user-attachments/assets/4f2656a4-9c06-4256-bd13-20854d6044d4

Sarah4VT (approved):

misama-ct (approved):
Thank you for figuring a solution out on your own, I know this stacking issues are no fun to debug 🙇‍♂️

kterry1:
Thanks @valoriecarli!

PR #3123, opened 2025-05-27
FCT-1500: Adjust legacy CSS reset

misama-ct:
To make sure the nimbus styles take precedence over existing ui-kit styles, the reset styles of ui-kit will have to be wrapped in an @layer directive that nimbus is using as well, otherwise those reset styles will be applied on-top of all nimbus styles, which already lead to links missing their underlines.


misama-ct:
[preview_deployment]

ddouglasz (approved):

PR #3143, opened 2025-06-26
feat(filters): allow auto menu width for filter menu

ddouglasz:
This pull request introduces a new feature to control the width of filter menus dynamically and updates related components to accommodate this functionality. The changes primarily involve adding a new `allowAutoMenuWidth` property and modifying the `menuStyles` function to respect this property.

### New Feature: Dynamic Menu Width

* `packages/components/filters/src/filter-menu/filter-menu.tsx`:
  - Added a new `allowAutoMenuWidth` property to the `TFilterMenuProps` type, allowing the menu width to adjust automatically.
  - Updated the `menuStyles` function to conditionally set the menu width to `auto` if `allowAutoMenuWidth` is true.
  - Modified the `FilterMenu` component to pass the `props` object to the `menuStyles` function for dynamic styling.

* `packages/components/filters/src/filters.tsx`:
  - Added the `allowAutoMenuWidth` property to the `TFilterConfiguration` type to propagate this feature through the filter configuration.
  - Passed the `allowAutoMenuWidth` property to the `FilterMenu` component in the `Filters` component to enable dynamic width control.
  - Ensured the default behavior of the menu width remains consistent by explicitly setting `allowAutoMenuWidth: false` in certain contexts.

### DIFF
--- a/packages/components/filters/src/filter-menu/filter-menu.tsx
+++ b/packages/components/filters/src/filter-menu/filter-menu.tsx
@@ -96,7 +96,7 @@ export type TFilterMenuProps = {
   /**
    * controls whether menu is expanded
    */
-  allowAutoMenuWidth?: boolean;
+  hasWideMenu?: boolean;
 };
 
 export const menuStyles = (props: Partial<TFilterMenuProps> = {}) => css`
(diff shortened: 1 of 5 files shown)


packages/components/filters/src/filter-menu/filter-menu.tsx
```diff
@@ -93,14 +93,18 @@ export type TFilterMenuProps = {
...
 };
 
-export const menuStyles = css`
+export const menuStyles = (props: Partial<TFilterMenuProps> = {}) => css`
   display: flex;
   flex-direction: column;
   align-items: flex-start;
   gap: ${designTokens.spacing30};
-  width: ${designTokens.constraint7};
+  width: ${props.allowAutoMenuWidth ? 'auto' : designTokens.constraint7};
```
ddouglasz, line 107: Using the `min-width` would have some weird effect on the menu when even in default cases, causing the width to be larger in menus with smaller(width) menu items .  Two options, either reduce the min-width to a smaller value or allow users set a value to change width to auto (`allowAutoMenuWidth`).
image

ByronDWall (approved):

PR #3147, opened 2025-06-30
Add missing prop in filter prop list storybook documentation

ddouglasz:
add missing prop in filter prop list storybook documentation

### DIFF
--- a/.changeset/two-lizards-stay.md
+++ b/.changeset/two-lizards-stay.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/filters': patch
+---
+
+add missing prop in filter prop list storybook documentation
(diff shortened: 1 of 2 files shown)


.changeset/two-lizards-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-uikit/filters': patch
+---
+
+add missing prop in filter prop list storybook documentation
```
ByronDWall, line 5: A change to the storybook does not necessitate a new release of the component does it?  I was under the impression that changesets are only necessary for changes that are included in the published packages.  Lmk if I'm wrong on that tho

ByronDWall (approved):
Thanks for noticing this and updating!!  One possible nit/question about the changeset, but nothing that would stop me from approving

.changeset/two-lizards-stay.md
ddouglasz: Correct! Sorry, merged before seeing the comment.

PR #3150, opened 2025-07-02
Fix DataTable cell alignment props not working

nima-ct:
#### Summary

This PR fixes DataTable cell alignment for row cell content.

We observed in list pages across the MC that `DataTable` cells aren’t vertically centered when there’s a mix of text, icons, or buttons.

You can also see this directly in Storybook, where changing vertical or horizontal alignment props has no effect.
https://uikit.commercetools.com/?path=/docs/components-datatable--props


misama-ct (approved):
Good catch, thanks for fixing.

PR #3223, opened 2026-03-13
FEC-365: Extend UI-kit MoneyInput with Non-Iso Currencies

nima-ct:
#### Summary

This PR adds support for non-ISO currency variants (HUF0, KZT0,...) in the currency lookup table from `MoneyInput`

This change is needed so that non-ISO currencies are correctly recognized and handled in the Merchant Center money inputs.

https://commercetools.atlassian.net/browse/FEC-354

### DIFF
--- a/packages/components/inputs/money-input/src/money-input.tsx
+++ b/packages/components/inputs/money-input/src/money-input.tsx
@@ -379,7 +379,7 @@ const createEmptyMoneyValue = (currencyCode: TCurrencyCode): TMoneyValue => ({
   type: 'centPrecision',
   currencyCode,
   centAmount: NaN,
-  fractionDigits: 2,
+  fractionDigits: allCurrencies[currencyCode].fractionDigits,
 });
 
 const getAmountAsNumberFromMoneyValue = (moneyValue: TMoneyValue) =>
--- a/packages/components/inputs/money-input/src/money-input.tsx
+++ b/packages/components/inputs/money-input/src/money-input.tsx
@@ -379,7 +379,7 @@ const createEmptyMoneyValue = (currencyCode: TCurrencyCode): TMoneyValue => ({
   type: 'centPrecision',
   currencyCode,
   centAmount: NaN,
-  fractionDigits: allCurrencies[currencyCode].fractionDigits,
+  fractionDigits: allCurrencies[currencyCode].fractionDigits ?? 2,
 });
 
 const getAmountAsNumberFromMoneyValue = (moneyValue: TMoneyValue) =>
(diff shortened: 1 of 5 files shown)


packages/components/inputs/money-input/src/money-input.tsx
```diff
@@ -379,7 +379,7 @@ const createEmptyMoneyValue = (currencyCode: TCurrencyCode): TMoneyValue => ({
   type: 'centPrecision',
   currencyCode,
   centAmount: NaN,
-  fractionDigits: allCurrencies[currencyCode].fractionDigits,
+  fractionDigits: allCurrencies[currencyCode].fractionDigits ?? 2,
```
ByronDWall, line 382: It's more verbose, but I think it would make sense to get the CLDR fraction digits for a currency code, I think the error state if the currency code isn't ISO standard when it has no fractionDigits would be very helpful.  Also, I could see future issues if we naively default to 2.

```suggestion
  let fractionDigits = allCurrencies[currencyCode].fractionDigits;
  if(!fractionDigits){
    const formatter = new Intl.NumberFormat(locale, {                                                                                            
      style: "currency",                                                                                                                          
      currencyCode                                                                                                                            
    });                                                                                                                                           
    // KRW → 0, USD → 2, BHD → 3                                                                                                  
    const { minimumFractionDigits } = formatter.resolvedOptions(); 
    let fractionDigits = minimumFractionDigits; 
  }
  const createEmptyMoneyValue = (currencyCode: TCurrencyCode): TMoneyValue => ({
    type: 'centPrecision',
    currencyCode,
    centAmount: NaN,
    fractionDigits
  });
    
```

ByronDWall (approved):
This is a good safeguard and I'm going to approve, but I think it's worth considering a less naive fallback for the `fractionDigits` than always defaulting to `2`.

nima-ct:
> This is a good safeguard and I'm going to approve, but I think it's worth considering a less naive fallback for the `fractionDigits` than always defaulting to `2`.

@ByronDWall Thx for pointing this out
Looking closer, I see that the component assumes in several other places known currency codes that can only be resolved via the lookup table.
I think addressing it in one spot won’t solve the overall issue. Unknown currency would throw in other places.
A proper solution would require a broader refactor. Let me know what you think.

misama-ct (approved):

packages/components/inputs/money-input/src/money-input.tsx
misama-ct: It does make sense, I mean it changes _a_ default to the _correct_ default, but it also changes existing behavior. Personally I can't foresee possible side-effects, but I can imagine there are some so this is probably better tackled in its own ticket.

PR #3224, opened 2026-03-13
CRAFT [FEC] -2082 - Update to appkit v27/ eslint9

valoriecarli:
Migrate to ESLint 9 flat config
                                                                                                                                                                                                                
Migrates the root workspace from ESLint 8 legacy config to ESLint 9 flat config, following the migration guide shipped with @commercetools-frontend/eslint-config-mc-app@27.0.0.                            

  Changes:
  - Replaced `.eslintrc.js` + `eslintignore` with `eslint.config.js` (ignore patterns inlined, overrides converted to flat config objects)
  - Bumped `eslint` 8.x → ^9.0.0 and `eslint-config-mc-app` to 27.0.0
  - Removed cliOptions.rules from `jest-runner-eslint.config.js` - not supported in flat config mode; both rules were already covered by the base config
  - Added globals.jest for *`*/*.visualspec.js` and  `**/*.bundlespec.js` - the base config only injects Jest globals for 
 `*.{spec,test}.*`
  - Removed stale @typescript-eslint/no-empty-function eslint-disable comments in test/setup-tests.js  - @typescript-eslint is not registered for .js files in flat config, making those comments invalid


eslint.config.js
```diff
@@ -0,0 +1,85 @@
+process.env.ENABLE_NEW_JSX_TRANSFORM = 'true';
+
+const globals = require('globals');
+const mcAppConfig = require('@commercetools-frontend/eslint-config-mc-app');
+
+module.exports = [
+  // Ignores replace .eslintignore (directory patterns end with /)
```
misama-ct: I tried reading this and got a stroke, can you rephrase this?

ByronDWall (approved):
Lints good, looks good, tests good. 

Only thing to note is that `jest-runner-eslint v2.3.0` declares `"eslint": "^7 || ^8"` as a peer dependency, not `^9`. The actual code is compatible but yarn may print a peer dep warning in CI.

@ByronDWall ah good call. i caught that in some, but clearly not here. on it. thanks.

ByronDWall (approved):

misama-ct (approved):
LGTM 💐

eslint.config.js
valoriecarli: @misama-ct ofc, but only because you asked so nicely.

valoriecarli:
> Lints good, looks good, tests good.
> 
> Only thing to note is that `jest-runner-eslint v2.3.0` declares `"eslint": "^7 || ^8"` as a peer dependency, not `^9`. The actual code is compatible but yarn may print a peer dep warning in CI.
> 
> @ByronDWall ah good call. i caught that in some, but clearly not here. on it. thanks.

I got ahead of myself, but I will leave this for now.
Regarding this stale metadata, Yarn's `packageExtensions` can't override an existing peer dep range ( only add missing ones), so there's no clean way to silence the warning without either patching the package or suppressing it via `logFilters`. I didn't care for that. Ideally [jest-runner-eslint](https://www.npmjs.com/package/jest-runner-eslint?activeTab=readme) ships a v9-aware release soon.

eslint.config.js
valoriecarli: this has been removed. deemed unnecessary.

PR #3228, opened 2026-04-01
refactor: migrate from Yarn 3.8.7 to pnpm

misama-ct:
> [!CAUTION]
> This PR was created to try out the freshly added AI Agent tooling. Do not merge without thorough review and CI validation.

## Summary

- Replace Yarn 3.8.7 with pnpm 10.11.0 as the package manager
- Remove all Yarn artifacts (`.yarnrc.yml`, `.yarn/`, `yarn.lock`, `constraints.pro`)
- Add pnpm equivalents (`pnpm-workspace.yaml`, `.npmrc`, `pnpm-lock.yaml`)
- Update root `package.json` (`packageManager`, `engines`, `pnpm.overrides`, `pnpm.onlyBuiltDependencies`, scripts)
- Migrate all shell scripts, git hooks, CI workflows, config files, and documentation from `yarn` to `pnpm`
- Replace `manypkg run/exec` with native `pnpm --filter` / `pnpm -r exec`
- Replace Yarn's Prolog-based `constraints.pro` with a Node.js validation script
- Fix phantom dependency (`yaml` in design-system) exposed by pnpm's stricter isolation
- Fix pre-existing incorrect deep import (`@commercetools-uikit/constraints/src`)
- Regenerate all icon components with updated template comment

## Known trade-offs

- `shamefully-hoist=true` is enabled in `.npmrc` to match Yarn's flat hoisting behavior — many cross-workspace type references rely on it. Can be tightened later with targeted `public-hoist-pattern[]` entries.
- `pnpm.onlyBuiltDependencies` allowlist may need tuning if build script warnings persist in CI.

## Test plan

- [x] `pnpm install` succeeds
- [x] `pnpm build` succeeds
- [x] `pnpm typecheck` passes (0 errors)
- [x] `pnpm test` passes (1402 tests, 125 suites)
- [x] `pnpm lint` passes (1307 files)
- [x] `node scripts/check-workspace-constraints.js` passes
- [x] No `yarn` references in source files
- [ ] CI pipeline passes
- [ ] Changesets publish flow works
- [ ] Percy visual regression tests pass

---

## Review Navigator

**270 files changed** across 7 commits; ~25 are substantive, ~38 are README `yarn add` → `pnpm add` replacements, ~17 are regenerated icon comment updates, and ~20 are AI tooling files from the base branch (not part of the migration).

> Review in the logical chunks below — start with the core config (commits 1-2), then CI, then spot-check the bulk changes.

### File Relationships & Review Order

```mermaid
[...]

PR #3239, opened 2026-05-13
refactor(FEC-924): migrate from Yarn 3 to pnpm

misama-ct:
## TL;DR

Migrates `commercetools/ui-kit` from **Yarn 3.8.7** to **pnpm 10.33.4**, aligning the repository with the rest of the MCF organisation. No public API change, no consumer-facing behaviour change; ships as a minor bump in the 20.x line.

Tracks Jira [FEC-924]. Follow-up to closed PR #3228 (April spike) — same mechanical migration, now coordinated end-to-end: CI green, Percy approved, byte-parity documented, canary validated against `merchant-center-application-kit`.

## Why

`ui-kit` was the org's last frontend repo on Yarn 3. The original spike (#3228) proved the migration was technically feasible but closed without merging because none of the gating signals (CI green end-to-end, Changesets publish flow, Percy, internal-consumer canary) had been exercised. This PR is the coordinated execution.

## What changes

| Yarn surface                                         | pnpm equivalent                                                                                                        |
| ---------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| `yarn.lock`                                          | `pnpm-lock.yaml`                                                                                                       |
| `.yarnrc.yml` + `.yarn/`                             | `pnpm-workspace.yaml` + `.npmrc` (`shamefully-hoist=true`)                                                             |
| `packageManager: yarn@3.8.7` + `engines.yarn`        | `packageManager: pnpm@10.33.4` (no `engines.yarn`)                                                                     |
| `constraints.pro` (Prolog rules)                     | `scripts/check-workspace-constraints.js` (same checks: license / repository / publishConfig)                           |
| `resolutions` (37 entries, yarn-specific key syntax) | `pnpm.overrides` (ported verbatim; pnpm accepts the same `package@range` syntax)                                       |
| `yarn workspace …` / `manypkg run` / `manypkg exec`  | `pnpm --filter … run …` / `pnpm -r exec …`                                                                             |
| `yarn-plugin-postinstall` fork                       | npm `postinstall` lifecycle hook wired to `./scripts/postinstall.sh`                                                   |
| GitHub Actions: `actions/setup-node` + yarn cache    | `pnpm/action-setup@v4` + setup-node@v6 (`package-manager-cache: false`) + explicit `actions/cache@v5` for `pnpm store` |
| Vercel preview: default installer                    | `installCommand` / `buildCommand` overridden, `ENABLE_EXPERIMENTAL_COREPACK=1`                                         |

`shamefully-hoist=true` is deliberate, per FEC-924's "Resolved decisions" — tightening it to targeted `public-hoist-pattern[]` is a separate ticket. The publint gate below mitigates the risk that comes with it.

## What surfaced along the way

The mechanical conversion is well-trodden ground. The interesting work was the set of failures encountered on the way to green CI and a working canary. Each item below is a one-time discovery; none of them changes anything consumers can observe:

- **Puppeteer's Chrome download gets silently skipped.** pnpm's `side-effects-cache` caches postinstall side-effects in the store. When CI restores the store from the GitHub Actions cache, pnpm skips re-running puppeteer's postinstall — but Chrome lives outside `node_modules` / the pnpm store, so the binary is missing on the runner. Yarn's eager postinstall masked this. Resolved by an explicit `pnpm exec puppeteer browsers install chrome` step before the VRT job.

- **`design-system` had three phantom devDependencies** (`prettier`, `rcfile`, `yaml`) used by its build scripts but never declared on the package. Yarn's hoisted layout resolved them transitively; pnpm's strict isolation does not. Declared explicitly. The same risk class — for **publishable** packages — motivated the publint gate below.

- **One phantom import in a story file:** `packages/components/dropdowns/dropdown-menu/src/dropdown-menu.stories.tsx` imported `@commercetools-uikit/constraints/src` (a deep import into the source folder rather than the published entry point). Surfaced by pnpm's stricter isolation, unrelated to the lockfile flip. Normalised to the top-level export.

- **`setup-node@v6` + `registry-url:` + the runner's pre-installed `yarn 1.22.22` is a footgun.** v6's new `package-manager-cache: true` default auto-detects the package manager; combined with `registry-url`, it picks up the runner's pre-installed yarn 1 instead of pnpm and aborts at `yarn config get cacheFolder` before any of the workflow's own steps run. The error message yarn-1 emits — `"packageManager": "yarn@pnpm@10.33.4"` — is a yarn-1 message-construction quirk, not a malformed field. Resolved by `package-manager-cache: false` on the affected workflows; the pnpm store cache is managed explicitly via `actions/cache@v5` below the setup-node step.

- **`@babel/runtime` resolves to a newer patch under a fresh pnpm install** (`7.27/7.28` → `7.29`, both within the existing `^7.20.13` range). Documented in the byte-parity check below; output is strictly smaller / more modern but semantically identical. Normal lockfile-refresh effect, not migration-specific.

- **The `[preview_deployment]` PR comment couldn't run against this branch pre-merge.** GitHub resolves `uses: ./.github/workflows/...` references against the _default branch_ on `issue_comment` events. So `[preview_deployment]` was always going to load main's (still-yarn) workflow and run it against this branch's pnpm-only code — guaranteed to fail. Resolved by landing the pnpm version of `preview-release-on-comment.yml` as a precursor PR (#3240) on main first, then triggering `[preview_deployment]` against this branch. Once this PR merges, the precursor's content is just part of main; no orphan state.

- **`patch-package` silently doesn't apply under pnpm**, taking `pnpm generate-readmes` down with it. patch-package writes to `node_modules/<pkg>/`; under pnpm's strict layout that path is a symlink into `.pnpm/<pkg>@<ver>/node_modules/<pkg>/`, and patch-package doesn't follow the symlink to patch the canonical file. The lone patch in this repo (`typescript-react-function-component-props-handler+1.1.1.patch` — an `Array.isArray` guard before `.params.length` inside `react-docgen`'s prop extractor) therefore never applied, and `pnpm generate-readmes` threw `TypeError: Cannot read properties of undefined (reading 'length')`. Resolved by migrating to native `pnpm patch` — patches are now wired via `pnpm.patchedDependencies` and applied directly into the `.pnpm` store at install time. `patch-package` removed from devDependencies, postinstall step trimmed. `pnpm generate-readmes` now runs to completion.

## Validation evidence

Three independent angles, summarised below; detailed artefacts (per-package tarballs, extracted trees, `diff -r` outputs, publint snapshots) are preserved locally for inspection.

**publint parity vs the npm registry.** Pulled the published `20.5.0` tarball of every non-private workspace package, packed every package locally on this branch with pnpm, diffed normalised `publint --pack pnpm` output. **97/97 packages identical.** Zero migration-introduced findings. `shamefully-hoist=true` did not let any undeclared import slip through. The new `pnpm lint:publint` CI step locks this property in going forward.

**`npm pack` byte-parity vs main.** Built `main` in a worktree under yarn 3.8.7, packed every non-private package, packed the same set on this branch under pnpm 10.33.4, extracted both, ran `diff -r` on the `package/` directories.

- 9 packages **byte-identical**: aggregator / re-export packages unaffected by babel-runtime resolution.
- 88 packages **differ**, in well-understood categories:

  - **11**: README only — install snippet `yarn add` → `pnpm add` (deliberate).
  - **16**: Bundle only — newer `@babel/runtime[-corejs3]` patch (`7.27/7.28` → `7.29`, range `^7.20.13` already permits this). Concretely: `_taggedTemplateLiteral` helper + `_templateObject` declarations gone; `concat` polyfill replaced by template literals; internal `_context` variable numbering shifts. Strictly smaller, more modern output; semantically identical.
  - **59**: README + bundle (Class A ∪ Class B).
  - **1** (`design-system`): `build:tokens:watch` script wording (dev-only, never executed by consumers) + the three declared `devDependencies` mentioned above (not installed by consumers).
  - **1** (`data-table-manager`): content-derived chunk-hash rename, content delta itself is Class B.

  No code-generation drift, no API or runtime behaviour change, no unintended file additions / removals.

**`merchant-center-application-kit` canary CI.** PR [commercetools/merchant-center-application-kit#4008](https://github.com/commercetools/merchant-center-application-kit/pull/4008) pinned every `@commercetools-uikit/*` dep — direct and transitive — to the published snapshot `0.0.0-migration-pnpm-20260513114959` (tag `migration-pnpm`) via `pnpm.overrides`. Required because the `0.0.0-` prefix doesn't satisfy `^20.4.0` semver ranges; the override resolves the whole dep graph to one consistent canary version. **All 20+ checks green** on the first run: `lint_and_test`, `test_visual`, all 4 starter-template tests (js/ts × custom-app/custom-view) plus their `_installation` variants, `test_playground`, both Percy builds, all 4 Orca scans, both Vercel previews. **Percy reported zero visual changes** between current `20.5.0` and the migration snapshot — confirms the byte-parity prediction that the babel-runtime patch bump is semantically identical at the rendered-output level. PR closed after validation.

## Local + CI validation

- [x] `pnpm install` / `pnpm build` / `pnpm typecheck` / `pnpm test` (125 suites, 1402 tests) / `pnpm lint` (1307 files)
- [x] `pnpm lint:publint` — 97 published packages, zero publint errors
- [x] `node scripts/check-workspace-constraints.js`
- [x] `rg '\byarn\b'` returns **zero** hits in source / scripts / CI (excluding `.yarn/`, `node_modules`, `CHANGELOG*`, `pnpm-lock.yaml`)
- [x] GitHub Actions main workflow — green end-to-end, no suppressions
- [x] Percy — approved by the PR author
- [x] Vercel preview deploy — green
- [x] Changesets snapshot publish (`migration-pnpm` tag) — green
- [x] mc-app-kit canary CI — green, zero Percy diff

## Commit structure & merge strategy

Seven commits, organised by topic so reviewers can read in chunks:

| #   | Commit                                                                                                                                             | Notes                                                                                                                                                                                                                                                                                     |
| --- | -------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | [`0f3eb87`](https://github.com/commercetools/ui-kit/pull/3239/commits/0f3eb87e9) `chore: bulk text replace yarn → pnpm in docs and tooling`        | Mechanical `s/yarn/pnpm/g` across 95 files (READMEs, generator templates, jest configs, lint-staged, svgr). Low cognitive load.                                                                                                                                                           |
| 2   | [`5390457`](https://github.com/commercetools/ui-kit/pull/3239/commits/53904575b) `refactor: replace yarn tooling with pnpm at the workspace level` | The heart of the migration: lockfile, workspace config, packageManager pin, 37 `resolutions` → `pnpm.overrides`, `constraints.pro` → JS validator, root + per-workspace `package.json`, shell scripts, husky, phantom-import fix, includes the `immutable@^4.3.8` pin for CVE-2026-29063. |
| 3   | [`2416423`](https://github.com/commercetools/ui-kit/pull/3239/commits/2416423c9) `ci: migrate GitHub Actions workflows and Vercel preview to pnpm` | `main.yml` + `publish-release.yml` + `vercel.json`, including the setup-node@v6 fix and the Puppeteer Chrome install step.                                                                                                                                                                |
| 4   | [`c42e11f`](https://github.com/commercetools/ui-kit/pull/3239/commits/c42e11fb3) `feat(ci): gate phantom-dep regressions via publint`              | `scripts/check-publint.js` + the CI step calling it.                                                                                                                                                                                                                                      |
| 5   | [`c465531`](https://github.com/commercetools/ui-kit/pull/3239/commits/c465531d2) `chore: regenerate icons`                                         | 161 generated-icon files, output of `pnpm generate-icons`.                                                                                                                                                                                                                                |
| 6   | [`a7d16a3`](https://github.com/commercetools/ui-kit/pull/3239/commits/a7d16a304) `chore(changeset): pnpm migration`                                | Anchor changeset for the fixed-version group.                                                                                                                                                                                                                                             |
| 7   | [`06b4e58`](https://github.com/commercetools/ui-kit/pull/3239/commits/06b4e5838) `refactor: migrate patch-package to native pnpm patch`            | Replaces the silently-broken `patch-package` flow with `pnpm.patchedDependencies`; restores `pnpm generate-readmes` as a side effect.                                                                                                                                                     |

**Merge strategy: squash-merge** per the FEC-924 plan. The chunking exists so reviewers can read the migration in pieces; the final history on `main` is one commit.

## Out of scope (follow-up tickets)

- **Tighten hoisting**: replace `shamefully-hoist=true` with targeted `public-hoist-pattern[]`.
- **Pre-existing publint warnings/suggestions**: per-package findings about `pkg.exports`, `pkg.type`, `pkg.engines.node`, `.esm.js` interpreted as CJS, `repository.url` formatting. Identical on yarn-built main and pnpm-built migration; CI gate doesn't fail on them.
- **Regenerate stale package READMEs**: the docgen-based generator now runs again (see commit 7 above). Running it produces real prop-table updates in ~10 components — out of scope here, separate maintenance task.
- **pnpm catalogs** adoption.
- **`bundlewatch` vs `bundlesize`** reconciliation.

## Refs

- Jira: [FEC-924]
- Precedent: #3228 (April spike, closed without merge)
- Precursor: #3240 (preview-release workflow → pnpm, merged ahead of this PR)
- Canary: [commercetools/merchant-center-application-kit#4008](https://github.com/commercetools/merchant-center-application-kit/pull/4008) (closed after validation)
- Snapshot tag: `migration-pnpm` (resolves to `0.0.0-migration-pnpm-20260513114959`)
- Org precedent: `merchant-center-application-kit` migrated to pnpm in PR #2976 (May 2023)

[FEC-924]: https://commercetools.atlassian.net/browse/FEC-924


misama-ct:
[preview_deployment]

misama-ct:
[preview_deployment]

misama-ct:
[preview_deployment]

.npmrc
```diff
@@ -0,0 +1,3 @@
+# Hoist all dependencies to root node_modules (matches previous Yarn behavior).
+# TODO: tighten with public-hoist-pattern[] for specific packages later.
+shamefully-hoist=true
```
tdeekens, line 3: I think it's good to not mix this in.

.changeset/pnpm-migration.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-uikit/design-system": minor
```
tdeekens, line 29: Would it make sense to public every package as a fix version?

tdeekens (approved):

.changeset/pnpm-migration.md
misama-ct: it does — [`5563e55`](https://github.com/commercetools/ui-kit/pull/3239/commits/5563e5568a5c6f320edf3d5ee2d0836a1afc22d6)

scripts/check-workspace-constraints.js
ByronDWall, line 1: The old constraints.pro had 4 rules. The replacement scripts/check-workspace-constraints.js only implements 3. The missing rule:
```
% Enforces that a dependency doesn't appear in both `dependencies` and `devDependencies`
  gen_enforced_dependency(WorkspaceCwd, DependencyIdent, null, 'devDependencies') :-
    workspace_has_dependency(WorkspaceCwd, DependencyIdent, _, 'devDependencies'),
    workspace_has_dependency(WorkspaceCwd, DependencyIdent, _, 'dependencies').
```
This constraint prevents the same package from appearing in both dependencies and devDependencies of a single workspace — a real correctness guard. It should be added to the new script.

ByronDWall (approved):
Overall PR looks really good. Just one comment about recreating the last rule from `constraints.pro`, but that concern isn't enough to also stop me from approving.

scripts/check-workspace-constraints.js
misama-ct: Fixed in [`46665f5`](https://github.com/commercetools/ui-kit/commit/46665f5011973f22a5ccf02bb3666c6db9350d76)

PR #3245, opened 2026-05-18
refactor(FEC-938): ui-kit post-pnpm tooling polish (workspace:^, bundlewatch, catalogs)

misama-ct:
## Summary

Follow-up to FEC-924 (Yarn 3 → pnpm migration). Closes the post-migration tooling polish in one PR.

**[FEC-938](https://commercetools.atlassian.net/browse/FEC-938)**

### 1. Internal deps switched to `workspace:^`

Rewrote 706 internal `@commercetools-uikit/*` and `@commercetools-frontend/ui-kit` specifiers from concrete versions to `workspace:^` across 97 workspaces. The `.npmrc` workaround from PR #3243 (`link-workspace-packages=true` / `prefer-workspace-packages=true`) is no longer needed and was dropped. Lockfile collapses ~5500 lines as duplicated per-internal-package tarball trees become workspace links.

### 2. Bundle-size script renamed

Only `bundlewatch` is installed; the `bundlesize` script name was a vestige. Renamed across `package.json`, the config file (`bundlesize.config.json` → `bundlewatch.config.json`), and the CI workflow.

### 3. External deps consolidated under `pnpm-workspace.yaml`

`pnpm-workspace.yaml` is now the single source of truth for every external dep version — 160 install pins + 8 peer-compatibility ranges. The earlier audit's "version drift" cases (react / react-intl / react-router-dom / react-dom / moment) turned out to be install-pin vs peer-range section differences, not real drift; the two catalog axes (install pin vs peer range) cleanly model both.

### 4. Catalogs organized by toolchain role

Install pins are split across 7 named catalogs + a default catalog for singletons. Grouping rule: deps that must (or strongly want to) move together share a catalog.

| Catalog       | Entries | Scope                                                                     |
|---------------|--------:|---------------------------------------------------------------------------|
| `react`       |      24 | React framework + React-bound libraries                                   |
| `build`       |      27 | compile/bundle/style toolchain (babel, vite, svgr, emotion, postcss, ts)  |
| `test`        |      20 | jest, testing library, percy, puppeteer, jsdom                            |
| `rich-text`   |      16 | slate + remark/unified/vfile (rich-text editor stack)                     |
| `repo`        |      12 | monorepo plumbing & release tooling (changesets, manypkg, husky, …)       |
| `lint`        |       8 | code-quality enforcers (eslint, stylelint, prettier, typescript-eslint)   |
| `storybook`   |       6 | storybook dev environment (must move in lockstep)                         |
| `catalog:`    |      47 | leaf utilities / singletons with no cohort partner                        |
| `peer`        |       8 | consumer compatibility ranges for public packages                         |

Entries within each named catalog are grouped by sub-cohort (e.g. `react` splits into core / @types / React Router / i18n / React-bound utilities) with `#` header comments and blank-line separators — visual aid for maintainers adding deps. The default catalog stays alphabetical since its members are singletons by definition.

### 5. Validator enforces all of the above

`scripts/check-workspace-constraints.js` runs in `postinstall` and CI. In addition to the prior per-package metadata checks it now enforces:

- Every dep in the default `catalog:` must be consumed via `catalog:` in `dependencies` / `devDependencies`.
- Every dep in a named catalog `catalogs.<name>` must be consumed via `catalog:<name>`.
- Every dep in `catalogs.peer` must be consumed via `catalog:peer` in `peerDependencies`.
- Literal versions on any cataloged dep are an error.
- Drift: an uncataloged external dep used at two or more distinct specifiers across workspaces is an error — the fix is to add it to a catalog so the version is centrally controlled. Install (deps + devDeps) and peer drift are checked separately.

The YAML parser tolerates blank lines and `#` comment-only lines anywhere inside a catalog block, so the sub-cohort layout doesn't trip enforcement.

### 6. Changeset

Patch-level changeset on `@commercetools-frontend/ui-kit` so the next release exercises the production flow with `workspace:^` in place. The `fixed` group propagates the bump.

## Why this resolves PR #3236

`manypkg check` in `postinstall` was failing on the Version Packages PR with 5 specifier-drift errors on the root `ui-kit` workspace. `changeset version` bumps workspace versions and cross-workspace specifiers in *public* workspaces, but the root is `private: true` and was skipped — leaving its 5 internal deps pinned at the prior release. With `workspace:^`, the root auto-tracks each workspace's `version`, so the bump no longer creates drift. PR #3243's `.npmrc` workaround only covered install-time resolution; manypkg's static check needed the structural fix.

## Test plan

- [x] CI green: install / typecheck / lint / test / publint / constraints / bundlewatch
- [x] Percy: subpixel anti-aliasing diff noted on first run, confirmed non-issue
- [x] Changesets canary publishes successfully when merged
- [x] Reviewer sanity-checks `pnpm-workspace.yaml` (single source of truth) and `scripts/check-workspace-constraints.js` (default + 7 named + peer enforcement + drift detection)

## Out of scope (follow-ups)

- Publint warnings (FEC-936)

## Bonus points

Two tidy-ups landed on top of the main work:

### Centralized pnpm config in `pnpm-workspace.yaml`

`pnpm.overrides` (37 Dependabot/security pins), `pnpm.onlyBuiltDependencies` (postinstall allow-list), and `pnpm.patchedDependencies` (one `patches/` entry) moved from root `package.json` → `pnpm-workspace.yaml`, alongside the catalogs. All entries ported verbatim including yarn-style `package@range` keys (`path-to-regexp@^1.7.0`, `picomatch@^2.2.1`, etc.). `pnpm install` reports "Lockfile is up to date" — pnpm 10 honors all three fields from `pnpm-workspace.yaml` with a byte-identical resolution graph.

### Cyclic workspace dependencies eliminated

`pnpm install` was emitting two `cyclic workspace dependencies` warnings. Both resolved without any structural restructuring.

**SCC A** (`design-system`, `hooks`, `presets/ui-kit`, `presets/fields`, `date-field`, `calendar-utils`) traced to a single back-edge: `@commercetools-uikit/hooks` declared `@commercetools-frontend/ui-kit` as a `devDependency` solely for `Spacings.Stack` + `PrimaryButton` scaffolding in two spec files. Swapped for native `<button aria-label=...>` (the specs test hooks, not UI Kit components) and dropped the devDep.

**SCC B** (`select-input` ↔ `select-utils`, originally surfaced as `checkbox-input` ↔ `select-utils`) had two contributing edges: `checkbox-input → select-utils` was a phantom dep with no source/test reference, and `select-utils → select-input` was a single type-only import (`TSelectInputProps['appearance']`, a 3-value string literal). Dropped the phantom dep and inlined the literal — `select-utils` no longer depends on `select-input` at all.

`pnpm install` no longer emits a `WARN  There are cyclic workspace dependencies` line.

[FEC-938]: https://commercetools.atlassian.net/browse/FEC-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


misama-ct:
[preview_deployment]

misama-ct:
> Release workflow succeeded ✅\nSee details: [Workflow Run](https://github.com/commercetools/ui-kit/actions/runs/26032912837)

YASS

My work here is done. 💅 💃

tdeekens (approved):
Good stuff!

tylermorrisford (approved):

ByronDWall (approved):

pnpm-workspace.yaml
```diff
@@ -11,3 +11,336 @@ packages:
   - packages/components/spacings/*
   - presets/*
   - visual-testing-app
+
```
ByronDWall, line 14: @misama-ct what are your thoughts on implementing a `catalog` in all our repos?  To me it seems worth it, if for no other reason than it gives us a single file in which to change package versions instead of having to change versions in each lib/app's `package.json`

misama-ct: Yes, that would be the next logical step. The catalogs drastically simplify maintenance concerns. I initially wanted to do the mc first, as it would have the best bangs-for-bucks ratio, but I think starting with the satellites makes more sense. The catalogs can also be used as upgrade-cohorts for dependabot with minimal additional configuration.

PR #3247, opened 2026-05-19
debug: jest --detectOpenHandles to find CI leak

misama-ct:
Throwaway debug PR. Adds `--detectOpenHandles` to the test step so the Linux CI runner reveals what handles are leaking (not reproducible on macOS). Once we know the culprit, this PR gets closed and the real fix goes into #3246.


misama-ct:
Throwaway debug PR — revealed the real failure (jest-environment-node hoisting conflict). Fix landed on #3246.

PR #3252, opened 2026-05-20
refactor(FEC-936): publint metadata cleanup (engines, type, repository.url)

misama-ct:
## TL;DR

Closes 3 of the 4 publint findings on all 97 published packages — pure metadata, no consumer-API or build-output change. The 4th finding (`pkg.exports`) is not being tackled here because the only working paths to fix it carry real breaking-change risk for downstream consumers (subpath imports, TypeScript type resolution, module resolution), and ui-kit is approaching EOL — consumers asked to spend cycles fixing breakage on a deprecated package would reasonably resent that instead of investing in the Nimbus migration.

## Summary

Closes [FEC-936](https://commercetools.atlassian.net/browse/FEC-936) (partial). Four commits, one per finding for review, plus a `docs:` commit that rewrites the changeset to follow the Nimbus changeset conventions (consumer-facing, not internal):

1. **`engines.node` floor `>=22`** declared on every published package. Runtime floor decoupled from the build env (`>=24` at root).
2. **`fix: preserve engines.node` in the package-json generator template** — the pre-commit hook regenerates package.json from a hardcoded template; without this fix the engines field would have been stripped back out on the next unrelated commit.
3. **`repository.url` normalization** — `git+https://…` everywhere; `REPO_URL` constant in the constraints script flipped in lockstep.
4. **`type: "commonjs"`** declared on every published package, matching what Preconstruct's `.cjs.js` / `.esm.js` output already required Node to assume.

`scripts/check-workspace-constraints.js` gains rules for both `engines.node = PUBLISHED_NODE_FLOOR` and `type = "commonjs"` so future packages can't drop them silently.

## Why `pkg.exports` is not in this PR

The cost isn't engineering effort — with AI in the loop the work itself is cheap. The cost is **risk to consumers we can't fully survey**:

- Adding an `exports` field via Preconstruct's stable config silently degrades TypeScript types for any consumer using `moduleResolution: "Bundler"` or `"NodeNext"`. TS walks the conditions, picks `module → .esm.js`, finds no sibling `.esm.d.ts`, and drops types to `any`. Preconstruct's experimental `importsConditions` flag emits a `types` condition but the `.d.ts` re-export chain still breaks under a real `pnpm build`.
- Deep-import surveys across the maintained/* repos already turned up consumer patterns we had to preserve (`@commercetools-uikit/i18n/compiled-data/*.json`, `@commercetools-uikit/design-system/materials/*.css`, plus a trailing-slash typo in mc-frontend). There are almost certainly more we haven't seen.
- ui-kit is approaching EOL — Nimbus is the successor. Shipping a non-trivial breaking-change risk on a sunsetting package is a bad trade: consumers reasonably expect a stable runway through migration, not new breakage to absorb.

If this surfaces again post-EOL or via a Nimbus migration tool, it's worth revisiting with a proper consumer-impact analysis.

## Test plan

- [x] `node scripts/check-workspace-constraints.js` — green
- [x] `pnpm generate-package-json --all-workspace-packages` — round-trip stable
- [x] `pnpm typecheck` — clean (no regressions from this PR)
- [x] `pnpm lint:publint` — 3 findings closed at source, 1 remaining (the deferred exports suggestion) does not fail the gate
- [ ] CI green

[FEC-936]: https://commercetools.atlassian.net/browse/FEC-936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/fec-936-engines-node.md
+++ b/.changeset/fec-936-engines-node.md
@@ -1,0 +1,13 @@
+---
+'@commercetools-frontend/ui-kit': patch
+---
+
+Declare `engines.node` on all published packages.
+
+Adds `"engines": { "node": ">=22" }` to every published `@commercetools-uikit/*`
+and `@commercetools-frontend/ui-kit` package. Closes one of the publint
+suggestions tracked in FEC-936. Pure metadata — no runtime, build, or
+consumer-API change.
+
+Consumers installing on Node <22 may see a non-fatal `npm install` engine
+warning. Node 22 is the current LTS line; pick a newer Node to silence it.
--- a/.changeset/fec-936-engines-node.md
+++ b/.changeset/fec-936-engines-node.md
@@ -2,12 +2,31 @@
 '@commercetools-frontend/ui-kit': patch
 ---
 
-Declare `engines.node` on all published packages.
+**FEC-936 — publint metadata cleanup across published packages.** Closes
+three of the four publint findings that every published package
+emitted. All changes are pure metadata — no build output, runtime, or
+consumer-API impact.
 
-Adds `"engines": { "node": ">=22" }` to every published `@commercetools-uikit/*`
-and `@commercetools-frontend/ui-kit` package. Closes one of the publint
-suggestions tracked in FEC-936. Pure metadata — no runtime, build, or
-consumer-API change.
+- `"engines": { "node": ">=22" }` declared on every published package.
+  Consumers installing on Node <22 will see a non-fatal `npm install`
+  engine warning.
+- `"type": "commonjs"` declared on every published package, matching
+  what Preconstruct's `.cjs.js` / `.esm.js` output already required Node
+  to assume.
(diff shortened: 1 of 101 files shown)


misama-ct:
Closing — costs outweigh the benefits.

Walking back through the four findings with the EOL context in mind, none of them produce consumer-visible value:

- `engines.node`, `type`, `repository.url` are all metadata that consumers will never observe.
- `pkg.exports` is the only one with real consumer impact, and it carries breaking-change risk (TypeScript type resolution, deep-import patterns we can't fully survey) on a repo whose consumers are already being asked to migrate to Nimbus.

The remaining value was "the publint CI step looks tidier," which doesn't justify 101 files of reviewer load on a sunsetting package. Closing FEC-936 along with this.

PR #3260, opened 2026-06-15
FEC-1010: add `positionFixed` prop and resize-based reflow to prevent viewport clipping

tylermorrisford:
### Problem

When a `TooltipWrapperComponent` renders the tooltip body into a React portal
(e.g. `ReactDOM.createPortal(..., document.body)`), tooltips near the bottom of the
page render partially outside the visible viewport.

Two root causes were identified:

**1. `position: absolute` coordinate drift**
Popper.js defaults to `position: absolute`. When the scroll container is not
`<body>` (e.g. the MC app scrolls an inner div), the absolute coordinates Popper
computes diverge from the viewport, causing the portal tooltip to paint at the
wrong location.

**2. Position computed before content has loaded**
Popper.js calculates the tooltip's position when the popper node is first mounted.
If the body contains an `` that has not yet loaded, the container reports 0
height at that moment. Once the image loads and the container grows, Popper.js never
recomputes — the tooltip extends below the visible area.

---

### Fix

**`positionFixed` prop**

A new optional `positionFixed?: boolean` prop is threaded through to `usePopper`.
When `true`, Popper.js uses `position: fixed` for the popper element instead of
`position: absolute`. Fixed positioning is always viewport-relative and bypasses any
scrolled ancestor that would otherwise shift absolute coordinates.

Consumers should pair this with:
```tsx
modifiers={{
  flip: { enabled: false },
  preventOverflow: { boundariesElement: 'viewport' },
}}
```
`flip` disabled prevents Popper from switching placement (e.g. `"top"` → `"bottom"`)
near the page edge. `preventOverflow` with `boundariesElement: 'viewport'` aligns the
boundary with the fixed-position coordinate system so the tooltip is clamped correctly.

**`ResizeObserver` reflow**

A `ResizeObserver` is attached to the popper element whenever the tooltip is open.
If the element's dimensions change after the initial position was computed (the common
case when an `` loads lazily inside the body), `scheduleUpdate()` is called so
Popper.js recomputes with the correct height and `preventOverflow` can clamp it within
the viewport. The observer is disconnected when the tooltip closes.

---

### Files changed

- `packages/components/tooltip/src/tooltip.tsx`

---

### Backward compatibility

- `positionFixed` is optional and defaults to `false`, preserving existing behavior
  for all current consumers.
- No other public props or the component API are changed.

### DIFF
--- a/.changeset/tiny-bananas-beam.md
+++ b/.changeset/tiny-bananas-beam.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-uikit/tooltip': patch
+---
+
+Prevents image clipping when Tooltip component is hovered.
--- a/packages/components/tooltip/src/tooltip.tsx
+++ b/packages/components/tooltip/src/tooltip.tsx
@@ -107,6 +107,12 @@ export type TTooltipProps = {
    */
   modifiers?: Modifiers;
   /**
+   * Use CSS `position: fixed` for the popper element instead of `position: absolute`.
+   * Recommended when the tooltip body is rendered in a portal (e.g. via `TooltipWrapperComponent`)
+   * to prevent clipping caused by scrolled ancestors.
+   */
+  positionFixed?: boolean;
+  /**
    * Customize the appearance of certain elements of the tooltip.
    */
   components?: TComponents;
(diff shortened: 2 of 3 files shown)


valoriecarli (approved):
thanks for keeping at it with this adventure!!

PR #3273, opened 2026-07-28
fix(calendar-utils):  forward menu ref onto DOM node in CalendarMenu

kterry1:
### Summary
Fixes `TypeError: t.contains is not a function` when a date-input / date-range-input / date-time-input calendar closes on blur/click-outside under downshift ≥ 9.3.4 (React 19).

### Description
The three inputs spread downshift's `getMenuProps()` ref onto the shared `CalendarMenu` a class component, so the ref was the class instance (no .contains()). downshift 9.3.4 dropped the guard that used to tolerate non-DOM refs, so 9.3.4+ throws. ui-kit was pinned to 9.3.3 so it never crashed itself, but consumers on ≥ 9.3.4 did.

### Fix
Make `CalendarMenu` a `forwardRef` function component that forwards the ref onto the `<div>` it already renders. One shared file fixes all three inputs. No DOM/role/ARIA/behavior change; all existing specs pass unchanged. Also unpins downshift 9.3.3 → 9.4.0 (still exact-pinned).

### Testing
Regression test added (open → click-outside asserts no throw); fails with the crash when the fix is reverted, passes with it. Verified live in Storybook on 9.4.0. pnpm typecheck clean.

### Note
`pnpm-workspace.yaml` also sets `jest-circus: true` in allowBuilds; a leftover placeholder that was blocking pnpm start. Unrelated to the crash; happy to split out.

_This video shows the crash with the original class instance, before the forwardRef fix. It's demoed here on downshift 9.4.0 rather than 9.3.6, but the cause is identical: the crash appears on any downshift ≥ 9.3.4 (the version that removed the internal guard)._

https://github.com/user-attachments/assets/3e503c6b-5390-4f19-b4ed-c838e27416c4

_With forward ref fix in place._

https://github.com/user-attachments/assets/c543c221-689d-414a-aba8-9c28e164488c

### DIFF
--- a/.changeset/date-calendar-menu-forwardref.md
+++ b/.changeset/date-calendar-menu-forwardref.md
@@ -1,0 +1,18 @@
+---
+'@commercetools-uikit/calendar-utils': patch
+'@commercetools-uikit/date-input': patch
+'@commercetools-uikit/date-range-input': patch
+'@commercetools-uikit/date-time-input': patch
+---
+
+Fix `TypeError: t.contains is not a function` crash thrown when a date-input,
+date-range-input, or date-time-input calendar closes on blur / click-outside
+under `downshift` >= 9.3.4 (verified against `downshift` 9.4.0).
+
+`CalendarMenu` (shared by all three inputs) received downshift's menu ref via
+`getMenuProps()`, but as a class component the ref pointed at the class instance
+— which has no `.contains()`. downshift 9.3.4 removed the guard that previously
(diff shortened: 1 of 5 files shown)


ByronDWall (approved):
Runs good, tests good, looks good. I looked into the `jest-circus` thing, and it's a no-op so I'm ok with it. But I do want to be really wary when adding packages to that list, because it allows them to execute arbitrary code in our dev environment

PR #3277, opened 2026-08-03
FEC - 1219 Wire up Chromatic

valoriecarli:
Covers FEC 1219, 1223m 1224, 1225 & 1226

ui-kit's VRTs run on Percy today. 
Each component has a `*.visualroute.jsx` page plus a `*.visualspec.js` puppeteer script. 

This branch stands up Chromatic alongside it & converts the first six route files into Storybook stories. 
Percy stays running & untouched & nothing is deleted 

***

### What's in this PR?

  **Chromatic setup**
  - `.github/workflows/chromatic.yml` with TurboSnap & a changed-files gate, so a PR only snapshots what it touched
  - `storybook/chromatic.config.json` & `disableSnapshot: true` in `preview.tsx`  so capture is **opt-in** per story rather than covering all 92 demo stories
  - `VisualSpec` / `VisualSpecGroup` helpers, the successors to Percy's `Spec`
  - A padding decorator, since Chromatic crops to content & edge-painted focus rings would clip

  **`/visualroute-to-story` skill** in `.agents/skills/`, which reads a route file & its spec & emits the story. Migration tooling with a bounded life: it gets deleted with the Percy files at teardown.

  Completed eight conversions, 8 Percy snapshots → 14 stories, 209 states: primary-button, secondary-button, icon-button, secondary-icon-button, flat-button, stamp, tag (split across Tag and TagList), &  text (a composite that splits across its six sub-components).

  ### Reviewing

  All 13 stories are opted into Chromatic, so the build will show them as **new snapshots to accept**, not diffs. There are no baselines on `main` yet.

  Conversions are parity-only: same states, same props, same order as the Percy  route files. States whose  `percySnapshot` call is commented out were skipped rather than revived.

  ### Docs

`.agents/skills/visualroute-to-story/README.md` covers what the migration is &  how to convert a component. Decisions, counts & the per-component list live in the planning doc.


packages/components/buttons/icon-button/src/icon-button.visualroute.jsx
```diff
@@ -168,7 +168,7 @@ export const component = () => (
       <IconButton
         icon={<InformationIcon />}
         shape="square"
-        size="40"
+        size="30"
```
valoriecarli, line 171: The state labelled sizes. When '30' was passing size="40", identical to the '40' state right below it, so size 30 has never been visually covered. Corrected here & in the matching story at icon-button.stories.tsx (ln205)

Ibidapo (approved):

PR #3272, opened 2026-07-20
fix(supply-chain): pin deps and actions, extend shared renovate config

tylermorrisford:
## Summary

Implements the supply chain hardening items from [FEC-968](https://commercetools.atlassian.net/browse/FEC-968) across the ui-kit repo. No production code or published package APIs were changed — this is entirely tooling and CI infrastructure.

## Changes

**pnpm v11 upgrade**

Bumps the `packageManager` field in `package.json` from `pnpm@10.33.4` to `pnpm@11.10.0` (matching the version already adopted in nimbus). The sha512 integrity hash is included so Corepack can verify the binary before running it.

**`onlyBuiltDependencies` → `allowBuilds` migration**

pnpm v11 deprecates `onlyBuiltDependencies` in favour of a unified `allowBuilds` map. The existing 7-entry allowlist (`@percy/core`, `@swc/core`, `core-js`, `core-js-pure`, `esbuild`, `puppeteer`, `unrs-resolver`) is migrated to the new format with no entries added or removed.

Also adds `strictDepBuilds: true`, which causes `pnpm install` to fail immediately if any dependency not in the `allowBuilds` map attempts to run a lifecycle script — this is the strict enforcement mode the ticket calls for. Any new dep that needs a build script will surface as a hard install failure rather than silently succeeding, forcing an explicit review before it's added to the list.

**Renovate config**

Adds `renovate.json` extending the shared `github>commercetools/renovate-config` preset. This brings ui-kit into the standard commercetools Renovate setup (matching nimbus and other repos).

**GitHub Actions pinned to commit SHAs**

All third-party actions across the four affected workflow files are now pinned to their commit SHA with the version tag preserved as a comment. Pinning prevents a compromised or mutated tag from silently changing what runs in CI.

| Action | Tag | SHA |
|---|---|---|
| `actions/checkout` | v6 | `d23441a4` |
| `pnpm/action-setup` | v4 | `b906affc` |
| `actions/setup-node` | v6 | `24997072` |
| `actions/cache` | v5 | `caa29612` |
| `actions/stale` | v10 | `1e223db2` |
| `tibdex/github-app-token` | v2.1.0 | `3beb63f4` |
| `xt0rted/pull-request-comment-branch` | v1 | `653a7d5c` |
| `changesets/action` | v1.7.0 | `6a0a831f` |

[...]

PR #3282, opened 2026-08-05
fix(renovate): remove root config that shadowed .github/renovate.json

ByronDWall:
## Problem

ui-kit has **two** Renovate config files on `main`:

- `renovate.json` (root) — added 2026-07-22 by #3272
- `.github/renovate.json` — the original

Renovate reads the **first** file it finds in a fixed search order, and root `renovate.json` sorts before `.github/renovate.json`. So since #3272 landed, `.github/renovate.json` has been dead — silently, with no warning or error in the run log.

The root file is only:

```json
{ "extends": ["github>commercetools/renovate-config"] }
```

which resolves to `default.json` **alone** — not `composite/component-library`. For the past two weeks ui-kit has therefore been running without:

- every `allowedVersions` guard — `html-loader <2.0.0`, `markdown-loader <8.0.0`, `@percy/puppeteer <=2.0.0`, `react-textarea-autosize <8.4.1`, `react-from-dom <=0.6.2`, `node-fetch <3.0.0`
- `ignoreDeps` for `slate`, `slate-html-serializer`, `slate-react`
- the svgr and stylelint `groupName` groupings
- `rebaseWhen: "never"`
- the storybook / emotion / react / typescript / testing presets from the composite

#3272 touched `renovate.json`, four workflows, `.npmrc`, `package.json` and `pnpm-workspace.yaml` — it never deleted `.github/renovate.json`, so this reads as an accident rather than a deliberate migration.

## Change

Delete the root `renovate.json`, keep `.github/renovate.json`. That matches where the rest of the org keeps it — 7 of the 11 commercetools repos that have a Renovate config use `.github/renovate.json`, including every merchant-center repo.

**The supply-chain hardening from #3272 is not affected.** `composite/component-library` extends `default.json`, which already pulls in `config:best-practices` (`:pinDevDependencies`, `docker:pinDigests`) and `helpers:pinGitHubActionDigests`. Pinning behaviour is identical either way.

## Not in scope

- **ui-kit is in Renovate's `silent` mode**, set on the Mend side, so every update — including the `dompurify`, `immutable` and `postcss` security patches — waits on a manual checkbox. This PR doesn't change that; it needs a Mend portal change. Worth doing, since patch/minor automerge from the shared config can't fire while silent mode is on.
- ui-kit doesn't extend `composite/auto-merging`, unlike merchant-center-frontend / -prices / -services. Probably should, but that's a policy call for a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,0 @@
-{
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["github>commercetools/renovate-config"]
-}


valoriecarli (approved):

tylermorrisford:
@valoriecarli Thanks for beating me to the approval! @ByronDWall Thank you for finding this! Coupla gems, you two

tylermorrisford (approved):
🚀

PR #3291, opened 2026-08-10
chore(renovate): hold pnpm at 11.14.0 until deploy --legacy regression is fixed

ByronDWall:
## Summary

Adds a Renovate `packageRules` entry that disables updates to `pnpm` past our intended pin of `11.14.0`.

## Why

pnpm/pnpm#13618 is a regression, first appearing in pnpm 11.19.0 and still present in the current published latest (11.20.0), where a workspace dependency that has peerDependencies is left as a symlink escaping the deploy directory instead of being packed in when running `pnpm deploy --legacy`. In a typical Docker `COPY ./deploy .` flow this creates a dangling symlink that crashes the app at runtime.

The fix (pnpm/pnpm#13755) merged upstream but has not shipped in a published pnpm release yet.

I did not find any usage of `pnpm deploy --legacy` in this repo's Docker builds or scripts, so this hold is precautionary rather than fixing a currently broken build. It keeps Renovate from proposing a bump past our intended `11.14.0` target while the regression is unresolved upstream, so we do not pick it up unintentionally later.

This mirrors the guard already in place on [commercetools/nimbus](https://github.com/commercetools/nimbus/blob/main/.github/renovate.json), which holds pnpm at `11.14.0` for the same reason.

This PR is independent of and complements #3290 (the `packageManager` bump to `pnpm@11.14.0`). It only affects Renovate's future update proposals and pairs with whatever pin is currently in `package.json`.

## Validation

- `jq . .github/renovate.json` passes.
- `npx --yes -p renovate renovate-config-validator .github/renovate.json` reports the config validated successfully.

## Follow-up

Remove this rule once a released pnpm version includes the fix from pnpm/pnpm#13755.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -40,6 +40,11 @@
     {
       "matchPackageNames": ["node-fetch"],
       "allowedVersions": "<3.0.0"
+    },
+    {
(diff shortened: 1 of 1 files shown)


ByronDWall:
Superseded by a consolidated PR bumping to pnpm 11.17.0 and correcting this rule's version reference.

PR #3295, opened 2026-08-10
FEC - 1227 -> 1232 -> Percy to Chromatic

valoriecarli:
Build results: https://www.chromatic.com/build?appId=6a6cf2304a0c917990e1a8aa&number=22

Converts 25 Percy visual routes into Storybook stories that Chromatic captures.
The goal is **parity**: the same states, same props and same order the Percy routes already have, so Chromatic can take over without changing what is covered.

## Description

| Ticket   | Scope                                                            |
| -------- | ---------------------------------------------------------------- |
| FEC-1227 | `icons`, `spacings`, `dropdown-menu` — the unconventional routes |
| FEC-1228 | Select inputs                                                    |
| FEC-1229 | Text inputs                                                      |
| FEC-1230 | Localized inputs                                                 |
| FEC-1231 | Date and time inputs                                             |
| FEC-1232 | Numeric and choice inputs                                        |

Each component gains an `AllVariants` story appended to its existing stories file, tagged `vrt` and opted into capture with `chromatic: { disableSnapshot: false }`. Demo stories are untouched and stay
uncaptured.

**Parity, not new coverage.** States whose `percySnapshot` is commented out are skipped — no baseline exists for them today, so converting them would add coverage rather than migrate it.

**Nothing Percy is deleted.** Route files, specs and the Percy CI step all stay until parity is signed off.

### Deliberate deviations from Percy

- `icons` color stories are hidden from the sidebar with `!dev`; Chromatic still captures them.

### Follow-up

**A final cleanup round happens once all snapshots are gathered.** Presentation details are deliberately not being polished component by component: label formatting and alignment, whether the demo stories fold into the captured set, and the Percy teardown all land in one pass at the end, when the full set of baselines exists to judge them against.

PR #3296, opened 2026-08-11
FEC - 1237 - 1246 Chromatic wrap up

valoriecarli:
ui-kit's VRTs run on Percy, and we're replacing it with Chromatic. 
Each` *.visualroute.jsx `+ `*.visualspec.js` pair becomes a captured Storybook story. 
Though some questionable presentation tweaks were made, this migration is **PARITY ONLY.** Same states, same props, same order, so Chromatic takes over w/o changing what's covered.

This is the last conversion batch that includes the remaining fields, inputs and stragglers, plus theme-provider. 
It also tightens the Chromatic comments and skill docs.

The numbers aren't 1:1 for reasons, but every **_live_** Percy snapshot now has a story.

Percy is running concurrently, nothing is deleted at this time.

References:
- #3277 — Chromatic setup and the first 8 routes
- #3295 — 25 routes
- [FEC-1247](https://commercetools.atlassian.net/browse/FEC-1247) — Percy teardown
- Chromatic builds for [uikit](https://www.chromatic.com/builds?appId=6a6cf2304a0c917990e1a8aa)
- Percy builds for [uikit](https://percy.io/commercetools-GmbH/web/ui-kit/)

[FEC-1247]: https://commercetools.atlassian.net/browse/FEC-1247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.agents/skills/visualroute-to-story/README.md
+++ b/.agents/skills/visualroute-to-story/README.md
@@ -15,35 +15,43 @@ converted and the gap is recorded in the planning doc.
 ## Where we deviate on purpose
 
 Parity is about which states are covered, not matching Percy pixel for pixel.
-Four departures, none changing what is under test:
+The scaffolding differences are in
+[What a converted story changes](#what-a-converted-story-changes). Two more
+departures, neither changing what is under test:
 
-- **Padding.** A global decorator adds `1rem` inside each story: Chromatic crops
-  to rendered content, so anything painted at the edge would clip.
-- **Layout.** Label beside the component, smaller per-state min-height, so a long
-  frame stays reviewable. A state whose content overflows its container needs
-  `overflow: hidden` on that container, or the overflow draws over the label.
-- **Grouping.** Runs of states sharing an axis may get a `VisualSpecGroup`
-  heading, only where it removes ambiguity.
+- **Padding.** A global decorator adds `1rem` inside each story, except where a
+  story sets `layout: 'fullscreen'`: Chromatic crops to rendered content, so
+  anything painted at the edge would clip.
 - **Light fixes.** A prop the component's types reject gets corrected, not cast,
   and called out in the PR. If the fix moves pixels, the route file gets it too,
   so Percy and Chromatic stay comparable while both run.
 
+One consequence of the label moving above the component: a state whose content
+overflows its container needs `overflow: hidden` on that container, or the
+overflow draws over a neighboring label.
+
 ## Percy vs Storybook + Chromatic
 
-| Concern                 | Percy                                                  | Storybook + Chromatic                                    |
-| ----------------------- | ------------------------------------------------------ | -------------------------------------------------------- |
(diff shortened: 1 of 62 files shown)


.agents/skills/visualroute-to-story/README.md
```diff
@@ -50,126 +47,86 @@ The same job, split across different pieces:
...
-A converted story renders the same states, from the same props, in the same
-source order. What changes is the scaffolding Percy wrapped around them.
-
-Percy's `Spec` (`test/percy/spec.jsx`) becomes `VisualSpec`
-(`storybook/src/helpers/visual-spec.tsx`):
+Same states, same props, same source order. Only Percy's scaffolding differs:
 
 |                      | Percy `Spec`                             | `VisualSpec`               |
 | -------------------- | ---------------------------------------- | -------------------------- |
 | Layout               | column: label bar, prop table, component | row: component, then label |
```
tylermorrisford: Should we update this? The table still says "row: component, then label" but `visual-spec.tsx` now renders a column with the label on top. The old description matched the grid layout; the new one is `flex-direction: column` with `<Label>` before `<Box>`.

packages/components/fields/creatable-select-field/src/creatable-select-field.stories.tsx
```diff
@@ -171,3 +171,132 @@ BasicExample.args = {
...
+          name="form-field-name"
+          value={visualValue}
+          onChange={() => {}}
+          options={visualOptions}
+          horizontalConstraint={7}
+          /* CreatableSelectField types isReadOnly as ConstrainBooleanParameters,
+             unlike its four siblings which use boolean. The component only does
+             `if (!this.props.isReadOnly)`, so this object is truthy and renders
+             exactly as the route file's `true` did. */
+          isReadOnly={{ exact: true }}
```
tylermorrisford, line 252: The comment explains this well — just flagging it so we don't lose track. `CreatableSelectField` types `isReadOnly` as `ConstrainBooleanParameters` while its four siblings use a plain boolean. Worth a follow-up ticket to align the prop type, separate from this migration.

tylermorrisford (approved):
The theme-provider.spec.tsx is a genuinely good test file — covers the DOM-writing behavior, scoped vs root targeting, overrides, custom properties, and the MutationObserver-based reactivity all in one place. A lot of care went into this one.

Two nits inline.

.agents/skills/visualroute-to-story/README.md
valoriecarli: you right, updated in https://github.com/commercetools/ui-kit/pull/3296/commits/cb28b00ae5837d8cf34b7d35ed1cacb207535140

packages/components/fields/creatable-select-field/src/creatable-select-field.stories.tsx
valoriecarli: for reasons i will politely decline. thank you, though!

